PR #49, opened 2021-05-07
Add core infra terraform templates

deniscep:
# Add core infra terraform templates

## What is being addressed

- Implement a set of terraform templates to create TRE Management core as described in infra diagram

## How is this addressed

- Added terraform modules for VNET, Subnets, Keyvault, Firewall, S AppGateway, ACR, private DNS, and API Web App (22 resources and 1 resource group provisioned)
- Subnets are delegated/dedicated as specified in infra
- Private endpoints are provisioned for resources on private link, and private DNS is associated with VNET
- Standardized naming convention to "resource abbreviation"-"prefix"-"environment"-"unique-random-id" (e.g. rg-tre-dev-0001)
- Route table is provisioned and associated with Shared subnet, with default 0.0.0.0 route pointing to firewall
- Added tfvars template with sample defaults

Out of scope for this PR:
- hooking up app gateway to app service (pending decision on deployment implementation)
- no firewall rules are added at this time (pending decision on exceptions)
- deployment/composition service resources are not provisioned (pending decision on Porter)
- Gitea & Nexus are not provisioned (pending decision on ACI vs VM)

Closes 5465 in AzDO


templates/core/terraform/.terraform.lock.hcl
```diff
@@ -0,0 +1,38 @@
+# This file is maintained automatically by "terraform init".
```
marrobi, line 1: Should this be in .gitignore and not in the PR?

templates/core/terraform/api-webapp/api-webapp.tf
```diff
@@ -0,0 +1,215 @@
...
+  }
+
+  https_only = true
+  app_settings = {
+
+    "APPINSIGHTS_INSTRUMENTATIONKEY" = azurerm_application_insights.core.instrumentation_key
+  }
+
+  site_config {
+    app_command_line            = "gunicorn -w 2 -k uvicorn.workers.UvicornWorker main:app"
```
marrobi, line 39: What's this doing at this point? Not required if using a docker image. Guess need to decide on an approach.

templates/core/terraform/api-webapp/api-webapp.tf
```diff
@@ -0,0 +1,215 @@
...
+  target_resource_id         = azurerm_app_service.management_api.id
+  log_analytics_workspace_id = var.log_analytics_workspace_id
+
+  log {
+    category = "AppServiceHTTPLogs"
+    enabled  = true
+
+
+    retention_policy {
+      days    = 1
```
marrobi, line 123: Logging policies probably need further consideration. Another issue maybe?

templates/core/terraform/appgateway/appgateway.tf
```diff
@@ -0,0 +1,61 @@
...
+
+resource "azurerm_application_gateway" "agw" {
+  name                = "agw-${var.resource_name_prefix}-${var.environment}-${var.tre_id}"
+  resource_group_name = var.resource_group_name
+  location            = var.location
+
+  sku {
+    name     = "Standard_v2"
+    tier     = "Standard_v2"
+    capacity = 2
```
marrobi, line 17: Do we need 2? Is 1 not sufficient?

templates/core/terraform/appgateway/appgateway.tf
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  gateway_ip_configuration {
+    name      = "gateway-ip-configuration"
+    subnet_id = var.app_gw_subnet
+  }
+
+  frontend_port {
+    name = local.frontend_port_name
+    port = 80
```
marrobi, line 27: Need to create an issue for this - shouldn't be exposing port 80. Should be 443, but will need a certificate.

templates/core/terraform/appgateway/appgateway.tf
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  backend_address_pool {
+    name = local.backend_address_pool_name
+  }
+
+  backend_http_settings {
+    name                  = local.http_setting_name
+    cookie_based_affinity = "Disabled"
+    port                  = 80
```
marrobi, line 42: All traffic needs to be encrypted point to point.

templates/core/terraform/firewall/firewall.tf
```diff
@@ -0,0 +1,38 @@
...
+
+  route {
+    name           = "DefaultRoute"
+    address_prefix = "0.0.0.0/0"
+    next_hop_type  = "VirtualAppliance"
+    next_hop_in_ip_address = azurerm_firewall.fw.ip_configuration.0.private_ip_address
+  }
+}
+
+resource "azurerm_subnet_route_table_association" "rt_shared_subnet_association" {
```
marrobi, line 35: Should this live with the subnet tf, rather than the firewall?

templates/core/terraform/network/locals.tf
```diff
@@ -0,0 +1,8 @@
+locals {
+  core_services_vnet_subnets            = cidrsubnets(var.address_space, 2, 4, 4, 2, 2)
```
marrobi, line 2: Has this been checked as appropriate? /16 is likely too large for our requirements, and not sure about these allocations.

templates/core/terraform/keyvault/keyvault.tf
```diff
@@ -0,0 +1,39 @@
...
+  resource_group_name   = var.resource_group_name
+  private_dns_zone_name = azurerm_private_dns_zone.vaultcore.name
+  virtual_network_id    = var.core_vnet
+}
+
+resource "azurerm_private_endpoint" "kvpe" {
+  name                = "pe-kv-${var.resource_name_prefix}-${var.environment}-${var.tre_id}"
+  location            = var.location
+  resource_group_name = var.resource_group_name
+  subnet_id           = var.shared_subnet
```
marrobi, line 26: Is the shared subnet appropriate? Is the key vault really shared by workspace users, or is it actually just used by composition service and for certs?

marrobi (commented):
Added some comments from a quick review of tf on my mobile. Some things might need pushing into other issues.

Have naming conventions been considered?

Great to see this making progress.

templates/core/terraform/.terraform.lock.hcl
deniscep: Since Terraform 0.14 dependency lock file is a recommendation to commit to git as per:
https://www.terraform.io/docs/language/dependency-lock.html

templates/core/terraform/api-webapp/api-webapp.tf
deniscep: Since terraform 0.14 dependency lock file is recommended to be committed to git, as per:
https://www.terraform.io/docs/language/dependency-lock.html

deniscep: It's essentially cutting a step from github action on CI/CD, once we hook it up. This can be replaced once the container approach is built.

templates/core/terraform/.terraform.lock.hcl
marrobi: Cool. Didn't know that.

templates/core/terraform/api-webapp/api-webapp.tf
deniscep: Yeah, we should have a story on what logging policies in general are. @christoferlof would you mind adding a story and then resolving this with it?

templates/core/terraform/appgateway/appgateway.tf
deniscep: Yes, Standard_v2 SKU is required by Standard public IP. (Old) Standard SKU only allows for a Basic public IP.

marrobi: Not the SKU, capacity unit.

templates/core/terraform/appgateway/appgateway.tf
deniscep: Agree, right now it's base-configured to 80 but not hooked up to anything (so nothing is exposed). This is because if we base-configure to 443, we do need to have a certificate first. Since we don't have a process on acquiring a certificate, we need to have a separate story on certificate provisioning for app gateway @christoferlof if you can add, and then resolve this one.

templates/core/terraform/appgateway/appgateway.tf
deniscep: same as above

templates/core/terraform/appgateway/appgateway.tf
marrobi: Maybe make a variable with a default of 1?

templates/core/terraform/network/locals.tf
deniscep: No, but I haven't seen any specs on this, so defaulting to something. We can add a story to specify the subnet space @christoferlof

templates/core/terraform/firewall/firewall.tf
marrobi: I was more thinking about when we do workspaces, the route associations will need to be with the network, as is no firewall module in those templates.

templates/core/terraform/keyvault/keyvault.tf
deniscep: Valid question, everything with PE is right now hookedup into the Shared. I would run with this as a initial, and refine as we go along.

marrobi: @christoferlof can we have an issue for this. Given private endpoints don't obey nsg rules at present, I tend to keep them separate, and need to ensure traffic is blocked outbound to that subnet from everything else.

Also what's the NSG setup in this PR? Need another issue for that?

templates/core/terraform/api-webapp/api-webapp.tf
```diff
@@ -0,0 +1,215 @@
...
+    http_logs {
+      file_system {
+        retention_in_days = 7
+        retention_in_mb   = 100
+      }
+    }
+  }
+}
+
+resource "azurerm_user_assigned_identity" "management_api" {
```
sachinkundu, line 71: This now expects AAD owner isnt it? Can an unpriveleged account create these? Should the identities be created in a different process and we can import them using data block?

templates/core/terraform/routetable/routetable.tf
```diff
@@ -0,0 +1,18 @@
+resource "azurerm_route_table" "rt" {
+  name                          = "rt-${var.resource_name_prefix}-${var.environment}-${var.tre_id}"
+  resource_group_name           = var.resource_group_name
+  location                      = var.location
+  disable_bgp_route_propagation = false
+
+  route {
+    name           = "DefaultRoute"
```
sachinkundu, line 8: Do the packets intended for the subnet also take a hop from the firewall?

templates/core/terraform/api-webapp/api-webapp.tf
marrobi: This could possibly be a system assigned identity too? Reason it was user assigned in this is example is we were using the same identity for multiple apps to simplify assignment of permissions. However think in the architecture we are looking at system assigned will work?

deniscep: I think that it requires only "Managed Identity Contributor" https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-manage-ua-identity-portal

deniscep: @marrobi good point. we can for now remove this part and rely on system identity, and then add this if we need it later on.

templates/core/terraform/routetable/routetable.tf
deniscep: AFAIK no, @marrobi if you want to elaborate?

marrobi: Traffic on the same subnet will not need to be routed so no.

Also route tables are evaluated most specific to least specific - so this is  only for traffic without a more specific route defined - https://docs.microsoft.com/en-us/azure/virtual-network/virtual-networks-udr-overview#how-azure-selects-a-route

So get something like: 

image

templates/core/terraform/api-webapp/api-webapp.tf
christoferlof: Please see #50 .

templates/core/terraform/appgateway/appgateway.tf
christoferlof: Please see #51

templates/core/terraform/network/network.tf
```diff
@@ -0,0 +1,53 @@
+resource "azurerm_virtual_network" "core" {
+  name                = "vnet-${var.resource_name_prefix}-${var.environment}-${var.tre_id}"
+  location            = var.location
+  resource_group_name = var.resource_group_name
+  address_space       = [var.address_space]
+}
+
+resource "azurerm_subnet" "bastion" {
+  name                 = "AzureBastionSubnet"
```
askew, line 9: Are we using Azure Bastion? I thought not.

templates/core/terraform/network/network.tf
```diff
@@ -0,0 +1,53 @@
+resource "azurerm_virtual_network" "core" {
```
askew, line 1: We should have NSGs on the subnets.

templates/core/terraform/network/network.tf
deniscep: @marrobi requested we have subnet created for it; comments?

templates/core/terraform/network/network.tf
deniscep: A new story #53 is created to define NSGs defaults.

templates/core/terraform/keyvault/keyvault.tf
christoferlof: Please see #53

templates/core/terraform/network/network.tf
marrobi: It might be needed for administration - I usually se it deployed.

As it needs a dedicated subnet with specific name, my view is put the subnet in, as to add it post deployment is a headache to manage address space wise.

christoferlof: The scenario here is for troubleshooting etc and not for the Virtual Desktop service.

christoferlof (approved):

templates/core/terraform/appgateway/appgateway.tf
askew: The way I've don't this, is to create the listener with a self-signed cert `CN=temp` which allows you to configure the routing correctly. You then replace the cert with `certbot`, or whatever you need to use. I have an example the uses `certbot` I can share.

templates/core/terraform/appgateway/appgateway.tf
```diff
@@ -0,0 +1,61 @@
...
+    port = 80
+  }
+
+  frontend_ip_configuration {
+    name                 = local.frontend_ip_configuration_name
+    public_ip_address_id = azurerm_public_ip.appgwpip.id
+  }
+
+  backend_address_pool {
+    name = local.backend_address_pool_name
```
askew, line 36: Why not hook up the app service that is also created with these templates.

deniscep: We need https to expose API, but we don't have a process for creation of certificates used by the gateway.

deniscep: A new story has been created for this as well.

TessFerrandez (approved):

PR #68, opened 2021-05-11
Add private link for state store cosmosdb

sachinkundu:
# PR for issue #5292 -> #5549

## What is being addressed

Enable private link for statestore

## How is this addressed

Terraform for private dns zone, virtual network link and private endpoint.

### DIFF
--- a/templates/core/terraform/state-store/statestore.tf
+++ b/templates/core/terraform/state-store/statestore.tf
@@ -26,15 +26,15 @@ resource "azurerm_cosmosdb_sql_database" "tre-db" {
   throughput          = 400
 }
 
-resource "azurerm_private_dns_zone" "statestore" {
-  name                = "privatelink.statestore.azure.net"
+resource "azurerm_private_dns_zone" "cosmos" {
+  name                = "privatelink.documents.azure.com"
   resource_group_name = var.resource_group_name
 }
 
-resource "azurerm_private_dns_zone_virtual_network_link" "statestorelink" {
-  name                  = "statestorelink"
+resource "azurerm_private_dns_zone_virtual_network_link" "cosmos_documents_dns_link" {
+  name                  = "cosmos_documents_dns_link"
   resource_group_name   = var.resource_group_name
-  private_dns_zone_name = azurerm_private_dns_zone.statestore.name
+  private_dns_zone_name = azurerm_private_dns_zone.cosmos.name
   virtual_network_id    = var.core_vnet
 }
 
@@ -46,7 +46,7 @@ resource "azurerm_private_endpoint" "sspe" {
 
   private_dns_zone_group {
     name                 = "private-dns-zone-group"
-    private_dns_zone_ids = [azurerm_private_dns_zone.statestore.id]
+    private_dns_zone_ids = [azurerm_private_dns_zone.cosmos.id]
   }
 
   private_service_connection {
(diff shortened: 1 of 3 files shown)


templates/core/terraform/state-store/statestore.tf
```diff
@@ -25,3 +25,34 @@ resource "azurerm_cosmosdb_sql_database" "tre-db" {
   account_name        = azurerm_cosmosdb_account.tre-db-account.name
   throughput          = 400
 }
+
+resource "azurerm_private_dns_zone" "statestore" {
+  name                = "privatelink.statestore.azure.net"
```
marrobi, line 30: This should be `privatelink.documents.azure.com.`.

templates/core/terraform/state-store/statestore.tf
```diff
@@ -25,3 +25,34 @@ resource "azurerm_cosmosdb_sql_database" "tre-db" {
   account_name        = azurerm_cosmosdb_account.tre-db-account.name
   throughput          = 400
 }
+
+resource "azurerm_private_dns_zone" "statestore" {
```
marrobi, line 29: The DNS zone resource isn't necessarily specific to statestore - it's all cosmos. Can the name reflect that?

templates/core/terraform/state-store/statestore.tf
```diff
@@ -25,3 +25,34 @@ resource "azurerm_cosmosdb_sql_database" "tre-db" {
...
+  name                = "privatelink.statestore.azure.net"
+  resource_group_name = var.resource_group_name
+}
+
+resource "azurerm_private_dns_zone_virtual_network_link" "statestorelink" {
```
marrobi, line 34: Again, can we name `cosmos_documents_dns_link` or similar s isn't state store specific.

marrobi (changes requested):
Good progress, couple of comments around naming and private link DNS address.

templates/core/terraform/state-store/statestore.tf
sachinkundu: I can change it. How did you come up with the name though? Is this documented?

templates/core/terraform/state-store/statestore.tf
sachinkundu: Do you think that is a better name? statestore is cosmos in our domain and hence isn't it better to be more concrete than being more general.

templates/core/terraform/state-store/statestore.tf
sachinkundu: Same comment as above. I don't mind changing but I think statestore is more concrete than cosmos.

templates/core/terraform/state-store/statestore.tf
marrobi: https://docs.microsoft.com/en-us/azure/private-link/private-endpoint-dns

There are a few for Cosmos - not sure which is the one we need.

templates/core/terraform/state-store/statestore.tf
marrobi: I guess it's more convention. Becomes more of an issue when have multiple resources sharing the same DNS zone. So might have multiple web apps sharing a azure website private link DNS zone, so would call it `azurewebsites_private_dns_zone`, rather than `management_api_dns_zone`. 

Are there any DNS zones already in the templates?

templates/core/terraform/state-store/statestore.tf
sachinkundu: Thanks½

templates/core/terraform/state-store/statestore.tf
sachinkundu: Cool, thanks!

marrobi (approved):

Lybecker (approved):
Looks good to me

PR #75, opened 2021-05-12
API dockerfile and workflow

marrobi:
# PR for issue 70 - Put management API into a docker container

## What is being addressed

- Create dockerfile for API
- Create workflow to upload to GitHub package registry

## How is this addressed

- Docker image based on `python:3.8`, API runs as none root user
- Workflow triggers on push to `main` or `develop`, or PR. Image tagged with branch name or PR branch name. Image pushed to this repo's package registry.

Created #78 to clean up old images.


.github/workflows/build_and_push_docker_images.yml
```diff
@@ -0,0 +1,43 @@
...
+          DOCKER_BUILDKIT: 1
+          REPOSITORY_NAME: docker.pkg.github.com/microsoft/azuretre/management-api
+        run: |
+          echo $GITHUB_REF
+          if [ $GITHUB_EVENT_NAME == 'push' ] && [ $GITHUB_REF == 'refs/heads/develop' ]; then
+            TAG='develop-latest'
+          elif [ $GITHUB_EVENT_NAME == 'push' ] && [ $GITHUB_REF == 'refs/heads/main' ]; then
+            TAG='main-latest'
+          else
+            TAG=$(jq --raw-output .pull_request.number $GITHUB_EVENT_PATH)
```
jjcollinge, line 39: Does this differentiate image builds on a given PR or should we include a git sha or index? Just thinking there might be multiple images built for a PR given feedback and iteration etc. and it'll be better to be able to pin point which image refers to which iteration and also ensure that any PR test infra picks up a specific image.

core/README.md
```diff
@@ -34,6 +34,22 @@ Run the API Tests locally
...
+From the root of the repository, build the container image:
+
+```cmd
+cd core/api
+docker build -t tre-management-api .
+```
+
+To run:
+```cmd
+docker run -it -e PORT=8000 -p 8000:8000 tre-management-api
```
jjcollinge, line 48: What's the use case for having a variable internal port? Can we not fix the API to listen on 8000 and just allow them to provide a different mapping at the docker layer?

.github/workflows/build_and_push_docker_images.yml
marrobi: No, all images in the PR are tagged the same, so overwrites. True, need to avoid caching and trigger a change. Will change to git sha.

core/README.md
marrobi: Ah, this is left over from when linux web apps was in preview, they passed a PORT variable that varied (and still do - but is always 80 now unless otherwise specifed).

https://docs.microsoft.com/en-us/azure/app-service/configure-custom-container?pivots=container-linux#configure-port-number

I'll leave it to 8000, and set WEBSITES_PORT in the app service.

deniscep (approved):
Looks good, clean and straightforward, action passes, container available.

jjcollinge (approved):
LGTM

PR #77, opened 2021-05-12
Add a workspace service template

deniscep:
# Add a workspace service template

## What is being addressed

- Adds a storage service provisioning sample 

## How is this addressed

- Adds storage template with private link, connecting to workspace subnet


christoferlof (approved):

marrobi (approved):
LGTM. 

Only comment would be is this not "storage_service" rather than base_service"? But a good starter to enable further work either way.

deniscep:
> "storage_service"



> LGTM.
> 
> Only comment would be is this not "storage_service" rather than base_service"? But a good starter to enable further work either way.

Good point, changing name

PR #79, opened 2021-05-12
Add docker image to management API

marrobi:
# PR for issue 73

## What is being addressed

Added docker image to management API and exposed through app gateway

## How is this addressed

- Added docker settings to Azure Web App
- Added path mapping for /api to app gateway to web app back end
- upped TF provider version to more recent version

image


templates/core/terraform/appgateway/appgateway.tf
```diff
@@ -14,7 +14,7 @@ resource "azurerm_application_gateway" "agw" {
   sku {
     name     = "Standard_v2"
     tier     = "Standard_v2"
-    capacity = 2
+    capacity = 1
```
TessFerrandez, line 17: Is this for the docker container? or if not, can you mention why this change is made?

TessFerrandez (commented):
We should also update the dev setup documentation on how this is used

TessFerrandez (approved):

templates/core/terraform/appgateway/appgateway.tf
marrobi: Ah, not really part of this PR, we only need 1 - half the cost - can put it back or in another PR if think I should.

TessFerrandez: No - it's cool, was just curious :)

deniscep (approved):

PR #85, opened 2021-05-13
Add NSG rules to workspace services subnet, add bastion

marrobi:
# PR for issue 53

## What is being addressed

Restrict ability to access external network locations from the workspace services subnet, but retain access to the shared subnet.

## How is this addressed

- Added NSG configuration below:
image

- To troubleshoot had to add a bastion. Might want to make this optional with a variable in future. Will create an issue. Can see could access API in shared subnet, but not the internet.
image

Created issue to create network flow diagram: https://github.com/microsoft/AzureTRE/issues/87

### DIFF
--- a/templates/core/terraform/bastion/bastion.tf
+++ b/templates/core/terraform/bastion/bastion.tf
@@ -1,5 +1,5 @@
 resource "azurerm_public_ip" "bastion" {
-  name                  = "pip-bas-${var.resource_name_prefix}-${var.environment}-${var.tre_id}"
+  name                = "pip-bas-${var.resource_name_prefix}-${var.environment}-${var.tre_id}"
   resource_group_name = var.resource_group_name
   location            = var.location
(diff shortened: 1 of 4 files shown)


templates/core/terraform/bastion/bastion.tf
```diff
@@ -0,0 +1,20 @@
+resource "azurerm_public_ip" "bastion" {
+  name                  = "pip-bas-${var.resource_name_prefix}-${var.environment}-${var.tre_id}"
+  resource_group_name = var.resource_group_name
```
jjcollinge, line 3: nit: alignment

sachinkundu (approved):

templates/core/terraform/bastion/bastion.tf
marrobi: done :)

jjcollinge (approved):
lgtm

PR #91, opened 2021-05-17
Adds persistent remote terraform state in Azure

### LINKED ISSUE #71: Persist terraform state in Azure
## Description

As a cloud administrator
I want to persist the terraform state in azure
So that the terraform state is in a common location and can be used collaboratively (among other things - see link)

## Acceptance critiera

- [x] Script to provision storage for terraform state
- [x] Terraform scripts (for core) updated to store state in azure

https://docs.microsoft.com/en-us/azure/developer/terraform/store-state-in-azure-storage

deniscep:
# Adds persistent remote terraform state in Azure

## What is being addressed

Changes locally stored terraform state to a common storage location in Azure so that it can be used collaboratively.

## How is this addressed

- Adds a script for creation of terraform state RG, STG and container
- Adds remote block to main terraform scripts
- Alters documentation to reflect the change

Closes #71


marrobi (changes requested):
I've added some comments, but stopped as I'm slightly confused what are we trying to achieve here?

Who is this state setup for? 

The main challenges I see are:
- State implementation is leaking into the project code
- The quick start is becoming less of a quick start - we shouldn't be documenting how to do remote TF state on Azure.

I would usually expect whoever is deploying the solution to make the choices about where to store their TF state, be it local, Azure, or somewhere else, such as TF cloud. We can provide bootstrap scripts for a dev experience, or for our CD pipelines, but shouldn't fix the state config as part of the TF templates. Also worth noting the workspace services state will have to be managed by the composition service - rather than being a static backend config.

This blog by Ben Coleman @benc-uk (in one of the UK crews)  is a good approach: https://dev.to/cse/terraform-bootstrap-backend-state-in-azure-5b73 (ignoring the fact its AzDO focused).

It has two parts:
1. A bootstrap script to create the deployment dependencies - tf state storage - key vault for secrets, ACR if you need a container registry for deployment (as per our chat earlier) etc.
2. Then a deploy script - the backend config passed in here: https://github.com/benc-uk/terraform-mgmt-bootstrap/blob/bf284169841445eca232b6225dbeebef877af6ae/deploy.sh#L53

Feel like that would be a good way forward. Let me know if want to discuss.

deniscep:
Closing as the task isn't clear, re-assign to @marrobi

Issue #66, opened 2021-05-11
TRE developer creates SSL certificate for development purposes


marrobi:
## Description

As a developer I want to acquire a certificate that can be used when I deploy the TRE.

So that I can provide the certificate when deploying the TRE #51 

## Acceptance criteria

- [ ] Set of commands to create certificate request and process certificate request
- [ ] Works cross platform
- [ ] Document the commands
- [ ] Wildcard certificate to enable subdomains to be used on workspaces
- [ ] Mirrors process of acquiring a certificate from corporate IT as close as possible. I.e. Create request, process request. So that the same process can be used to create a request to give to corporate IT to process.

askew:
I've got scripts that will use certbot to request a cert and upload it to KeyVault. We don't need to try to guess what a customers CA workflow might be, it will vary. But we should be able to assume a customer is able to store and renew certs in KeyVault.

askew:
Working draft #103

Issue #94, opened 2021-05-19
Initial deployment fails with error 'Public IP <app gw name> was not found.


christoferlof:
**Describe the bug**
When following the steps in the quickstart to deploy a new Azure TRE instance, the deployment fails on the first run. Running the deployment again seems to resolve the issue.

Error message
```
module.appgateway.data.azurerm_public_ip.appgwpip_data: Still reading... [11m0s elapsed]
module.state-store.azurerm_private_endpoint.sspe: Creation complete after 8m25s [id=/subscriptions/<id>/resourceGroups/<resource group>/providers/Microsoft.Network/privateEndpoints/pe-ss-<name>-<env>-<id>]
╷
│ Error: Error: Public IP "pip-agw-<name>-<env>-<id>" (Resource Group "rg-<name>-<env>-<id>") was not found
│ 
│   with module.appgateway.data.azurerm_public_ip.appgwpip_data,
│   on appgateway/appgateway.tf line 82, in data "azurerm_public_ip" "appgwpip_data":
│   82: data "azurerm_public_ip" "appgwpip_data" {
│ 
╵
```

**Steps to reproduce**

1. Follow the instructions in quickstart doc.
1. Wait for `make tre-deploy`
1. Review error message

PR #103, opened 2021-05-24
Setup Application Gateway to use SSL

askew:
# PR for issue #66 

## What is being addressed

This PR sets up the application gateway with an HTTPS listener and a default redirect from HTTP to HTTPS.
It has a default backend of a static website in a storage account, to be used to host future UI website. It also provides a mechanism to host the ACME challenge response needed to acquire a certificate from LetsEncrypt.

The `/api/*` route will pass though to the App Service hosting the management API.

The LetsEncrypt certificate is requested using the scripts which are design to be run automatically from the DevOps pipeline.

Once the core resources are deployed a LetsEncrypt certificate can be obtained with
```
make letsencrypt
```


askew:
I need to check, but I think app gateway may not be able to retrieve certificates from KeyVault with a private endpoint. We've need a second KeyVault in this case.

limorl:
@askew would be good to document the commands in the develop quickstart - It's also one of the success criteria :)

devops/scripts/letsencrypt.sh
```diff
@@ -0,0 +1,111 @@
...
+        --account-name "${STORAGE_ACCOUNT}" \
+        --auth-mode login \
+        --container-name '$web' \
+        --file index.html \
+        --name index.html \
+        --no-progress \
+        --only-show-errors
+
+    # Wait a bit for the App gateway health probe to notice.
+    echo "Waiting 30s for health probe"
```
limorl, line 56: would be good to use the health probe interval parameter (times a factor), in case it changes. I understand it's not in the context of this script, but can be made available if we think it safer. Or do we have confident that 30s will always be enough?

templates/core/terraform/appgateway/appgateway.tf
```diff
@@ -18,83 +24,166 @@ resource "azurerm_application_gateway" "agw" {
...
+  # We don't want Terraform to revert certificate cycle changes. We assume the certificate will be renewed in keyvault.
+  lifecycle {
+    ignore_changes = [
+      ssl_certificate
+    ]
+  }
+
 }
 
 data "azurerm_public_ip" "appgwpip_data" {
```
limorl, line 186: `make all` fails to setup this one:

	Error: Public IP "pip-agw-limorltre-lldev-9808" (Resource Group "rg-limorltre-lldev-9808") was not found
	
It also expects the certificate in keyvault, yet according to the (offline) guidance, male letsencrypt should be run afterwards. It's a confusing dev experience IMO.

christoferlof: @limorl , please see #94 which captures above error

Makefile
```diff
@@ -41,12 +41,18 @@ tre-deploy:
 	&& . ./devops/scripts/load_env.sh ./templates/core/terraform/.env \
 	&& cd ./templates/core/terraform/ && ./deploy.sh 
 
+letsencrypt:
```
christoferlof, line 44:
```suggestion
.PHONY: letsencrypt
letsencrypt:
```

christoferlof (changes requested):
Need to rebase from `develop` and add step to quickstart.md, thanks.

askew:
> @askew would be good to document the commands in the develop quickstart - It's also one of the success criteria :)

I've updated [quickstart.md](https://github.com/microsoft/AzureTRE/blob/develop/docs/quickstart.md) with instructions for running `make letsencrypt`.

devops/scripts/letsencrypt.sh
askew: The [default probe](https://docs.microsoft.com/azure/application-gateway/application-gateway-probe-overview#default-health-probe) timing is 30sec and it only needs a single good response, so 30sec wait seems okay.

templates/core/terraform/appgateway/appgateway.tf
askew: I'll mark this as resolved here, as this PR doesn't change the IP definition. We should try to resolve this error as as fix for #94.

Makefile
askew: Merged with `develop`, updated [docs/quickstart.md](https://github.com/microsoft/AzureTRE/blob/develop/docs/quickstart.md), and added `letsencrypt` to the `.PHONY` list at the start of the __makefile__.

docs/quickstart.md
```diff
@@ -91,9 +91,18 @@ Copy [/templates/core/terraform/.env.sample](../templates/core/terraform/.env.sa
...
 
-The deployment of the Azure TRE is done via Terraform.
+The deployment of the Azure TRE is done via Terraform. Run:
+```
+make tre-deploy
+```
 
-- Run `make tre-deploy`
+The Azure TRE is initially deployed with an invalid self-signed SSL certificate. This certificate is stored in the deployed KeyVault. To update 
+the certificate in KeyVault needs to be repaced with one valid for the configured domain name. To use a certificate from [Let's Encrypt][letsencrypt],
```
christoferlof, line 100:
```suggestion
the certificate in KeyVault needs to be repaced with one valid for the configured domain name. To use a certificate from [Let's Encrypt](https://letsencrypt.org/), 
```

christoferlof (approved):
Link to Let's Encrypt website is broken, otherwise looks good to me 👍

Issue #111, opened 2021-05-26
Bootstrap the Cosmos DB


TessFerrandez:
- [x] Cosmos bootstrapped with required container(s) for Workspaces
- [x] Cosmos bootstrapped with required container(s) for Schemas
- [x] Schemas added to the database

PR #125, opened 2021-05-28
Bootstrap the Cosmos DB #111

sachinkundu:
# PR for issue 111

## What is being addressed

This patch adds a method for making bundle specs in the state store during startup.
It creates the container and items once and then checks if they exist and exits if they do.

Eventually we should add an API which makes it possible to register bundle specs during authoring experience

Tests are not added for temporary functionality. 
Should all BundleSpec Repository when needed(another patch)


management_api_app/db/events.py
```diff
@@ -10,14 +11,56 @@ async def connect_to_db(app: FastAPI) -> None:
...
 
     try:
-        cosmos_client = CosmosClient(config.STATE_STORE_ENDPOINT, config.STATE_STORE_KEY)
+        cosmos_client = CosmosClient(config.STATE_STORE_ENDPOINT, config.STATE_STORE_KEY, connection_verify=False)
         app.state.cosmos_client = cosmos_client
         logging.debug("Connection established")
     except Exception as e:
         logging.debug(f"Connection to state store could not be established: {e}")
 
+# Bootstrapping is temporary while the API does not have a register spec api implemented.
```
TessFerrandez, line 20: If this is only for adding samples for testing purposes - there is a script in the scripts folder that we can use for this.

IMHO we should not add samples as part of startup. 

We should however bootstrap the DB on startup with the base bundle specs, but from disc.

management_api_app/db/events.py
```diff
@@ -10,14 +11,56 @@ async def connect_to_db(app: FastAPI) -> None:
...
+
+async def create_bundle_registry(database: DatabaseProxy):
+    """
+    Creates a bundle spec container if one does not exist and populate it with a canonical spec.
+    :param database: DatabaseProxy for STATE_STORE_DATABASE
+    :returns: None
+    """
+    bundle_spec = {
+        "url": "blah/blah:v1",
+        "id": str(uuid.uuid4()),
```
TessFerrandez, line 31: Haven’t had a chance to run this yet. But… will this not add a new sample each time you start the API (as the UUID will make it a new record)

TessFerrandez: This would also be solved by loading from disc as the sample would then have a predetermined UUID

TessFerrandez (changes requested):

management_api_app/db/events.py
sachinkundu: Umm it's not testing purpose. This is a fixed entry till we have the authoring experience designed and done.

management_api_app/db/events.py
sachinkundu: no before creation it checks if the container exists and if it does it skips. I did check it out and it works.

management_api_app/db/events.py
sachinkundu: Could load it from disk but this is clear what is being loaded instead of reading it from another place. Could move the json(dict) to another file if that makes it more elegant. Writing a json file to read into a dict is not needed for such a small case. If the dict goes larger then yes could put it into a file but don't think that will happen.

TessFerrandez: So are you saying that this fixed entry will be here later too, but look different? or is it only here now, to be able to develop and test the other code? or is this the long term bootstrapping of the base bundle?

sachinkundu: While the authoring experience is undecided we can use this fixed entry to code POST /api/workspaces and use this fixed entry for bundle_url(let's change it when a real bundle exists on ACR). This will get removed as soon as we add another API POST /api/bundlespecs and make a wrapper on top of porter publish to also add bundle.json content to the store using the API.

deniscep: Hey, my 5c: I would also split this out into a sample provisioning, just to keep this clean of any sample code. So essentially when mare runs push_bundle, we can fire off a sample python that adds these.

management_api_app/db/events.py
```diff
@@ -10,14 +11,56 @@ async def connect_to_db(app: FastAPI) -> None:
...
+# Bootstrapping is temporary while the API does not have a register spec api implemented.
+
+
+async def create_bundle_registry(database: DatabaseProxy):
+    """
+    Creates a bundle spec container if one does not exist and populate it with a canonical spec.
+    :param database: DatabaseProxy for STATE_STORE_DATABASE
+    :returns: None
+    """
+    bundle_spec = {
```
TessFerrandez, line 29: Discussing with @Lybecker @jjcollinge @askew and they brought up that we want to decouple the API from the backend tech we use - so perhaps resource_spec and /ResourceSpecs would be an name that allows us to decouple, and then re-thinking the url - we could maybe make the URL the id - or similar

sachinkundu: resource spec is a good idea. Best to deep the domain model separate. id I am not so sure about. Think about explaining id to someone. Well the id is the url where the resource spec exists and then so why not use url? It's like name for docker container one specifies the full docker path to it including the registry name so it is a full url

TessFerrandez (approved):
Approving with the changes suggested in our daily

- Change ResourceSpecs to BundleSpecs
- Read from disc (optional)

PR #126, opened 2021-05-28
Minor fix to type in documentation that was missed as part of #103

askew:
Follow-up PR for #66 to fix broken link in documentation that was missed as part of PR #103


Lybecker (approved):
:)

PR #127, opened 2021-05-28
Improvements to Dev Container

askew:
This PR just contains a few changes to the Dev Container configuration that I found helpful working with it.

 * The container now mounts the ~/.azure directory, so that it uses the same Azure CLI profile that is configured on the host. (means you don't have to keep doing an `az login --use-device-code`.
 * The container mounts the bash history file from the host, so command history is preserved and specific to the project.
 * The Dev container now runs as the __vscode__ user, not __root__.
   - When using WSL this mean that new files are not created as root, so you still have access when outside the container.
   - It's also good practice not to develop as root.
  * The `pip install requirements.txt` is moved to the end of the Docker file so that changes to __requirements.txt__ doesn't cause all the `apt install`s to run again.
  * The running container is named `tre`. This is a bit of a personal preference and does have a trade-off.
    - The benefit is it make it easy to set up a profile in Window Terminal to run `docker exec -it tre bash` so you can have dev container shell outside VS Code.
    - The slight downside is the fixed name means you can only have one instance of the dev container running.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -9,15 +9,11 @@ ARG INSTALL_NODE="true"
 ARG NODE_VERSION="lts/*"
 RUN if [ "${INSTALL_NODE}" = "true" ]; then su vscode -c "umask 0002 && . /usr/local/share/nvm/nvm.sh && nvm install ${NODE_VERSION} 2>&1"; fi
 
-# Install requirements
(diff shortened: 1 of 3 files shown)


Lybecker (approved):
Looks good to me. Nice enhancements 👍

TessFerrandez (approved):

Issue #69, opened 2021-05-11
Trigger the Workspace Bundle when Workspace is created through the API


Lybecker:
## Description

As a Researcher/TRE admin, 
I want the notify the Workspace Bundle executor when changing a Workspace or Workspace Service, 
so the Workspace Bundle can perform the requested changes

When a change is requested through the Management API, a command (message) with information about the change is added to a queue. The message is picked up by Resource Processor and passed to the Workspace Bundle.

It should be implemented as part of the API in #62.

The command (message) should contain the Workspace schema information.

## Acceptance criteria

- [x] Message successfully placed on the queue

christoferlof:
Removing the requirement of sb sessions for now as we only see this important when handling concurrent service deployments in the future.
Closing.

PR #134, opened 2021-05-31
[POST] /workspaces

### LINKED ISSUE #112: Implement [POST] /workspaces
As a AzureTRE Admin, I want to create a new request for a workspace, to have a starting point for the workspace creation workflow.

**Acceptance criteria:**

- [x] Workspace resource created in cosmos DB

TessFerrandez:
# PR for issue #112 POST /workspaces

## What is being addressed

Create a new workspace with the following input
    - resourceSpec { name, version } 
    - parameters

Should add the workspace to the cosmos DB

## Try it locally

From `management_api_app` run `uvicorn main:app` and navigate to `/docs`

Remember that you have to have an azure cosmos DB or a cosmos db emulator configured in your .env

## How is this addressed

- Added POST endpoint for /workspaces - 202 if post succeeds
- Create new resources document
- Add error handling for database failures - throws 503 if backend is unavailable - have tested this both if the db is not available on startup or if the db goes missing, but please do your own testing too, as we don't have integration tests yet
- Add test fixture (for all tests) such that it avoids connecting to the database
- Changed the GETs so they get if not isDeleted instead of status="Deleted" as per discussion in integration meeting
- Changed description strings for the API to be cleaner and use resource strings

## Items that are not addressed or will be addressed in other stories:

- No tests are added as there is no logic that doesn't require the database to be present (we need to do a bigger take on how we should address this)
-  Sending the message to the service bus will be addressed in #69 

Closes #112

### DIFF
--- a/management_api_app/api/routes/workspaces.py
+++ b/management_api_app/api/routes/workspaces.py
@@ -4,21 +4,18 @@ from api.dependencies.workspaces import get_repository, get_workspace_by_workspa
 from db.repositories.workspaces import WorkspaceRepository
 from models.domain.resource import Resource
 from models.schemas.resource import ResourcesInList, ResourceInResponse
+from resources import strings
 
 
 router = APIRouter()
 
 
-@router.get("/workspaces", response_model=ResourcesInList, name="workspaces:get-active-workspaces")
-async def retrieve_active_workspaces(
-        workspace_repo: WorkspaceRepository = Depends(get_repository(WorkspaceRepository))
-) -> ResourcesInList:
+@router.get("/workspaces", response_model=ResourcesInList, name=strings.API_GET_ALL_WORKSPACES)
+async def retrieve_active_workspaces(workspace_repo: WorkspaceRepository = Depends(get_repository(WorkspaceRepository))) -> ResourcesInList:
     workspaces = workspace_repo.get_all_active_workspaces()
     return ResourcesInList(resources=workspaces)
 
 
-@router.get("/workspaces/{workspace_id}", response_model=ResourceInResponse, name="workspaces:get-workspace")
-async def retrieve_workspace_by_workspace_id(
-        workspace: Resource = Depends(get_workspace_by_workspace_id_from_path)
-) -> ResourceInResponse:
+@router.get("/workspaces/{workspace_id}", response_model=ResourceInResponse, name=strings.API_GET_WORKSPACE_BY_ID)
+async def retrieve_workspace_by_workspace_id(workspace: Resource = Depends(get_workspace_by_workspace_id_from_path)) -> ResourceInResponse:
     return ResourceInResponse(resource=workspace)
--- a/management_api_app/api/routes/workspaces.py
+++ b/management_api_app/api/routes/workspaces.py
@@ -1,7 +1,8 @@
-from fastapi import APIRouter, Depends
+from fastapi import APIRouter, Depends, HTTPException
 from starlette import status
 
 from api.dependencies.workspaces import get_repository, get_workspace_by_workspace_id_from_path
+from db.errors import UnableToAccessDatabase
 from db.repositories.workspaces import WorkspaceRepository
 from models.domain.resource import Resource
 from models.schemas.resource import ResourcesInList, ResourceInResponse, ResourceInCreate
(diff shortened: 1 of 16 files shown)


management_api_app/models/domain/resource.py
```diff
@@ -18,7 +17,10 @@ class ResourceType(str, Enum):
...
 class Resource(AzureTREModel):
-    id: UUID
-    description: str
+    id: str
+    resource_name: str
+    resource_version: str
+    resource_parameters: dict
     resourceType: ResourceType
     status: Status
+    isDeleted: bool = False
```
tompaana, line 26: Shouldn't this be `is_deleted` to match the convention?

tompaana: Ok. I'm fine with either way as long as *all* the variable names follow the same convention. `resource_name` and `resource_version` vs `resourceType` and `isDeleted`.

TessFerrandez: Updated so that it is consistently using camelCase (per teams discussion)

tompaana (approved):

management_api_app/api/routes/workspaces.py
```diff
@@ -1,24 +1,35 @@
...
 router = APIRouter()
 
 
-@router.get("/workspaces", response_model=ResourcesInList, name="workspaces:get-active-workspaces")
-async def retrieve_active_workspaces(
-        workspace_repo: WorkspaceRepository = Depends(get_repository(WorkspaceRepository))
-) -> ResourcesInList:
-    workspaces = workspace_repo.get_all_active_workspaces()
-    return ResourcesInList(resources=workspaces)
+@router.get("/workspaces", response_model=ResourcesInList, name=strings.API_GET_ALL_WORKSPACES)
```
sachinkundu, line 15: instead of doing error checking here you should use a middleware and fail from there if state store is down. Then you can remove all error handling from here and make code simpler and remove duplication

TessFerrandez: @sachinkundu Can you provide a suggestion of how we could do this.

management_api_app/db/repositories/workspaces.py
```diff
@@ -1,43 +1,60 @@
...
-            self._container = self.database.create_container_if_not_exists(id=STATE_STORE_RESOURCES_CONTAINER, partition_key=PartitionKey(path="/appId"))
-        else:
-            self._container = None
+        super().__init__(client, STATE_STORE_RESOURCES_CONTAINER)
 
     @property
     def container(self) -> ContainerProxy:
         return self._container
 
     def get_all_active_workspaces(self) -> List[Resource]:
```
sachinkundu, line 22: potentially dangerous function - if there are lots of workspaces loading them all in memory could kill the server. IS there no pagination support?

management_api_app/db/repositories/workspaces.py
```diff
@@ -1,43 +1,60 @@
...
+                resource_name=workspace_create.name,
+                resource_version=workspace_create.version,
+                resource_parameters=workspace_create.parameters,
+                resourceType=ResourceType.Workspace,
+                status=Status.NotDeployed
+            )
+            self.container.create_item(body=workspace.dict())
+            return workspace
+        except Exception:
+            raise EntityDoesNotExist
```
sachinkundu, line 59: What entity does not exist is being raised? It's the create so error should be create failed or something like that

management_api_app/db/repositories/workspaces.py
TessFerrandez: @sachinkundu if you are talking about get_all_active_workspaces - that was done in a PR that is already merged... we can create another issue about adding pagination if needed, but would prefer keeping that out of this POST PR.

management_api_app/db/repositories/workspaces.py
TessFerrandez: That was supposed to raise UnableToAccessDatabase, will change

management_api_app/db/repositories/workspaces.py
TessFerrandez: Created issue #139 so that we can prioritize this separately

scripts/bootstrap_db_for_local_tests.py
```diff
@@ -16,18 +16,31 @@
 
 
 def create_workspace_resource(resource_id: str):
```
sachinkundu, line 18: this file if for doing manual testing?

management_api_app/db/repositories/workspaces.py
```diff
@@ -1,43 +1,60 @@
...
     @property
     def container(self) -> ContainerProxy:
         return self._container
 
     def get_all_active_workspaces(self) -> List[Resource]:
-        workspaces = []
-        if self.container:
+        try:
             query = f'SELECT * from c ' \
                     f'WHERE c.resourceType = "{strings.RESOURCE_TYPE_WORKSPACE}" ' \
```
sachinkundu, line 25: IF you are always going to use isDeleted = false you should write a query builder which decorates your query string with an AND clause instead of remembering to put it everywhere

TessFerrandez: @sachinkundu could you add a code suggestion or point to a sample

scripts/bootstrap_db_for_local_tests.py
TessFerrandez: Yes, to create a few records, if you don't want to manually sit and do posts... added it to the scripts directory so it's outside the API

management_api_app/api/routes/workspaces.py
sachinkundu: ``
@app.middleware("http")
async def add_check_statestore(request: Request, call_next):
    try:
       check_state_store()
   except:
       503
    return await call_next(request)
``

From https://fastapi.tiangolo.com/tutorial/middleware/

TessFerrandez: Looks neat - will add this...

management_api_app/db/repositories/workspaces.py
sachinkundu: def deleted_records(query):
      return query + 'AND deleted';
def basic_query():
      return blah
def default_query():
     return deleted_records(basic_query())

Then always use default query. Since the interface is fluent on string one could add other clauses to build a DSL. A suggestion for clarity and reducing mistakes but I'll leave it to you to think about it.

deniscep (approved):

management_api_app/db/repositories/workspaces.py
TessFerrandez: Created #146

management_api_app/api/routes/workspaces.py
TessFerrandez: Created #145

PR #151, opened 2021-06-01
Resource request message sender implemented and integrated to API

### LINKED ISSUE #69: Trigger the Workspace Bundle when Workspace is created through the API

tompaana:
Closes #69 - Nice!

## What is being addressed

Implements a class to send service bus (resource request) messages to a queue and integrates that into the management API.

## How is this addressed

- Adds new `ServiceBus` class with a function to send messages to the resource request queue
- Makes the create workspace API send the resource request message to the queue


TessFerrandez (approved):
LGTM!

management_api_app/service_bus/service_bus.py
```diff
@@ -0,0 +1,27 @@
+from azure.identity.aio import DefaultAzureCredential
```
sachinkundu, line 1: you should call this service_bus_sender because soon we could have service_bus_receiver

management_api_app/api/routes/workspaces.py
```diff
@@ -23,12 +24,22 @@ async def retrieve_active_workspaces(workspace_repo: WorkspaceRepository = Depen
 
 @router.post("/workspaces", status_code=status.HTTP_202_ACCEPTED, response_model=ResourceInResponse, name=strings.API_CREATE_WORKSPACE)
 async def create_workspace(workspace_create: ResourceInCreate, workspace_repo: WorkspaceRepository = Depends(get_repository(WorkspaceRepository))) -> ResourceInResponse:
```
sachinkundu, line 26: instead of creating service bus here could you not also use Depends like we have for repository?

management_api_app/api/routes/workspaces.py
```diff
@@ -23,12 +24,22 @@ async def retrieve_active_workspaces(workspace_repo: WorkspaceRepository = Depen
...
+
     try:
         workspace = workspace_repo.create_workspace(workspace_create)
-        return ResourceInResponse(resource=workspace)
     except UnableToAccessDatabase:
         raise HTTPException(status_code=status.HTTP_503_SERVICE_UNAVAILABLE, detail=strings.STATE_STORE_ENDPOINT_NOT_RESPONDING)
 
+    try:
+        service_bus = ServiceBus()
+        await service_bus.send_resource_request_message(str(workspace.dict()))
```
sachinkundu, line 36: any particular reason why a dict is converted to str before passing to another class?

management_api_app/service_bus/service_bus.py
```diff
@@ -0,0 +1,27 @@
...
+    """
+
+    async def send_resource_request_message(self, resource_request_message: str):
+        """
+        Sends the given message to the resource request queue.
+
+        :param resource_request_message: The message to send.
+        :type resource_request_message: str
+        """
+        credential = DefaultAzureCredential()
```
sachinkundu, line 19: should credentials be read in class initialize?

tompaana: Credential instance is closed in the end of the method call so that would render the ServiceBus instance nonfunctional after one-time use.

management_api_app/api/routes/workspaces.py
tompaana: The ServiceBusMessage constructor likes to ingest strings.

management_api_app/service_bus/service_bus.py
tompaana: Let's see when we get there. There still room for more methods in the class.

management_api_app/api/routes/workspaces.py
tompaana: I have an opinion about these types of dependency management features. I will not disclose it here.

In its current state I don't see any benefit using it for ServiceBus class.

management_api_app/.env.tmpl
```diff
@@ -1,9 +1,14 @@
-# Api configuration
-# ------------------------
+# API configuration
+# -----------------
 # PROJECT_NAME=""
 # DEBUG=False
 
 # State store configuration
-# ------------------------
+# -------------------------
```
Lybecker, line 7: This is an important change :)

Issue #63, opened 2021-05-10
Document the TRE logical data model


christoferlof:
## User story

As a TRE developer
I need to know the TRE logical data model - at least in its initial version - 
So that I can start implementing the API and the resource processors.

## Acceptance criteria

- [x] Schema defined for workspace, workspace service and shared service
- [x] Schema documentation created incl any supporting diagrams

PR #158, opened 2021-06-02
TRE Logical data model

msebragge:
# PR for issue #63

## What is being addressed

Document TRE logical data model.

## How is this addressed

- JSON schema files.
- Document file with diagrams.


docs/logical-data-model.md
```diff
@@ -0,0 +1,80 @@
...
+
+![Resource](./assets/logical-data-model-1.png)
+
+## Resource
+
+| Key | Value description |
+| --- | ----------------- |
+| `id` | A unique resource id. |
+| `parentId` | The parent resource id. |
+| `version` | Resource version. |
```
TessFerrandez, line 11: In the API we have a 
ResourceSpec {
   name: the porter package name... could also be a generic resource-spec name
   version: resource spec version>
}

Do we want to add this as well?

TessFerrandez, line 11: Is this the version of the resource or the resource specification? I think it is the resource, i.e. that we are versioning the resource request, but just wanted to understand.

docs/logical-data-model.md
```diff
@@ -0,0 +1,80 @@
...
+| `title` | The name (title) of the resource. Rendered in the UI. |
+| `description` | The description of the resource. Rendered in the UI. |
+| `resourceType` | Resource type. Workspace or Service. |
+| `status` | Resource status. |
+| `markedForDeletion` | Is the resource marked for deletion. |
```
TessFerrandez, line 16: we currently have an isDeleted in the API - is this the same? i.e. is this for deleting the resource request or for deleting the underlying resources?

docs/logical-data-model.md
```diff
@@ -0,0 +1,80 @@
...
+| `properties` | Resource properties list.|
```
TessFerrandez, line 17: Currently we only have name/value pairs sent in (as requested by the spec) - so in the current implementation this is a dict (name/value). But we can leave as is, this might be where we need to go eventually

TessFerrandez (commented):
I had some questions but feel free to resolve, we can take them offline -- my bigger question is, this seems to deviate a little bit from the physical data model, but maybe that is ok

docs/logical-data-model.md
msebragge: I changed to isDeleted

docs/logical-data-model.md
msebragge: Added

msebragge: Now separate fields;
- specificationVersion
- resourceVersion

christoferlof:
There are some slight differences on the resource model on what is implemented versus what's described here but it's a match on the conceptual level. Operations and resource assignments/auth is yet to be implemented and this is a good start to build from. Let's get this one in and then we continue to iterate on it.

christoferlof (approved):

Issue #165, opened 2021-06-02
[Task] Update /workspaces API to spec (input/output)


TessFerrandez:
Story: #161 

**Description**

Inputs and outputs for /GET /GET by id and /POST should be as defined by the story comment

Issue #161, opened 2021-06-02
Refine API contract based on recent demo comments


christoferlof:
## Description

As we now have the first version of the API with request and response formats for the /workspaces resource, we've identified a few areas of improvements for increased usability. 

## Acceptance criteria

- [x] Environment variables added for the API #164 
- [x] API follows spec per comment #165
- [x] Workspace ID (4 digit) is calculated #166
- [x] Resource spec uses latest version #167

## Topics to decide on

- [x] Desired request and response formats for `/workspaces` resource. 
   -  Version, friendly name, resourceSpec vs workspaceSpec etc.
- [x] Should we query by sequential (human friendly) workspace id/name or system generated resource?
- [x] Bundle parameters that are not required/expected to be provided by end-user should not be part of request?
   -  `workspace_id` should be generated and not provided by the user
   -  `core_id` is static per TRE instance

TessFerrandez:
the workspace resource should look as follows:

    {
	"id": GUID #generated by the API
	"friendlyName": "Tess workspace",	# shown in the UI
	"workspaceURL": ""	# default entry point to the workspace for the researcher (not filled in for right now)
	"status": "not_deployed" # status of the workspace deployment
	"description": "blabla" #provided by the user at request time
	"resourceSpec": {
             "name" : "bundlename", # provided by the user
             "version": "0.1.0", # latest version CALCULATED BY API
             "parameters" : {
                      "location" : "westeurope"   # for now default TRE location (provided with ENV VAR)
                      "workspace_id" : "0001"     # 4 char string - continuous number - calculated by API
                      "core_id": "mytre-dev-3142" # ENV VAR (same for all workspaces in TRE instance)
                      "address_space": "10.2.1.0/24" # hard-coded for now - calculated by API in later issue
             }
        }
        "isDeleted": false # whether the request is deleted and should be cleaned up (ex. if service bus fails) 
        "resourceType": "workspace"
    },

- GET /workspaces - returns a list of all the resources as is (if not isDeleted)
- GET /workspace/{workspace_id} - returns a single workspace resource (if not isDeleted) - workspace_id here is the GUID
- POST /workspace
    - INPUT: friendlyName, workspaceType (resourceSpec/name), description
    - OUTPUT: 202 - id (GUID)

Issue #167, opened 2021-06-02
[Task] Resource spec calculates latest version


TessFerrandez:
Story: #161

**Description**

Calculate latest version of resource spec (to fill in post request)

PR #171, opened 2021-06-02
update API interface

TessFerrandez:
# PR for issue #165 

## What is being addressed

Update the API interface for /workspaces per the description in story #161 

## How is this addressed

- Input parameters for POST simplified to friendlyName, workspaceType and description
- Output parameters for POST simplified to only the ID
- core_id, and location are accessed from environment variables
- Workspaces structure changed to match Porter - Parameters are inside the ResourceSpec and stored as a list (name/value)
- Tests updated
- Examples (that will be shown in the API/docs) - added or updated to reflect the input/output schema


TessFerrandez:
Will update with documentation before moving to PR

management_api_app/tests/test_api/test_routes/test_workspaces.py
```diff
@@ -80,20 +92,26 @@ async def test_workspaces_id_get_returns_422_if_workspace_id_is_not_a_uuid(get_w
 @patch("api.dependencies.workspaces.WorkspaceRepository.get_workspace_by_workspace_id")
 async def test_workspaces_id_get_returns_workspace_if_found(get_workspace_mock, app: FastAPI, client: AsyncClient):
     sample_workspace = {
-        "id": "afa000d3-82da-4bfc-b6e9-9a7853ef753e",
+        "id": "933ad738-7265-4b5f-9eae-a1a62928772e",
         "resourceSpec": {
-            "name": "tre-workspace-vanilla",
+            "name": "My workspace",
```
marrobi, line 97: Just to make sure I'm not getting confused - this is the bundle name? And friendly name is the workspace name?

Maybe we should have them as `bundleName` and `workspaceName` to be explicit.

marrobi:
Looking good I will build and deploy and try creating and retrieving some workspaces.

management_api_app/tests/test_api/test_routes/test_workspaces.py
TessFerrandez: Good catch - this was a copy paste from the example schema - will change both here and in the schema to tre-workspace-vanilla

management_api_app/db/repositories/workspaces.py
```diff
@@ -34,13 +35,20 @@ def get_workspace_by_workspace_id(self, workspace_id: UUID4) -> Resource:
...
+    def create_workspace(self, workspace_create: WorkspaceInCreate) -> Workspace:
+        workspace = Workspace(
             id=str(uuid.uuid4()),
-            resourceSpec=ResourceSpec(name=workspace_create.resourceSpec.name, version=workspace_create.resourceSpec.version),
-            parameters=workspace_create.parameters,
-            resourceType=ResourceType.Workspace,
-            status=Status.NotDeployed
+            friendlyName=workspace_create.friendlyName,
+            description=workspace_create.description,
+            resourceSpec=ResourceSpec(
```
sachinkundu, line 43: Hmm the actual spec should have come from the ResourceSpecs container using name and version. Is that done somewhere else or later?

TessFerrandez: can you elaborate -- the workspaceType here is coming in from the user as the resourceSpec name (e.g. tre-vanilla-...), the version will be calculated as the latest for that spec in #167

Do you mean the parameters?

It becomes a little difficult as the parameters are both parameters that are fixed - that should be provided by the API
(the user can't give neither the workspace_id, location, core_id or really the address_space.) and potentially some that are not fixed... perhaps we should rethink this and make the ones that should be provided by the API move outside the params section.

management_api_app/tests/test_api/test_routes/test_workspaces.py
sachinkundu: you should read it as ``resourceSpec_name``

management_api_app/api/routes/workspaces.py
```diff
@@ -3,23 +3,23 @@
 
 from api.dependencies.workspaces import get_repository, get_workspace_by_workspace_id_from_path
 from db.repositories.workspaces import WorkspaceRepository
-from models.domain.resource import Resource
-from models.schemas.resource import ResourcesInList, ResourceInResponse, ResourceInCreate
+from models.domain.workspace import Workspace
+from models.schemas.workspace import WorkspaceInCreate, WorkspaceIdInResponse, WorkspacesInList, WorkspaceInResponse
```
sachinkundu, line 7: I am a bit confused with the meaning of schema here. IS this request-response schemas and is it a fast api convention to call these schemas?

sachinkundu (approved):

management_api_app/db/repositories/workspaces.py
sachinkundu: I am saying that when creating a workspace you receive name and version of the bundle. We should get the spec(which was pre registered with us) from the cosmos container ResourceSpecs using the given name and version. Then we know what spec to fill and ideally we would validate the presence of input parameters using this spec(but we are not doing that right now) Spec is something to read from the database.

TessFerrandez: right... but it gets complicated, let's discuss in standup

management_api_app/api/routes/workspaces.py
TessFerrandez: yes and yes

management_api_app/db/repositories/workspaces.py
TessFerrandez: Resolving per offline discussions - moving this to a more flat structure to differentiate between ResourceSpec (the specification) and what we send as input (like in this case)

PR #179, opened 2021-06-04
ResourceSpecs have a latest field and they can be read from the repository

sachinkundu:
# PR for issue #167

## What is being addressed

Add latest to bootstrapping json for ResourceSpecs

Add repository for ResourceSpecs


management_api_app/db/repositories/resource_specs.py
```diff
@@ -0,0 +1,37 @@
...
+    @property
+    def container(self) -> ContainerProxy:
+        return self._container
+
+    def _query(self, query: str):
+        resource_specs = list(self.container.query_items(query=query, enable_cross_partition_query=True))
+        return resource_specs
+
+    def get_by_name(self, name: str, latest=False) -> List[ResourceSpec]:
+        by_name = f"SELECT * FROM ResourceSpecs rs where rs.name = {name}"
```
TessFerrandez, line 23: Can we use the query builder?

management_api_app/tests/test_db_repository/test_resource_spec_repo.py
```diff
@@ -0,0 +1,32 @@
+import db.repositories.resource_specs
+from unittest.mock import patch, MagicMock
+
+
+@patch('azure.cosmos.CosmosClient')
+def test_get_by_name_exisiting_document_returns_document(cosmos_client):
+    sut = db.repositories.resource_specs.ResourceSpecRepository(cosmos_client)
+    sut.container.query_items = MagicMock()
```
TessFerrandez, line 8: Can we use the same mocking as we have used earlier? to keep the code base consistent (see the other tests)

TessFerrandez (approved):

management_api_app/db/repositories/resource_specs.py
```diff
@@ -0,0 +1,37 @@
...
+        if latest:
+            by_name += " AND rs.latest = true"
+        return self._query(by_name)
+
+    def get_by_name_and_version(self, name: str, version: str) -> ResourceSpec:
+        by_name_and_version = f"SELECT * FROM ResourceSpecs rs where rs.name = {name} AND rs.version = {version}"
+        specs = self._query(by_name_and_version)
+        assert(len(specs) == 1)  # name and version should have been unique
```
TessFerrandez, line 31: There might not be a template with a given name/version - do we really want to assert? better throw an EntityNotFound error

TessFerrandez (approved):
Approving as I need this to unblock me - I can make the refactors myself in the next PR if needed

management_api_app/db/repositories/resource_specs.py
sachinkundu: That is true. It should be
if specs:
   assert()
  return spec[0]
else:
  return None

I am not fond of throwing EntityNotFound in repository code. That should be done at controller level. If not found then result is None from the repo's perspective but if not found in controller it should be EntityNotFound and hence 404.

What do you think?

management_api_app/models/domain/resource.py
```diff
@@ -24,6 +25,14 @@ class ResourceType(str, Enum):
     Service = strings.RESOURCE_TYPE_SERVICE
 
 
+class ResourceSpec(AzureTREModel):
```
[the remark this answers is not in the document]
sachinkundu: We could. It was just in this file to start with so...

management_api_app/models/domain/resource.py
```diff
@@ -24,6 +25,14 @@ class ResourceType(str, Enum):
...
+    name: str
+    version: str
+    id: str
+    latest: str
+    fields: List[Dict]
```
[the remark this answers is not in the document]
sachinkundu: hehe naming naming...

management_api_app/db/repositories/resource_specs.py
sachinkundu: We could but isn't it easier to read this way right now? My intention is to make it easier to read so I suggest query builder for you since you had default clause and it would have to be remembered. Here the two query are different and sure I could refactor a bit more but I thought that comes at the expense of reading the code.

management_api_app/db/repositories/resource_specs.py
TessFerrandez: but... that is literally what the EntityNotFound is for -- its not an HTTP exception - its a db error (see db.errors)

management_api_app/db/repositories/resource_specs.py
TessFerrandez: you were the one who wanted the query builder :) I had added the queries to the repo code -- the only reason I created a query builder was because you requested it :D

TessFerrandez: AFAIC I think we should remove the query builder everywhere - it is overengineering - so let's skip it in all repositories.... I'll undo my work in the next PR

PR #186, opened 2021-06-06
Workspace Create picks up latest version of requested template

TessFerrandez:
# PR for issue #167 (alternative to existing PR)

## What is being addressed

- renames resource spec to resource template
- gets the current version of the requested template when creating a new workspace

## How is this addressed

- adds workspace template repository
- renames resource spec to resource template
- gets the current version of the requested template when creating a new workspace
- adds tests for template repo, and affected routes
- if template is not available in db - throws 400 bad request

This is a multistep rocket: it would replace PR #179 and builds on #185 (so this diff also contains 185)


sachinkundu (approved):
as discussed

Issue #181, opened 2021-06-04
[Task] Implement GET /workspace-templates


TessFerrandez:
Story: #138 

**Description**

Return a list of unique workspace template names in the DB (i.e. if multiple workspace templates with the same name exist, only return the name once)

PR #187, opened 2021-06-06
Implement GET /workspace-templates

TessFerrandez:
# PR for issue #181 

## What is being addressed

implement GET /workspace-templates

## How is this addressed

- Implement API endpoint
- Implement underlying repo functions
- Implement tests

This is a multi-step rocket - this PR builds on #186 so if it is reviewed before 186 - note that it will contain the diff for 186 as well


management_api_app/api/routes/workspace_templates.py
```diff
@@ -0,0 +1,15 @@
...
+from db.repositories.workspace_templates import WorkspaceTemplateRepository
+from models.schemas.workspace_template import WorkspaceTemplateNamesInList
+from resources import strings
+
+
+router = APIRouter()
+
+
+@router.get("/workspace-templates", response_model=WorkspaceTemplateNamesInList, name=strings.API_GET_WORKSPACE_TEMPLATES)
+async def retrieve_workspace_templates(workspace_template_repo: WorkspaceTemplateRepository = Depends(get_repository(WorkspaceTemplateRepository))) -> WorkspaceTemplateNamesInList:
```
sachinkundu, line 13: nit: why not just get instead of retrieve. Easier to read and means the same thing.

management_api_app/api/routes/workspaces.py
```diff
@@ -20,13 +23,18 @@ async def retrieve_active_workspaces(workspace_repo: WorkspaceRepository = Depen
...
+    try:
+        workspace = workspace_repo.create_workspace_item(workspace_create)
+    except ValueError as e:
+        raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=str(e))
 
     try:
         service_bus = ServiceBus()
         await service_bus.send_resource_request_message(workspace)
-    except Exception:
+        workspace_repo.save_workspace(workspace)
```
sachinkundu, line 34: This is now interesting. You are creating an item and only saving it after sending a service bus message? Any reason why not saving and then sending a message?

management_api_app/db/repositories/base.py
```diff
@@ -7,15 +8,21 @@
...
 
-    def get_container(self, container_name) -> ContainerProxy:
+    def _get_container(self, container_name) -> ContainerProxy:
         try:
             database = self._client.get_database_client(config.STATE_STORE_DATABASE)
             return database.create_container_if_not_exists(id=container_name, partition_key=PartitionKey(path="/appId"))
         except Exception:
             raise UnableToAccessDatabase
+
+    def _query(self, query: str):
```
sachinkundu, line 24: _methods are not used here it seems. Are they for overriding in child class. Do they then need to be _ ?

management_api_app/db/bootstrapping_data/resource_templates.json
```diff
@@ -0,0 +1,32 @@
...
+        },
+        {
+          "name": "address_space",
+          "type": "string",
+          "default": "10.2.1.0/24",
+          "description": "VNet address space for the workspace services"
+        }
+      ],
+      "resourceType": "workspace",
+      "isCurrent": true
```
sachinkundu, line 29: personal choice so take it or leave it. I think fields should be named as "current" and the getter method for that property as isCurrent.

management_api_app/db/repositories/workspace_templates.py
```diff
@@ -0,0 +1,43 @@
...
+
+    @staticmethod
+    def _workspace_template_by_name_query(name: str) -> str:
+        return f'SELECT * FROM c WHERE c.resourceType = "workspace" AND c.name = "{name}"'
+
+    def get_workspace_templates_by_name(self, name: str) -> List[ResourceTemplate]:
+        query = self._workspace_template_by_name_query(name)
+        return self._query(query=query)
+
+    def get_current_workspace_template_by_name(self, name: str) -> dict:
```
sachinkundu, line 23: should this not return a ResourceTemplate?

management_api_app/db/repositories/workspace_templates.py
```diff
@@ -0,0 +1,43 @@
...
+        query = self._workspace_template_by_name_query(name) + ' AND c.isCurrent = true'
+        workspace_templates = self._query(query=query)
+        if len(workspace_templates) != 1:
```
sachinkundu, line 26: if len == 0 then the exception holds true. But if len > 1 then there should be a different exception. That should never happen in correctly functioning system but it might. When we receive a new bundle registration we should supercede the last entry and mark the new one as current. That could fail and we might get two current. That is an error and we should report an assert on it.

management_api_app/db/repositories/workspace_templates.py
```diff
@@ -0,0 +1,43 @@
...
+            raise EntityDoesNotExist
+        return workspace_templates[0]
+
+    def get_workspace_template_by_name_and_version(self, name: str, version: str) -> dict:
+        query = self._workspace_template_by_name_query(name) + f' AND c.version = "{version}"'
+        workspace_templates = self._query(query=query)
+        print(len(workspace_templates))
```
sachinkundu, line 33: debugging prints? could remove them.

management_api_app/db/repositories/workspace_templates.py
```diff
@@ -0,0 +1,43 @@
...
+        print(workspace_templates)
+        if len(workspace_templates) != 1:
+            raise EntityDoesNotExist
+        return workspace_templates[0]
+
+    def get_workspace_template_names(self) -> List[str]:
+        query = 'SELECT * FROM c'
```
sachinkundu, line 40: why not database handle the projection??
SELECT {"Name":f.name} FROM WorkspaceTemplates f 
isn't that better we only get the stuff we need from the DB instead of loading the whole content in memory?

management_api_app/db/repositories/workspaces.py
```diff
@@ -17,27 +17,34 @@ class WorkspaceRepository(BaseRepository):
...
 
-    @property
-    def container(self) -> ContainerProxy:
-        return self._container
+    @staticmethod
+    def _active_workspaces_query():
+        return 'SELECT * FROM c WHERE c.resourceType = "workspace" AND c.isDeleted = false'
+
+    def _get_template_version(self, template_name):
+        wt_repo = WorkspaceTemplateRepository(self._client)
```
sachinkundu, line 25: nit: wt = workspace_template

management_api_app/db/repositories/workspaces.py
```diff
@@ -17,27 +17,34 @@ class WorkspaceRepository(BaseRepository):
...
+        template = wt_repo.get_current_workspace_template_by_name(template_name)
+        return template["version"]
 
     def get_all_active_workspaces(self) -> List[Workspace]:
-        query = QueryBuilder().select_active_resources(ResourceType.Workspace).build()
-        workspaces = list(self.container.query_items(query=query, enable_cross_partition_query=True))
-        return workspaces
+        query = self._active_workspaces_query()
```
sachinkundu, line 30: unnecessary temp variable

management_api_app/db/repositories/workspaces.py
```diff
@@ -17,27 +17,34 @@ class WorkspaceRepository(BaseRepository):
...
     def get_workspace_by_workspace_id(self, workspace_id: UUID4) -> Workspace:
-        query = QueryBuilder().select_active_resources(ResourceType.Workspace).with_id(workspace_id).build()
-        workspaces = list(self.container.query_items(query=query, enable_cross_partition_query=True))
-
-        if workspaces:
-            return workspaces[0]
-        else:
+        query = self._active_workspaces_query() + f' AND c.id="{workspace_id}"'
+        workspaces = self._query(query=query)
+        if len(workspaces) != 1:
```
sachinkundu, line 36: In this case I think writing it as 
if not workspaces:
   raise exception
is more elegant to read.

sachinkundu (commented):
Loving all the tests. Great work!

management_api_app/api/routes/workspaces.py
TessFerrandez: moving this

management_api_app/db/repositories/workspaces.py
TessFerrandez: per discussion - keeping the pattern

sachinkundu (approved):
as discussed

PR #243, opened 2021-06-10
Fix CD -> change env AZURE_LOCATION to LOCATION

deniscep:
Fixes the CD pipeline; vanilla bundle builder has env variable of LOCATION, and was expecting env variable of AZURE_LOCATION.

### DIFF
--- a/workspaces/vanilla/parameters.json
+++ b/workspaces/vanilla/parameters.json
@@ -13,7 +13,7 @@
     {
       "name": "azure_location",
       "source": {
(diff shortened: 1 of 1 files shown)


marrobi:
@deniscep looks like should fix it, have we got a test run - either by adding your branch to the workflow, or locally using `act`?

deniscep:
@marrobi locally with act looks fine

marrobi (approved):
LGTM

Issue #135, opened 2021-05-31
Resource processor reports bundle execution events to state store


christoferlof:
## Description

Resource processor needs to emit and report on bundle execution progress and report back to state store. Eventually this information will be retrieved by a caller through the management API.

## Acceptance criteria

- [x] Bundle execution events captured by resource processor
- [x] Bundle execution events persisted in state store
- [x] Resource updated with relevant deployment status, e.g., *deploying*, *succeeded*, *failed*, etc.

## Tasks
- [x] Read ACI Log #256 
- [x] Write to SB #257 
- [x] Read SB from background task #254
- [x] Write to Cosmos from background task #255
- [x] Verify return channel end to end #300

christoferlof:
Need design and decision on how processor gets the events persisted in state store.

Issue #13, opened 2021-03-09
Explore deployment service and management API and configuration store options


christoferlof:
This issue is to explore how to implement three components which are interconnected namely the:
* Deployment Service - reliable and auditable deployments of workspaces and services (aka resources)
* Configuration Store - metadata and desired configuration for resources
* Management API - API surface for apps and scripts to invoke the deployment service and read (and update) the configuration store.

![image](https://user-images.githubusercontent.com/166592/110518221-237e4300-810c-11eb-828f-86e042a3d30a.png)

## High-level requirements
* Repeatable and reliable deployment runs - Ability to re-run and retry in cases of intermittent errors etc.
* Auditable deployment runs - Need to know who initiated a deployment run when.
* Declarative resource definitions
* Open for extensibility - Allow others to extend with additional service types etc.
* Management API needs to retrieve configuration data to serve requests from callers - i.e *List services in workspace 'a'*
* Handle operations across several control planes such as identity provider, resource providers, resource data plane.
* Clean up and removal - i.e *delete a workspace*
* Deployment services are managed by an Azure operations generalist - monitor health and troubleshoot deployments.
* Also see #12 

## Edge cases we need to cater for
* Simultaneous updates of a shared resource - Example: Two workspace deployment runs are trying to update the same shared Firewall with their required rules
* Hide eventual consistency of the system to the users - show the actual resource state and not the desired state. Example: resource being updated or deleted.

## Implementation
- [ ] Review past implementations
- [ ] Create spike

## Ideas
Given above requirements, Terraform and GitHub actions would probably take us a long way. Let's discuss.

jjcollinge:
A few other requirements:
- Must support a mechanism to expose the deployment outputs back to the management API.
- Must support a mechanism to handle secrets securely i.e. return secret references rather than the secret in plaintext.
I'll probably add a few more as I think about them.

TessFerrandez:
Decision was made that at a high level we will re-use a lot of the design from an internal engagement by the Clarke team - but remove dependencies on CAF and moving to python where possible

PR #244, opened 2021-06-10
Add a sub doc in Resources for deployments

sachinkundu:
# PR for issue #135

Preparing for Resource processor reports bundle execution events to state store ---> Bundle execution events persisted in state store

This patch adds a deployment sub document which tracks the status of deployment and a possible error message which is received from the resource processor.

All tests pass

### DIFF
--- a/management_api_app/db/repositories/workspaces.py
+++ b/management_api_app/db/repositories/workspaces.py
@@ -5,12 +5,13 @@ from azure.cosmos import CosmosClient
 from pydantic import UUID4
 
 from core import config
+from resources import strings
 from db.errors import EntityDoesNotExist
-from db.repositories.base import BaseRepository
-from db.repositories.workspace_templates import WorkspaceTemplateRepository
-from models.domain.resource import Status
 from models.domain.workspace import Workspace
+from db.repositories.base import BaseRepository
+from models.domain.resource import Deployment, Status
 from models.schemas.workspace import WorkspaceInCreate
+from db.repositories.workspace_templates import WorkspaceTemplateRepository
 
 
 class WorkspaceRepository(BaseRepository):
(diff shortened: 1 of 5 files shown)


marrobi:
Question: When workspace/service update is implemented a resource can have multiple deployments queued/in progress - does that mean we need multiple deployment documents for a resource? A separate deployments collection?

Take the scenario user 1 sets `no_of_vms` to 5, then while that is deploying, user 2 sets `no_of_vms` to 10?

sachinkundu:
Might have to extend 

> Question: When workspace/service update is implemented a resource can have multiple deployments queued/in progress - does that mean we need multiple deployment documents for a resource? A separate deployments collection?
> 
> Take the scenario user 1 sets `no_of_vms` to 5, then while that is deploying, user 2 sets `no_of_vms` to 10?

Might have to extend the model with the upcoming stories - that is correct. It might be we need a list of deployment operations done on a resource. For now though I think we should go with the requirements we already know.

marrobi:
I believe we should plan for it, as will be required, is referenced on internal docs if want me to share, but also here: https://github.com/microsoft/AzureTRE/issues/13#issuecomment-795315816

sachinkundu:
> I believe we should plan for it, as will be required, is referenced on internal docs if want me to share, but also here: [#13 (comment)](https://github.com/microsoft/AzureTRE/issues/13#issuecomment-795315816)

Plan is - refactoring as we go along. Why design upfront? Refactoring is easier and more concrete and afaik the current choice does not prevent evolving the design.

marrobi (approved):
Approving with the understanding that this will need refactoring to handle simultaneous operations on a single resource.   cc. @christoferlof

PR #248, opened 2021-06-10
Adds service principal for deployment processor

### LINKED ISSUE #203: [Task] Add an SPN as part of automatic provisioning
Create a new SPN in a script, with Contributor rights to scope /subscription/<subid>. Store the secret and clientid in keyvault through main provisioning TF. Allow user identity (msi-) to read the secret from keyvault.

The secrets (SPN clientid and secret) will be read by function, and then passed as secure variables to Porter for deploying resources.

## Acceptance criteria

- [x] Service principal (SP) with contributor privileges created for InnerEye subscription
- [x] Service principal client ID and secret added to Key Vault in CD workflow
- [x] Managed identity can read the client ID and secret from the Key Vault
- [x] Service principal creation and related details documented

tompaana:
Closes #203 

## How is this addressed

- Documents SP creation as a preliminary manual step
- Adds a script that inserts SP's client ID and secret into the Key Vault
- Integrates the script to the deployment step in the Makefile
- Updates documentation to reflect the changes


templates/core/terraform/keyvault/keyvault.tf
```diff
@@ -1,10 +1,10 @@
...
-  resource_group_name = var.resource_group_name
-  sku_name            = "standard"
+  name                     = "kv-${var.tre_id}"
+  location                 = var.location
+  resource_group_name      = var.resource_group_name
+  sku_name                 = "standard"
   purge_protection_enabled = true
-  tenant_id           = var.tenant_id
+  tenant_id                = var.tenant_id
 }
```
deniscep, line 9: Can we add permissions for managed user identity to read KV as well.

deniscep (commented):
Looks good, one small addition (MSI KV permissions), otherwise all clear!

Makefile
```diff
@@ -51,7 +51,8 @@ tre-deploy:
 	&& . ./devops/scripts/check_dependencies.sh nodocker \
 	&& . ./devops/scripts/load_env.sh ./devops/terraform/.env \
 	&& . ./devops/scripts/load_env.sh ./templates/core/terraform/.env \
-	&& cd ./templates/core/terraform/ && ./deploy.sh
+	&& cd ./templates/core/terraform/ && ./deploy.sh \
+	&& cd ../../../ && ./devops/scripts/create_sp_for_deployer.sh
```
marrobi, line 55: Do we want a new SP creating each time we deploy? Do we need `make create_xxxx_sp` or something that is run separately?

marrobi (commented):
Quick look from mobile. Looks good, one comment would be good to address.

Makefile
marrobi: Also the Dev running make tre-deploy, might not have create SP permissions.

tompaana: ATM, no new SP is created each time, but if one exists, its secret will get updated. I could add a check in the script that if the SP exists, no action is taken.

marrobi: That would be good. 

However the scenario where dev can't create SPs - this is common on our projects (And even multiple deployments/TREs share the SP) - this needs to be catered for - so a separate step creates the SP. 

The deploy step then consumes the pre created credentials. This is the same SP that we have `ARM_xxx` variables for at present?

tompaana: How would you preserve the pre-created credentials? I had the SP creation in another step earlier, but it since there can be different processes running depending on the environment, the env variables don't always keep.

deniscep (approved):

Makefile
tompaana: This resolves as per separate discussion.

Issue #209, opened 2021-06-08
Consolidate variable formatting


marrobi:
We have a mixture of variables need defining to enable the quickstart CD setup. Some are the format `ACR_NAME`, others `TF_VAR_acr_name`. This is confusing and results in duplicate variables.

We should use the `ACR_NAME` standard, and translate to TF_VAR at the point needed within scripts.

All these variables should be clearly documented as part of the quickstart.

PR #262, opened 2021-06-14
Consolidate variables and make file

marrobi:
# PR for issue #209

## What is being addressed

The repository had lots of duplicate variables both across Porter and Terraform, and across various components such as core and workspaces. These have been consolidated.

## How is this addressed

- Remove all TF_VAR_ .env files. Create script to convert names such as `ACR_NAME` to `TF_VAR_acr_name`
- Update porter parameter files to use consistant naming
- Update all scripts to use consistent naming
- Update docs to match 
- consolidate make file to use `DIR=` rather than having many independent steps.
- 
Passes CD pipeline here: https://github.com/microsoft/AzureTRE/actions/runs/936799251


sachinkundu (approved):
TL;DR
As long as it works and it does according to the pipeline.

Issue #142, opened 2021-05-31
Validate API requests against workspaceTemplate parameters


christoferlof:
## Description

As a TRE workspace owner
I want to have the *new workspace service* request parameters validated
So that I can create new workspace services with confidence, getting quick feedback on whether my request will be accepted or not.

## Acceptance criteria

- [ ] workspaceTemplate name and version are validated
- [ ] Request parameters validated against the parameter definitions on the specified workspaceTemplate.
- [ ] In case of validation errors, HTTP 400 Bad Request is returned with a message describing which parameter(s) that failed the validation.

## Examples

Given that a workspace bundle is registered with the following resourceSpec:

```json
{
    "name": "tre-workspace-vanilla",
    "version": "0.1.0",
    "parameters" : [
    {
        "name": "location",
        "type": "string",
        "default": null,
        "applyTo": "All Actions",
        "description": "Azure location (region) to deploy to",
        "required": true
      },
      {
        "name": "address_space",
        "type": "string",
        "default": "10.2.1.0/24",
        "applyTo": "All Actions",
        "description": "VNet address space for the workspace services",
        "required": false
      }
    ]
}
```
When a request is made to `/workspaces` to create a new workspace with the below request body:

```json
{
    "resourceSpec": {
       "name": "tre-workspace-vanilla",
       "version": "0.1.0"
    },
    "parameters": {
       "address_space": "10.3.1.0/24"
    },
    "name": "my first workspace"
}
```

Then a HTTP 400 Bad Request is returned with a message saying that the parameter "location" is missing or incorrect:

```http
HTTP 400 Bad Request

{
    "resourceSpec": {
       "name": "tre-workspace-vanilla",
       "version": "0.1.0"
    },
    "validationErrors": [
        "Parameter name `location` is required"
    ]
}
```

christoferlof:
As part of implementing this, let's think about how to add more granular validation rules, like allowed locations etc.

TessFerrandez:
Me and @marrobi have started thinking about this a little bit - we may extract the json schema on the porter bundle and validate directly against it with pydantic (maybe) -- but we should create anther story with a deeper investigation there after this basic validation

marrobi:
@TessFerrandez I've done some research around this, and believe storing a standards compliant JSON Schema (as with a past implementation) within the Resource Template rather than a set of Porter parameters would be a good solution. 

[...]

PR #263, opened 2021-06-15
Validate API requests

msebragge:
# PR for issue #142 (Draft)

## What is being addressed

API request is validated against template.

## How is this addressed

- Need to add unit tests


management_api_app/validation/validate_request.py
```diff
@@ -0,0 +1,40 @@
...
+        if not validation_errors:
+            template_parameters = template.get(strings.VALIDATION_PARAMETERS)
+            validation_errors = ValidateRequest.validate_parameters(template_parameters, workspace_create.parameters)
+
+        if validation_errors:
+            raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST,
+                                detail={strings.VALIDATION_ERRORS: validation_errors})
+
+    @staticmethod
+    def validate_parameters(template_parameters: [dict], workspace_create_parameters: dict):
```
TessFerrandez, line 32: See my last PR - the template parameters are not a list of dict, they are a List[Parameter]

management_api_app/validation/validate_request.py
```diff
@@ -0,0 +1,40 @@
...
+        validation_errors = []
+        for parameter in template_parameters:
+            name = parameter.get(strings.VALIDATION_NAME)
+            create_parameter = workspace_create_parameters.get(name)
+            if create_parameter is None and parameter.get(strings.VALIDATION_REQUIRED):
```
TessFerrandez, line 37: Since these are Parameter items - we can use the parameter.required instead and parameter.name to validate.
Also since workspace_create_parameters are a dict we can get it with workspace_create_parameters[name]

management_api_app/validation/validate_request.py
```diff
@@ -0,0 +1,40 @@
...
+                validation_errors.append(strings.PARAMETER_X_IS_REQUIRED.format(name))
```
TessFerrandez, line 38: Please rewrite and validate that this works (I am just writing a suggestion from memory here)
```suggestion
        for parameter in template_parameters:
            name = parameter.name
            create_parameter = workspace_create_parameters[name]
            if parameter.required and not create_parameter:
                validation_errors.append(strings.PARAMETER_X_IS_REQUIRED.format(name))
```

management_api_app/validation/validate_request.py
```diff
@@ -0,0 +1,40 @@
+from fastapi import HTTPException
+from starlette import status
+
+from db.errors import EntityDoesNotExist
+from db.repositories.workspaces import WorkspaceTemplateRepository
+from models.schemas.workspace import WorkspaceInCreate
+from resources import strings
+
+
+class ValidateRequest:
```
TessFerrandez, line 10: Does this need to be a class? seems like it could just be a single function

management_api_app/validation/validate_request.py
```diff
@@ -0,0 +1,40 @@
...
+    """
+    Implements methods to validate API requests.
+    """
+
+    @staticmethod
+    def validate_request(workspace_create: WorkspaceInCreate, workspace_template_repo: WorkspaceTemplateRepository):
```
TessFerrandez, line 16:
```suggestion
    def validate_workspace_request(workspace_create: WorkspaceInCreate, workspace_template_repo: WorkspaceTemplateRepository):
```

TessFerrandez (changes requested):

TessFerrandez (changes requested):
This might only be intended as partial implementation - but it’s missing checks to see that all given parameters are valid parameters. 

note also that some required may have default values - but we should create a separate issue to add default values if not given.

And while we cant have extended validation yet we should have basic type validation (string/int)

management_api_app/api/routes/workspaces.py
```diff
@@ -22,7 +23,12 @@ async def retrieve_active_workspaces(workspace_repo: WorkspaceRepository = Depen
 
 
 @router.post("/workspaces", status_code=status.HTTP_202_ACCEPTED, response_model=WorkspaceIdInResponse, name=strings.API_CREATE_WORKSPACE)
-async def create_workspace(workspace_create: WorkspaceInCreate, workspace_repo: WorkspaceRepository = Depends(get_repository(WorkspaceRepository))) -> WorkspaceIdInResponse:
+async def create_workspace(workspace_create: WorkspaceInCreate,
+                           workspace_repo: WorkspaceRepository = Depends(get_repository(WorkspaceRepository)),
+                           workspace_template_repo: WorkspaceTemplateRepository = Depends(get_repository(WorkspaceTemplateRepository))) -> WorkspaceIdInResponse:
+
+    ValidateRequest.validate_request(workspace_create, workspace_template_repo)
```
TessFerrandez, line 30: validation is done outside try/catch and also requires extra db calls - this would be better placed inside the create_workspace_item code so that the template can be reused.

management_api_app/validation/validate_request.py
```diff
@@ -0,0 +1,40 @@
...
+            template = workspace_template_repo.get_current_workspace_template_by_name(workspace_create.workspaceType)
+        except EntityDoesNotExist:
+            validation_errors.append(strings.WORKSPACE_TEMPLATE_X_DOES_NOT_EXIST.format(workspace_create.workspaceType))
+
+        if not validation_errors:
+            template_parameters = template.get(strings.VALIDATION_PARAMETERS)
+            validation_errors = ValidateRequest.validate_parameters(template_parameters, workspace_create.parameters)
+
+        if validation_errors:
+            raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST,
```
TessFerrandez, line 28: suggest this should throw a ValidationError to be caught in the route so the route can return the appropriate HTTPException

TessFerrandez (changes requested):

management_api_app/validation/validate_request.py
msebragge: Changed

management_api_app/api/routes/workspaces.py
TessFerrandez, line 30: The way this is written it will require the user to provide the parameters that are auto-populated by the API

TessFerrandez (changes requested):

management_api_app/api/routes/workspaces.py
```diff
@@ -2,14 +2,17 @@
 
 from fastapi import APIRouter, Depends, HTTPException
 from starlette import status
+from pydantic import ValidationError
 
 from api.dependencies.database import get_repository
 from api.dependencies.workspaces import get_workspace_by_workspace_id_from_path
-from db.repositories.workspaces import WorkspaceRepository
+from api.errors.validation_error import ValidationError
```
TessFerrandez, line 9: suggestion: import validation error straight from pydantic

TessFerrandez (changes requested):

msebragge:
Obsolete

PR #264, opened 2021-06-15
Adds ACI status update writing to SB

deniscep:
# Adds ACI status update writing to SB

## What is being addressed

- Read ACI log
- Write deployment events to SB

## How is this addressed

- Add return channel queue to service bus
- Read ACI log from deployed container, wait for either error or success and write to Service Bus


processor_function/requirements.txt
```diff
@@ -1,4 +1,6 @@
 azure-functions
 azure-mgmt-containerinstance
 azure-mgmt-network
-msrest
\ No newline at end of file
+msrest
+azure-identity==1.6.0
```
tompaana, line 5: Nit: Alphabetical order: identity becomes before mgmt

processor_function/shared/cnab_builder.py
```diff
@@ -113,9 +116,31 @@ def deploy_aci(self):
...
+        service_bus.send_status_update_message(self._id, "ACI container deployed", "")
+
+        logs = aci_client.container.list_logs(self._resource_group_name, self._container_group_name, self._container_group_name)
+
+        while "Error" not in logs.content and "Success" not in logs.content:
+            time.sleep(5)
+            logs = aci_client.container.list_logs(self._resource_group_name, self._container_group_name, self._container_group_name)
+            if "Error" in logs.content:
+                service_bus.send_status_update_message(self._id, "Deployment failed", logs.content)
+                print(logs.content.split("Error", 1)[1])
```
tompaana, line 139: Now that #272 has been merged, these print statements should be replaced with logging.<log level>(). This comment applies to all print statements.

templates/core/terraform/processor_function/function.tf
```diff
@@ -29,6 +29,7 @@ resource "azurerm_function_app" "procesorfunction" {
     CNAB_AZURE_REGISTRY_PASSWORD          = var.docker_registry_password
     REGISTRY_SERVER                       = var.docker_registry_server
     servicebusconnection                  = var.servicebus_connection_string
-    queueName                             = var.workspacequeue
+    SERVICE_BUS_RESOURCE_REQUEST_QUEUE    = var.workspacequeue
```
tompaana, line 32: Nit: Convention broken

templates/core/terraform/processor_function/variables.tf
```diff
@@ -12,4 +12,5 @@ variable "docker_registry_username" {}
 variable "docker_registry_password" {}
 variable "docker_registry_server" {}
 variable "servicebus_connection_string" {}
-variable "workspacequeue" {}
\ No newline at end of file
+variable "workspacequeue" {}
+variable "service_bus_deployment_status_update_queue" {}
```
tompaana: Nit: Unify naming (prefix): `servicebus` vs nothing vs `service_bus`. My suggestion is to prefix all of these with `service_bus`. Need to be careful to update the naming change everywhere.

tompaana (changes requested):
Good work. I have few, mainly convention related comments that I'd like to be addressed.

processor_function/shared/cnab_builder.py
```diff
@@ -113,9 +116,31 @@ def deploy_aci(self):
 
         credential = AzureIdentityCredentialAdapter()
         aci_client = ContainerInstanceManagementClient(credential, self._subscription_id)
+
+        service_bus = ServiceBus()
+        service_bus.send_status_update_message(self._id, "ACI Deployment started", "Deploying ACI container: " + self._container_group_name)
```
sachinkundu, line 121: I think we should make a Status Enum(which can be a copy of Staus Enum in API)

sachinkundu: Also strings should be put into resources and used with a well defined NAME_LIKE_THIS. Right now it is very hard to read.

processor_function/shared/cnab_builder.py
```diff
@@ -113,9 +116,31 @@ def deploy_aci(self):
...
+
         result = aci_client.container_groups.create_or_update(self._resource_group_name, self._container_group_name,
                                                               group)
 
         while result.done() is False:
```
sachinkundu, line 126: consider writing while not result.done() which is more pythonic

processor_function/shared/cnab_builder.py
```diff
@@ -113,9 +116,31 @@ def deploy_aci(self):
...
             logging.info('-- Deploying -- ' + self._container_group_name + " to " + self._resource_group_name)
             time.sleep(1)
+
+        service_bus.send_status_update_message(self._id, "ACI container deployed", "")
```
sachinkundu, line 130: no message?

processor_function/shared/cnab_builder.py
```diff
@@ -113,9 +116,31 @@ def deploy_aci(self):
...
+
+        logs = aci_client.container.list_logs(self._resource_group_name, self._container_group_name, self._container_group_name)
```
sachinkundu, line 132: Is it not possible to have streaming logs generator expression from the SDK? This will load all the logs and that might lead to buffer problems. Could crash the server too.

processor_function/shared/cnab_builder.py
```diff
@@ -113,9 +116,31 @@ def deploy_aci(self):
...
+
+        while "Error" not in logs.content and "Success" not in logs.content:
```
sachinkundu, line 134: consider extracting the expression into a function like still_working() and then it reads like while still_working()

processor_function/shared/cnab_builder.py
```diff
@@ -113,9 +116,31 @@ def deploy_aci(self):
...
+            time.sleep(5)
+            logs = aci_client.container.list_logs(self._resource_group_name, self._container_group_name, self._container_group_name)
+            if "Error" in logs.content:
+                service_bus.send_status_update_message(self._id, "Deployment failed", logs.content)
```
sachinkundu, line 138: we probably dont want to send all the contents on the service bus

processor_function/shared/service_bus.py
```diff
@@ -0,0 +1,28 @@
...
+        Sends a resource request message for the resource to the service bus
+        :param resource: resource to deploy
+        """
+        resource_request_message = json.dumps({
+            "id": id,
+            "status": status,
+            "message": message
+        })
+
+        service_bus_client = ServiceBusClient.from_connection_string(conn_str=os.environ["servicebusconnection"])
```
sachinkundu, line 22: the processor function does not use MSI to talk to service bus like the API does??

sachinkundu (commented):
I don't see the ENV var for the queue being exposed to the web app.
How does the processor recover from failures? If the processor crashes then can it still recover and send the messages back to the API for deployments that are going on?

processor_function/requirements.txt
deniscep: fixed

processor_function/shared/cnab_builder.py
deniscep: AFAIK, I don't see a way to get the streaming logs from the SDK

processor_function/shared/cnab_builder.py
deniscep: Well, it's complicated :) How much should we send? The log is just a dump from ACI command line, which might contain a dump from TF failure. So essentially ~20 lines up from last line might be error information. Should we send just the last line to the cosmos, and the rest to LogAnalytics?

processor_function/shared/service_bus.py
deniscep: No, it's using the same full connection string as with trigger.

processor_function/shared/cnab_builder.py
deniscep: At this point there is no message as the CNAB Azure ACI has just kicked off at this point.

processor_function/shared/cnab_builder.py
sachinkundu: We should send the reason for failure. This message is essentially shown to the end user when they query a workspace. We don't want TF and porter errors in their full glory being shown to the API user.

deniscep: Let's spin this into another stand-alone task (Refine error message reporting from Porter); we don't know the exact reason for failure at this point.

templates/core/terraform/processor_function/function.tf
deniscep: fixed

templates/core/terraform/processor_function/variables.tf
deniscep: unified all names to common pattern

processor_function/shared/cnab_builder.py
deniscep: changed

processor_function/shared/cnab_builder.py
deniscep: Added strings; let's split enum into own task -> don't want to copy-paste code from API, and requires alignment on what values should actually be there (how granular the status should be reported)

processor_function/shared/cnab_builder.py
deniscep: fixed

processor_function/shared/cnab_builder.py
deniscep: fixed

tompaana (approved):
Grrreat!

PR #272, opened 2021-06-15
Composition service monitoring enabled

### LINKED ISSUE #174: Composition service logs and events forwarded to App Insights
## Description

As a cloud administrator
I want all logs and custom events emitted by the AzTRE composition services to be sent to application insights
So that I have one place to query for information in case of failures.

## Acceptance criteria

- [x] Application Insights configured for both the API and the processor function
- [x] [Opencensus Python SDK included in API app and Function apps to send logs](https://docs.microsoft.com/en-us/azure/azure-monitor/app/opencensus-python#introducing-opencensus-python-sdk)
- [x] [Logger configured to export logs to Application Insights](https://docs.microsoft.com/en-us/azure/azure-monitor/app/opencensus-python#instrument-with-opencensus-python-sdk-with-azure-monitor-exporters)
- [x] Log entries and events can be correlated from API through Functions in a query to get the end-to-end log stream of a *create workspace request* -- This will be implemented in #271

tompaana:
Implements 3/4 of the acceptance criteria of #174 

## How is this addressed

- Configures Application Insights (instrumentation key) for the management API and the processor function
- Implements logging initialization with App Insights for the management API and the processor function


Lybecker (approved):

management_api_app/main.py
```diff
@@ -27,10 +29,30 @@ def get_application() -> FastAPI:
...
 
 
+def initialize_logging(logging_level: int):
+    """
+    Adds the Application Insights handler for the root logger and sets the given logging level.
+
+    :param logging_level: The logging level to set e.g., logging.WARNING.
+    """
+    logger = logging.getLogger()
+    app_insights_instrumentation_key = os.getenv("APPINSIGHTS_INSTRUMENTATIONKEY")
```
TessFerrandez, line 39: can we use the config instead - that picks up the env variables but then we have them all represented in one place

management_api_app/main.py
```diff
@@ -27,10 +29,30 @@ def get_application() -> FastAPI:
...
+    try:
+        logger.addHandler(AzureLogHandler(connection_string=f"InstrumentationKey={app_insights_instrumentation_key}"))
+    except ValueError:
+        logger.error("Application Insights instrumentation key missing")
+
+    logging.basicConfig(level=logging_level)
+    logger.setLevel(logging_level)
+
+
 if DEBUG:
```
TessFerrandez, line 50: I know this is not part of the PR - but... since you're here - there is also a config.DEBUG that we could use instead of a random DEBUG flag, if you would be so kind and use that here.

management_api_app/main.py
tompaana: Fixed

Issue #213, opened 2021-06-08
[Task] Add more descriptive error messages if workspaces/workspace templates can't be created/retrieved 


TessFerrandez:
Add more descriptive error messages if workspaces/workspace templates can't be created/retrieved 

They mostly return state_store_unavailable, would be great to have more specific errors

- [x] Add try/catch on all API calls that access the DB
- [x] Add descriptive/actionable error messages (but don't reveal secrets in API returns - or TMI)
- [x] Error messages should be defined in resources/strings

PR #273, opened 2021-06-15
Ensure the repository functions return the right types

TessFerrandez:
## What is being addressed

Currently the repository functions return dictionaries when they say they return Workspace or ResourceTemplate objects. This PR casts them to the expected return type so that we get validation exceptions at the repository level if the data coming back from the db is malformed. (So that those errors are catchable later)

## How is this addressed

- Cast return types in the repository classes appropriately
- Update related tests

This is a prerequisite for #213


msebragge (approved):

tompaana (approved):
LGTM AFAICS

PR #282, opened 2021-06-16
[Obsolete] Validate API request [See #323]

msebragge:
# PR for issue #142

Closing #142

## What is being addressed

Workspace create API request is validated against template.

## How is this addressed

New functionality added to WorkspaceRepository.


management_api_app/db/repositories/workspaces.py
```diff
@@ -22,10 +23,55 @@ def __init__(self, client: CosmosClient):
...
+            if parameter not in template_parameters_by_name:
+                extra_parameters.append(parameter)
+            else:
+                template_parameter = template_parameters_by_name[parameter]
+                supplied_parameter_type = WorkspaceRepository._convert_type_name(type(supplied_request_parameters[parameter]).__name__)
+                if supplied_parameter_type != template_parameter.type:
+                    wrong_type.append({"parameter": parameter, "expected_type": template_parameter.type, "supplied_type": supplied_parameter_type})
+
+        if missing_required:
+            errors['Missing required parameters'] = missing_required
```
TessFerrandez, line 63: can we use resource strings here?

TessFerrandez (approved):
LGTM - just small comment about resource strings

management_api_app/tests/test_db/test_repositories/test_workpaces_repository.py
```diff
@@ -100,3 +110,56 @@ def test_save_workspace_saves_the_items_to_the_database(cosmos_client_mock):
     workspace_repo.save_workspace(workspace)
 
     workspace_repo.container.create_item.assert_called_once_with(body=workspace)
+
+
+@patch('azure.cosmos.CosmosClient')
```
TessFerrandez, line 115: Nice tests 👍

TessFerrandez (approved):

management_api_app/db/repositories/workspaces.py
```diff
@@ -22,10 +23,55 @@ def __init__(self, client: CosmosClient):
...
+        if s == 'int':
+            return 'integer'
+        return 'invalid'
+
+    @staticmethod
+    def _validate_workspace_parameters(template_parameters: List[Parameter], supplied_request_parameters: dict):
+        # verify all required parameters are given
+        errors = {}
+        missing_required = []
+        required_parameters = [parameter for parameter in template_parameters if parameter.required]
```
TessFerrandez, line 44: The templates contain some parameters that are provided down the line by the deployment service - we should ignore those -- something like this could be a solution (please test before committing)

If we don't exclude these - the user will need to provide them which we don't want

```suggestion
        system_provided = ["acr_name", "porter_driver", "tfstate_container_name", "tfstate_resource_group_name", "tfstate_storage_account_name"]
        required_parameters = [parameter for parameter in template_parameters if parameter.required and parameter.name not in system_provided]
```

TessFerrandez (changes requested):
Requesting a change for the system provided parameters so we don't end up in a situation where we force the user to provide these.

management_api_app/tests/test_db/test_repositories/test_workpaces_repository.py
msebragge: Thanks!

management_api_app/db/repositories/workspaces.py
msebragge: strings moved to resources/strings.py

msebragge:
See PR #323

Issue #255, opened 2021-06-14
[Task] Store deployment status from Service Bus to CosmosDB in background task


deniscep:
Read deploymentstatus queue and update cosmosDB

PR #296, opened 2021-06-17
Store the status of deployment received from service bus into state store

sachinkundu:
# PR for issue #255 

This patch now saves the received message into the state store.

An async generator pulls the message from the service bus and yields them to the function which write them to the state store. Unfortunately the cosmos db sdk for python is not async so this blocks!

Edge cases are handled and if the message processing is not successful then it will be retried again when it comes back(10 times in all). If we receive a bad message, error is logged and the message is removed from queue as it will never pass anyways.

Tests written.


management_api_app/service_bus/deployment_status_update.py
```diff
@@ -1,5 +1,106 @@
...
+            receiver = service_bus_client.get_queue_receiver(queue_name=config.SERVICE_BUS_DEPLOYMENT_STATUS_UPDATE_QUEUE)
+            async with receiver:
+                received_msgs = await receiver.receive_messages(max_message_count=10, max_wait_time=5)
+                for msg in received_msgs:
+                    result = True
+                    try:
+                        message = json.loads(str(msg))
+                        result = (yield parse_obj_as(DeploymentStatusUpdateMessage, message))
+                    except (json.JSONDecodeError, ValidationError):
+                        logging.error(strings.DEPLOYMENT_STATUS_MESSAGE_FORMAT_INCORRECT)
```
Lybecker, line 48: I suggest logging the message too, so it is easier to debug.

management_api_app/service_bus/deployment_status_update.py
```diff
@@ -1,5 +1,106 @@
...
+    """
+    result = False
+    try:
+        workspace = workspace_repo.get_workspace_by_workspace_id(message.id)
+        workspace_repo.update_workspace(create_updated_deployment_document(workspace, message))
+        result = True
+    except EntityDoesNotExist:
+        # Marking as true as this message will never succeed anyways and should be removed from the queue.
+        result = True
+        logging.error(strings.DEPLOYMENT_STATUS_ID_NOT_FOUND)
```
Lybecker, line 86: I would add the Id to the log

tompaana (approved):

Issue #297, opened 2021-06-17
[BUG] Key Vault script in tre-deploy target fails due to missing environment variable


tompaana:
`/devops/scripts/set_contributor_sp_secrets.sh` is run as part of the `tre-deploy` target in Makefile. If run **manually**, the script will fail, because `TRE_ID` environment variable is not defined. The Makefile target loads environment variables from .env files, but injects `TF_VAR_` prefix so although TRE ID is in the environment variables, it has a name the script is not expecting.

**A** solution is to also use `load_env.sh` to process the `/templates/core/.env` file in `tre-deploy` target.

limorl:
@tompaana  I'll update the makefile and test so will add the line in bold:

tre-deploy:
	echo -e "\n\e[34m»»» 🧩 \e[96mDeploying TRE\e[0m..." \
	&& . ./devops/scripts/check_dependencies.sh nodocker \
	**&& . ./devops/scripts/load_env.sh ./templates/core/.env \**
	&& . ./devops/scripts/load_terraform_env.sh ./devops/.env \
	&& . ./devops/scripts/load_terraform_env.sh ./templates/core/.env \
	&& cd ./templates/core/terraform/ && ./deploy.sh \
	&& cd ../../../ && ./devops/scripts/set_contributor_sp_secrets.sh

limorl:
The variables are loaded correctly but the script fails on this line and actually crashes my terminal:
az keyvault secret set --name deployment-processor-azure-client-id --vault-name $key_vault_name --value $CONTRIBUTOR_SP_CLIENT_ID

christoferlof:
When I add the above two calls to load the required env variables, I'm instead seeing auth issues and Terraform terminates.
Given the error message, it seems additional env variables are picked up but also a change in behaviour - probably because of the existence of the ARM_ variables? ARM_TENANT_ID should be optional according to the quickstart.

```plaintext
[...]

Issue #280, opened 2021-06-16
[BUG] Running make build-api-image fails on docker not installed


limorl:
**Describe the bug**
When going over quickstart on a clean environment (on Mac). 
make bootstrap completes successfully but then `make build-api-image` fails on docker not installed.

The way to workaround it is to run the following commands manually, including docker build using sudo:
build-api-image:
	echo -e "\n\e[34m»»» 🧩 \e[96mBuilding API Image\e[0m..." \
	&& . ./devops/scripts/check_dependencies.sh \
	&& . ./devops/scripts/load_env.sh ./devops/.env \
	&& docker build -t "$${ACR_NAME}.azurecr.io/microsoft/azuretre/management-api:$${IMAGE_TAG}" ./management_api_app/

 

**Steps to reproduce**

1. run `make bootstrap` successfully
2. run `make mgmt-deploy`
2. run `make build-api-image`
3. Get the following error:
»»» ✅ Checking pre-reqs...

Checking for Azure CLI...

**Checking for Docker...
»»» **⚠️ Docker is not installed! 😥 Please go to https://docs.docker.com/engine/install/ to set it up****

4. Run the commands manually
     ./devops/scripts/check_dependencies.sh \
     ./devops/scripts/load_env.sh ./devops/.env \
     *sudo* docker build -t "$${ACR_NAME}.azurecr.io/microsoft/azuretre/management-api:$${IMAGE_TAG}"

limorl:
None issue (missing mgmt build)

marrobi:
@limorl  I can't replicate this, but could be my docker setup.

Can you try:

https://itprohelper.com/run-docker-without-sudo-linux/

limorl:
PR #299 merged

PR #299, opened 2021-06-17
Fixing Bug 280 + Bug 297 + documentation typos

limorl:
# PR for issue #280 #297

## What is being addressed
Bug 280 - az and docker commands in the Makefile scripts fail on permissions
Bug 297 - set_contributor_sp_secrets.sh fails since the varialbes weren't loaded properly. 

Describe the current behavior you are modifying

## How is this addressed

- Bug 280:  When running from dev container (on Mac), az and sudo commands require running in sudo mode.
- Bug 297 - load_env was ran prior to this script to load the correct variables


limorl:
submitted prematurely

sachinkundu:
Instead of sudo for docker everywhere have you tried adding $USER to the docker group?

limorl:
> Instead of sudo for docker everywhere have you tried adding $USER to the docker group?

Yes I have tried it Sachin and it didn;t work well with dev containers. Hence I added inline.

TessFerrandez (approved):

limorl:
> > Instead of sudo for docker everywhere have you tried adding $USER to the docker group?
> 
> Yes I have tried it Sachin and it didn;t work well with dev containers. Hence I added inline.

To be more accurate, adding user is done differently on MAC and wasn't a very good option.
I did follow these guidelines but it didn;t work for me: https://code.visualstudio.com/docs/remote/containers

PR #315, opened 2021-06-18
Aligns return channel between API and Processor function

deniscep:
# Aligns return channel between API and Processor function

## What is being addressed

Modifies messages and completes return channel e2e 

## How is this addressed

- Modify return messages
- Add API variables
- Add permissions to MSI


sachinkundu (approved):
We should refactor the message model into a shared module and use it between processor function and management api app. I will let this one in for now.

PR #318, opened 2021-06-18
Quickstart clarifications

christoferlof:
## What is being addressed

While running through the quickstart again, I took the opportunity to remove few things that were inaccurate and add few clarifications for an improved user experience.

## How is this addressed

- Remove requirement of GitHub Personal Access Token (PAT)
- Clarify that the management infrastructure can be used to manage multiple TRE instances
- Add an example of `.env` file since not all parameters are required.
- Add additional instructions on how to validate the deployment by hitting the `/api/health` endpoint.


Lybecker (approved):
LGTM

devops/.env.sample
```diff
@@ -1,12 +1,17 @@
+# Management infrastructure
 ARM_TENANT_ID=__CHANGE_ME__
 ARM_SUBSCRIPTION_ID=__CHANGE_ME__
-ARM_CLIENT_ID=__CHANGE_ME__
-ARM_CLIENT_SECRET=__CHANGE_ME__
```
marrobi, line 4: @christoferlof these are needed for deploying bundles, such as:

https://github.com/microsoft/AzureTRE/blob/develop/workspaces/azureml_devtestlabs/azure.json

marrobi (changes requested):
Just adding this until clarified why removing these.

devops/.env.sample
christoferlof: They are still in the .sample but further down, in an attempt to group the variables based on their usage. If we're requiring two SPs for the quickstart scenario, we need to add that obviously.

marrobi (approved):

PR #319, opened 2021-06-18
Add better api error messages

TessFerrandez:
# PR for issue #213 

## What is being addressed

- Unhandled errors should not leak information
- Make it easier to debug errors without leaking info

## How is this addressed

- Add a generic error handler that logs information, but also displays in the browser if DEBUG=True
- Change the HTTP and ValidationError output so that it is more human readable
- Update related tests

NOTE: Because I built this on an existing branch - for some reason the changes that have already been merged still show up in this diff - but you can ignore any changes in resource_templates, workspace_templates and the tests for those files - they have already been merged

### DIFF
--- a/management_api_app/api/errors/generic_error.py
+++ b/management_api_app/api/errors/generic_error.py
@@ -1,0 +1,19 @@
+import logging
+
+from starlette.requests import Request
+from starlette.responses import PlainTextResponse
+
+from core import config
+from resources import strings
+
+
+async def generic_error_handler(_: Request, exception: Exception) -> PlainTextResponse:
+    logging.debug("=====================================")
+    logging.debug("=====================================")
+    logging.exception(exception)
+    logging.debug("=====================================")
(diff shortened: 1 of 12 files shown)


management_api_app/api/errors/generic_error.py
```diff
@@ -0,0 +1,19 @@
...
+from resources import strings
+
+
+async def generic_error_handler(_: Request, exception: Exception) -> PlainTextResponse:
+    logging.debug("=====================================")
+    logging.debug("=====================================")
+    logging.exception(exception)
+    logging.debug("=====================================")
+    logging.debug("=====================================")
+    if config.DEBUG:
```
sachinkundu, line 16: how about
error_string = exception if config.DEBUG else UNABLE_TO_PROCESS_REQUEST
return PlainTextResponse(error_string, status_code=500)

sachinkundu (approved):
Nice work with the middleware for error handling!

PR #321, opened 2021-06-20
Set docker.sock permission

msebragge:
# PR for issue #280

## What is being addressed

When devcontainer is deployed to codespaces the docker group is not associated with docker.sock file

## How is this addressed

Script file to do the association.


Lybecker (approved):
LGTM

PR #334, opened 2021-06-22
Correlation ID to Service Bus and logs

### LINKED ISSUE #271: Operation ID for logging
A correlation/operation ID is needed for the workspace creation flow in order to associate log entries with the operation.

## Acceptance criteria

- [x] ID generated by the system
- [x] ID used when logging
- [x] ID passed between the required components of the composition service as needed e.g., from the API to the processor function
- [x] Verification: ID used successfully to inspect traces in Log Analytics

tompaana:
Closes #271 

## What is being addressed

Adds correlation ID for workspace deployment to Service Bus messages and manual injects that in logs.

## How is this addressed

- Adds correlation ID based on the generated resource ID to the Service Bus messages
- Logs the correlation ID in selected places manually
- Refactors resource request sending feature to match the deployment status updater
- Updates Management API unit tests


Lybecker (approved):
Looks good for now. Right now the developer has to remember to include the correlation id to each log method.

In the future I would be better with tighter integration with OpenCensus

- Accepting W3C Trace-Context from client HTTP requests
- creating span for each component (API, Resource Processer ...)
- store the Span Context in contextvars
- Make the application map work in App Insights
- Seamless propagate SpanContext through Azure Service Bus

Issue #95, opened 2021-05-21
Register API with Azure AD to enable auth


christoferlof:
## User story

As a Cloud Administrator, 
I want to be able to register an instance of Azure TRE in Azure AD,
so that I can assign users permissions to access the Azure TRE API.

## Acceptance criteria

- [ ] Script to create the application registration.
- [ ] Script to grant consent to the application.
- [ ] App role `TRE Administrator` created.

christoferlof:
App registration for cross workspace operations at TRE level.

marrobi:
@askew  I have a script to do this, will need you to review happy for me to pick this up while waiting for some deployments to run?

PR #346, opened 2021-06-22
API App Registration Script

askew:
# PR for issue #95

Shell script to create the app registrations for the API and the Swagger UI in Azure AD.

Uses the `az rest` rather than `az ad app create` as the CLI seems to be out-of-date and uses the old AAD API.

Note: The `-a` option to perform the admin consent seems to have stopped working. Need further investigation but may be due to a recent change in AAD and the Azure CLI has not caught up. In the meantime use the Azure Portal to grant admin consent.

It may be worth porting this from Bash to Python at some point.

### DIFF
--- a/scripts/aad-app-reg.sh
+++ b/scripts/aad-app-reg.sh
@@ -1,0 +1,324 @@
+#!/bin/bash
+
+# Setup Script
+set -euo pipefail
+
+usage()
+{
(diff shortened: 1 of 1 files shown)


TessFerrandez (approved):

askew:
Update to the -a option not working: Seems to be an Azure AD issue (https://github.com/Azure/azure-cli/issues/18603)

PR #320, opened 2021-06-18
Initial authorization on workspace and workspace-templates

askew:
# PR for issue #310
(ignore branch name!)

## What is being addressed

This PR adds the requirement for a JWT Bearer token to authorize calls to `/workspaces/*` and `/workspace-templates/*`

This implementation supports the Azure AD v2 endpoints.

Notes:
- The token is validated and admin role is validated but access to user workspaces is not implemented in this PR
- The necessary API ENV vars will be set in a separate PR along with a script to create the necessary AAD Applications

Closes #96 - Enforce AUTH on POST /workspaces - only TRE Admin
Closes #310 - Implement AUTH token validation for API
Closes #311 - Enforce AUTH on GET /workspaceTemplates - only TRE Admin
Closes #312 - Enforce AUTH on POST /workspaceTemplates - only TRE Admin
Closes #313 - Enforce AUTH on GET /workspaceTemplates/{template_name} - only TRE Admin
Closes #314 - Configure swagger to use OpenID AAD AUTH

This PR also sets the required environment variables in the Action that deploys to the Dev environment.

PR #354, opened 2021-06-23
add app info to workspace request - NOTE: depends on #320

### LINKED ISSUE #306: Update POST /workspace to accept an auth_config
As a TRE admin
I want to supply the information about the AAD App when creating a workspace
so that we can validate against this data if a user should have access to see the resource

This will be used to validate the access for a user to the workspace

**NOTE:** data here is different for each provider so it should be given as a dict 

## Acceptance criteria:

- [ ] The workspace post should accept an auth_config as part of the request with the following format
"auth_config": {"provider": "AAD", "data": {"app_id": "blabla"}} -- **NOTE:** the data portion is different for each provider so make this a dict
- [ ] Validates that auth configuration is provided.
- [ ] For AAD validates that the APP has the roles "owner" and "researcher"
- [ ] Stores the APP Role IDs for the owner and researcher roles as part of the workspace request 
- [ ] Stores the SP object ID as part of the resource request

TessFerrandez:
# PR for issue #306 

## What is being addressed

- User is required to add AAD app info to POST /workspaces for authentication
- App info is validated to ensure it exists and contains the right roles
- SP and role IDs are added to the workspace request to be used for user/workspace lookup

## How is this addressed

- New class AccessService - this will act as an interface (abstract base class) for all AccessServices (if someone wants to change auth method)
- Implementation of AADAccessService that validates the auth_config (e.g. app - and returns relevant info to be stored in workspace request)
- POST Workspaces input takes auth_config which in the case of the AAD provider is app_id -- this is intentionally a dict so it can be different for different auth providers
- POST Workspaces calls validation/extraction - this class in turn picks the relevant AccessService based on the provider supplied
- Relevant tests updated
- Tests added for AccessService

NOTE: This PR depends on ENV VARS that will be added by @askew as part of #320

Closes #306


management_api_app/core/config.py
```diff
@@ -30,3 +30,12 @@
...
 APP_INSIGHTS_INSTRUMENTATION_KEY: str = config("APPINSIGHTS_INSTRUMENTATIONKEY", default="")
+
+# Authentication
+API_CLIENT_ID: str = config("API_CLIENT_ID", default="")
+API_CLIENT_SECRET: str = config("API_CLIENT_SECRET", default="")
+SWAGGER_UI_CLIENT_ID: str = config("SWAGGER_UI_CLIENT_ID", default="")
+AAD_TENANT_ID: str = config("AAD_TENANT_ID", default="")
+
+AAD_INSTANCE: str = config("AAD_INSTANCE", default="https://login.microsoftonline.com")
+API_AUDIENCE: str = config("API_AUDIENCE", default=f"api://{API_CLIENT_ID}")
```
tompaana, line 41: Might be a good idea to document these.

tompaana (approved):
Missing some documentation, but otherwise looks nice and clean!

management_api_app/core/config.py
TessFerrandez: Indeed - good catch!

TessFerrandez: Added #357 to document these and all other ENV VARs needed for local runs of API - will pick that up next

PR #375, opened 2021-06-26
Adds a network flow diagram

deniscep:
# Adds a network flow diagram

## What is being addressed

Adds a diagram of current setup of TRE components as they reside on network, documenting all restrictions between workspace, core and public network.

## How is this addressed

- An architecture diagram is added


Lybecker (approved):
LGTM

Issue #307, opened 2021-06-18
Implement AUTH component for AAD


TessFerrandez:
## Description

As a TRE Admin
I want to ensure that only the researchers/owners that have access to a workspace can see the workspace info
So that information is not leaked

This story creates an auth component for AAD that can be used by authentication/authorization

## Acceptance critiera

- [ ] The component can be dependecy injected into auth

## Tasks

- [ ] Given a user - retrieve the workspaces a user has access to #308
- [ ] Given a user and a workspace - retrieve the users permissions on this workspace #309

Issue #221, opened 2021-06-08
Enforce auth on /workspaces/{workspace_id} endpoint


christoferlof:
## Description

As a researcher
I want to be able to access the information about a specific workspace
So that I can connect to it and use its deployed services

## Acceptance criteria

- [ ] Access token validated for Oidc audience matching the Oidc audience set in the workspace config. (We have sample in spike, contact @christoferlof or @marrobi for pointer)
- [ ] General validity of access token checked
- [ ] Access token needs to contain `Workspace owner` or `Researcher` role claim for access.

Depends on #307 to be finished

Issue #208, opened 2021-06-08
Retrieve 'my' workspaces


christoferlof:
## Description

As a researcher
I want to see the list of workspaces that I have access to
So that I more easily can find them rather than seeing all workspaces within the organization

## Acceptance criteria

- [ ] `GET /workspaces` only returns workspaces that the current user is assigned `researcher` or `workspace owner` role to
- [ ] `GET /workspaces` returns all workspaces for users that have the role `TRE Administrator`

## Note on implementation

We'll rely on the identity provider's capability to list applications a user has access to. For Azure AD we should query the Graph to get [appRoleAssignments](https://docs.microsoft.com/en-us/graph/api/user-list-approleassignments?view=graph-rest-1.0&tabs=http) for the current user, and then filter/join by resourceId to match with oidcApplicationId on each workspace.
This will likely require us to create an abstraction or client for different Idp's. Atm we'll only implement Azure AD.

Depends on #307 to be finished

PR #376, opened 2021-06-28
Implement conditional workspace authorization

TessFerrandez:
# PR for issue #307, #221 and #208

## What is being addressed

- API should check if user is either owner or researcher on the workspaces to determine if the user should be able to see the workspace info

## How is this addressed

- When the user is authenticated - the users role assignments are retrieved and stored in the user object
- When the user attempts to retrieve info about a single workspace, its role assignments are checked against the workspaceowner and workspaceresearcher roles on the workspace
- When the user attempts to retrieve info about multiple workspaces, its role assignments are checked against the workspaces to filter out which ones the user can see
- Tests related to the this functionality are created and updated


tompaana (approved):
LVGTMTHX

TessFerrandez:
> LVGTMTHX

WTYVMSFRTPR

Issue #340, opened 2021-06-22
[Task] Create non-privileged tests to test routes don't allow lower priviledge 


TessFerrandez:
Story: #305 

**Description**
- [ ] Create tests that check that users are authenticated when auth should be required
- [ ] Create tests that check that all routes requiring admin roles require admin roles
- [ ] Create tests that check that all routes requiring specific users are only accessed by specific users

PR #381, opened 2021-06-29
Run tests as non priv user and use admin user explicitly when needed

sachinkundu:
# PR for issue #340 

## What is being addressed

Run tests as user without TREAdmin role. This is the default.

For workspacetemplate tests which are all run under admin context, use class level _prepare fixture.
For workspaces where some routes are under admin context, do a just in time patch and use admin user instead.


TessFerrandez (approved):
LGTM!

PR #395, opened 2021-06-30
Fixes processor function App Insights configuration

### LINKED ISSUE #387: Application Insights configuration for processor function broken
The `createNewWorkspace` function monitoring blade indicates that Application Insights integration is not enabled.
image

At a first glance, the config seems to be there in [processor_function/function.tf](https://github.com/microsoft/AzureTRE/blob/develop/templates/core/terraform/processor_function/function.tf#L24)
```plaintext
    APP_INSIGHTS_INSTRUMENTATION_KEY      = var.app_insights_instrumentation_key
```
however, the setting key name is incorrect.

Manually enabling the integration in the portal adds additional config keys:

`APPLICATIONINSIGHTS_CONNECTION_STRING`
`APPINSIGHTS_INSTRUMENTATIONKEY`

Need to add above keys with values to function app config.

[Reference](https://docs.microsoft.com/en-us/azure/azure-monitor/app/resource-manager-function-app)

tompaana:
Fixes #387 

## How is this addressed

- Adds the configuration implicitly expected by the Azure Function resource:
  - Application Insights connection string added
  - Application Insights instrumentation key environment variable name changed to match the expected one

In addition, the logging code changed to use the connection string instead of the instrumentation key now that the aforementioned is available as an environment variable.

## Receipts

From "from scratch" deployment:
Function App Insights blade

Looking at Monitor blade of the function (compared to the one in bug report) after fresh deployment:
Monitor blade

### DIFF
--- a/processor_function/shared/logging.py
+++ b/processor_function/shared/logging.py
@@ -18,7 +18,7 @@ def initialize_logging(logging_level: int, correlation_id: str) -> logging.Logge
     """
     logger = logging.getLogger()
     logger.addHandler(logging.StreamHandler())  # For logging into console
-    app_insights_instrumentation_key = os.getenv("APP_INSIGHTS_INSTRUMENTATION_KEY")
+    app_insights_instrumentation_key = os.getenv("APPINSIGHTS_INSTRUMENTATIONKEY")
 
     try:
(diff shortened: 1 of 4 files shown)


christoferlof (approved):
Looking good to me. Will verify end-to-end once available in dev environment after merge.

christoferlof:
Verified. Works as expected 👍

Issue #350, opened 2021-06-22
[Task] Create InnerEye Deep Learning and Inference Bundle


marrobi:
Story: #132

**Description**
Build upon the inner eye bundle by adding the inference service.

PR #404, opened 2021-07-01
Refactor porter bundles to simplify dependencies.

marrobi:
# PR for issue #350

## What is being addressed

Refactor porter bundles to simplify dependencies.

## How is this addressed

- Move to using scripts as proved out with the innereye inference bundle
- Remove workspace outputs no longer needed.

### DIFF
--- a/workspaces/azureml_devtestlabs/install_workspace_vanilla.sh
+++ b/workspaces/azureml_devtestlabs/install_workspace_vanilla.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+set -e
+
+porter install tre-workspace-vanilla --reference "${ACR_NAME}.azurecr.io/tre-workspace-vanilla:v0.1.0" \
+    --cred ./azure.json \
+    --parameter-set ./parameters_workspace_vanilla.json \
+    --debug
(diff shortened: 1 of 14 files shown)


workspaces/azureml_devtestlabs/install_workspace_vanilla.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -e
+
+porter install tre-workspace-vanilla --reference "${ACR_NAME}.azurecr.io/tre-workspace-vanilla:v0.1.0" \
```
sachinkundu, line 4: consider making the code DRY as the file above is almost the same and exactly the same if the parameters are abstracted as inputs

sachinkundu (commented):
I think you should write one bash script which takes args and does what you need done instead of replicating the same commands in many files. You can make a dict of images and parameters etc and then only one place needs to be changed when eventually changes do happen.

marrobi:
@sachinkundu agree, although would still need to be one in each bundle as are separate docker contexts.

The challenge comes with when need to retrieve different outputs for different bundles. Any suggestions?

I'm hoping these scripts becomes redundant once we have the API configured to deploy services.

sachinkundu:
> @sachinkundu agree, although would still need to be one in each bundle as are separate docker contexts.
> 
> The challenge comes with when need to retrieve different outputs for different bundles. Any suggestions?
> 
> **I'm hoping these scripts becomes redundant once we have the API configured to deploy services.**
Hence approved.

sachinkundu (approved):

PR #416, opened 2021-07-06
Update workspace docs and add devtest labs scripts.

marrobi:
# PR for issue 350

## What is being addressed

Update readme files for all services adn add script to create dev test labs VM and expose through the firewall.

## How is this addressed

- Update .env.sample files
- Update readme files for each workspace and service
- Add script to create a dev test labs VM, and create a nat rule through the firewall as a workaround until we have a virtual desktop service.


workspaces/services/azureml/readme.md
```diff
@@ -0,0 +1,37 @@
+# Azure machine Learning Serivce bundle
```
Lybecker, line 1:
```suggestion
# Azure machine Learning Service bundle
```

workspaces/services/azureml/readme.md
```diff
@@ -0,0 +1,37 @@
...
+
+URLs:
+
+- graph.windows.net
+- ml.azure.com
+- login.microsoftonline.com
+- aadcdn.msftauth.net
+- graph.microsoft.com
+- management.azure.com
+- viennaglobal.azurecr.io
```
Lybecker, line 17: I suggest describing why each of these are needed, as I would expect questions about this from a network administrator

workspaces/services/devtestlabs/readme.md
```diff
@@ -0,0 +1,43 @@
+# Azure DevTest Labs Service bundle
```
Lybecker, line 1: I suggest adding a description of why Dev Test Labs are used and the advantage

workspaces/services/innereye_deeplearning/readme.md
```diff
@@ -0,0 +1,56 @@
+# InnerEye DeepLearning Service Bundle
+
+See: [https://github.com/microsoft/InnerEye-DeepLearning](https://github.com/microsoft/InnerEye-DeepLearning)
+
+## Firewall Rules
+
+Please be aware that the following Firewall rules are opened for the workspace when this service is deployed:
```
Lybecker, line 7:
```suggestion
Please be aware that the following Firewall rules are opened for the workspace when this service is deployed. These are all dependencies needed by InnerEye to be able to develop and train models:
```

Lybecker (approved):
LGTM.

A few comments.

I suggest describing that the documentation states that the Workspace_ID should be the same for all bundles.

Issue #390, opened 2021-06-29
[BUG] Tre-deploy fails with current quickstart on the last step


msebragge:
**Describe the bug**
Missing policy for user login

**Steps to reproduce**

Follow quickstart

*az login*
...
*make tre-deploy*

**result:**
»»» 🤖 Creating (or updating) service principal ID and secret to Key Vault...
The user, group or application 'appid=04b07795-8ddb-461a-bbee-02f9e1bf7b46;oid=<USER_ID>;numgroups=98;iss=https://sts.windows.net/72f988bf-86f1-41af-91ab-2d7cd011db47/' does not have secrets set permission on key vault 'kv-<TRE_ID>;location=<LOCATION>'. For help resolving this issue, please see https://go.microsoft.com/fwlink/?linkid=2125287
make: *** [Makefile:58: tre-deploy] Error 1

**Workaround:**

*az login --service-principal --username APP_ID --password PASSWORD --tenant TENANT_ID*

using the previously created service principal for ARM_CLIENT_ID

Issue #373, opened 2021-06-24
[BUG] Confusion between TRE Deployment SP and ResourceProcessor SP


limorl:
**Describe the bug**
Quickstart.md indicates the ARM_SUBSCRIPTION_ID, ARM_TENANT_ID, ARM_CLIENT_ID and ARM_CLIENT_SECRET are optional and required only from deployment from CD or when deploying workspace.
However they are required also when running locally.

![image](https://user-images.githubusercontent.com/1086572/123336925-313ac580-d54f-11eb-9199-3ea2ed23de1b.png)

**Steps to reproduce**

1. Go to .devops/.env
2. Comment out ARM_SUBSCRIPTION_ID
3. run `make bootstrap`
4. Get the following error:
![image](https://user-images.githubusercontent.com/1086572/123326292-bcad5a00-d541-11eb-9f2d-0c3cae391243.png)
5. Do the same for the rest of the variables, and run make tre-deploy

christoferlof:
It seems the Terraform related env variables (`ARM_*`) are injected into the CNABs by the processor function, why they are now required. We need to update the quickstart to flag them as required and also to describe what service principal(s) that are required.

Lybecker:
The problem seems to be due to a mix-up of Service Principals. The solution uses two:

1. Owner SP to create the environment.
2. Contributor SP for the Resource Processor of the Composition Service to dynamically provision environments.

The first SP should be optional but is used by the Resource Processor function.

**Research**

The Resource Processor uses the SEC_ARM_CLIENT_* environment variables:
https://github.com/microsoft/AzureTRE/blob/e9c74900d2ac832199b52fcbecee5ef6f1aeb654/templates/core/terraform/processor_function/function.tf#L45-L46

As the CNAB_Builder cuts the _SEC_ section of the variable name before setting it for the ACI container
https://github.com/microsoft/AzureTRE/blob/e9c74900d2ac832199b52fcbecee5ef6f1aeb654/processor_function/shared/cnab_builder.py#L65

The values are set to the Owner SP here:
https://github.com/microsoft/AzureTRE/blob/e9c74900d2ac832199b52fcbecee5ef6f1aeb654/templates/core/terraform/main.tf#L129-L130

There is a problem with naming in the Terraform scripts - it would make it easier to read if they had names like OwnerSP/ContributorSP or ResourceProcessorSP... something along those lines

limorl:
Here's a summary of our. discussion.
[...]

PR #428, opened 2021-07-07
Fixes required to make deploying vanilla workspace in a new subscription

sachinkundu:
#373 #390 #412


devops/.env.sample
```diff
@@ -6,15 +6,10 @@ TERRAFORM_STATE_CONTAINER_NAME=tfstate
...
-# Service Principal used by the Composition Service to provision workspaces
-CONTRIBUTOR_SP_CLIENT_ID=__CHANGE_ME__
-CONTRIBUTOR_SP_CLIENT_SECRET=__CHANGE_ME__
-
 # Azure Resource Manager credentials used in CI/CD pipeline and local development scenarios
-ARM_TENANT_ID=__CHANGE_ME__
 ARM_SUBSCRIPTION_ID=__CHANGE_ME__
-ARM_CLIENT_ID=__CHANGE_ME__
-ARM_CLIENT_SECRET=__CHANGE_ME__
+RESOURCE_PROCESSOR_CLIENT_ID=__CHANGE_ME__
```
tompaana, line 11: I suppose there is a good reason for changing the name here, but isn't this SP used outside of the processor function too?

tompaana (approved):

devops/.env.sample
sachinkundu: Having variable ARM_CLIENT_SECRET confused terraform locally and it was deemed that since the SP is used in resource processor it should be marked as such. I have changed the variables accordingly in TF but not all the ways inside the chain. It works!

Issue #411, opened 2021-07-05
[Story] Documentation on how to build and publish a porter bundle to the Azure TRE bundle registry


marrobi:
## Description

As a Cloud Administrator
I want to publish porter bundles to the Azure TRE registry 
So that the bundles are available for deployment

## Acceptance criteria

- [x] Documentation on how to register a workspace bundle
- [x] Registered bundle returned by the get workspace templates operation.

Issue #431, opened 2021-07-08
Documentation refined and restructured


Lybecker:
The Azure TRE needs to have **a better structure** and should have the following content added:

- [x] An overview of what Azure TRE is and what are its benefits
- [x] A guide for cloud administrators to provision
- [x] A guide to provision new workspaces
- [x] A guide for developers to setup the development environment
- [x] A guide where to find what and what is the content in each folder
- [x] Architectural overview with description of each component
- [x] Network diagram with rules, boundaries, ingress/egress

Related task:

- [x] #378 Add instruction to set auth related vars in Quickstart

ktakeda1:
Please could we add higher-level documentation for business decision-makers? Happy to help with those, could include researcher user stories

Lybecker:
Of cause... I will need some help :)

Lybecker:
@ktakeda1 I've added a section for business decision makers here: https://github.com/microsoft/AzureTRE/tree/docs/docs#business-decision-makers

If you don't mind, then either post it here or a PR.

PR #433, opened 2021-07-08
Add Registering Resource Templates document

marrobi:
# PR for issue #411

## What is being addressed

Add Registering Resource Templates document
## How is this addressed

- Add document describing how to register templates.


tompaana (approved):
Looks good to me. The new document should be cross-linked with `authoring-workspaces.md`, but that can be done in #431 too.

Issue #442, opened 2021-07-12
[BUG] make mgmt-deploy needs to be run before publish_bundles (missing ACR) 


msebragge:
**Describe the bug**
Running workflow the first time in a clean environment fails on step **publish_bundles** because ACR is not yet created.
ACR is created in the **make mgmt-deploy** step. 

**Steps to reproduce**

```bash
»»» 🧩 Publishing ./workspaces/vanilla bundle...

╔══════════════════════════════════════╗
║          Azure TRE Makefile          ║
╚══════════════════════════════════════╝

»»» ✅ Checking pre-reqs...

Checking for Azure CLI...

Checking for Docker...

Checking for porter...
»»» 🔨 Azure details from logged on user 
»»»   • Subscription: anlybeck-ebragge
»»»   • Tenant:       72f988bf-86f1-41af-91ab-2d7cd011db47

WARNING: The resource with name '***' and type 'Microsoft.ContainerRegistry/registries' could not be found in subscription 'anlybeck-ebragge (d167531c-1631-46c9-9edc-0486b659e7c0)'.
Using '***.azurecr.io' as the default registry login server.
ERROR: Could not connect to the registry login server '***.azurecr.io'. Please verify that the registry exists and the URL 'https://***.azurecr.io/v2/' is reachable from your environment.
Try running 'az acr check-health -n *** --yes' to diagnose this issue.
make: *** [Makefile:131: porter-publish] Error 1
Error: Process completed with exit code 2.
```

msebragge:
Pull request #443

PR #443, opened 2021-07-12
Add make mgmt-deploy to deploy_mgmt_infra

msebragge:
# PR for issue #442

Closing #442 

## What is being addressed

Running workflow first time fails is publish_bundles step because of missing ACR.

## How is this addressed

Add step **make mgmt-deploy** to job **deploy_mgmt_infra**


marrobi (approved):
Good catch, sorry!

Issue #448, opened 2021-07-13
[BUG] Second run of deploy_tre removed gateway access to keyvault


msebragge:
**Describe the bug**

Running deploy_tre second time removes msi-agw-**** from keyvault. 

**Steps to reproduce**

1. Run a clean workflow
2. Verify that msi-agw-**** is assigned to keyvault
3. Re-run workflow
4. msi-agw-**** is not assigned anymore

msebragge:
PR #451

PR #451, opened 2021-07-13
Unify keyvault access policies to fix bug 448

msebragge:
# PR for issue #448 

Closing #448

## What is being addressed

Second run of deploy_tre removed access policies from keyvault.

## How is this addressed

All access policies are now separate resources (removed from inside keyvault resource).


marrobi (approved):
LGTM

sachinkundu:
Access policies within keyvault block and access policies for the same kv outside behave diffrently? sounds like a bug in azure terraform provider

PR #454, opened 2021-07-13
Add resource processor VNET, Firewall rules, and optional VMSS for resource processor

marrobi:
# PR for #452

## What is being addressed

To enable the option of running porter jobs on a VNet with managed identity a VNet and VMSS is required.

## How is this addressed

- Add resource processor VNet
- Add firewall rules for shared and resource processor subnets that allow package installation
- Add VMSS for resource processor with optional deployment based on `RESOURCE_PROCESSOR_TYPE` env var
- tidied core subnets with smaller address space


sachinkundu (approved):
TL;DR If it works LGTM.

Issue #444, opened 2021-07-12
[Task] Integrate e2e tests into GitHub workflow


marrobi:
Story: XX

**Description**
A clear and concise description of what the this task is about.

PR #460, opened 2021-07-14
Enable E2E tests 

sachinkundu:
# PR for issue #444 

run e2e tests within the workflow.

Add and refactor tests to report full status

Fix issue that status might be written over after a deployed status is received which should never happen,


marrobi (approved):
One nit, happy to clean up later if easier.

Great work.

PR #464, opened 2021-07-14
VMSS Processor

marrobi:
# PR for issue #456

## What is being addressed

Create a resource processor that runs on VMSS

## How is this addressed


marrobi:
New PR to be created when ready by @sachinkundu

PR #467, opened 2021-07-15
Fixes for E2E tests

sachinkundu:
# PR for issue #444 

Correct if block for github
Sleep 1m before deleting resources to prevent terraform lock
Remove CNAB resources - comment lines if need to debug later
always try to delete the created resources even if tests fail


marrobi (approved):
LGTM

Issue #406, opened 2021-07-02
[BUG] make letsencrypt not working


msebragge:
**Describe the bug**
1) Documentation outdated
    - Missing info around STORAGE_ACCOUNT

2) Network rules changed

**Steps to reproduce**

```bash
vscode ➜ /workspaces/AzureTRE (develop ✗) $ export STORAGE_ACCOUNT=stwebebra01
vscode ➜ /workspaces/AzureTRE (develop ✗) $ alias exit=return
vscode ➜ /workspaces/AzureTRE (develop ✗) $ make letsencrypt
echo -e "\n\e[34m»»» 🧩 \e[96mRequesting LetsEncrypt SSL certificate\e[0m..." \
&& . ./devops/scripts/check_dependencies.sh nodocker,certbot \
&& chmod 755 ./devops/scripts/letsencrypt.sh ./devops/scripts/auth-hook.sh ./devops/scripts/cleanup-hook.sh \
&& . ./devops/scripts/get-coreenv.sh \
&& ./devops/scripts/letsencrypt.sh

»»» 🧩 Requesting LetsEncrypt SSL certificate...

╔══════════════════════════════════════╗
║          Azure TRE Makefile          ║
╚══════════════════════════════════════╝

»»» ✅ Checking pre-reqs...

Checking for Azure CLI...

Checking for Certbot...
»»» 🔨 Azure details from logged on user 
»»»   • Subscription: xxxxx
»»»   • Tenant:       xxxxx

RESOURCE_GROUP=rg-ebra01
APPLICATION_GATEWAY=agw-ebra01
STORAGE_ACCOUNT=stwebebra01
KEYVAULT=kv-ebra01
FQDN=ebra01.westeurope.cloudapp.azure.com
Static Web hosting already enabled on storage account
Checking for index.html file in storage account
ERROR: 
The request may be blocked by network rules of storage account. Please check network rule set using 'az storage account show -n accountname --query networkRuleSet'.
If you want to change the default action to apply when no rule matches, please use 'az storage account update'.
                    
Uploading index.html file
[...]

PR #468, opened 2021-07-15
Modify lets encrypt configuration and add TF wrapper

marrobi:
# PR for issue #406

## What is being addressed

Lets Encrypt was failing as account is locked down. 

Build failign on TF locks.

## How is this addressed

- Moved static web storage to a private endpoint
- modified scripts to whitelist client ip do the lets encrypt stuff and remove
- moved the scripts to live with core templates
- added a tf wrapper that waits for state file to be unlocked


sachinkundu (approved):
1. What if the ip address is added but terraform crashes and ip is not removed?
2. No doc changes?

marrobi:
Not sure it affects the docs, will double check. Fair point about the crash. Let's chat later.

Will merge as is and create issues ss we decide fit.

PR #473, opened 2021-07-16
Fix missing deploy_gitea var

marrobi:
# PR for issue #463 

## What is being addressed

CD failing due to missing variable and TF lock file inconsistency.

## How is this addressed

- Add deploy_gitea variable
- Add updated lock file
- Change TF command to run plan first

green run: Run: https://github.com/microsoft/AzureTRE/actions/runs/1037051556


.github/workflows/deploy_tre.yml
```diff
@@ -133,6 +133,7 @@ jobs:
             export TF_VAR_resource_processor_client_id=$RESOURCE_PROCESSOR_CLIENT_ID
             export TF_VAR_resource_processor_client_secret=$RESOURCE_PROCESSOR_CLIENT_SECRET
             export TF_VAR_debug="false"
+            export TF_VAR_deploy_gitea=true
```
tamirkamara, line 136: This means that we always want to deploy that service, right? Shouldn't this be an option a user need to choose?

marrobi: This is only in our CD pipeline. The person deploying can set it to false if they do not want deploy.

marrobi: Also if not clear, this is a shared service, so decision at deploy time, rather than a service within a workspace.

anatbal: @marrobi, the "DEPLOY_GITEA" var should also be documented in the docs under both manual deployment and workflows

anatbal (commented):
@marrobi , the "DEPLOY_GITEA" var should also be documented in the docs under both manual deployment and workflows

tamirkamara (approved):

Issue #445, opened 2021-07-12
[Story] Create resource processor supporting managed identity and vnets


marrobi:
## Description

As a Cloud Administrator
I want an resource processor that uses managed identities, can be deployed on a VNET and supports docker in docker
So that I do not have to provide a service principal for deployments, can carry out data plane operations, and can deploy nested porter bundles

## Acceptance criteria

- [x] Resource processor deployed on a vnet with access to workspace data plane
- [x] Resource processor uses managed identities to deploy bundles
- [x] azureml-devtestlabs workspace successfully deploys via API
- [x] #492

## Tasks
- [x] #452
- [x] #480
- [x] #456
- [x] #457
- [x] #496
- [x] #506
- [x] #533

PR #491, opened 2021-07-22
Enable vmss_porter as the default resource processor

sachinkundu:
# PR for issue #445 

Now VMSS porter is the default resource processor.

Build runs green and tests pass here https://github.com/microsoft/AzureTRE/actions/runs/1055300642


marrobi (approved):
Lgtm

PR #493, opened 2021-07-23
Switch resource processor to use MSI

marrobi:
# PR for issue 457

## What is being addressed

Configure resource processor to use MSI

## How is this addressed

- Add Owner rights to VMMS MSI, See for future: #495 
- Update runner.py to set client_secret to blank value if not set
- Remove porter_driver parameters

### DIFF
--- a/processor_function/.env.sample
+++ b/processor_function/.env.sample
@@ -29,3 +29,8 @@ SEC_ARM_CLIENT_SECRET=__CHANGE_ME__
 param_tfstate_resource_group_name=__CHANGE_ME__
 param_tfstate_container_name=__CHANGE_ME__
 param_tfstate_storage_account_name=__CHANGE_ME__
+
+# The below are needed for local development of the VMSS resource processor
+ARM_USE_MSI=false
+RESOURCE_PROCESSOR_CLIENT_ID=__CHANGE_ME__
+RESOURCE_PROCESSOR_CLIENT_SECRET=__CHANGE_ME__
--- a/processor_function/vm_porter/runner.py
+++ b/processor_function/vm_porter/runner.py
@@ -5,7 +5,7 @@ import socket
 import asyncio
 import logging
 from shared.logging import disable_unwanted_loggers, initialize_logging  # pylint: disable=import-error # noqa
-from resources import strings
+from resources import strings # pylint: disable=import-error # noqa
 from contextlib import asynccontextmanager
 from azure.servicebus import ServiceBusMessage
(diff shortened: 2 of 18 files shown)


processor_function/vm_porter/runner.py
```diff
@@ -188,11 +189,19 @@ def read_env_vars():
...
         # Needed for running porter
+        "arm_use_msi": os.environ["ARM_USE_MSI"],
         "arm_subscription_id": os.environ['ARM_SUBSCRIPTION_ID'],
         "arm_client_id": os.environ["ARM_CLIENT_ID"],
-        "arm_client_secret": os.environ["ARM_CLIENT_SECRET"],
-        "arm_tenant_id": os.environ["ARM_TENANT_ID"],
+        "arm_tenant_id": os.environ["ARM_TENANT_ID"]
     }
+
+    if env_vars["arm_use_msi"] == "false":
```
sachinkundu, line 198: env_vars["arm_client_secret"]  = os.environ["ARM_CLIENT_SECRET"] if env_vars["arm_use_msi"] == "false" else "not_required"

processor_function/.env.sample
```diff
@@ -28,4 +28,9 @@ SEC_ARM_CLIENT_ID=__CHANGE_ME__
 SEC_ARM_CLIENT_SECRET=__CHANGE_ME__
 param_tfstate_resource_group_name=__CHANGE_ME__
 param_tfstate_container_name=__CHANGE_ME__
-param_tfstate_storage_account_name=__CHANGE_ME__
\ No newline at end of file
+param_tfstate_storage_account_name=__CHANGE_ME__
+
+# The below are needed for local development of the VMSS resource processor
+ARM_USE_MSI=false
+RESOURCE_PROCESSOR_CLIENT_ID=__CHANGE_ME__
```
sachinkundu, line 35: Is this really still needed for VMSS processor when working locally? I think we can use local login credentials to do even porter deployments as long as the user has the right set of permissions.

processor_function/vm_porter/runner.py
marrobi: trying to get head round that - so thats a single line? any docs/python 101 I can read?!

processor_function/.env.sample
marrobi: Its the TF inside the bundles that needs them. As the bundles run inside a container, can't see how would access the local creds?

sachinkundu (approved):

Issue #496, opened 2021-07-23
Ensure azureml-devtestlabs and other bundles deploy.


marrobi:

marrobi:
As the more complex workspace bundles now rely on MSI - for ACR login and modifying firewall rules the cannot be deployed using `make`. They likely need to be moved to e2e tests to be deployed via the API.

PR #497, opened 2021-07-23
Enable Azure ML DevTest Labs bundle to deploy

marrobi:
# PR for issue #496

## What is being addressed

Enable azureml_devtestlabs bundle to deploy

## How is this addressed

- Switch workspaces to use MSI for child services
- Use msi for docker logic
- Include az cli firewall extension in base images
- Remove azurecr.io from az login command to remove warning
- Add register-bundle to make file
- Comment out tests for "complex" bundles as they are executing on the github runner so cannot use managed identity. They need adding as e2e tests deploying using the API. #511


sachinkundu (approved):
TL;DR

Issue #477, opened 2021-07-19
[BUG] - bootstrap.sh scripts depends on the absence of azure cli defaults (rg name)


eladiw:
Remember to label the issue with the appropriate area, and also assign it to yourself if you plan to work on the issue in the near future.

**Describe the bug**
Using the devcontainer option, ran 'make all' and got the following error:
_(ResourceGroupNotFound) Resource group 'xyz-redacted-xyz' could not be found._
After investigating I noticed that the bootstrap.sh command is not receiving the resource group argument, which  means that if you have any default value already set it will use a wrong rg.

**Steps to reproduce**

1. set a default value for the rg using 'az config set' (set to a non-existing rg name)
2. use devcontainer env
3. run 'make all' 
4. it should fail as it is looking for the wrong rg

PR #505, opened 2021-07-27
Adding explicit resource group name to bootstrap script

eladiw:
# PR for issue

#477 

The bootstrap script was missing a rg name. in some scenarios it works but if the one deploying the solution has a default rg defined locally the script will fail.

## How is this addressed

- Explicitly passed the rg name to the command


marrobi (approved):
Thanks.

Issue #508, opened 2021-07-28
When registering a bundle, the associated JSON schema is stored


sachinkundu:

TessFerrandez:
@sachinkundu is this closed?

Issue #420, opened 2021-07-06
[Story] Provide JSON schema for API validation and UI generation


marrobi:
## Description

As a TRE Developer
I want to receive an JSON Schema that represents the input parameters for the resource being deployed
So that I can create a UI using a library such as https://jsonforms.io/

As an API
I want to use a JSON Schema for a resource template
So that I can validate input parameters using a library such as https://python-jsonschema.readthedocs.io/en/stable/

As a TRE service integrator
I want to provide a JSON Schema document with my templates
So that I get appropriate values for resource deployment

## Acceptance criteria

- [x] #509
- [x] #526
- [x] #508

## Tasks
- [x] Create workspace and vanilla bundle json schema files
- [x] Enable storage of template json schema when template registered 
- [x] On get workspace template merge the workspace schema (see below) and bundle schema and return to user
- [x] On create workspace carry out validation against the merged json schema and store parameters
- [x] Ensure workspace deploys (only selected parameters will be required)

marrobi:
Each resource type has a standard set of parameters provided by the system and a standard set provided by the user, and then custom parameters:

Workspaces have the following system parameters, each bundle must use as part of deployment:

```json
{
            "azure_location": "",
            "workspace_id": "",
            "tre_id": ""
}
```
The following user provided parameters are not used for bundle deployment, but are required for all workspaces:

```json
{
            "display_name": "",
            "description":""
}
```

Workspace authors can also request additional parameters for use by the bundle:

```json
{
            "address_space": "",
            "no_of_vms":""
}
```

Then there are also auth parameters that need to be provided by the user dependent on the the configured auth provider:
```json
{
    "appRegId": ""
}

marrobi:
The GET /workspace-template/{template_name} API should return these as a **single** schema that can be used create a UI using a library such as https://jsonforms.io/.

The returned values by POST /workspaces, should be validated against the schema and stored as resource parameters in the state store.

PR #513, opened 2021-07-28
JSON forms based schema for templates POST and GET

sachinkundu:
# PR for issue #420 #508 #509 

When registering a template a payload like 

```json
{
  "name": "my-tre-workspace-6",
  "version": "0.0.1",
  "current": "true",
  "json_schema": {
    "$schema": "http://json-schema.org/draft-07/schema",
    "$id": "https://github.com/microsoft/AzureTRE/templates/workspaces/myworkspace/workspace.json",
    "type": "object",
    "title": "My Workspace Template Custom Parameters",
    "description": "These parameters are specific to my workspace template",
    "required": [
      "blah",
      "blah_2"
    ],
    "properties": {
      "blah": {
        "type": "string",
        "title": "VM size",
        "description": "Size of the VMs in my workspace",
        "default": "Standard_A1",
        "enum": [
          "Standard_A1",
          "Standard_A2",
          "Standard_A3"
        ]
      },
      "blah_2": {
        "type": "integer",
        "title": "Number of VMs",
        "description": "Number of virtual machines to be deployed in the workspace",
        "default": 0
      }
    }
  }
}
```
Notice the presence of **json_schema** sub document.

This will be stored as a **ResourceTemplate** in state store as

```json
{
    "id": "2b3f98a7-1152-4a24-8b9b-6513b8bb23e7",
    "name": "my-tre-workspace-6",
    "description": "These parameters are specific to my workspace template",
    "version": "0.0.1",
    "resourceType": "workspace",
    "current": true,
    "type": "object",
    "required": [
        "blah",
        "blah_2"
    ],
    "properties": {
        "blah": {
            "type": "string",
            "title": "VM size",
            "description": "Size of the VMs in my workspace",
            "default": "Standard_A1"
        },
        "blah_2": {
            "type": "integer",
            "title": "Number of VMs",
            "description": "Number of virtual machines to be deployed in the workspace",
            "default": 0
        }
    },
    "_rid": "q24aAPVBQEIHAAAAAAAAAA==",
    "_self": "dbs/q24aAA==/colls/q24aAPVBQEI=/docs/q24aAPVBQEIHAAAAAAAAAA==/",
    "_etag": "\"00000000-0000-0000-83da-95af04e001d7\"",
    "_attachments": "attachments/",
    "_ts": 1627495286
}
```

When GET ``/api/workspace-templates/my-tre-workspace-6`` is done we will get an enriched response like

```json
{
  "id": "2b3f98a7-1152-4a24-8b9b-6513b8bb23e7",
  "name": "my-tre-workspace-6",
  "description": "These parameters are specific to my workspace template",
  "version": "0.0.1",
  "resourceType": "workspace",
  "current": true,
  "type": "object",
  "required": [
    "display_name",
    "description",
    "app_id",
    "blah",
    "blah_2"
  ],
  "properties": {
    "display_name": {
      "type": "string",
      "title": "Name for the workspace",
      "description": "The name of the workspace to be displayed to users",
      "default": null
    },
    "description": {
      "type": "string",
      "title": "Description of the workspace",
      "description": "Description of the workspace",
      "default": null
    },
    "address_space": {
      "type": "string",
      "title": "Address space",
      "description": "Network address space to be used by the workspace",
      "default": null
    },
    "app_id": {
      "type": "string",
      "title": "App Registration ID",
      "description": "AAD App registration of the workspace",
      "default": null
    },
    "blah": {
      "type": "string",
      "title": "VM size",
      "description": "Size of the VMs in my workspace",
      "default": "Standard_A1"
    },
    "blah_2": {
      "type": "integer",
      "title": "Number of VMs",
      "description": "Number of virtual machines to be deployed in the workspace",
      "default": 0
    }
  },
  "system_properties": {
    "tre_id": {
      "type": "string",
      "title": "",
      "description": "",
      "default": null
    },
    "workspace_id": {
      "type": "string",
      "title": "",
      "description": "",
      "default": null
    },
    "azure_location": {
      "type": "string",
      "title": "",
      "description": "",
      "default": null
    }
  }
}
```
Notice **required** **properties** and **system_properties** field. This is JSON schema compatible result and should be possible to generate auto UI from.

**NOTE** Validation tests have been removed as it makes little sense to fix tests which will be removed anyways since we are going to use JSON schema validation directly.


marrobi:
Have you considered the pros/cons of having the schema as a subdocument in the registration payload vs the payload actually being the schema?

So maybe:

```json
{
    "$schema": "http://json-schema.org/draft-07/schema",
    "$id": "https://github.com/microsoft/AzureTRE/templates/workspaces/myworkspace/workspace.json",
    "type": "object",
    "title": "my-tre-workspace-6",
    "version": "0.0.1",
    "current": "true",
    "description": "My workspace template",
    "required": [
      "blah",
      "blah_2"
    ],
    "properties": {
      "blah": {
        "type": "string",
        "title": "VM size",
        "description": "Size of the VMs in my workspace",
        "default": "Standard_A1",
        "enum": [
          "Standard_A1",
          "Standard_A2",
          "Standard_A3"
        ]
      },
      "blah_2": {
        "type": "integer",
        "title": "Number of VMs",
        "description": "Number of virtual machines to be deployed in the workspace",
        "default": 0
      }
    }
}
````

Not sure  which makes most sense.

sachinkundu:
> Not sure which makes most sense.

Yes would have tried that but TBH could not figure out a method where $schema could be used as a key name in a dictionary. Probably some quoting stuff in python should enable it but was not straightforward. Also schema being a reserved keyword in pydantic had to use json_schema. I don't think the current request makes it too inconvenient either so did not want to spend more time on it. If you think it should be changed I can try to dig in

marrobi:
@sachinkundu I'm happy with the JSON - but think this will break deployments as are currently relying on the porter parameters to know which fields to pass to porter.

So the parameters required by the bundle need to be either stored at registration time, or alternatively the runner could use `porter explain` to extract the required parameters and then provide the required values.

marrobi:
I'd also take `system_properties` out of the JSON response. I don't see why we need to expose them as they are internal.

sachinkundu:
> I'd also take `system_properties` out of the JSON response. I don't see why we need to expose them as they are internal.

I will remove this in the next patch. Just something extra and it was being reported before too.

marrobi (approved):
Happy from a "what it does" perspective, and understanding will need to be follow ups.

PR #525, opened 2021-07-30
Use JSON schema validation instead of custom logic during POST workspace

sachinkundu:
# PR for issue #420 #526

We accept JSON schema as workspace template so we can verify the incoming payload against this schema using jsonschema validate.

Hence removing all custom validation. Updating tests.

porter bundle must be given only those parameters which are defined in its own manifest hence changing the runner code to filter out extra parameters by calling porter explain on the reference to be installed and removing parameters not needed.


marrobi (approved):
Functionality verified and works.

Issue #519, opened 2021-07-29
[Task] Add Nexus readme


tamirkamara:
Story: #437 

**Description**
The readme should explain the service and any steps the TRE admin (or researchers) need/should take.

Issue #518, opened 2021-07-29
[Task] Install Nexus as a shared service


tamirkamara:
Story: #437 

**Description**
We want to setup Nexus as a shared service and enable it by default.
Note that there are storage considerations due to admin access.

PR #529, opened 2021-08-01
Nexus shared service

tamirkamara:
# PR for issue #518

## What is being addressed

Enabling workspace users to access external software packages.

## How is this addressed

- New shared service for Nexus, enabled by default.
- Docs are still pending #519

### DIFF
--- a/templates/shared_services/sonatype-nexus/terraform/deploy.sh
+++ b/templates/shared_services/sonatype-nexus/terraform/deploy.sh
@@ -1,8 +1,3 @@
-# is this required?
-export TF_VAR_docker_registry_server="$TF_VAR_acr_name.azurecr.io"
-export TF_VAR_docker_registry_username=$TF_VAR_acr_name
-export TF_VAR_docker_registry_password=$(az acr credential show --name ${TF_VAR_acr_name} --query passwords[0].value -o tsv | sed 's/"//g')
-
 export TF_LOG=""
 
 terraform init -input=false -backend=true -reconfigure \
--- a/templates/shared_services/sonatype-nexus/terraform/deploy.sh
+++ b/templates/shared_services/sonatype-nexus/terraform/deploy.sh
@@ -1,0 +1,14 @@
+# is this required?
+export TF_VAR_docker_registry_server="$TF_VAR_acr_name.azurecr.io"
+export TF_VAR_docker_registry_username=$TF_VAR_acr_name
+export TF_VAR_docker_registry_password=$(az acr credential show --name ${TF_VAR_acr_name} --query passwords[0].value -o tsv | sed 's/"//g')
+
+export TF_LOG=""
+
+terraform init -input=false -backend=true -reconfigure \
+    -backend-config="resource_group_name=$TF_VAR_mgmt_resource_group_name" \
+    -backend-config="storage_account_name=$TF_VAR_mgmt_storage_account_name" \
+    -backend-config="container_name=$TF_VAR_terraform_state_container_name" \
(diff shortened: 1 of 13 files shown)


templates/shared_services/sonatype-nexus/terraform/deploy.sh
```diff
@@ -0,0 +1,14 @@
+# is this required?
```
eladiw, line 1: need to check

templates/shared_services/sonatype-nexus/terraform/webapp.tf
```diff
@@ -0,0 +1,193 @@
...
+
+    WEBSITES_ENABLE_APP_SERVICE_STORAGE = false
+  }
+
+  site_config {
+    linux_fx_version            = "DOCKER|sonatype/nexus3"
+    remote_debugging_enabled    = false
+    scm_use_main_ip_restriction = true
+
+    # why do we need this?
```
eladiw, line 27: check before merge (and remove comment)

templates/shared_services/sonatype-nexus/terraform/webapp.tf
```diff
@@ -0,0 +1,193 @@
...
+    category = "AppServiceAppLogs"
+    enabled  = true
+
+    retention_policy {
+      days    = 1
+      enabled = false
+    }
+  }
+
+  log {
```
eladiw, line 129: all these should be set individually?

eladiw (approved):
some minor comments.

templates/shared_services/sonatype-nexus/terraform/webapp.tf
tamirkamara: yes :-(

templates/shared_services/sonatype-nexus/terraform/deploy.sh
marrobi: I've been using these for running the TF locally using the make file. They should probably be consolidated into a single script used by all, outside of the terraform directories.

templates/shared_services/sonatype-nexus/terraform/webapp.tf
marrobi: Not sure do need them?

templates/shared_services/sonatype-nexus/terraform/webapp.tf
```diff
@@ -0,0 +1,193 @@
...
+    retention_policy {
+      enabled = false
+    }
+  }
+}
+
+resource "azurerm_storage_share" "nexus" {
+  name = "nexus-data"
+  storage_account_name = data.azurerm_storage_account.nexus.name
+  quota                = 1024 # 1TB
```
marrobi, line 192: Can see this being hit if we actually mirror anything rather than proxy. Might be worth adding an issue to track/make a variable so can easily be increased.

marrobi (approved):
Couple of comments, but looks good, can't wait to try it :)

templates/shared_services/sonatype-nexus/terraform/webapp.tf
tamirkamara: Added a var

templates/shared_services/sonatype-nexus/terraform/webapp.tf
tamirkamara: removed this as it's not is use

templates/shared_services/sonatype-nexus/terraform/deploy.sh
tamirkamara: Removed from here in the meantime

Issue #504, opened 2021-07-27
[Story] Create target resource service - template and bundle (vm)


adishachar:
## Description

As a TRE administrator
I want to have a target resource service vm template and a target resource vm  service bundle
So that I can use it to register a target resource vm service template for a TRE workspace.

## Acceptance critiera

- [ ] Criteria 1 - Given a Register API for target resource service, TRE administrator can register target resource vm service template to a TRE workspace

PR #530, opened 2021-08-01
Adding a guacamole user resource service template (a Windows DSVM)

eladiw:
#504 

This PR introduces a new service ( a target resource workspace service). This is a Win DSVM workspace service to be used with the guacamole workspace service

## How is this addressed

- Added TF scripts
- Added porter definitions
- Added to the TRE networking (TF) a new subnet for VMs


LizaShak (approved):

workspaces/services/guacamole-azure-dsvm/terraform/main.tf
```diff
@@ -0,0 +1,33 @@
...
+
+data "azurerm_client_config" "current" {}
+
+data "azurerm_resource_group" "ws" {
+  name = "rg-${var.tre_id}-ws-${var.workspace_id}"
+}
+
+data "azurerm_virtual_network" "ws" {
+  name                = "vnet-${var.tre_id}-ws-${var.workspace_id}"
+  resource_group_name = "rg-${var.tre_id}-ws-${var.workspace_id}"
```
tamirkamara, line 26: can this reuse the data element of the rg defined right above instead of redefining the name?

workspaces/services/guacamole-azure-dsvm/terraform/main.tf
```diff
@@ -0,0 +1,33 @@
+# Azure Provider source and version being used
+terraform {
+  required_providers {
+    azurerm = {
+      source  = "hashicorp/azurerm"
+      version = "=2.61.0"
```
tamirkamara, line 6: Version...

workspaces/services/guacamole-azure-dsvm/terraform/dsvm.tf
```diff
@@ -0,0 +1,128 @@
...
+}
+
+resource "azurerm_network_security_rule" "deny-all-inbound-override-dsvm" {
+  access                      = "Deny"
+  destination_address_prefix  = "*"
+  destination_port_range      = "*"
+  direction                   = "Inbound"
+  name                        = "deny-inbound-override"
+  network_security_group_name = azurerm_network_security_group.dsvmnsg.name
+  priority                    = 900
```
tamirkamara, line 40: Lower number = higher priority, isn't it?
Doesn't this blocks the rule allowing access through port 3389?

workspaces/services/guacamole-azure-dsvm/terraform/dsvm.tf
```diff
@@ -0,0 +1,128 @@
...
+  source_address_prefix       = "*"
+  source_port_range           = "*"
+}
+
+resource "azurerm_network_interface_security_group_association" "dev_vm_nsg_association" {
+  network_interface_id      = azurerm_network_interface.internal.id
+  network_security_group_id = azurerm_network_security_group.dsvmnsg.id
+}
+
+resource "random_string" "username" {
```
tamirkamara, line 66: Do we need a random username if we know that only 1 user will use it?

workspaces/services/guacamole-azure-dsvm/terraform/dsvm.tf
```diff
@@ -0,0 +1,128 @@
...
+  min_special      = 1
+  override_special = "_%@"
+}
+
+resource "azurerm_virtual_machine" "dsvm" {
+  name                             = local.vm_name
+  location                         = data.azurerm_resource_group.ws.location
+  resource_group_name              = data.azurerm_resource_group.ws.name
+  network_interface_ids            = [azurerm_network_interface.internal.id]
+  vm_size                          = "Standard_DS1_v2"
```
tamirkamara, line 94: Seems very small machine for the use-case.

workspaces/services/guacamole-azure-dsvm/terraform/dsvm.tf
```diff
@@ -0,0 +1,128 @@
...
+  delete_os_disk_on_termination    = false
+  delete_data_disks_on_termination = false
+
+  storage_image_reference {
+    publisher = "microsoft-dsvm"
+    offer     = "dsvm-windows"
+    sku       = "server-2016"
+    version   = "latest"
```
tamirkamara, line 102: Looks like an older image. I think this is the newest:
                    "imageReference": {
                        "publisher": "microsoft-dsvm",
                        "offer": "dsvm-win-2019",
                        "sku": "server-2019-gen2",
                        "version": "latest"
                    }

workspaces/services/guacamole-azure-dsvm/terraform/dsvm.tf
```diff
@@ -0,0 +1,128 @@
...
+
+  os_profile_windows_config {
+  }
+
+  identity {
+    type = "SystemAssigned"
+  }
+
+  tags = {
+    parent_service_id = var.guacamole_parent_service_id
```
tamirkamara, line 126: Are you doing anything with this apart from the Tags?

workspaces/services/guacamole-azure-dsvm/terraform/dsvm.tf
eladiw: Thanks. good catch.

workspaces/services/guacamole-azure-dsvm/terraform/dsvm.tf
eladiw: one e2e will work we will upgrade. no reason to do it now

workspaces/services/guacamole-azure-dsvm/terraform/dsvm.tf
eladiw: no. its more of a 'todo' item for me to make sure that this bundle install does get the parent id .

workspaces/services/guacamole-azure-dsvm/terraform/main.tf
eladiw: done (and in all other services)

workspaces/services/guacamole-azure-dsvm/terraform/dsvm.tf
eladiw: talked offline

workspaces/services/guacamole-azure-dsvm/.env.sample
```diff
@@ -0,0 +1,2 @@
+WORKSPACE_ID=__CHANGE_ME__
+GUACAMOLE_PARENT_SERVICE_ID=__CHANGE_ME__
```
marrobi, line 2: is there a reason this cannot be just `SERVICE_ID` - that way it's generic for all workspace service resources?

workspaces/services/guacamole-azure-dsvm/porter.yaml
```diff
@@ -0,0 +1,83 @@
...
+      clientVersion: 0.15.4
+
+install:
+  - terraform:
+      description: "Deploy Guacamole Target Resource Service (Windows DSVM)"
+      vars:
+        workspace_id: "{{ bundle.parameters.workspace_id }}"
+        tre_id: "{{ bundle.parameters.tre_id }}"
+        arm_client_id: "{{ bundle.credentials.azure_client_id }}"
+        arm_client_secret: "{{ bundle.credentials.azure_client_secret }}"
```
marrobi, line 54: We should be using MSI for deploying. If look at https://github.com/microsoft/AzureTRE/blob/develop/workspaces/vanilla/porter.yaml - have an ARM_USE_MSI variable. nothing is being passed into the TF command.

This might pose a challenge if running locally?

workspaces/services/guacamole-azure-dsvm/readme.md
```diff
@@ -0,0 +1,27 @@
+# Guacamole Target Resource Service bundle (Windows DSVM)
+
+This is a Target Resource Service template. It contains a Windows DSVM to be used by TRE researchers and to be connected using a [Guacamole server](https://guacamole.apache.org/).
```
marrobi, line 3: Not sure "target resource" is the right term we should be using. It doesn't make sense to be. "Service Child Resource Template" maybe?!

workspaces/services/guacamole-azure-dsvm/terraform/deploy.sh
```diff
@@ -0,0 +1,8 @@
+export TF_LOG=""
+terraform init -input=false -backend=true -reconfigure \
+    -backend-config="resource_group_name=$TF_VAR_mgmt_resource_group_name" \
+    -backend-config="storage_account_name=$TF_VAR_mgmt_storage_account_name" \
+    -backend-config="container_name=$TF_VAR_terraform_state_container_name" \
+    -backend-config="key=${TF_VAR_tre_id}${TF_VAR_workspace_id}guacamolewindsvm"
```
marrobi, line 6: This should probably have parent service ID in it? Only used for running locally/testing at present, so don't think essential.

workspaces/services/guacamole-azure-dsvm/terraform/main.tf
```diff
@@ -0,0 +1,33 @@
...
+data "azurerm_resource_group" "ws" {
+  name = "rg-${var.tre_id}-ws-${var.workspace_id}"
+}
+
+data "azurerm_virtual_network" "ws" {
+  name                = "vnet-${var.tre_id}-ws-${var.workspace_id}"
+  resource_group_name = data.azurerm_resource_group.ws.name
+}
+
+data "azurerm_subnet" "dsvms" {
```
marrobi, line 29: Is there a reason for a new subnet? Can we just put them on the `ServicesSubnet` please? Otherwise each time someone adds a service VNET configurations will need changing and won't be possible to add the service to existing vnets.

workspaces/services/guacamole-azure-dsvm/terraform/dsvm.tf
```diff
@@ -0,0 +1,128 @@
...
+  resource_group_name = data.azurerm_resource_group.ws.name
+
+  ip_configuration {
+    name                          = "primary"
+    subnet_id                     = data.azurerm_subnet.dsvms.id
+    private_ip_address_allocation = "Dynamic"
+  }
+}
+
+resource "azurerm_network_security_group" "dsvmnsg" {
```
marrobi, line 13: Can it not reuse the workspace NSG - it allows all traffic within the `ServicesSubnet` (see below).

workspaces/services/guacamole-azure-dsvm/terraform/dsvm.tf
```diff
@@ -0,0 +1,128 @@
...
+  name                        = "deny-inbound-override"
+  network_security_group_name = azurerm_network_security_group.dsvmnsg.name
+  priority                    = 900
+  protocol                    = "*"
+  resource_group_name         = data.azurerm_resource_group.ws.name
+  source_address_prefix       = "*"
+  source_port_range           = "*"
+}
+
+resource "azurerm_network_security_rule" "deny-outbound-override-dsvm" {
```
marrobi, line 47: This will block all outbound access? Traffic needs to be allowed out - to the Internet in the NSG - but routed using the default route on the subnet to direct through the firewall which then restricts the traffic. As above would use the existing nsg for the workspace and services subnet.

workspaces/vanilla/terraform/network/locals.tf
```diff
@@ -1,6 +1,7 @@
 locals {
-  ws_services_vnet_subnets       = cidrsubnets(var.address_space, 1, 1)
+  ws_services_vnet_subnets       = cidrsubnets(var.address_space, 1, 1, 1)
```
marrobi, line 2: We need to try leave the vanilla (base) workspace generic.

marrobi (changes requested):
Thanks, good start, few things we should probably talk offline about.

workspaces/services/guacamole-azure-dsvm/readme.md
eladiw: Child-resource service template?

marrobi: Not sure the VM is a service. "User Resource Template" or "Service Resource Template" maybe.

workspaces/services/guacamole-azure-dsvm/porter.yaml
eladiw: added the arm_use_msi
I think for local use we can have arm_use_msi equal to false and supply the client id + secret

workspaces/services/guacamole-azure-dsvm/readme.md
eladiw: User Resource Template

workspaces/services/guacamole-azure-dsvm/terraform/dsvm.tf
eladiw: removed and moved to the ws nsg and subnet

workspaces/services/guacamole-azure-dsvm/porter.yaml
```diff
@@ -0,0 +1,87 @@
...
+    type: string
+    description: "The name of the Terraform state storage account"
+  - name: tfstate_container_name
+    env: tfstate_container_name
+    type: string
+    default: "tfstate"
+    description: "The name of the Terraform state storage container"
+  - name: arm_use_msi
+    env: ARM_USE_MSI
+    default: true
```
marrobi, line 43: Is this defaulting to true on the other bundles? I think it is false, unless it is explicitly set to false in the make file when running locally.  Don't mind either way as long as all consistent.

workspaces/services/guacamole-azure-dsvm/readme.md
marrobi: Lets stick with `User Resource Template` and `User Resource`. Agree?!

workspaces/services/guacamole-azure-dsvm/porter.yaml
eladiw: others are false. I thought true is needed as it usually will work with MSI. 
I will switch

marrobi (approved):
One comment around MSI, other than that looks good.

Issue #520, opened 2021-07-29
[Task] Disable default repos in Nexus


tamirkamara:
Story: #437 

**Description**
Nexus has a set of default repos defined which is a security problem as it enables different workspaces to share data.
We should investigate how to disable this if possible. A lead can be found here: https://community.sonatype.com/t/disable-the-creation-of-the-default-repositories/5261

PR #532, opened 2021-08-01
Nexus without default repositories

tamirkamara:
# PR for issue #520, #519

## What is being addressed

Nexus is normally created with a default set of repositories but those might be a problem since they allow data to be shared across workspaces. This PR changes that to have our Nexus created empty.

## How is this addressed

- Include a nexus.properties file to skip the default repositories and upload it before Nexus starts up.
- Add a readme for the service


marrobi (approved):
LGTM

PR #552, opened 2021-08-05
Add E2E tests for other bundles than vanilla.

### LINKED ISSUE #511: [Task] Add API tests to verify bundle registration and the deployment of all workspace bundles
Story: #445 

**Description**
Add API tests to verify bundle registration and the deployment of all workspace bundles (is currently just vanilla)

sachinkundu:
Mark tests as smoke and extended.

Modify pipeline to only run smoke tests. Other tests should be enabled when a good delete mechanism in in place.

Add an option to run with verify=false for testing purpose when using self signed TLS certificates.

Closes #511


marrobi (approved):
LGTM

Issue #348, opened 2021-06-22
[Task]  Configure Gitea registration preferences


marrobi:
Story: #544 

**Description**
- Need to restrict open registration to Gitea to prevent pushing of code.
- Ideally and admin user created automatically on deploy or AD integration for TRE administrators (only).

Useful - can be set using env vars: https://docs.gitea.io/en-us/config-cheat-sheet/#security-security

anatbal:
Hi @marrobi , I have a few questions about the task:
1. Is the admin user that supposed to manage the gitea instance is the TRE administrator? or can it also be a TRE owner?
2. Will it be okay to have a static username for the gitea admin (e.g gitea_admin)? Since an admin user cannot be created with an env var, we will have to create it separately one the gitea container is up.

Issue #542, opened 2021-08-03
[Task] Set Gitea admin user, password and email


anatbal:
Story: #544 

**Description**
- Need the user to define the specific username and password of the gitea admin

PR #559, opened 2021-08-05
Configure Gitea registration preferences

anatbal:
# PR for issue #348 

## What is being addressed

Configure Gitea registration preferences

## How is this addressed
- Block normal users registrations by GITEA__service__DISABLE_REGISTRATION in the gitea config
- Add an admin user in order for the TRE admin to be able to push git repositories


Makefile
```diff
@@ -2,7 +2,7 @@
 
 SHELL:=/bin/bash
 
-all: bootstrap mgmt-deploy build-api-image push-api-image build-resource-processor-vm-porter-image push-resource-processor-vm-porter-image tre-deploy
+all: bootstrap mgmt-deploy build-api-image push-api-image build-resource-processor-vm-porter-image push-resource-processor-vm-porter-image push-gitea-image tre-deploy
```
marrobi, line 5: For the other images we have a build and push stage - might use build if want to test locally without push - can we do the same for gitea?

templates/core/terraform/main.tf
```diff
@@ -255,6 +255,10 @@ module "gitea" {
   source   = "../../shared_services/gitea/terraform"
   tre_id   = var.tre_id
   location = var.location
+  management_api_image_tag =  var.management_api_image_tag
```
marrobi, line 258: nit - formatting (prevents churn later)

templates/shared_services/gitea/terraform/gitea-webapp.tf
```diff
@@ -10,8 +18,15 @@ resource "azurerm_app_service" "gitea" {
...
-    TRE_ID            = var.tre_id
-    RESOURCE_LOCATION = var.location
+     # Settings for private Container Registires  
+    "DOCKER_REGISTRY_SERVER_USERNAME"            = var.docker_registry_username
+    "DOCKER_REGISTRY_SERVER_URL"                 = "https://${var.docker_registry_server}"
+    "DOCKER_REGISTRY_SERVER_PASSWORD"            = var.docker_registry_password
+   
+    # TBD, Username and password should not be defined here: #542
+    GITEA_USERNAME                 = "giteaadmin"
+    GITEA_PASSWD                   = random_password.gitea_passwd.result
```
marrobi, line 28: Can we save the password in keyvault, as per https://github.com/microsoft/AzureTRE/blob/b80eb68503e5b2f4f5368f0a36e247ec6b3ffb6b/templates/core/terraform/resource_processor/vmss_porter/main.tf#L40

marrobi (approved):
Looks good, couple of suggestions that would be good to address as part of this PR or later.

templates/shared_services/gitea/terraform/gitea-webapp.tf
anatbal: @marrobi Sure, we are working on it as part of a different task (#542 ) just in order to make it in a separate PR

Makefile
anatbal: Fixed.

templates/core/terraform/main.tf
anatbal: Fixed.

templates/core/terraform/main.tf
```diff
@@ -221,10 +221,13 @@ module "bastion" {
...
 module "gitea" {
   count = var.deploy_gitea == true ? 1 : 0
-
-  source   = "../../shared_services/gitea/terraform"
-  tre_id   = var.tre_id
-  location = var.location
+  source                                     = "../../shared_services/gitea/terraform"
+  tre_id                                     = var.tre_id
+  location                                   = var.location
+  management_api_image_tag                   = var.management_api_image_tag
```
tamirkamara, line 227: need to change this "management_api" prefix. I think we can just remove it.

tamirkamara (changes requested):

tamirkamara (approved):
LGTM

Issue #550, opened 2021-08-04
[Task] Implement PATCH /api/workspaces to disable/enable workspaces


TessFerrandez:
Story: #547 

**Description**
Implement PATCH /api/workspaces to disable/enable workspaces

Issue #547, opened 2021-08-04
[Story] Add API functionality to disable/enable a workspace


TessFerrandez:
## Description

**As a TRE Administrator** 
I want to be able to disable/enable workspaces
So that I can control if they are accessible to workspace owners and researchers

## Acceptance critiera

- [x] Admin can disable a workspace - PATCH /api/workspaces/{workspace_id}/ -- posting { "enabled": False }
- [x] Admin can enable a workspace - PATCH /api/workspaces/{workspace_id}/ -- posting { "enabled": True }
- [x] disabled/enabled status should be visible on GET /api/workspaces/ and GET /api/workspaces/{workspace_id}
- [x] Workspace resources for disabled workspaces will still be accessible with direct URIs to VMs and through the Azure Portal

Tasks:

- [x] #550

PR #564, opened 2021-08-06
add API functionality to enable/disable workspaces

TessFerrandez:
# PR for issue #550 and story #547 

## What is being addressed

- API User can enable/disable workspace by sending a PATCH /api/workspaces/{workspace_id} - {"enabled": false}

## How is this addressed

- Add new property to the workspace (and workspace.json) called enabled - so this can be set also on create -- by default enabled=True
- Removed workspace parameters (display name, description, enabled) from the resource template parameters when stored to db - to avoid duplicate data
- Add new route PATCH /api/workspaces/{workspace_id}
- Tests to validate new functionality

## Additional

- Girl scout cleanup of other test names and ordering of imports - this is just cosmetic and can be ingored


management_api_app/models/schemas/workspace.py
```diff
@@ -88,3 +93,14 @@ class Config:
                 "workspaceId": "49a7445c-aae6-41ec-a539-30dfa90ab1ae",
             }
         }
+
+
+class WorkspacePatch(BaseModel):
```
sachinkundu, line 98: Is this class itself going to get extended when we add new fields which can be patched? If not then we could rename this WorkspaceEnabledPatch. It could read better and if we add new patch classes existing tests will not need changing everytime we add new fields.

TessFerrandez: @sachinkundu My plan was for it to be extended. This is the schema for the general patch.

If you want to have different WorkspaceEnabledPatch and WorkspaceDescriptionPatch, you would need to have different PATCH endpoints, is that what you envision? i.e. that we should instead have PATCH /api/workspace/{workspace_id}/enabled? or can you explain a bit more how you envision this working?

sachinkundu: Ah right. That's a bummer. I would like to have a single endpoint for patch but a more nuanced payload control. Is it possible to have a Union type? That is payload is of type WorkspacePatchEnabled or WorkspacePatchDisplayname

If not then go ahead.

TessFerrandez: Looks like you can union - so I have changed it to WorkspacePatchEnabled for now and will union once there is a need for it

sachinkundu (approved):

PR #566, opened 2021-08-08
Gitea reference passwords from keyvault

anatbal:
# PR for issue #542 

## What is being addressed

The passwords for both gitea_db and giteeadmin user are being refrenced to keyvault instead of in the app configuration.
How to access the user and password is being documented in the README. 

## How is this addressed

- The gitea admin user has the default name of "giteaadmin" and the password is being saved in keyvault.
- The same for the gitea database


tamirkamara (approved):
LGTM

Issue #347, opened 2021-06-22
[Task] Configure Gitea to connect to mySQL via SSL


marrobi:
Story: #544 

**Description**
Gitea has limited TLS options and I have been unable to get Gitea to connect to mySQL using `SSL_MODE=skip-verify`.

The docs example provisions a custom PEM as root CA. https://docs.microsoft.com/en-us/azure/mysql/howto-configure-ssl#golang

Can we get mysql working, or maybe another DB server type?

PR #573, opened 2021-08-08
Configure Gitea to connect to mySQL via SSL

anatbal:
# PR for issue #347 

## What is being addressed
 
Enabling Gitea to connect with SSL to the mysql database using the right certificate.
Additional changes regarding better connection to the acr - we now use managed identity to fetch the docker image.

## How is this addressed

- Enabling the ssl connection in Gitea
- Updating the Gitea image with the certificate


tamirkamara (commented):
Looks good @anatbal.
Can you please link the issue and also add an "additional" item in the description to note that we now use managed identity to fetch the docker image?
I'll let someone else approve since I co-authored this.

marrobi (approved):
Providing all works, LGTM!

Issue #502, opened 2021-07-26
[Story] Add Auth extension for Guacamole


eladiw:
## Description

As a TRE developer 
I want to be able to use a custom guacamole image with a custom Auth extension
So that when users use this guacamole server the authentication against kv will be done seamlessly

This user story will be written in three parts:
1. The authentication without password injection
2. add password injection
3. handle session timeouts.
4. 
## Acceptance critieria

- [ ] Given a vm's user credentials stored in kv, Guacamole with the custom extension can authenticate and log into the vm using the kv credentials

PR #577, opened 2021-08-08
Guacamole auth extension draft

eladiw:
# PR for issue #502 

## What is being addressed

Adding the auth extension that allows to authorize and log into the remote vm

## How is this addressed

1. Created the Auth extension. including the dockerfile change. it will be installed implicitly with the Guacamole bundle

## Todos:
1. Replace mock calls to 'get vms' and 'get vm creds' with actual calls to the upcoming api calls.
2. implement a token refresh mechanism 

## How to test:

1. Install the bundle
2. Add the 3 temp values in the app service config section
TEMP_USERNAME
TEMP_PASSWORD
TEMP_HOSTNAME


marrobi (approved):
As a draft, looks fine, I suggest when we get a working version in we get a Java expert to review it.

PR #586, opened 2021-08-10
upgrading all hcl lock files + script

eladiw:
# PR for issue

## What is being addressed

Lock files are out of sync. this fix addresses it.

## How is this addressed
Ran a script to upgdate all lock files


marrobi (approved):
Nice solution.

PR #587, opened 2021-08-10
Add API functionality to add a workspace-service to an existing workspace

### LINKED ISSUE #553: [Story] Add API functionality to add a workspace-service to an existing workspace
## Description

As a TRE WorkspaceOwner
I want to add a new workspace-service to an existing workspace
So that I can work with the service (ex. Guacamole)

## Acceptance criteria

- [ ] Only workspace owners can add new workspace-services to the workspace
- [ ] The workspace-service links to a matching parent workspace
- [ ] WorkspaceOwner adds the workspace-service using POST /api/workspaces/{workspace_id}/workspace-services and a payload like the following

```json
{
  "workspaceServiceType": "guacamole",
  "properties": {}
}
```
- [ ] The request body is validated against the template
- [ ] POST returns { "workspaceServiceId": "<GUID>" } if successful or errors if not

# Implementation

- [x] #540

LizaShak:
# PR for issue

## What is being addressed

Resolves #553 by adding API functionality to add a workspace-service to an existing workspace

## How is this addressed

- Add new route for workspace service installation
- Handle deployment status update for both workspaces and workspace services


management_api_app/api/routes/workspaces.py
```diff
@@ -55,6 +57,42 @@ async def create_workspace(workspace_create: WorkspaceInCreate, workspace_repo:
...
+                                   ) -> WorkspaceServiceIdInResponse:
+    access_service = get_access_service()
+    if access_service.get_workspace_role(user, workspace) != WorkspaceRole.Owner:
+        raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail=strings.ACCESS_USER_IS_NOT_OWNER)
+
+    try:
+        workspace_service = workspace_service_repo.create_workspace_service_item(workspace_create, workspace.id)
+    except (ValidationError, ValueError) as e:
+        logging.error(f"Failed create workspace service model instance: {e}")
+        raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=str(e))
```
TessFerrandez, line 76: ideally we don't want to return the error directly to the user (as it can contain implementation specific details that could be used for security breaches) - do we know that this error is safe to return? I can see that it is done the same way in the workspaces api so depending on the answer, please change in both places.

management_api_app/api/routes/workspaces.py
```diff
@@ -55,6 +57,42 @@ async def create_workspace(workspace_create: WorkspaceInCreate, workspace_repo:
...
+        workspace_service_repo.save_workspace_service(workspace_service)
+    except Exception as e:
+        logging.error(f"Failed save workspace service instance in DB: {e}")
+        raise HTTPException(status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
+                            detail=strings.STATE_STORE_ENDPOINT_NOT_RESPONDING)
+
+    try:
+        await send_resource_request_message(workspace_service)
+    except Exception as e:
+        # TODO: Rollback DB change, issue #154
```
TessFerrandez, line 88: I wasn't a big fan of this TODO in the workspaces API :) - rather than a TODO - can we delete the created item from the DB here?

management_api_app/db/repositories/resources.py
```diff
@@ -0,0 +1,26 @@
+from azure.cosmos import CosmosClient
+from pydantic import UUID4
+
+from core import config
+from db.errors import EntityDoesNotExist
+from db.repositories.base import BaseRepository
+from jsonschema import validate
```
TessFerrandez, line 7:
```suggestion
from azure.cosmos import CosmosClient
from jsonschema import validate
from pydantic import UUID4

from core import config
from db.errors import EntityDoesNotExist
from db.repositories.base import BaseRepository
```

management_api_app/db/repositories/resources.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class ResourceRepository(BaseRepository):
+    def __init__(self, client: CosmosClient):
+        super().__init__(client, config.STATE_STORE_RESOURCES_CONTAINER)
+
+    @staticmethod
+    def _validate_resource_parameters(workspace_create, workspace_template):
```
TessFerrandez, line 15: since this is a generic resource - should we have resource_create and resource_template?

management_api_app/db/repositories/resources.py
```diff
@@ -0,0 +1,26 @@
...
+        validate(instance=workspace_create["properties"], schema=workspace_template)
+
+    def get_resource_dict_by_id(self, workspace_id: UUID4) -> dict:
```
TessFerrandez, line 18: This doesn't look like a get_resource_dict_by_id but rather like a get_workspace_dict_by_id -- only time this makes sense is if it is a workspace (or should it say resource_id)

management_api_app/db/repositories/resources.py
```diff
@@ -0,0 +1,26 @@
...
+        query = self._active_workspaces_query() + f' AND c.id="{workspace_id}"'
```
TessFerrandez, line 19: the ResourceRepository class does not appear to have a self._active_workspaces_query() -- and if it should have, it should probably be called _active_resources_query()

the rest of this function also heavily references workspaces.  If all of this is workspace specific perhaps it is better to leave it in the sub class

management_api_app/models/domain/resource.py
```diff
@@ -40,3 +40,12 @@ class Resource(AzureTREModel):
     resourceTemplateParameters: dict = Field({}, title="Resource template parameters", description="Parameters for the deployment")
     deployment: Deployment = Field(Deployment(status=Status.NotDeployed, message=""), title="Deployment", description="Fields related to deployment of this resource")
     deleted: bool = Field(False, title="Is deleted", description="Marks the resource request as deleted (NOTE: this is not the deployment status)")
+
+    def get_resource_request_message_payload(self) -> dict:
```
TessFerrandez, line 44: why is this moved here instead of in the service bus code? This code already exists there -- or is it because it is different depending on type? 

It would be great if we can keep this in the service bus class - otherwise we will need to duplicate code for this for other actions etc. (I am working with this specific thing in delete -- see the branch I shared with you earlier) - it doesn't feel like the repositories responsibility.

management_api_app/models/schemas/workspace.py
```diff
@@ -24,7 +24,7 @@ def get_sample_workspace(workspace_id: str, spec_workspace_id: str = "0001") ->
             "status": "not_deployed",
             "message": "This resource is not yet deployed"
         },
-        "isDeleted": False,
+        "deleted": False,
         "resourceType": "workspace",
```
TessFerrandez, line 28: I know this is not in the changed code, but as you are changing adjacent code - we should replace workspace by the enum (so that we change it only in one place when we change it)

management_api_app/models/schemas/workspace_service.py
```diff
@@ -0,0 +1,59 @@
...
+        "resourceTemplateParameters": {
+            "displayName": "my workspace service",
+            "description": "some description",
+         },
+        "deployment": {
+            "status": "not_deployed",
+            "message": "This resource is not yet deployed"
+        },
+        "deleted": False,
+        "resourceType": "workspace-service",
```
TessFerrandez, line 21: same here -- let's use the enum (and the same applies for status above) -- less places to change when we change the "magic string"

management_api_app/models/schemas/workspace_service.py
```diff
@@ -0,0 +1,59 @@
+from pydantic import BaseModel, Field
+
+from models.domain.workspace import Workspace
+
+
+def get_sample_workspace(workspace_id: str, spec_workspace_id: str = "0001") -> dict:
```
TessFerrandez, line 6: is this supposed to be get_sample_workspace_service? and workspace_service_id?
The spec_workspace_id doesn't seem to be used

management_api_app/models/schemas/workspace_service.py
```diff
@@ -0,0 +1,59 @@
...
+            "message": "This resource is not yet deployed"
+        },
+        "deleted": False,
+        "resourceType": "workspace-service",
+        "workspaceURL": "",
+        "authInformation": {}
+    }
+
+
+class WorkspaceInResponse(BaseModel):
```
TessFerrandez, line 27: Is this WorkspaceServiceInResponse? - and workspace_service below?

management_api_app/models/schemas/workspace_service.py
```diff
@@ -0,0 +1,59 @@
...
+        schema_extra = {
+            "example": {
+                "workspace": get_sample_workspace("933ad738-7265-4b5f-9eae-a1a62928772e")
+            }
+        }
+
+
+class WorkspaceServiceInCreate(BaseModel):
+    workspaceServiceType: str = Field(title="Workspace service type", description="Bundle name")
+    properties: dict = Field({}, title="Workspace service parameters", description="Values for the parameters required by the workspace resource specification")
```
TessFerrandez, line 40:
```suggestion
    properties: dict = Field({}, title="Workspace service parameters", description="Values for the parameters required by the workspace service resource specification")
```

management_api_app/service_bus/deployment_status_update.py
```diff
@@ -56,7 +56,7 @@ async def receive_message():
                         await receiver.complete_message(msg)
 
 
-def create_updated_deployment_document(workspace: Workspace, message: DeploymentStatusUpdateMessage):
+def create_updated_deployment_document(workspace: dict, message: DeploymentStatusUpdateMessage):
```
TessFerrandez, line 59: why are we changing to a dict here? why not a resource? -- this used to return a workspace - but now it returns a dict and the callers expect a workspace/resource type -- the class should also have a return type

management_api_app/service_bus/deployment_status_update.py
```diff
@@ -66,10 +66,10 @@ def create_updated_deployment_document(workspace: Workspace, message: Deployment
     Returns:
         [Workspace]: Workspace with the deployment sub doc updated
```
TessFerrandez, line 67: see comment above - function no longer returns a workspace

management_api_app/service_bus/deployment_status_update.py
```diff
@@ -86,8 +86,8 @@ def update_status_in_database(workspace_repo: WorkspaceRepository, message: Depl
     result = False
 
     try:
-        workspace = workspace_repo.get_workspace_by_workspace_id(message.id)
-        workspace_repo.update_workspace(create_updated_deployment_document(workspace, message))
+        workspace = workspace_repo.get_resource_dict_by_id(message.id)
```
TessFerrandez, line 89: It is not super ideal that we now have to return less specific items because we are making this very generic -- we are missing a lot of pydantic validation here and class.field access -- perhaps the WorkspaceRepository could still have a get_workspace_by_workspace_id that would call get_resource_dict_by_id and parse_obj_as Workspace on return?

management_api_app/service_bus/deployment_status_update.py
```diff
@@ -86,8 +86,8 @@ def update_status_in_database(workspace_repo: WorkspaceRepository, message: Depl
...
+        workspace_repo.update_resource_dict(create_updated_deployment_document(workspace, message))
```
TessFerrandez, line 90: same comment as above

TessFerrandez (changes requested):
In general, looks good

Need to change some copy/paste issues and some suggestions on how we can avoid making the code less specific only to reuse code. 

Tip: we have pre-commit hooks that will catch the linting errors on commit so you can catch them before PR

management_api_app/service_bus/deployment_status_update.py
LizaShak: this is due to the fact that now we get deployment status update both for workspace and workspace service (and soon user resource) those objects have different properties. And to reuse the existing  code I had to make it generic.

management_api_app/service_bus/deployment_status_update.py
LizaShak: same here , because it can be workspace and workspace service (and user resource in the future) ... we want to retrieve it and update the deployment properties

management_api_app/api/routes/workspaces.py
LizaShak: Those are ValidationError and ValueError telling the user about missing parameters and parameter value errors...  I think it is ok to return them, they inform the user about missing/invalid information that the template requires, and I think it is safe, because the user can call the template api by name and see what are the required parameters that are set there.

TessFerrandez: cool with me

management_api_app/service_bus/deployment_status_update.py
TessFerrandez: I had a suggestion for a solution, do you feel that this would not work?

management_api_app/service_bus/deployment_status_update.py
TessFerrandez: per our talk earlier we okd the dict - but should it not be resource then instead of workspace?

TessFerrandez (changes requested):
still have a few unresolved comments from earlier and one new - but know you said you resolved all - did you miss a commit perhaps?

management_api_app/models/domain/resource.py
```diff
@@ -46,3 +46,12 @@ def is_enabled(self) -> bool:
         if "enabled" not in self.resourceTemplateParameters:
             return True     # default behavior is enabled = True
         return self.resourceTemplateParameters["enabled"] is True
+
+    def get_resource_request_message_payload(self, action: str) -> dict:
```
TessFerrandez, line 50: should action be an enum - RequestAction as in the original merged code? so we can control what actions are possible

TessFerrandez (changes requested):

management_api_app/db/repositories/resources.py
TessFerrandez: Thanks for fixing the function name -- can we change workspace_id to be resource_id for consistency?

TessFerrandez (approved):

PR #593, opened 2021-08-11
API to use Managed Identity to authenticate against Cosmos DB

### LINKED ISSUE #345: [Task] API to use Managed Identity to authenticate against Cosmos DB
**Is your feature request related to a problem?

Currently the API use an API key `STATE_STORE_KEY` to authenticate against Cosmos DB. The API uses Managed Identity to authenticate against Azure Service Bus.

**Describe the solution you'd like

Using the same simplified access pattern to authenticate from the API to Azure resource.

Note: looks like Cosmos data-plane doesn't really support managed identity (in Python at least) but there might be a way to do this:
https://github.com/MicrosoftDocs/azure-docs/issues/29762
https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/tutorial-linux-vm-access-cosmos-db

deniscep:
# API to use Managed Identity to authenticate against Cosmos DB

## What is being addressed

Switched API authentication to CosmosDB from using key stored in config to actively pulling key via mgmt API with MSI credentials.

## How is this addressed

- Added DefaultCredenital auth to database.py
- Added needed variables to TF (subscription_id, resource_group_name, cosmosdb_account_name)
- Removed key passing from TF to app config
- Modified docs where needed.

Closes #345

### DIFF
--- a/management_api_app/.env.sample
+++ b/management_api_app/.env.sample
@@ -17,7 +17,9 @@ TRE_ID=mytre-dev-3142                           # The Azure TRE instance name - 
 # State store configuration
 # -------------------------
 STATE_STORE_ENDPOINT=https://localhost:8081     # The Cosmos DB endpoint - keep localhost if using an emulator. Otherwise https://<your_cosmos_db>.documents.azure.com:443/
-STATE_STORE_KEY=__CHANGE_ME__                   # The Cosmos DB access key
+COSMOSDB_ACCOUNT_NAME=__CHANGE_ME__            # The Cosmos DB account name
+SUBSCRIPTION_ID=__CHANGE_ME__                   # The subscription id where Cosmos DB is located
+RESOURCE_GROUP_NAME=__CHANGE_ME__               # The resource group name where Cosmos DB is located
 
 # Service bus configuration
 # -------------------------
--- a/management_api_app/.env.sample
+++ b/management_api_app/.env.sample
@@ -17,6 +17,7 @@ TRE_ID=mytre-dev-3142                           # The Azure TRE instance name - 
 # State store configuration
 # -------------------------
 STATE_STORE_ENDPOINT=https://localhost:8081     # The Cosmos DB endpoint - keep localhost if using an emulator. Otherwise https://<your_cosmos_db>.documents.azure.com:443/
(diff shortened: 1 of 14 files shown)


management_api_app/.env.sample
```diff
@@ -17,7 +17,9 @@ TRE_ID=mytre-dev-3142                           # The Azure TRE instance name -
 # State store configuration
 # -------------------------
 STATE_STORE_ENDPOINT=https://localhost:8081     # The Cosmos DB endpoint - keep localhost if using an emulator. Otherwise https://<your_cosmos_db>.documents.azure.com:443/
-STATE_STORE_KEY=__CHANGE_ME__                   # The Cosmos DB access key
+COSMOSDB_ACCOUNT_NAME=__CHANGE_ME__            # The Cosmos DB account name
```
TessFerrandez, line 20: Does this mean we are now requiring an Azure based cosmos? so emulator won't work anymore?

deniscep: Good question, the way that the key retrieval works it uses the RG name and cosmos db name to get the keys. I'm not sure same can be queried against a local instance of cosmosdb. Should we add to the local env template the store_key, and if it exists, it uses it instead of the getting it from credentials?

deniscep: @TessFerrandez fixed now

management_api_app/services/health_checker.py
```diff
@@ -10,11 +12,24 @@ def create_state_store_status() -> (StatusEnum, str):
...
     except exceptions.ServiceRequestError:
         status = StatusEnum.not_ok
         message = strings.STATE_STORE_ENDPOINT_NOT_RESPONDING
     except:     # noqa: E722 flake8 - no bare excepts
         status = StatusEnum.not_ok
         message = strings.UNSPECIFIED_ERROR
     return status, message
+
+
+def get_store_key() -> str:
```
TessFerrandez, line 26: This looks like a dupe of the same function in management_api_app/api/dependencies/database.py - can we remove this and import that one instead? or did I miss some difference?

TessFerrandez (approved):
@deniscep looks good - but added a suggestion to reuse code if you can take a look before merge

management_api_app/services/health_checker.py
deniscep: true, fixed

PR #594, opened 2021-08-11
Add API functionality to register a new user-resource template

### LINKED ISSUE #482: [Story] Add API functionality to register a new user-resource template
## Description

**As a TRE Admin**
I want to be able to register a template for a new user-resource
so that TRE Researchers can add them to their workspaces

(a user-resource requires a specific workspace-service, ex. **Guacamole VM** that can be deployed in a **Guacamole Service**)

## Acceptance criteria

- [ ] **TRE Admin** can register a template by calling POST /api/service-templates/{template_name}/user-resource-templates -- with a workload like the following

```json
{
  "name": "guacamole-vm",
  "parentWorkspaceService": "guacamole-service",
  "version": "0.0.1",
  "current": "true",
  "json_schema": {
    "$schema": "http://json-schema.org/draft-07/schema",
    "$id": "https://github.com/microsoft/AzureTRE/templates/workspaces/myworkspace/user-resource.json",
    "type": "object",
    "title": "My User-Resource Template Custom Parameters",
    "description": "These parameters are specific to my user-resource template",
    "required": ["vm_size"],
    "properties": {
      "vm_size": {
        "$id": "#/properties/vm_size",
        "type": "string",
        "title": "VM size",
        "description": "Size of the VM",
        "default": "Standard_A1",
        "enum": [
          "Standard_A1",
          "Standard_A2",
          "Standard_A3"        ]
      }
    }
  }
}
```

- [ ] Only admins can access the API
- [ ] Fails if same name/version already exists
- [ ] If current: true - sets this to current - and removes any previous current version
- [ ] Enriched user-resource schema is returned
- [ ] API is listed in swagger as register
- [ ] Swagger documentation, and post defaults apply to the template
- [ ] user-resource.json created

Technical Note:
- GET will be added in a separate story

LizaShak:
# PR for issue

## What is being addressed

Resolves #482 by adding API functionality to register a new user-resource template

## How is this addressed

- Add api for new user-resource template
- Validate the parent workspace service template the user resource is created under


management_api_app/api/routes/workspace_service_templates.py
```diff
@@ -29,3 +34,21 @@ async def create_workspace_service_template(
...
+async def register_user_resource_template(
+        user_resource_template_create: UserResourceTemplateInCreate,
+        user_resource_template_repo: UserResourceTemplateRepository = Depends(get_repository(UserResourceTemplateRepository)),
+        workspace_service_template: ResourceTemplate = Depends(get_workspace_service_template_by_name_from_path)
+) -> UserResourceTemplateInResponse:
+    try:
+        template_created = create_user_resource_template(user_resource_template_create,
+                                                         user_resource_template_repo,
+                                                         workspace_service_template.name)
+        enriched_template = enrich_workspace_service_schema_defs(template_created)
```
TessFerrandez, line 51: should we not save the enriched version in the database? - or do we always have to enrich on retrieval?

TessFerrandez (approved):
LGTM - just the one question

management_api_app/api/routes/workspace_service_templates.py
LizaShak: We always enrich on retrieval

sachinkundu: we don't have to but we do. We can put an lru cache on it if speed is of essence. Don't think is should go to the DB because there is no enriched model in the domain. We can create one ofcourse but it does not have a physical meaning as the enrichment if purely for implementation needs.

Issue #575, opened 2021-08-08
[Task] Verify docker credentials are not passed/stored


tamirkamara:
Story: #345 

**Description**
Whenever possible we should pull docker images with managed-identities instead for the docker registry admin credentials.
Check out gitea shared service for an example (not of the difference in managed identity type)

PR #609, opened 2021-08-12
Removing docker registry username and pass from API config

deniscep:
# Removing docker registry username and pass from API config

## What is being addressed

Docker credentials are pushed to API, but have already been replaced by MSI downstream. Removing the credentials.

## How is this addressed

- Removed the credentials from TF so they are not pushed to API config

Closing #575


TessFerrandez (approved):
LGTM

Issue #611, opened 2021-08-12
[Task] Implement GET /api/workspace-service-templates/{template_name}/user-resource-templates


TessFerrandez:
Story: #558 

**Description**
Implement GET /api/workspace-service-templates/{template_name}/user-resource-templates

PR #614, opened 2021-08-12
implement get user resource templates by service type

TessFerrandez:
# PR for issue #611 

## What is being addressed

Implements API for getting the user resource templates that can be created for a given workspace service

## How is this addressed

- Implement API
- Implement related tests


sachinkundu (approved):
we have to figure out if we are doing one line or multi line arguments :)

TessFerrandez:
> we have to figure out if we are doing one line or multi line arguments :)

indeed :)

PR #628, opened 2021-08-15
Remove redundant variables from Guacamole and fetch image with managed identity

### LINKED ISSUE #617: [Task] Guacamole - ACR name should be fetched from mgmt acr
Story: #483 

**Description**
ACR name should be fetched without explicitly specifying

### LINKED ISSUE #618: [Task] Guacamole - acr_use_managed_identity_credentials should be used
Story: #483 

**Description**
azurerm 2.71 now supports acr_use_managed_identity_credentials. it should be used

### LINKED ISSUE #619: [Task] register-bundle doesn't work for guacamole
Story: #483 

**Description**
When running make register-bundle for guacamole / guacamole-vm the script fails because it doesn't account for the different directory.

tamirkamara:
# Resolves #617, Resolves #618, Resolves #619

## What is being addressed

- Instead of passing the ACR name, we get it from the mgmt acr object
- removed other irrelevant params
- Allow make register-bundle to work for any directory

### DIFF
--- a/workspaces/services/guacamole/terraform/locals.tf
+++ b/workspaces/services/guacamole/terraform/locals.tf
@@ -7,11 +7,7 @@ resource "random_string" "unique_id" {
 locals {
   service_id                   = random_string.unique_id.result
   service_resource_name_suffix = "${var.tre_id}-ws-${var.workspace_id}-svc-${local.service_id}"
+  webapp_name                  = "guacamole-${local.service_resource_name_suffix}"
   core_vnet                    = "vnet-${var.tre_id}"
   core_resource_group_name     = "rg-${var.tre_id}"
-}
-
-data "azurerm_container_registry" "mgmt_acr" {
-  name                = var.acr_name
-  resource_group_name = var.mgmt_resource_group_name
 }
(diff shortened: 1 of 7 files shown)


workspaces/services/guacamole/terraform/locals.tf
```diff
@@ -7,11 +7,7 @@ resource "random_string" "unique_id" {
 locals {
   service_id                   = random_string.unique_id.result
   service_resource_name_suffix = "${var.tre_id}-ws-${var.workspace_id}-svc-${local.service_id}"
+  webapp_name                  = "guacamole-${local.service_resource_name_suffix}"
```
anatbal, line 10: Why put webapp name in locals and not just state it in the web_app.tf? 
If we take the azureml service for reference it's not the convention there

anatbal (approved):
Looks good, only one question

workspaces/services/guacamole/terraform/locals.tf
tamirkamara: It's just a "shorthand" since we keep using it in many places then I define it once in locals and then is able to reference it everywhere I need to.

PR #633, opened 2021-08-15
Fix resource processor duplicate logs

### LINKED ISSUE #627: [BUG] Resource processor outputs duplicate log items
**Describe the bug**
The VMSS resource processor outputs many duplicate log items (some twice, some 3 times, 4, etc.)

**Steps to reproduce**

1. Make a query in app insights and look for rolename=runner.py
2. Make actions that require the RP to do something - create a workspace for instance
3. See that some of the logs entries appear multiple times

tamirkamara:
# Closes #627 

## What is being addressed

Resource processor generated multiple entries of the same logged item.

## How is this addressed

- Avoid calling initialize_logging multiple times which is actually adding multiple handlers on the default logger resulting in the duplicate outputs-


sachinkundu (approved):
Nice!

PR #635, opened 2021-08-15
Adding Keyvault for Guacamole service

anatbal:
# PR for issue #630 #631

## What is being addressed

In order to implement credentials retrieval in Guacamole auth extension, we need a dedicated keyvault and for the credentials of VMs to be stored in the keyvault.

## How is this addressed

- Adding keyvault in terraform for the guacamole service
- Adding both access policies for the deploy processor and the guacamole managed identity
- Saving the vm username + password in the keyvault

### DIFF
--- a/workspaces/services/guacamole-azure-win10vm/terraform/main.tf
+++ b/workspaces/services/guacamole-azure-win10vm/terraform/main.tf
@@ -38,6 +38,6 @@ data "azurerm_subnet" "services" {
 }
 
 data "azurerm_key_vault" "kv" {
-  name                = "kv-${var.tre_id}-${var.workspace_id}"
+  name                = "kv-guacamole-${var.tre_id}-${var.workspace_id}"
   resource_group_name = data.azurerm_resource_group.ws.name
 }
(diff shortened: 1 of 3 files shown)


workspaces/services/guacamole-azure-win10vm/terraform/main.tf
```diff
@@ -36,3 +36,8 @@ data "azurerm_subnet" "services" {
   virtual_network_name = data.azurerm_virtual_network.ws.name
   resource_group_name  = data.azurerm_resource_group.ws.name
 }
+
+data "azurerm_key_vault" "kv" {
+  name                = "kv-${var.tre_id}-${var.workspace_id}"
```
tamirkamara, line 41: Looks like this isn't the same name you use when creating the resource.

tamirkamara (changes requested):

workspaces/services/guacamole-azure-win10vm/terraform/main.tf
anatbal: Fixed

tamirkamara (approved):

PR #640, opened 2021-08-16
Changes e2e get workspace template test to match new JSON from API

deniscep:
# PR for issue

## What is being addressed

Get Workspace template used to return templateNames json with a list of template names, when now it returns a name-description pair of templates.

## How is this addressed

- Switched JSON reading to match new JSON format

### DIFF
--- a/e2e_tests/test_workspace_creation.py
+++ b/e2e_tests/test_workspace_creation.py
@@ -101,9 +101,8 @@ async def test_get_workspace_templates(template_name, token, verify) -> None:
             f"https://{config.TRE_ID}.{config.RESOURCE_LOCATION}.cloudapp.azure.com{strings.API_WORKSPACE_TEMPLATES}",
             headers=headers)
 
-        all_templates = [value for item in response.json()["templates"]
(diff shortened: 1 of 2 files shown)


e2e_tests/test_workspace_creation.py
```diff
@@ -100,8 +100,10 @@ async def test_get_workspace_templates(template_name, token, verify) -> None:
         response = await client.get(
             f"https://{config.TRE_ID}.{config.RESOURCE_LOCATION}.cloudapp.azure.com{strings.API_WORKSPACE_TEMPLATES}",
             headers=headers)
-
-        assert (template_name in response.json()["templateNames"]), f"No {template_name} template found"
+        
+        all_templates = [value for item in response.json()["templates"]
```
TessFerrandez, line 104: we should probably only check against names

TessFerrandez (approved):

e2e_tests/test_workspace_creation.py
deniscep: fixed

Issue #639, opened 2021-08-16
[Task] Refactor API code


TessFerrandez:
Story: #558 

**Description**
- Clean up code style
- Make template/resource code more DRY (lots of repeated code)
- Clean up template fixtures etc. and add missing tests
- Clean up repositories for templates and resources (names etc. that didn't get changed in copies)

PR #646, opened 2021-08-16
Refactor APIs

TessFerrandez:
# PR for issue #639 

Apologies for the large PR - most changes are readability (variable names, formatting etc.) which causes it to look larger than it is.

## What is being addressed

- Code style
- Duplicate code
- Combining resources to make code more readable/maintainable
- Clean-up in tests

Only functional change is to fix GET /.../user-resource-templates to require a logged in user
All other changes are non-functional

## How is this addressed

- Single-lining functions and function calls (for readability) - code style
- Moving scattered functions to repositories (ResourceTemplateRepository)
- Combining copy/pasted code to single functions with parameters to keep the code DRY
- Cleaning up functions in repositories so that common db functionality is moved to the base repository
- Reducing function names in ResourceTemplateRepository to their essentials to make the code more readable/shorter
- Cleaning up code in ResourceTemplateRepository that is now general (applies to all ResourceTypes) so that it doesn't refer to specific user types
- Updating tests, and general code cleanup in tests/fixtures etc.

TODO: some more refactoring of API code could help remove duplication - but will wait with this as @marrobi may have a suggestion for simplifying the API that will reduce this code duplication.


deniscep (approved):
LGTM

TessFerrandez:
Walked through this with @deniscep

PR #652, opened 2021-08-17
Fix resource processor version to buster

deniscep:
# PR for issue

## What is being addressed

Building resource processor fails for python 3.8 image; switch to buster
Added also the acr_user_managed_identity_client_id to api_web_app

### DIFF
--- a/resource_processor/vmss_porter/Dockerfile
+++ b/resource_processor/vmss_porter/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:buster
+FROM python:3.8-buster
 
 # Azure CLI
 RUN apt-get update \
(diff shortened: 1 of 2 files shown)


resource_processor/vmss_porter/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.8
+FROM python:3.8.10
```
marrobi, line 1: Can we use a base image specifically tagged with `buster`, that way we know will have az cli version available as this might get its base image changes at some point. 

https://hub.docker.com/_/python?tab=tags&page=1&ordering=last_updated&name=buster

deniscep: switched to 3.8-buster

marrobi (approved):
lgtm

Issue #663, opened 2021-08-17
[Task] Align porter outputs in guacamole


tamirkamara:
Story: #483 

**Description**
Make sure the outputs defined in porter are aligned and working for install (and other actions as needed)

PR #664, opened 2021-08-17
Update outputs in guacamole porter/terraform

tamirkamara:
# PR for issue #663 

## What is being addressed

Syncing both terraform and porter outputs to allow them to be fetched by the resource processor (separate task)

## How is this addressed

- Rename outputs to be exactly the same in porter and terraform
- Have outputs defined in porter with the right action (install)
- Upgrade TF to 1.0.4 
- Remove ACR from core RG


marrobi (approved):
LGTM

Are a couple of things we should think about - such as output naming we need a convention - for example always use `connection_uri` as when building the UI need to be consistent across resources.

workspaces/services/guacamole-azure-win10vm/porter.yaml
```diff
@@ -69,6 +70,7 @@ install:
         container_name: "{{ bundle.parameters.tfstate_container_name }}"
         key: "{{ bundle.parameters.tre_id }}-ws-{{ bundle.parameters.workspace_id }}-svc-guacamole-winwin10vm"
       outputs:
+      - name: guacamole_win10vm_address
```
tamirkamara, line 73: Here too?

workspaces/services/guacamole/porter.yaml
```diff
@@ -41,32 +40,34 @@ parameters:
     default: false
 
 outputs:
-  - name: guacamole_workspace_name
+  - name: guacamole_url
```
tamirkamara, line 43: connection_uri?

workspaces/services/guacamole-azure-win10vm/porter.yaml
marrobi: What's this used for?

marrobi (approved):
LGTM

Issue #612, opened 2021-08-12
[Task] Implement GET /api/workspace-service-templates/{template_name}/user-resource-templates/{template_name}


TessFerrandez:
Story: #558 

**Description**
Implement GET /api/workspace-service-templates/{template_name}/user-resource-templates/{template_name}

Issue #558, opened 2021-08-05
[Story] Implement GET /api/workspace-service-templates/{service_template_name}/user-resource-templates


TessFerrandez:
## Description

As a TRE User
I want to list all user-resource-templates available to add to a given workspace-service
So that I can create user-resources

## Acceptance critiera

- [x] GET /api/workspace-service-templates/{template_name}/user-resource-templates Returns name, description
- [x] GET /api/workspace-service-templates/{template_name}/user-resource-templates/{template_name} returns enriched template
- [x] all endpoints available to TRE users

## Tasks

- [x] #611
- [x] #639 
- [x] #612

TessFerrandez:
@marrobi please check if this looks ok - added to backlog per yesterdays discussion

marrobi:
@TessFerrandez  Do you mean list user-resource-templates?

PR #669, opened 2021-08-18
get user resource template by name 

TessFerrandez:
# PR for issue #612 story #558

## What is being addressed

Implement GET /workspace-service-templates/{service_template_name}/user-resource-templates/{user_resource_template_name}

## How is this addressed

- Refactored a bit of code to avoid having to copy code
- Implemented API endpoint
- Implemented related tests
- Refactored existing tests in the area to avoid code duplication
- Renamed parameters for template apis to des-ambiguate between the different template names

### DIFF
--- a/management_api_app/db/repositories/resource_templates.py
+++ b/management_api_app/db/repositories/resource_templates.py
@@ -5,7 +5,7 @@ from azure.cosmos import CosmosClient
 from pydantic import parse_obj_as
 
 from core import config
-from db.errors import EntityDoesNotExist, EntityVersionExist
+from db.errors import DuplicateEntity, EntityDoesNotExist, EntityVersionExist
 from db.repositories.base import BaseRepository
 from models.domain.resource import ResourceType
 from models.domain.resource_template import ResourceTemplate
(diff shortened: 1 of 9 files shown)


management_api_app/db/repositories/resource_templates.py
```diff
@@ -51,6 +51,16 @@ def get_current_template(self, template_name: str, resource_type: ResourceType)
             raise EntityDoesNotExist
         return parse_obj_as(ResourceTemplate, templates[0])
 
+    def get_current_user_resource_template(self, user_resource_template_name, parent_service_name) -> UserResourceTemplate:
+        """
+        Returns full user_resource_template for the current version of the template_name, parent_service_template_name combo
+        """
+        query = self._template_by_name_query(user_resource_template_name, ResourceType.UserResource) + f' AND c.parentWorkspaceService = "{parent_service_name}" AND c.current = true'
+        templates = self.query(query=query)
+        if len(templates) != 1:
```
msebragge, line 60: Should we handle cases where l**en(templates) == 0** and **len(templates) > 1** separately?

msebragge (approved):

management_api_app/db/repositories/resource_templates.py
TessFerrandez: Good point @msebragge

In reality - anything above 1 should never happen (as we only have one current) but never is a dangerous word :) will split them up so we have a good troubleshooting path, in case it happens

PR #670, opened 2021-08-18
Fix for failed workspace deployment when parameter value has space

### LINKED ISSUE #585: [BUG] Failed workspace deployment: only one positional argument may be specified, the installation name, but multiple were received
**Describe the bug**
Workspace deployment failed with:

```
only one positional argument may be specified, the installation name, but multiple were received [mrtre17-9c1d X secure project]
```

**Steps to reproduce**

Deployed a workspace with following values:

```
"azure_location":"westeurope",
"workspace_id":"9c1d",
"tre_id":"mrtre17",
"address_space":"192.168.10.0/24",
"display_name":"Project X",
"description":"Very secure project"
```

deniscep:
# Fix for failed workspace deployment when parameter value isn't a single string

## What is being addressed

The Runner was returning error "only one positional argument may be specified".

## How is this addressed

- Enclosed parameter values in quotes
- Enclosed installation_id in quotes
- Added porter run command to returning error message

Closes #585


.github/workflows/deploy_tre.yml
```diff
@@ -5,7 +5,7 @@ on:
     # 1am each night https://crontab.guru/#0_1_*_*_*
     - cron: "0 1 * * *"
   push:
-    branches: [develop, main]
+    branches: [develop, main, deniscep/bug-585-runner]
```
marrobi, line 8: Needs removing.

marrobi (approved):
Need to clean the test trigger. Test looks good.

.github/workflows/deploy_tre.yml
deniscep: ah, good catch, thx, removed

Issue #629, opened 2021-08-15
[Task] Porter sends back outputs


tamirkamara:
Story: #483 

**Description**
Porter output variables include values that are important for other parts of the solution and should be sent back on the deployment status message so that the API could save to cosmos.
Example: a vm is deployed and its host name is returned to be later used when a user wants to connect to it.

PR #676, opened 2021-08-19
Deployment returns outputs

tamirkamara:
# PR for issue #629 

## What is being addressed

Deployment via porter expose outputs we need to persist 

## How is this addressed

- Extract porter outputs and return in the deployment status message
- Change the way we prepare parameters for porter to allow for bundles without Terraform and also prioritize how the value sources
- Bundle meant to be deployed into workspaces (e.g. guacamole/vm) should be using the resource id as the Terraform store key
- Change guacamole/vm templates to make use special params (like workspace_id) automatically

### Additional
- Remove ANSI control chars from the shell outputs - will make it easier to read on app insights.

### DIFF
--- a/resource_processor/vmss_porter/runner.py
+++ b/resource_processor/vmss_porter/runner.py
@@ -5,7 +5,7 @@ import socket
 import asyncio
 import logging
 
-from shared.logging import disable_unwanted_loggers, initialize_logging, get_message_id_logger  # pylint: disable=import-error # noqa
+from shared.logging import disable_unwanted_loggers, initialize_logging, get_message_id_logger, shell_output_logger  # pylint: disable=import-error # noqa
 from resources import strings  # pylint: disable=import-error # noqa
 from contextlib import asynccontextmanager
 from azure.servicebus import ServiceBusMessage
@@ -89,9 +89,7 @@ async def get_porter_parameter_keys(msg_body, env_vars):
         return porter_parameter_keys
     if stderr:
         result_stderr = stderr.decode()
-        logger_adapter.info('[stderr]')
-        for string in result_stderr.split('\n'):
-            logger_adapter.info(str(string))
+        shell_output_logger(result_stderr, '[stderr]', logger_adapter, logging.WARN)
 
 
 def get_installation_id(msg_body):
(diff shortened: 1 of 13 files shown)


resource_processor/vmss_porter/runner.py
```diff
@@ -82,40 +85,64 @@ async def filter_parameters_not_needed_by_porter(msg_body, env_vars):
...
 async def build_porter_command(msg_body, env_vars):
+    porter_parameter_keys = await get_porter_parameter_keys(msg_body, env_vars)
     porter_parameters = ""
 
-    porter_keys = await filter_parameters_not_needed_by_porter(msg_body, env_vars)
-    for parameter in porter_keys:
-        porter_parameters = porter_parameters + f" --param {parameter}=\"{msg_body['parameters'][parameter]}\""
+    for parameter_name in porter_parameter_keys:
+        # first try to find the param in the msg body parameters collection
+        parameter_value = msg_body['parameters'].get(parameter_name)
```
sachinkundu, line 108: parameter_value = msg_body['parameters'].get(parameter_name, env_vars.get(parameter_name, msg_body.get(parameter_name))) I think would do the same without all the checking for None.

resource_processor/vmss_porter/runner.py
```diff
@@ -82,40 +85,64 @@ async def filter_parameters_not_needed_by_porter(msg_body, env_vars):
...
+        # if not found, try to get it from the msg body root (e.g. id of the resource)
+        if parameter_value is None:
+            parameter_value = msg_body.get(parameter_name)
 
-    porter_parameters = porter_parameters + f" --param tfstate_container_name={env_vars['tfstate_container_name']}"
-    porter_parameters = porter_parameters + f" --param tfstate_resource_group_name={env_vars['tfstate_resource_group_name']}"
-    porter_parameters = porter_parameters + f" --param tfstate_storage_account_name={env_vars['tfstate_storage_account_name']}"
-    porter_parameters = porter_parameters + f" --param arm_use_msi={env_vars['arm_use_msi']}"
+        # if still not found, might be a special case
+        if parameter_value is None:
```
sachinkundu, line 119: You know that the parameter_value is None so why not move this whole block at the start and break from the loop if parameter_name in ["mgmt_acr_name", "mgmt_resource_group_name"] as those are special. Infact could refactor method out to special_param_handling to make the intention clear.

sachinkundu (commented):
Isn't the whole parameter thing UNION(param_set1, param_set2...).INTERSECT(porter_params)

resource_processor/vmss_porter/runner.py
tamirkamara: I wanted to give an option for the api to supply those values in the msg params, which is why I left it at the end.

marrobi (approved):
Looks good. A few things that might affect existing bundles/we should be consistent across bundles.

Can you create an issue to sight on things like mgmt_acr_name, and any other can think of.

Thanks.

Issue #657, opened 2021-08-17
[Story] Implement GET API to get a workspace-service and a list of all workspace-services of a TRE deployment


adishachar:
## Description

As a TRE owner
I want to be able to get a specific workspace-service and to be able to get ALL workspace-services of a TRE deployment
So that I can use the workspace-service or know which workspace-services instances are available.

## Acceptance criteria

- [x] An API Endpoint GET /workspace-services/{workspace_service_id} - to get single workspace-service - available to workspace owner/researchers for the associated workspace.
- [x] An API Endpoint GET /workspaces/{workspace_id}/workspace-services - to get ALL workspace-services for a given workspace - available to workspace owner/researchers for the associated workspace.

## Tasks

- [x] #672
- [x] #673

PR #684, opened 2021-08-20
get single workspace service

TessFerrandez:
# PR for issue #673 story #657

## What is being addressed

Add endpoint to get a single workspace by ID - verifies that user is owner or researcher in associated workspace

## How is this addressed

- Implement endpoint
- Implement tests

### DIFF
--- a/management_api_app/api/dependencies/workspaces.py
+++ b/management_api_app/api/dependencies/workspaces.py
@@ -12,11 +12,15 @@ from models.domain.workspace_service import WorkspaceService
 from resources import strings
 
 
-async def get_workspace_by_workspace_id_from_path(workspace_id: UUID4 = Path(...), workspaces_repo=Depends(get_repository(WorkspaceRepository))) -> Workspace:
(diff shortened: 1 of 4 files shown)


management_api_app/api/dependencies/workspaces.py
```diff
@@ -12,13 +12,17 @@
 from resources import strings
 
 
-async def get_workspace_by_workspace_id_from_path(workspace_id: UUID4 = Path(...), workspaces_repo=Depends(get_repository(WorkspaceRepository))) -> Workspace:
+def get_workspace_by_id(workspace_id: UUID4, workspaces_repo) -> Workspace:
```
Lybecker, line 15: I guess this method should be async too?

TessFerrandez: This method doesn't need to be async and since there is nothing to await and we call it directly, making it async causes issues -- technically the other one doesn't need to be either but the depends methods tend to have that signature. 

If you want to change the other one to not be async, feel free to play around with that to make sure it works and check in the changes before merging.

sachinkundu (approved):

Issue #679, opened 2021-08-20
[BUG] Info about user-resource owner is not attached to the User Resource


TessFerrandez:
Info about the user that created the user resource is not attached to the user resource object.
This is necessary for future gets/deletes/updates etc. as access should be restricted to the workspace owner or user who created the resource

PR #685, opened 2021-08-20
add user info to user resources

TessFerrandez:
# PR for issue #679 

## What is being addressed

Adds owner info to user resources so that this can be checked on get

## How is this addressed

- Add an ownerId property on user resources
- update tests


marrobi (approved):
LGTM

Issue #668, opened 2021-08-18
[BUG] workspaces should still be visible from the API until they are deleted in Azure


TessFerrandez:
**Describe the bug**

Currently workspaces are marked as deleted as soon as a delete starts and resources marked as deleted (in the db) are not returned on GET.

Instead - we should start the delete process and set the deployment status to delete initiated -- and let them still be visible on GET 

(before starting this story - double check with @marrobi about exact details of how it is expected to work

PR #699, opened 2021-08-23
Change from Deleting to Deleted

sachinkundu:
# PR for issue #668 

Moving from setting status to Deleting instead of Deleted. This means that workspaces which are asked to be deleted will be shown in the list of workspaces till the deleting is complete to deleted.

In resource processor send appropriate signals through the service bus.

Change the rudimentary state machine to only make valid transitions.

Update the tests and add new tests to check the state machine transitions.


tamirkamara (approved):
LGTM

Issue #689, opened 2021-08-22
[Task] API persists outputs returned in deployment status message


tamirkamara:
Story: #483 

**Description**
Now porter sends back the outputs in the deployment status message and we should persist them into the item in the db.

PR #702, opened 2021-08-23
Persist outputs returned on deployment status

tamirkamara:
# PR for issue #689 

## What is being addressed

Outputs should be persisted.

## How is this addressed

- API now reads the outputs returned in deployment status messages and saves to the resource on cosmos.

### DIFF
--- a/management_api_app/service_bus/deployment_status_update.py
+++ b/management_api_app/service_bus/deployment_status_update.py
@@ -74,6 +74,12 @@ def create_updated_deployment_document(resource: dict, message: DeploymentStatus
     resource["deployment"]["status"] = message.status
     resource["deployment"]["message"] = message.message
 
+    # although outputs are likely to be relevant when resources are moving to "deployed" status,
+    # lets not limit when we update them and have the resource process make that decision.
+    output_dict = {output.Name: output.Value.strip("'").strip('"') for output in message.outputs}
+
+    resource["resourceTemplateParameters"].update(output_dict)
+
     return resource
 
 
--- a/management_api_app/tests_ma/test_service_bus/test_deployment_status_update.py
+++ b/management_api_app/tests_ma/test_service_bus/test_deployment_status_update.py
@@ -25,6 +25,16 @@ test_sb_message = {
     "message": "test message"
 }
 
+test_sb_message_with_outputs = {
+    "id": "59b5c8e7-5c42-4fcb-a7fd-294cfc27aa76",
+    "status": Status.Deployed,
+    "message": "test message",
+    "outputs": [
+        {"Name": "name1", "Value": "value1", "Type": "type1"},
+        {"Name": "name2", "Value": "\"value2\"", "Type": "type2"}
+    ]
+}
+
 
 class ServiceBusReceivedMessageMock:
     def __init__(self, message: dict):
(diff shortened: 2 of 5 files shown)


management_api_app/service_bus/deployment_status_update.py
```diff
@@ -74,8 +74,26 @@ def create_updated_deployment_document(resource: dict, message: DeploymentStatus
...
+    output_dict = {}
+    for output in message.outputs:
+        output_dict[output.Name] = dequote(output.Value)
+
+    original_params_dict = dict(resource["resourceTemplateParameters"])
+    resource["resourceTemplateParameters"] = {**original_params_dict, **output_dict}
+
     return resource
 
+def dequote(s):
```
sachinkundu, line 88: does string.strip("'").strip('"') not work?

management_api_app/service_bus/deployment_status_update.py
```diff
@@ -74,8 +74,26 @@ def create_updated_deployment_document(resource: dict, message: DeploymentStatus
     resource["deployment"]["status"] = message.status
     resource["deployment"]["message"] = message.message
 
+    # although outputs are likely to be relevant when resources are moving to "deployed" status,
+    # lets not limit when we update them and have the resource process make that decision.
+    output_dict = {}
```
sachinkundu, line 79: consider using dict comprehension 
`output_dict = { output.Name : output.Values for output in message.outputs }`

I don't understand the intention though. Seems like you trying to merge dictionaries? You could use update or the merge operator |=

sachinkundu (commented):
As for tests, you are right the tests create a deployment message so not having a new field in that message does not change anything and hence the tests pass.

The coverage of tests for deployment status update is thin in any case.

You should write a test that the parameters finally saved are indeed the merged parameters.

management_api_app/service_bus/deployment_status_update.py
tamirkamara: It's almost the same, although using strip doesn't check that the start and end of the string are the same.
I don't know what will produce this string but: 
word" 
will become:
word

I can change that.

management_api_app/service_bus/deployment_status_update.py
tamirkamara: Fixed your first point.

On the second, I think the "|" operator was introduced on python 3.9 and we're on 3.8.
I changed the code to use an update instead of creating a new dictionary object. Let me know if this is better.

management_api_app/service_bus/deployment_status_update.py
tamirkamara: It won't produce the same results every time, but maybe that's ok...
Consider this string: word" (just 1 quote) 
Your process will remove that last quote despite it not having a matching open one.

Anyway, I changed the code with your suggestion.

management_api_app/service_bus/deployment_status_update.py
```diff
@@ -74,6 +74,14 @@ def create_updated_deployment_document(resource: dict, message: DeploymentStatus
     resource["deployment"]["status"] = message.status
     resource["deployment"]["message"] = message.message
 
+    # although outputs are likely to be relevant when resources are moving to "deployed" status,
+    # lets not limit when we update them and have the resource process make that decision.
+    output_dict = {output.Name: output.Value.strip("'").strip('"') for output in message.outputs}
+
+    params_dict = dict(resource["resourceTemplateParameters"])
```
sachinkundu, line 81: resource["resourceTemplateParameters"] is already a dict afaik and check the code.
one line temp variables are not needed in any case. 
could write `resource["resourceTemplateParameters"].update(output_dict)`

management_api_app/tests_ma/test_service_bus/test_deployment_status_update.py
```diff
@@ -185,3 +195,54 @@ async def test_state_transitions_from_deployed_to_delete_failed(app, sb_client,
     await receive_message_and_update_deployment(app)
 
     repo().update_item_dict.assert_called_once_with(expected_workspace.dict())
+
+
+@patch('service_bus.deployment_status_update.ResourceRepository')
+@patch('logging.error')
+@patch('service_bus.deployment_status_update.ServiceBusClient')
+@patch('fastapi.FastAPI')
+async def test_outputs_are_added_to_resource_item(app, sb_client, logging_mock, repo):
```
sachinkundu, line 204: could extract the common code into a pytest fixture. I think 2 lines are unique in the other test

sachinkundu (approved):
consider the suggestions. Otherwise LGTM

management_api_app/tests_ma/test_service_bus/test_deployment_status_update.py
tamirkamara: We spoke, IMO it would make it less readable and we can tackle it at a later stage if needed.

Issue #576, opened 2021-08-08
[Task] Replace hard coded properties of auth extension with config values


eladiw:
Story: #502 

**Description**
Currently properties such as 'allow copy paste' are hard coded enabled/disabled in the auth extension.
They should be configured when installing the bundle

PR #704, opened 2021-08-23
Modify Guacamole configuration with the right env variables

anatbal:
# PR for issue #576 

## What is being addressed

Replace Guacamole config with env variables and add the necessary variables for the Guacamole to start up correctly.

### DIFF
--- a/workspaces/services/guacamole/porter.yaml
+++ b/workspaces/services/guacamole/porter.yaml
@@ -78,45 +78,6 @@ parameters:
     type: string
     description: "An Id for this installation"
     env: id
-  - name: guac_disable_copy
-    type: boolean
-    default: true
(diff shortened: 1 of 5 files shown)


workspaces/services/guacamole/porter.yaml
```diff
@@ -35,12 +35,88 @@ parameters:
...
+  - name: aad_tenant_id
+    type: string
+    env: aad_tenant_id
+    description: "AAD tenant id"
   # the following are added automatically by the resource processor
   - name: id
     type: string
     description: "An Id for this installation"
     env: id
+  - name: guac_disable_copy
```
tamirkamara: seems like all this block is duplicated

tamirkamara (approved):

Issue #706, opened 2021-08-24
[BUG] API identity lost its access to cosmos


tamirkamara:
**Describe the bug**
As part of #691, the identity of the api doesn't have any access to cosmos.

PR #707, opened 2021-08-24
API permission to access Cosmos

tamirkamara:
# PR for issue #706 

## What is being addressed

Add permissions for the API identity to access Cosmos that were lost while remove its Contributor over the entire subscription.

### DIFF
--- a/templates/core/terraform/state-store/output.tf
+++ b/templates/core/terraform/state-store/output.tf
@@ -5,3 +5,7 @@ output "endpoint" {
 output "cosmosdb_account_name" {
   value = azurerm_cosmosdb_account.tre-db-account.name
 }
+
+output "id" {
+  value = azurerm_cosmosdb_account.tre-db-account.id
+}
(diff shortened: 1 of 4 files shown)


marrobi (approved):

templates/core/terraform/state-store/output.tf
```diff
@@ -4,4 +4,8 @@ output "endpoint" {
 
 output "cosmosdb_account_name" {
   value = azurerm_cosmosdb_account.tre-db-account.name
-}
\ No newline at end of file
+}
+
+output "id" {
```
tompaana, line 9: Nit picking: This could've been aligned with the "cosmosdb_account_name" either by adding "cosmosdb_account_" prefix or removing the prefix from the name output. Something to consider, but not worth blocking this PR over.

tompaana (approved):

templates/core/terraform/state-store/output.tf
tamirkamara: I agree, and thought about doing that. However didn't have much time to consider the implications and fix elsewhere, so left it as.

Issue #709, opened 2021-08-24
[BUG] Deployments to E2E env shouldn't run in parallel


tamirkamara:
**Describe the bug**
Running multiple deployments on the single E2E environment can break it and we should prevent that from happening (saving us valuable time to fix things).

**Steps to reproduce**
Preferable not to try to reproduce it, but...
1. Merge 2 PRs to develop with little time in between. 
2. One would probably fail and perhaps the env. will be unusable from that point

PR #710, opened 2021-08-24
Fix workflow concurrency

tamirkamara:
# PR for issue #709

## What is being addressed
Concurrent runs of the deployment/e2e workflow can produce unexpected results making the environment unstable. 

## How is this addressed

- Use the [concurrent ](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency)definition to control the run behavior
- Deployment / e2e workflow run 1 at a time waiting for their turn
- For all other workflows, a new trigger will stop any running/pending workflows (of the same type) in favor of running it over the most recent code
- [small] turn on DOCKER_BUILDKIT for the deployment workflow as well.

An example of the result when I pushed something to this very branch as the workflow was running: 
image


sachinkundu (approved):
Nice!

PR #712, opened 2021-08-25
Wait for delete in tests

sachinkundu:
# PR for issue #668 

## What is being addressed

Make the tests wait while delete happens just like we wait for install.

When delete completes the GET response sends 404 which needs to be handled correctly.


Lybecker (approved):
LGTM

Issue #711, opened 2021-08-25
[BUG] Gitea sceret might be accessed by TF before KV policy is in place


tamirkamara:
**Describe the bug**
Sometimes TF tries to access the Gitea secret before it saved the access policy for it, which results in 403 error.

**Steps to reproduce**

1. Try to provision a new TRE with Gitea and Nexus enabled. The error won't happen every time.

PR #713, opened 2021-08-25
Change gitea identity for better dependencies

tamirkamara:
# PR for issue #711 

## What is being addressed

Sometimes TF fails to deploy a TRE with Gitea due to order of execution.

## How is this addressed

- Make the gitea secrets depend on the keyvault policy. This resulted in a dependency cycle which meant we couldn't keep using system managed identity.

### Additional
- Add missing ignore tags
- Add module dependency to make sure gitea & nexus are deployed after the plan is in place (although didn't see that problem myself)


marrobi (approved):
LGTM - If feel appropriate can you add an issue to maybe move the app service plan out of the API and sort dependencies?

Issue #441, opened 2021-07-09
Lock down public endpoints on service bus


marrobi:
## Description

As a Cloud Administrator
I want no public access to the service bus
So that the solution attack surface is reduced

## Acceptance criteria

- [ ] Service bus cannot be connected to other than via the private endpoint
- [ ] Workspaces still deploy successfully

jjcollinge:
AFAIK service bus private endpoint requires premium SKU $$$ - we've made it optional and use the service bus firewall settings to restrict it to our CIDR to work around this for those who need to keep cost down.

marrobi:
Having spoken to a few people, given the usual spend on the solution - from a workspace compute perspective - the additional cost isn't a major issue at present so will stick with premium for now.

PR #734, opened 2021-08-30
Lock down public Service Bus endpoints

tompaana:
# PR for issue #441 

## What is being addressed and how

- Network rule added to Service Bus to block public endpoints


TessFerrandez (approved):

marrobi (changes requested):
@tompaana I'm on leave today (public holiday in the UK) but wanted to comment before merged.

Why are we suddenly using service endpoints when we have private link configured?

Happy to chat tomorrow.

tompaana:
> Why are we suddenly using service endpoints when we have private link configured?
> 
> Happy to chat tomorrow.

@marrobi I couldn't get it working without the service endpoints - most likely due to my lack of expertise when it comes to Terraform. Let's keep this PR open until you get back and let's have that chat tomorrow.

tompaana:
Service Bus Networking blade after the change:
image

Access to the Service Bus from the outside is blocked:
```plaintext
Failed to send a Service Bus message: ServiceBusAuthorizationError: Ip has been prevented to connect to the endpoint.
```

templates/core/terraform/servicebus/variables.tf
```diff
@@ -1,6 +1,6 @@
+variable "tenant_id" {}
```
marrobi, line 1: Is `tenand_id` and `resource_processor_subnet_id` used anywhere?

tompaana: `tenant_id` is not and is now removed.

`resource_processor_subnet_id` is used by the *Service Bus* private endpoint - which is a little strange:

```json
resource "azurerm_private_endpoint" "sbpe" {
  ...
  subnet_id           = var.resource_processor_subnet_id
```

marrobi (approved):
Looking good now.

Issue #615, opened 2021-08-12
[BUG] Registering two user-resource templates with the same name but for different workspace-service templates is handled incorrectly


TessFerrandez:
**Describe the bug**
1. If you try to create a new user-resource template (ex. vm) for workspace-service template (guacamole) and then later create another new user-resource template (also vm) but for workspace-service template (aml) - this fails if they have the same version
2. Same scenario as above but different versions -> the second user-resource template (vm for aml) will be set to current - but there will no longer be a current for user-resource template combo (vm for guacamole)

** Expected behavior**

A template author should be able to author two user-resource templates with the same name (ex. **vm**) for two different workspace-services (ex. **guacamole** and **aml**) and the user-resource-template-name + parent-service-template-name should be the unique key.

**Steps to reproduce**

In API
1. POST /api/workspace-service-templates/ with a template named "workspace-service-1"
2. POST /api/workspace-service-templates/ with a template named "workspace-service-2"
3. POST /api/workspace-service-templates/workspace-service-1/user-resource-templates/ with a template named "my-user-resource"
4. POST /api/workspace-service-templates/workspace-service-2/user-resource-templates/ with a template named "my-user-resource" (same version as previous)

This will fail saying the a template with the same version already exists (even though this is for another workspace-service)

5. POST /api/workspace-service-templates/workspace-service-2/user-resource-templates/ with a template named "my-user-resource" (change version)

Look in the database - the first template (for workspace-service-1) will no longer have current set.

PR #741, opened 2021-08-30
make user resource template unique to name + parent_workspace_service…

TessFerrandez:
…_name

# PR for issue #615 

## What is being addressed

User resource template should be unique to template name + parent template name

## How is this addressed

- validate against both template name + parent template name for user resource templates
- update related tests


tompaana (approved):
I like it - much cleaner now!

Issue #705, opened 2021-08-23
[Task] Allow Guacamole to be accessible via internet


tamirkamara:
Story: #483

**Description**
Guacamole needs to be accessed somehow by the end user. Sometimes that means over public internet and the deployed bundle should allow that (currently it doesn't)

PR #755, opened 2021-08-31
Guacamole is accessible via public network

tamirkamara:
# PR for issue #705 

## What is being addressed

Guacamole is currently accessible via the private network which is not suitable for most use cases.

## How is this addressed

- Add a parameter that allows Guacamole to be provisioned with/out public network access by enable/disable its private endpoint.

### DIFF
--- a/templates/workspace_services/guacamole/porter.yaml
+++ b/templates/workspace_services/guacamole/porter.yaml
@@ -73,6 +73,11 @@ parameters:
     type: string
     env: AAD_TENANT_ID
     description: "AAD tenant id"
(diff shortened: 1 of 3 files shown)


marrobi (approved):
LGTM - I'm not 100% sure on the variable naming and default. For example should the user have to tick a box to expose externally for example. We can always sort that later if feel its the wrong way round.

anatbal (approved):

tamirkamara:
@marrobi what here makes it harder to include a checkbox for the user to choose? The default?
Do you another name for the param?

Issue #748, opened 2021-08-30
Nexus - verify description and correct the required egress rules


Lybecker:
Make sure that below egress firewall rules are what is needed for the Nexus and the description of each of them is described in detail.

- [Documentation](https://github.com/microsoft/AzureTRE/blob/defaultroute/templates/shared_services/sonatype-nexus/readme.md#network-requirements)
- [Firewall settings](https://github.com/microsoft/AzureTRE/blob/defaultroute/templates/core/terraform/firewall/firewall.tf#L82-L145) - be aware these rules seems to be shared with Gitea

Please create a PR to the defaultroute branch or let me know what to change.

Lybecker:
@tamirkamara who can help me with this task?

tamirkamara:
@Lybecker what should be the output from this task? Should I change something in the branch? TF / Doc?

Lybecker:
@tamirkamara I would like you to check that the URLs/ServiceTags are needed and describe why each of these are needed, so I can include it in the documentation.

You can either share it here or commit directly to the `defaultroute` branch

Issue #747, opened 2021-08-30
Gitea - verify description and correct the required egress rules


Lybecker:
Make sure that below egress firewall rules are what is needed for the Gitea and the description of each of them is described in detail.

- [Documentation](https://github.com/microsoft/AzureTRE/blob/defaultroute/templates/shared_services/gitea/readme.md#network-requirements)
- [Firewall settings](https://github.com/microsoft/AzureTRE/blob/defaultroute/templates/core/terraform/firewall/firewall.tf#L82-L145) - be aware these rules seems to be shared with Nexus

Please create a PR to the defaultroute branch or let me know what to change.

Lybecker:
@tamirkamara who can help me with this task?

PR #763, opened 2021-09-02
Update shared services firewall rules

tamirkamara:
# PR for issue #748 #747 

## What is being addressed

Updating the outbound addressed allowed for Gitea and Nexus shared services.


templates/core/terraform/firewall/locals.tf
```diff
@@ -0,0 +1,9 @@
+locals {
```
marrobi, line 1: Can we pull these out to variables with defaults? I expect customers will need to add to these - especially for Nexus. 

Don't really want to have to edit the locals file?

Can make a separate issue on the backlog - not urgent.

marrobi (approved):
LGTM

templates/core/terraform/firewall/locals.tf
tamirkamara: @marrobi pushed an update for this. This isn't "wired" end to end with env variables but it's almost there for the TF part.

PR #764, opened 2021-09-02
Internally pass AAD params

anatbal:
# PR for issue #761 

## What is being addressed

Remove Azure AD app registration params (API_CLIENT_ID, AAD_TENANT_ID) from the api call and pass them internally through terraform variables.


tompaana (approved):
LGTM

Issue #771, opened 2021-09-05
Auto calculate address_space for workspace with default size


tamirkamara:
Story: #52 

**Description**
Allow the system to calculate an address_space for a workspace in a default size (/24) by find a free space within the pool available for the workspaces (and not allocated already)

Issue #773, opened 2021-09-05
Allow custom size for address_space when creating a workspace


tamirkamara:
Story: #52 

**Description**
There's a default size for a workspace address_space and also the option to provide a custom one address_space. But we'd like to have a middle option of allowing the user to specify the desired size of the workspace network without providing the full network.

PR #775, opened 2021-09-05
API calculates address_space automatically

tamirkamara:
# PR for issue #771 

## What is being addressed

Address spaces for workspaces should be calculated automatically without the user having to provide it.

## How is this addressed

- API allows creating a workspace without providing address_space in the payload which will result in an auto-generated one. 
- Auto-generated address space is a /24 free one within the overall TRE_ADDRESS_SPACE (new "core" param)
- No change in E2E which are still using custom addresses

### DIFF
--- a/.github/workflows/deploy_tre.yml
+++ b/.github/workflows/deploy_tre.yml
@@ -17,6 +17,8 @@ jobs:
     name: Deploy Management Infrastructure
     runs-on: ubuntu-latest
     environment: Dev
+    env:
+      TF_INPUT: 0 # interactive is off
     steps:
       - name: Checkout
         uses: actions/checkout@v2
@@ -73,6 +75,7 @@ jobs:
     environment: Dev
     env:
       DOCKER_BUILDKIT: 1
+      TF_INPUT: 0 # interactive is off
     steps:
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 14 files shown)


management_api_app/db/repositories/workspaces.py
```diff
@@ -60,6 +66,12 @@ def create_workspace_item(self, workspace_input: WorkspaceInCreate) -> Workspace
 
         return workspace
 
+    def get_new_address_space(self, cidr_netmask: int = 24):
```
marrobi, line 69: Can the 24 be an environment variable? Will be cases in the near future needing larger address spaces for a workspace.

marrobi (commented):
Looking great. Thanks!

management_api_app/db/repositories/workspaces.py
tamirkamara: We have another task #773 which might give the result you're looking for, but we can also have the default size as an environment variable.
TBH, I don't understand the flow... Is the user expected to know the size required for a workspace?
@marrobi

marrobi: #773 will cover it for now. There are a few different flows, I can forsee. 

1. TRE Admin wants to specify specify address space. (custom) 
2. Address space automatically configured based on size specified by azure admin. (env var? )
3. TRE Admin wants to select from a t-shirt size.

1 and 2 are the ones feel will be most useful initially.

tamirkamara: Resolving this as it will be handled in another task.

eladiw (approved):
looks good

tamirkamara:
@marrobi yes, I wanted to rename ADDRESS_SPACE to CORE_ADDRESS_SPACE, but that is used in so many different places so preferred to do that in a specific task #804.

Issue #770, opened 2021-09-03
[BUG] devcontainer Dockerfile results in outdated azurecli (2.18)


SvenAelterman:
The Dockerfile installs the Azure CLI for Debian using a shell script found at https://azurecliprod.blob.core.windows.net/$root/deb_install.sh. That shell script sets the package source for azurecli to "bullseye," which appears to only contain azurecli up to version 2.18.

This, in turn, results in an issue when trying to login to the Azure Container Registry during the running of `make all`.

For Debian, it appears (to this non-Linux user) that `buster` is a better repo.

Workaround: we were able to update the package list to `buster` and run `sudo apt-get upgrade` to get az cli version 2.27.2. We could then login to the ACR and successfully continue with the makefile.

**Steps to reproduce**

1. Build devcontainer.
1. Try to run `az acr login -n [somename]`
1. Observe error message
1. Run `az --version`

tompaana:
Fixed in #781

PR #781, opened 2021-09-07
Changes the devcontainer base image to buster

tompaana:
# PR for issue (bug) #770 

Changes the devcontainer base image from `bullseye` to `buster` to combat the Azure CLI problem with package version mismatch as described in bug report #770.


Lybecker (approved):
Thank you!

PR #793, opened 2021-09-08
Line break consistency between deploy.sh and destroy.sh

SvenAelterman:
## What is being addressed

Formatting only

## How is this addressed

- Add line breaks to destroy.sh to break the Terraform wrapper call across lines for readability and consistency with line breaks found in deploy.sh.


marrobi (approved):
Thanks!

PR #799, opened 2021-09-09
Tessferrandez/longer urls

TessFerrandez:
# PR for issue #796 

## What is being addressed

Change URLs to the longer form workspace/workspace-service/user-resource

## How is this addressed

- Change the URLs
- Update tests - and clean them up a bit

NOTE: we also need to validate the whole chain for all long form URLs - this is done in a separate issue that comes next #797


api_app/api/routes/workspaces.py
```diff
@@ -154,23 +153,21 @@ async def patch_workspace_service(workspace_service_patch: WorkspaceServicePatch
...
+        user_resources = [resource for resource in user_resources if resource.ownerId == user.id]
+
     return UserResourcesInList(userResources=user_resources)
 
 
-@user_resources_router.get("/user-resources/{resource_id}", response_model=UserResourceInResponse, name=strings.API_GET_USER_RESOURCE)
-async def retrieve_user_resource_by_id(user_resource=Depends(get_user_resource_by_id_from_path), user=Depends(get_current_user), workspace_repo=Depends(get_repository(WorkspaceRepository))) -> UserResourceInResponse:
-    workspace = get_workspace_by_id(user_resource.workspaceId, workspace_repo)
+@user_resources_router.get("/workspaces/{workspace_id}/workspace-services/{service_id}/user-resources/{resource_id}", response_model=UserResourceInResponse, name=strings.API_GET_USER_RESOURCE)
+async def retrieve_user_resource_by_id(workspace_id: str, service_id: str, resource_id: str, workspace=Depends(get_workspace_by_workspace_id_from_path), user_resource=Depends(get_user_resource_by_id_from_path), user=Depends(get_current_user)) -> UserResourceInResponse:
```
eladiw, line 170: why resource_id:str is needed?
you also have the user_resource arg

eladiw (approved):

api_app/api/routes/workspaces.py
TessFerrandez: The parameter order in swagger depends on the when the arguments are accessed so this forces that -- when I start validating I'll be using the resource_id

PR #811, opened 2021-09-10
Allow inbound port 80 rule to App Gateway NSG

### LINKED ISSUE #808: [BUG] NSG rules prevent lets encrypt cert creation
**Describe the bug**
The new NSG rules on the app gw subnet block inbound port 80 which is needed for lets encrypt validation.

Either the lets encrypt script should add and remove or leave them off. Not sure if/how cert renewal works.

tompaana:
# Fixes #808


Lybecker (approved):
port 80 it is :)

tompaana:
> port 80 it is :)

I haven't yet locally finished troubleshooting to verify that this fix works, but since it won't do any damage (famous last words), I'll merge ASAP. However, I will keep the bug open until it is confirmed fixed.

Issue #762, opened 2021-09-02
Change API URIs to long form


TessFerrandez:
## Description

As a user of the API
I want to see the hierarchy in the API URIs
So that I can see how the resources fit

## Acceptance criteria

- [x] Change GET /api/workspace-services/{services_id} to /api/workspaces/{workspace_id}/workspace-services/{services_id}
- [x] Change GET /api/workspace-services/{service_id}/user-resources to /api/workspaces/{workspace_id}/workspace-services/{service_id}/user-resources
- [x] Change GET /api/user-resources/{resource_id} to /api/workspaces/{workspace_id}/workspace-services/{service_id}/user-resources/{resource_id}
- [x] Validate the chain in all cases - i.e. validate that the resource_id actuall belongs to service_id and that service_id actually belongs to workspace_id -- else 404

## Tasks

- [x] #796
- [x] #797

TessFerrandez:
Suggested sprint candidate for next sprint @marrobi @christoferlof

christoferlof:
We can shorten the long form if we go with the initial design as proposed in #12 
`GET /workspaces/{workspace_id}/services/{workspace_service_id}/resources/{resource_id}`, for instance

TessFerrandez:
Looks nice and clean @christoferlof - let's check with @marrobi when he gets back -- I can do the longer version now, the switch to the shorter (services vs. workspace-services etc.) can be done in a separate PR pretty easily

PR #813, opened 2021-09-10
validate full routes

TessFerrandez:
# PR for issue #797 #762 

## What is being addressed

For all URIs with long form ex. /workspace/workspace-service/user-resource - this validates that the user resource actually belongs to the workspace service and the workspace .

This is to disallow for someone to provide a workspace id they have access to - but that is not related to the actual service or resource

## How is this addressed

- Query for all ids, ex. user resource with id AND workspace id AND parent service id, not just id
- Update routes and tests to pass in all parameters and update test validation

### DIFF
--- a/api_app/db/repositories/workspaces.py
+++ b/api_app/db/repositories/workspaces.py
@@ -5,7 +5,7 @@ from azure.cosmos import CosmosClient
 from pydantic import parse_obj_as
 
 from core import config
-from db.errors import ResourceIsNotDeployed
+from db.errors import ResourceIsNotDeployed, EntityDoesNotExist
 from db.repositories.resources import ResourceRepository
 from models.domain.resource import Deployment, Status, ResourceType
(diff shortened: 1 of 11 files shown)


api_app/db/repositories/workspaces.py
```diff
@@ -36,8 +36,11 @@ def get_deployed_workspace_by_workspace_id(self, workspace_id: str) -> Workspace
         return workspace
 
     def get_workspace_by_workspace_id(self, workspace_id: str) -> Workspace:
```
LizaShak, line 38: Similar to the change in api_app/api/dependencies/workspaces.py
Maybe it is worth also renaming it here to : get_workspace_by_id ?

TessFerrandez: @LizaShak Fixed -- let me know if there is something else or if you are good to approve.

LizaShak (approved):
Looks Good!

Issue #805, opened 2021-09-09
PR validation should include building images for all custom applications


tamirkamara:
**Description**
We currently only build an API image as part of the PR validation. We should include also: resource processor, guacamole, gitea.
In addition, stop pushing to ACR since not in use.

PR #818, opened 2021-09-12
Build all docker images in PR validation

tamirkamara:
# PR for issue #805 

## What is being addressed

All docker images should be built as part of PR validation

## How is this addressed

- Add build steps for all our custom docker images 
- Run tests as part of the docker build for the API. This required moving the python requirements files
- Remove api build steps from the 
- Update python package versions on api/resource-processor to match
- [x] rename the workflow files after this draft is approved
- [ ] update PR checks: add aggregate check and remove Publish Unit Tests Results


.github/workflows/build_and_push_docker_images.yml
```diff
@@ -17,26 +11,97 @@ concurrency:
...
+        uses: docker/build-push-action@v2
+        with:
+          context: ./resource_processor
+          file: ./resource_processor/vmss_porter/Dockerfile
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+
+        # TODO: Add testing (#817) before this
+      - name: "Build image: guacamole_server"
+        if: (steps.filter.outputs.api == 'true' || github.event_name == 'workflow_dispatch')
```
eladiw, line 70: fix 'api' typos (all across)

eladiw (changes requested):
typo in the yml file.
looks good other than that

.github/workflows/build_and_push_docker_images.yml
tamirkamara: done.

eladiw (approved):
lgtm

.devcontainer/Dockerfile
```diff
@@ -47,9 +47,9 @@ RUN export PORTER_HOME=/home/vscode/.porter \
 ENV PATH /home/vscode/.porter/:$PATH
 
 # Install requirements
-COPY ["requirements.txt", "requirements-dev.txt", "/tmp/pip-tmp/" ]
```
TessFerrandez, line 50: not sure in github how to comment on the move so I'm doing it here... 

I'm a bit confused here... the requirements-dev.txt is supposed to be a more generic python requirements file -- containing things like pre-commit etc. it is not api specific, so it should be in the root... and the root requirements one tied the api and the dev one together (eg. included the api specific and the dev general requirements).  If we have a specific one for the resource processor, the main (root one) should include links to both the dev requirements, the api requirements, and the resource processor requirements - e.g. a one stop shop for all requirements you need for development on your dev machine (installed on docker but also easily installable in a non-docker environment for those who prefer pycharm or other ides).

Also, did this now overwrite the root or the API requirements.txt?

TessFerrandez (changes requested):
Just a question, but want to make sure it gets answered before merging so marking as request changes

.github/workflows/build_and_push_docker_images.yml
```diff
@@ -17,26 +11,97 @@ concurrency:
...
+      - name: "Build image: resource_processor"
+        if: (steps.filter.outputs.resource_processor == 'true' || github.event_name == 'workflow_dispatch')
+        uses: docker/build-push-action@v2
+        with:
+          context: ./resource_processor
+          file: ./resource_processor/vmss_porter/Dockerfile
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+
+        # TODO: Add testing (#817) before this
```
TessFerrandez, line 68: If we don't have tests here - does this mean that this PR goes untested? Why were they removed?

TessFerrandez (changes requested):

.github/workflows/build_and_push_docker_images.yml
tamirkamara: There're tests but they aren't being triggered as part of the build. Anat is tasked to solve it.

TessFerrandez: ahh, some other tests, was thinking this was our regular pytests that were removed

.devcontainer/Dockerfile
TessFerrandez: per discussion

TessFerrandez (approved):
LGTM

Issue #580, opened 2021-08-09
[Story] Add API functionality to disable/enable a user-resource


adishachar:
## Description

As a workspace user
I want to be able to disable/enable user resource
So that I can control if they are active

## Acceptance criteria

- [ ] Workspace user can disable a user resource that it owns
- [ ] Workspace user can enable a user resource that it owns
- [ ] Workspace owner can disable any user resource within the workspace
- [ ] Workspace owner can enable any disabled user resource within the workspace
- [ ] disabled/enabled status should be visible on GET API
- [ ] user-resources will still be accessible with direct URIs to VMs and through the Azure Portal
- [ ] full path (all IDs) are verified to belong to the resource

PR #819, opened 2021-09-12
Adding API functionality to Enable/Disable User Resource

eladiw:
#580 

## What is being addressed

Adding API functionality to Enable/Disable User Resource

## How is this addressed

- Added the relevant API route
- Added unit tests

## How this PR was tested

- Manually using the Swagger UI
- Using the unit tests


api_app/api/routes/workspaces.py
```diff
@@ -234,3 +234,10 @@ async def delete_user_resource(workspace_id: str, service_id: str, resource_id:
     await send_uninstall_message(user_resource, user_resource_repo, previous_deletion_status, ResourceType.UserResource)
 
     return UserResourceIdInResponse(resourceId=user_resource.id)
+
+
+@user_resources_router.patch("/workspaces/{workspace_id}/workspace-services/{service_id}/user-resources/{resource_id}", response_model=UserResourceInResponse, name=strings.API_UPDATE_USER_RESOURCE)
+async def patch_user_resource(workspace_id: str, service_id: str, resource_id: str, user_resource_patch: UserResourcePatchEnabled, user_resource=Depends(get_user_resource_by_id_from_path), user_resource_repo=Depends(get_repository(UserResourceRepository)), user=Depends(get_current_user), workspace_service=Depends(get_deployed_workspace_service_by_id_from_path), workspace=Depends(get_workspace_by_workspace_id_from_path)) -> UserResourceInResponse:
+    validate_user_is_owner_or_researcher(user, workspace)
```
TessFerrandez, line 241: For user resources, if the user is researcher, we need to check that they are actually the owner of the resource. You should not be able to patch other researchers resources -- but the workspace owner can patch all... See the delete functionality for the checks

api_app/tests_ma/test_api/test_routes/test_workspaces.py
```diff
@@ -820,3 +832,74 @@ async def test_delete_user_resource_returns_resource_id(self, _, __, ___, get_us
...
+
+        response = await client.patch(app.url_path_for(strings.API_UPDATE_USER_RESOURCE, workspace_id=WORKSPACE_ID, service_id=SERVICE_ID, resource_id=USER_RESOURCE_ID), json=user_resource_service_patch)
+        patch_user_resource_mock.assert_called_once_with(user_resource_to_patch, user_resource_service_patch)
+
+        assert response.status_code == status.HTTP_200_OK
+
+    @ patch("api.dependencies.workspaces.UserResourceRepository.get_user_resource_by_id")
+    @ patch("api.dependencies.workspaces.WorkspaceServiceRepository.get_workspace_service_by_id")
+    @ patch("api.dependencies.workspaces.WorkspaceRepository.get_workspace_by_workspace_id")
+    async def test_patch_user_resources_without_roles_throws(self, get_workspace_mock, get_workspace_service_mock, get_user_resource_mock, app, client):
```
TessFerrandez, line 896: all access tests should be in test_api_access -- however, this particular test seems a little odd - it looks like it tests internal workings of the access control... 

Try mocking that out and test specific access of this resource instead, i.e. that you need to be either workspace owner or researcher + resource owner.

api_app/tests_ma/test_api/test_routes/test_workspaces.py
```diff
@@ -820,3 +832,74 @@ async def test_delete_user_resource_returns_resource_id(self, _, __, ___, get_us
...
+    @ patch("api.dependencies.workspaces.WorkspaceServiceRepository.get_workspace_service_by_id")
+    @ patch("api.dependencies.workspaces.WorkspaceRepository.get_workspace_by_workspace_id")
+    @ patch("api.routes.workspaces.UserResourceRepository.patch_user_resource", return_value=None)
+    async def test_patch_user_resources_patches_user_resource(self, patch_user_resource_mock, get_workspace_mock, get_workspace_service_mock, get_user_resource_mock, app, client):
+        auth_info_user_in_workspace_owner_role = {'sp_id': 'ab123', 'roles': {'WorkspaceOwner': 'ab124', 'WorkspaceResearcher': 'ab125'}}
+
+        user_resource_to_patch = sample_user_resource_object()
+        get_user_resource_mock.return_value = user_resource_to_patch
+        get_workspace_mock.return_value = sample_deployed_workspace(WORKSPACE_ID, auth_info_user_in_workspace_owner_role)
+        get_workspace_service_mock.return_value = sample_deployed_workspace_service()
```
TessFerrandez, line 884: Nit: we can mock this directly in the decorator to simplify the test

api_app/tests_ma/test_api/test_routes/test_workspaces.py
```diff
@@ -820,3 +832,74 @@ async def test_delete_user_resource_returns_resource_id(self, _, __, ___, get_us
...
+        assert response.status_code == status.HTTP_200_OK
+
+    @ patch("api.dependencies.workspaces.UserResourceRepository.get_user_resource_by_id")
+    @ patch("api.dependencies.workspaces.WorkspaceServiceRepository.get_workspace_service_by_id")
+    @ patch("api.dependencies.workspaces.WorkspaceRepository.get_workspace_by_workspace_id")
+    async def test_patch_user_resources_without_roles_throws(self, get_workspace_mock, get_workspace_service_mock, get_user_resource_mock, app, client):
+        user_resource_to_patch = sample_user_resource_object()
+        get_user_resource_mock.return_value = user_resource_to_patch
+        get_workspace_mock.return_value = sample_deployed_workspace(WORKSPACE_ID)
+        get_workspace_service_mock.return_value = sample_deployed_workspace_service()
```
TessFerrandez, line 900: same nit as above

api_app/db/repositories/user_resources.py
```diff
@@ -54,3 +54,7 @@ def get_user_resource_by_id(self, resource_id: str) -> UserResource:
 
     def get_user_resource_spec_params(self):
         return self.get_resource_base_spec_params()
+
+    def patch_user_resource(self, user_resource: UserResource, user_resource_patch: UserResourcePatchEnabled):
```
TessFerrandez, line 58: looks like we are missing tests for this function

TessFerrandez (changes requested):

api_app/api/routes/workspaces.py
```diff
@@ -234,3 +234,10 @@ async def delete_user_resource(workspace_id: str, service_id: str, resource_id:
     await send_uninstall_message(user_resource, user_resource_repo, previous_deletion_status, ResourceType.UserResource)
 
     return UserResourceIdInResponse(resourceId=user_resource.id)
+
+
+@user_resources_router.patch("/workspaces/{workspace_id}/workspace-services/{service_id}/user-resources/{resource_id}", response_model=UserResourceInResponse, name=strings.API_UPDATE_USER_RESOURCE)
+async def patch_user_resource(workspace_id: str, service_id: str, resource_id: str, user_resource_patch: UserResourcePatchEnabled, user_resource=Depends(get_user_resource_by_id_from_path), user_resource_repo=Depends(get_repository(UserResourceRepository)), user=Depends(get_current_user), workspace_service=Depends(get_deployed_workspace_service_by_id_from_path), workspace=Depends(get_workspace_by_workspace_id_from_path)) -> UserResourceInResponse:
```
TessFerrandez, line 240: why is this bringing down the workspace service, doesn't seem like it is used here? - even if it is used, it shouldn't need to be deployed

api_app/tests_ma/test_api/test_routes/test_workspaces.py
```diff
@@ -106,6 +107,17 @@ def sample_workspace_service(workspace_service_id=SERVICE_ID, workspace_id=WORKS
     )
 
 
+def sample_deployed_workspace_service(workspace_service_id=SERVICE_ID, workspace_id=WORKSPACE_ID):
```
TessFerrandez, line 110: as above, patching shouldn't require a deployed workspace service, so can probably make do with a sample workspace service if it is needed at all -- in reality patching should not need the workspace service -- the chain is verified in the PR I have open by just getting the resource by full ID

api_app/tests_ma/test_api/test_routes/test_workspaces.py
```diff
@@ -820,3 +832,74 @@ async def test_delete_user_resource_returns_resource_id(self, _, __, ___, get_us
...
+    @ patch("api.dependencies.workspaces.UserResourceRepository.get_user_resource_by_id")
+    @ patch("api.dependencies.workspaces.WorkspaceServiceRepository.get_workspace_service_by_id")
+    @ patch("api.dependencies.workspaces.WorkspaceRepository.get_workspace_by_workspace_id")
+    @ patch("api.routes.workspaces.UserResourceRepository.patch_user_resource", return_value=None)
+    async def test_patch_user_resources_patches_user_resource(self, patch_user_resource_mock, get_workspace_mock, get_workspace_service_mock, get_user_resource_mock, app, client):
+        auth_info_user_in_workspace_owner_role = {'sp_id': 'ab123', 'roles': {'WorkspaceOwner': 'ab124', 'WorkspaceResearcher': 'ab125'}}
+
+        user_resource_to_patch = sample_user_resource_object()
+        get_user_resource_mock.return_value = user_resource_to_patch
+        get_workspace_mock.return_value = sample_deployed_workspace(WORKSPACE_ID, auth_info_user_in_workspace_owner_role)
```
TessFerrandez, line 883: why deployed workspace?

api_app/tests_ma/test_api/test_routes/test_workspaces.py
```diff
@@ -820,3 +832,74 @@ async def test_delete_user_resource_returns_resource_id(self, _, __, ___, get_us
...
+
+        assert response.status_code == status.HTTP_200_OK
+
+    @ patch("api.dependencies.workspaces.UserResourceRepository.get_user_resource_by_id")
+    @ patch("api.dependencies.workspaces.WorkspaceServiceRepository.get_workspace_service_by_id")
+    @ patch("api.dependencies.workspaces.WorkspaceRepository.get_workspace_by_workspace_id")
+    async def test_patch_user_resources_without_roles_throws(self, get_workspace_mock, get_workspace_service_mock, get_user_resource_mock, app, client):
+        user_resource_to_patch = sample_user_resource_object()
+        get_user_resource_mock.return_value = user_resource_to_patch
+        get_workspace_mock.return_value = sample_deployed_workspace(WORKSPACE_ID)
```
TessFerrandez, line 899: see question above

TessFerrandez (changes requested):

api_app/api/routes/workspaces.py
```diff
@@ -234,3 +234,10 @@ async def delete_user_resource(workspace_id: str, service_id: str, resource_id:
     await send_uninstall_message(user_resource, user_resource_repo, previous_deletion_status, ResourceType.UserResource)
 
     return UserResourceIdInResponse(resourceId=user_resource.id)
+
+
+@user_resources_router.patch("/workspaces/{workspace_id}/workspace-services/{service_id}/user-resources/{resource_id}", response_model=UserResourceInResponse, name=strings.API_UPDATE_USER_RESOURCE)
```
TessFerrandez, line 239: workspace_id: str, service_id: str, resource_id: str, won't be needed once we have the patch in PR merged so might as well remove it from here so we don't leave it in by mistake when we merge

TessFerrandez (changes requested):

api_app/tests_ma/test_api/test_routes/test_workspaces.py
```diff
@@ -820,3 +832,74 @@ async def test_delete_user_resource_returns_resource_id(self, _, __, ___, get_us
...
+    @ patch("api.dependencies.workspaces.UserResourceRepository.get_user_resource_by_id", side_effect=EntityDoesNotExist)
+    async def test_patch_user_resources_returns_404_if_user_resource_does_not_exist(self, _, app, client):
+        response = await client.patch(app.url_path_for(strings.API_UPDATE_USER_RESOURCE, workspace_id=WORKSPACE_ID, service_id=SERVICE_ID, resource_id=USER_RESOURCE_ID), json='{"enabled": true}')
+        assert response.status_code == status.HTTP_404_NOT_FOUND
+
+    # [PATCH] /workspaces/{workspace_id}/workspace-services/{service_id}/user-resources/{resource_id}
+    @ patch("api.dependencies.workspaces.UserResourceRepository.get_user_resource_by_id", return_value=sample_user_resource_object())
+    @ patch("api.dependencies.workspaces.WorkspaceServiceRepository.get_workspace_service_by_id", side_effect=EntityDoesNotExist)
+    @ patch("api.dependencies.workspaces.WorkspaceRepository.get_workspace_by_workspace_id", return_value=sample_workspace())
+    async def test_patch_user_resources_returns_404_if_ws_service_does_not_exist(self, _, __, ___, app, client):
```
TessFerrandez, line 846: This test won't be needed once we remove the workspace service parameter from the UR patch

TessFerrandez (changes requested):

api_app/api/routes/workspaces.py
eladiw: a) you should provide the service id as part of the URI.
b) It won't make much sense to deploy a user resource for a service for a service which is not yet deployed. Hence the sanity check...

TessFerrandez: a) once we have the PR in flight that checks the whole user resource id chain - this will already be in effect - so no need for the extra check
b) agree if we would be deploying the resource - but here we are patching the resource -- let's talk tomorrow, but this is an unnecessary DB call, and in some cases we may want to patch and disable even if the service is not deployed

eladiw: talked f2f. agreed to remove the check

TessFerrandez (approved):

Issue #804, opened 2021-09-09
Rename ADDRESS_SPACE to CORE_ADDRESS_SPACE


tamirkamara:
Story: #52 

**Description**
Since we now have TRE_ADDRESS_SPACE and the existing ADDRESS_SPACE actually refers to the core network, it makes sense to rename those.

PR #823, opened 2021-09-12
Rename ADDRESS_SPACE to CORE_ADDRESS_SPACE 

tamirkamara:
# PR for issue #804 

## What is being addressed

Address space refers to the core network's address space and should be named as such.

Note that this requires local env config changes and/or external CICD secret updates.


TessFerrandez (approved):
Didn't test running it - but from visual inspection it looks good

Issue #836, opened 2021-09-14
[BUG] Core deployment keeps changing firewall rules due to a space char


tamirkamara:
**Describe the bug**
When deploying the core infrastructure, a rouge space (in the default gitea FQDNs) can cause large firewall replacements that takes a long time.

**Steps to reproduce**
https://github.com/microsoft/AzureTRE/runs/3594555874?check_suite_focus=true#step:5:2686

PR #837, opened 2021-09-14
Remove space from FQDNs

tamirkamara:
# PR for issue #836 

## What is being addressed

Spaces in FQDNs cause TF changes in the firewall rules.

## How is this addressed

- remove space from the FQDN strings


Lybecker (approved):
LGTM

PR #843, opened 2021-09-14
Remove duplicate job from CD workflow

tamirkamara:
## What is being addressed

The CD workflow (deploy_tre) deploys the mgmt infra twice - one directly and another from the call to "make all". This PR removes the first one (which saves 1.5 minutes).

### DIFF
--- a/.github/workflows/deploy_tre.yml
+++ b/.github/workflows/deploy_tre.yml
@@ -247,9 +247,6 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
-        with:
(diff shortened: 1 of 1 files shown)


marrobi (commented):
So I remember why this was/is like this.

We currently have a "dev" environment, the plan is/was to a have a demo/uat environment. That would share the same management infrastructure.

So maybe the dev environment should run `make` stages other than the bootstrap and mgmt infra deployment stages?

tamirkamara:
I see @marrobi. I suggest we'll add aliases to support this flow in the make file, I guess a "setup" one that includes everything up to the tre-deploy which is the environment specific.

Issue #826, opened 2021-09-13
[Task] Configure Resource Processor to use Porter Azure plugin 


tompaana:
Story: #536 

**Description**

Configure the Porter Azure plugin and verify that it the state backend is used properly.

Sub-tasks:

- [x] Container for Porter Azure plugin created in bootstrapping phase
- [x] Porter `config.toml` set correctly (Resource Processor file system)
- [x] Managed identity has privileges to use the Porter container

Issue #806, opened 2021-09-10
[Task] Porter Azure plugin installed for Resource Processor


tompaana:
Story: #536

Issue #536, opened 2021-08-02
[Story] Configure Resource Processor to use Azure storage for Porter remote state


marrobi:
## Description

As a TRE developer
I want to be able to access the state of previous porter deployments
So that I can uninstall bundles.

## Acceptance criteria

- [x] [Porter Azure plugin](https://porter.sh/plugins/azure/) installed for Resource Processor
- [x] Porter Azure plugin configured: Runner uses MGMT_STORAGE_ACCOUNT_NAME for porter state - https://github.com/getporter/azure-plugins (azure cli and MSI)
- [x] Functionality documented in Resource Processor README

christoferlof:
Currently Porter state is persisted on the VM within the scaleset. Will be lost.

tamirkamara:
@marrobi I think we talked about this plugin being very "fresh"... do you have an opinion about using an azure share instead?

marrobi:
@tamirkamara  The challenge with shares is that we are having to deploy bundles within bundles as Porter cannot pass parameters into dependencies:

https://github.com/getporter/porter/issues/990

We also need to be able to uninstall these dependent bundles.

This means mounting the share to the the bundle, which I'm not sure is possible, I was planning to use the blob storage plugin once sorted at the base level (although haven't tested this). 

Ideally the bundle dependencies would work and then it wouldn't matter.

PR #844, opened 2021-09-14
Porter Azure plugin enabled for Resource Processor

tompaana:
# PR for tasks #806 #826 and for the parent story #536

## What and how?

- Adds `porter` storage container creation to bootstrapping phase
- Installs and configures Porter Azure plugin (storage feature) for Resource Processor
- Updates documentation


resource_processor/vmss_porter/Dockerfile
```diff
@@ -1,22 +1,33 @@
...
-    && apt-get update && apt-get install -y terraform 
+    && apt-get update && apt-get install -y terraform
 
 # Install Porter
 RUN export PORTER_HOME=/root/.porter && curl -L https://cdn.porter.sh/latest/install-linux.sh | bash
 
+# Install Porter Azure plugin, see https://porter.sh/plugins/azure/
+RUN /root/.porter/porter plugin install azure --url https://cdn.porter.sh/plugins/azure
+
+# Generate configuration for Porter Azure plugin
```
marrobi, line 27: Is there a reason this is being done a build time rather than run time? 

Doing it at build time means the image cannot be reused by multiple customers/state configurations.

marrobi (changes requested):
Thanks. Good start.

Would prefer the porter config was generated at runtime, runner.py could even do it, but bash script that does the config then starts runner.py is fine.

Also can you separate out the docs changes from the porter changes so the PR  just matches the issue?

resource_processor/vmss_porter/Dockerfile
tompaana: Fixed

marrobi (approved):
Looks good. Look forward to testing it out.

Issue #846, opened 2021-09-14
Optimize env vars in CD workflow


tamirkamara:
**Description**
We using many "exports" in our CD scripts which can be removed and make our scripts easier to understand

PR #847, opened 2021-09-14
Reduce number of exports in CD workflow

tamirkamara:
# PR for issue #846 

## What is being addressed

Using environment variables in workflow effectively. 
A test run can be seen here: https://github.com/microsoft/AzureTRE/actions/runs/1234686969

## How is this addressed

- Remove unnecessary exports and rely more on github action's "env" section


Lybecker (approved):
LGTM

PR #849, opened 2021-09-14
Guacamole starts and fetches the VMs correctly

anatbal:
# PR for issue

## What is being addressed
 In this PR I refer to some bugs @marrobi has came across while deploying the Guacamole service:

- When get logged: [http-nio-8080-exec-15] ERROR o.a.g.rest.RESTExceptionMapper - Request could not be processed: Exception getting VMs: A JSONObject text must begin with '{' at 1 [character 2 line 1]. It happens since the API has changed.
- Cosmetics regarding the urls: add "https", "/guacamole"
- Missing Build and Push for the guacamole image

Also, in addition:
- Renamed "connection_url" to "connection_uri" for consistency
- Fixed a bug that reproduced a few time in which the guacamole failed to fetch the image from the acr.


templates/workspace_services/guacamole/terraform/web_app.tf
```diff
@@ -34,6 +34,12 @@ resource "azurerm_app_service" "guacamole" {
     KEYVAULT_URL = local.kv_url
     API_URL      = local.api_url
     SERVICE_ID   = "${var.tre_resource_id}"
+    WORKSPACE_ID = "${var.workspace_id}"
+
+    # Webapp with pe won't pull the guacamole image without these
+    DOCKER_REGISTRY_SERVER_URL      = "${data.azurerm_container_registry.mgmt_acr.login_server}"
+    DOCKER_REGISTRY_SERVER_USERNAME = "${data.azurerm_container_registry.mgmt_acr.admin_username}"
+    DOCKER_REGISTRY_SERVER_PASSWORD = "${data.azurerm_container_registry.mgmt_acr.admin_password}"
```
tamirkamara, line 42: I am confused by the comment at top and the requirement to provide user/pass as we're not doing it elsewhere (like the API).

anatbal: You are right, probably it not the best practice and since I am leaving I opened a new bug about it (#861) and left it out of this PR.

marrobi (approved):
Looks good. A couple of things will need fixing up, but are issues for these so suggest get merged.

templates/workspace_services/guacamole/terraform/web_app.tf
```diff
@@ -19,7 +19,7 @@ resource "azurerm_app_service" "guacamole" {
   https_only          = true
 
   site_config {
-    linux_fx_version                     = "DOCKER|${data.azurerm_container_registry.mgmt_acr.name}.azurecr.io/${var.image_name}:${var.image_tag}"
+    linux_fx_version                     = "DOCKER|${data.azurerm_container_registry.mgmt_acr.login_server}/microsoft/azuretre/${var.image_name}:${var.image_tag}"
```
msebragge, line 22: Is the explicit binding to **/microsoft/azuretre/** needed? How about forked repos?

Issue #853, opened 2021-09-15
[BUG] Make deploy TRE follow DRY


deniscep:
There is a bit of repetitive code around workspace and template registration, which should be written as a loop.

TessFerrandez:
@deniscep are you referring to a script or the workflow here?

tamirkamara:
Who ever picks this up can take a look at tamirkamara/register-matrix branch for some ideas. It's not done and I won't be able to finish it myself in the next few weeks...

deniscep:
@TessFerrandez why not both ;) ? Specifically I was talking about workflow, but now that you mention it, I remember there are things that can be cleaned up in the script itself as well.

Issue #838, opened 2021-09-14
[Task] Verify AML helloworld works with new NSG and Firewall settings


deniscep:
Story: #418 

Verify that AML helloworld can be runs successfully with with new NSG and Firewall rules on NoPIP cluster; adjust where needed

Issue #694, opened 2021-08-22
[Story] Register innereye workspace template


deniscep:
As a TRE Developer
I want to register InnerEye and InnerEye Inference workspace templates
So that I can use the templates to deploy InnerEye Workspaces. 

## Acceptance critiera

- [ ] InnerEye and InnerEye Inference templates registered and deployed to ACR
- [ ] Deployment either documented or automated (prefer automated)

christoferlof:
@deniscep does this relate to the e2e tests?

Issue #418, opened 2021-07-06
[Story] Verify minimum set of firewall rules required for Azure ML


marrobi:
Story: #350

**Description**
A set of firewall rules that enable Azure ML to function are deployed as part of the Azure ML bundle:

```
URLs:

- graph.windows.net
- ml.azure.com
- login.microsoftonline.com
- aadcdn.msftauth.net
- graph.microsoft.com
- management.azure.com
- viennaglobal.azurecr.io (Azure ML preview features)

Service Tags:

- Storage.`{AzureRegion}`
- AzureContainerRegistry
```

ref:
https://github.com/microsoft/AzureTRE/blob/ab01fa4bd8f8a3834591f41d515c7e000e0b4a51/workspaces/services/azureml/terraform/locals.tf#L10-L11

We do not believe all these are still required. Remove those that are no longer needed and replace those that can be done with [Service Tags](https://docs.microsoft.com/en-us/azure/virtual-network/service-tags-overview) wherever possible.

Tasks:
- [x] #598
- [x] #600
- [x] #651
- [x] #599
- [x] #601
- [x] #838

marrobi:
Do along with #584 ?

Lybecker:
Closing this one calls for a celebration :)

PR #854, opened 2021-09-15
Adjustments to bundles needed to run AML

### LINKED ISSUE #794: [Task] Modify InnerEye Deeplearning service so it can be registered with API
Story: #694 

**Description**
Modify InnerEye Deeplearning service so template can be registered with publish_register_bundle.sh

deniscep:
# Adjustments to bundles needed to run AML

## What is being addressed

Target for this PR is adjust workspace/service bundles and any other relevant services in order to have AML/InnerEye run successfully.

## How is this addressed

- All workspace services have been added to Deploy TRE action -> note, this and workspace registration will be cleaned up in follow-up story to make it DRY (#853)
- Removed double InnerEye template publish from Deploy TRE
- Added --force to porter runner -> Porter will only fetch new version if the version number has changed. However changing a version number requires a cascading update. With --force porter will compare SHA of published image vs the locally image, and pull newer.
- Added KeyVault to resource processor Firewall exceptions -> when resource processor TF creates a KeyVault in WS, before it places it on PE, it needs to check that it has created. This is however done over public network.
- Added blank template custom_parameters.json files in each of the workspace service bundles -> This file is used to generate a payload for registering the template with API
- Updated compute instance size to match the need
- Updated the compute cluster size so it can run InnerEye workload
- Allowing inbound traffic from resource processor to the service subnet on the NSG
- Setting inbound and outbound rules on AML NSG to match the needs of NoPIP compute (29877,29876,44224,445)

Closes #794 #838 #694 #418


templates/workspace_services/azureml/terraform/network.tf
```diff
@@ -0,0 +1,62 @@
...
+  protocol                    = "TCP"
+  resource_group_name         = data.azurerm_resource_group.ws.name
+  source_port_range           = "*"
+}
+
+resource "azurerm_network_security_rule" "allow-AzureMachineLearning-inbound-44224" {
+  access                      = "Allow"
+  destination_port_range      = "44224"
+  destination_address_prefix  = "VirtualNetwork"
+  source_address_prefix       = "*"
```
marrobi, line 40: Is there a tag so this can be narrowed down?

There is potential risk that a rogue service could be run on this port and a connection made (if can be routed - unlikely) from outside the VNet.

templates/workspace_services/azureml/terraform/network.tf
```diff
@@ -0,0 +1,62 @@
...
+  direction                   = "Inbound"
+  name                        = "AzureMachineLearning-inbound-44224"
+  network_security_group_name = data.azurerm_network_security_group.ws.name
+  priority                    = 202
+  protocol                    = "TCP"
+  resource_group_name         = data.azurerm_resource_group.ws.name
+  source_port_range           = "*"
+}
+
+resource "azurerm_network_security_rule" "allow-Outbound_Storage_445" {
```
marrobi, line 50: Does this risk data being extracted from the VNet by users to other storage accounts?

templates/workspaces/base/terraform/network/network.tf
```diff
@@ -187,6 +193,22 @@ resource "azurerm_network_security_rule" "allow-inbound-from-bastion" {
   source_port_range = "*"
 }
 
+resource "azurerm_network_security_rule" "allow-inbound-from-resourceprocessor" {
+  access                       = "Allow"
+  destination_address_prefixes = azurerm_subnet.services.address_prefixes
+  destination_port_range       = "*"
```
marrobi, line 199: Do we need all ports? Is it just 443?

templates/workspace_services/azureml/terraform/network.tf
deniscep: Tried it: error indicated that this can be either "Internet" or "AzureMachineLearning". However when put to "AzureMachineLearning", the terminal didn't open.

templates/workspace_services/azureml/terraform/network.tf
deniscep: Yes, we'll need to clarify with PG why this is needed, but it is checked by compute cluster that it has to be there (compute can't be created without it). Max limitation is region.

templates/workspaces/base/terraform/network/network.tf
deniscep: 443 should probably be enough; I'm at least not aware of any other ports that might be needed. I'll adjust.

templates/core/terraform/firewall/firewall.tf
```diff
@@ -216,7 +216,8 @@ resource "azurerm_firewall_network_rule_collection" "resource_processor_subnet"
       "AzureResourceManager",
       "AzureContainerRegistry",
       "AzureMonitor",
-      "Storage"
+      "Storage",
+      "AzureKeyVault"
```
Lybecker, line 220: Please add this to the https://github.com/microsoft/AzureTRE/blob/develop/resource_processor/vmss_porter/readme.md#network-requirements

templates/workspaces/base/terraform/network/network.tf
deniscep: Done

templates/core/terraform/firewall/firewall.tf
deniscep: Done

Issue #643, opened 2021-08-16
[Story] Add tests that create a new workspace service deployment


deniscep:
As a TRE Developer
I want to ensure that all workspace services are deploying correctly. 
So that I can react to any issues that might arise from system modifications

Task:
Add tests that creates a new workspace service deployment

BLOCKED: due to no service templates being deployed by the provisioning pipeline.
To unblock, add a service template setup to Deploy TRE pipeline.

christoferlof:
Predecessors to unblock: #569 , #693

PR #856, opened 2021-09-15
Workspace service tests

msebragge:
# PR for issue #643

## What is being addressed

E2E test for workspace service creation.

## How is this addressed

New test cases added to e2e_test suite.


TessFerrandez (approved):
As discussed F2F - let's remove the individual test to create workspace as this is already tested as part of this workspace + workspace service test (to avoid incurring build/resource cost + time) 

Other than that LGTM!

TessFerrandez (dismissed):
As discussed F2F - let's remove the individual test to create workspace as this is already tested as part of this workspace + workspace service test (to avoid incurring build/resource cost + time)

Other than that LGTM!

Please note though that this will fail until we fix the Guacamole Firewall Rule - so let's not merge until that is done

TessFerrandez (approved):

PR #890, opened 2021-09-19
Task 841 - Guacamole custom parameters

eladiw:
# PR for issue

## What is being addressed

Populating the custom parameters file of the Guacamole service, so that when using the UI, these params will be shown and will be editable (for creating a customized service)

## How is this addressed

- Added some of the params to the file, such as enable copy / paste, enable drive, etc.
- Params such as acr location / name were not added to the file

### DIFF
--- a/templates/workspace_services/guacamole/custom_parameters.json
+++ b/templates/workspace_services/guacamole/custom_parameters.json
@@ -1,88 +1,52 @@
 {
-    "$schema": "http://json-schema.org/draft-07/schema",
-    "$id": "https://github.com/microsoft/AzureTRE/templates/workspace_services/guacamole/custom_parameters.json",
-    "type": "object",
-    "title": "Guacamole service template",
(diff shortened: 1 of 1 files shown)


templates/workspace_services/guacamole/custom_parameters.json
```diff
@@ -1,11 +1,52 @@
...
+      "title": "Enable Drive",
+      "description": "Enable mounted drive"
+    },
+    "guac_drive_name": {
+      "$id": "#/properties/guac_drive_name",
+      "type": "string",
+      "title": "Mounted drive's name",
+      "description": "The name of the mounted drive"
+    },
+    "guac_drive_path": {
```
marrobi, line 33: Is a user ever going to want to change this?

marrobi (approved):
LGTM, had this been tested - in as much as do the other parameters all get provided by the system?  Thanks.

Issue #899, opened 2021-09-20
[Task] Expose Nexus via App Gateway


deniscep:
Story: #584

Nexus should be available through app gateway on public network

PR #902, opened 2021-09-21
Add Nexus to app gateway

deniscep:
# Expose Nexus via App Gateway #899

## What is being addressed

Nexus should be available through public connection.

## How is this addressed

- Added app gateway configuration for exposing Nexus


marrobi (approved):
LGTM

PR #913, opened 2021-09-22
Expose Gitea via App Gateway

### LINKED ISSUE #911: [Task] Expose Gitea via App Gateway
Story: #584

Gitea should be available through app gateway on public network.

deniscep:
# Expose Gitea via App Gateway

## What is being addressed

Gitea should be available through app gateway on public network.

## How is this addressed

- Added Gitea FQDN to app gateway

Closes #911


Lybecker (approved):
LGTM

Issue #914, opened 2021-09-22
[Task] Fix Porter version


tompaana:
Porter 1.0.0-alpha release will introduce breaking changes to Azure TRE. At the moment the Porter version installed on Resource Processor (see `/resource_processor/vmss_porter/Dockerfile`) is whatever is the latest. This behaviour should be changed to use currently neglected script, `/devops/scripts/install_porter.sh`, that specifies which version to install.

`install_porter.sh`, if used, should be moved under `resource_processor` folder.

PR #917, opened 2021-09-22
Fixes porter version

tompaana:
# PR for issue #914 

## What is being addressed

* Porter version fixed to v0.38.6 to prevent spontaneous breaking change when v1.0.0-alpha becomes the latest
* Removes unused and neglected Porter installation script
* Removes deprecated piece of documentation related to Porter mixins


Lybecker (approved):
LGTM

Issue #909, opened 2021-09-21
[BUG] Azure ML bundle creates DNS zones that may already exist


marrobi:
**Describe the bug**
The Azure ML workspace service creates DNS zones for storage and keyvault - they may already exist due to other services being installed. 

We should move these zones to core or workspace - and maybe the common resources into the base template too.

PR #919, opened 2021-09-23
Move all DNS zones to core (1)

marrobi:
# PR for issue #909

## What is being addressed
Move DNs zones from workspaces and services to core. As only one of each dns zone can be linked to a vnet. For example keyvault private endpoints need to be resolved from workspace and core networks. 

It makes sense to be consistent and have these created in core, and linked as required to workspaces and/or core.

There are going to be a number of PRs. I will split them out as is too much for one. this will break end to end tests so suggest we queue them up and when all ready and reviewed, merge them.


deniscep (approved):
LGTM

PR #921, opened 2021-09-23
Switch base workspace to new ID format, add keyvault, storage and link DNS zones

marrobi:
# PR for issue 908

## What is being addressed

- Switch base workspace to new ID format 
- Add keyvault, storage as we often do not need a dedicated kv per service
- Link DNS zones at workspace level as can only be linked once, not in each service

Tested alongside the core changes.


deniscep (approved):
LGTM

PR #923, opened 2021-09-23
Azure ML Workspace Service Refactor (3)

marrobi:
# PR for issue 908

## What is being addressed
- Move out KV and Storage to base
- Move out private DNS zones
- remove modules to simplify
- new resource id model.


deniscep (approved):
LGTM

Issue #926, opened 2021-09-23
[BUG]  Porter Azure Storage Plugin Issues


marrobi:
**Describe the bug**
When deploying a bundle get the error:

Error context message = Error: could not save the pending action's status, the bundle was not executed: Timed out waiting for the azure blob storage tags cache to warm up

**Steps to reproduce**

This happens randomly, but once happens cannot get to a good state.

Porter team suggest to switch to table storage for the state store. I have tried this and fixes the issue.

EDIT: Spoke to soon the log data is too large for table storage.

PR #927, opened 2021-09-23
Switch to table storage for porter state

marrobi:
# PR for issue #926

## What is being addressed

Issue with using blob storage as per #926


marrobi:
Hit an issue...

Error: 1 error occurred: \t* Data exceeds maximum length for table storage for item: outputs/ group=\"01FGA3V6CPN06GKDZY50T227CP\" 01FGA3V6CPN06GKDZY50T227CP-io.cnab.outputs.invocationImageLogs length: 71382

tompaana (approved):

marrobi:
Aware of still an issue but at least unblocks. Will try to reduce log length as temporary fix.

Issue #916, opened 2021-09-22
[BUG] Azure Firewall "randomly" destroyed by Terraform


marrobi:
**Describe the bug**
A couple of the team have seen Azure Firewall get destroyed by terraform. We are unclear as to why this happens.

It could be linked to https://github.com/hashicorp/terraform-provider-azurerm/issues/11043

I believe I saw it when no dependent resources have changed. However we are using `data` items to retrieve network details so maybe the data changed?

**Things to consider:**
As a first step lets prevent destroy of the firewall (unless in DEBUG mode)? Also lets create some Azure resource locks where appropriate.

@deniscep feel free to add to this.

marrobi:
Closing as now have management locks. If we see the issue again will be able to troibleshoot.

PR #932, opened 2021-09-24
Add prevent destroy to cosmos and firewall

marrobi:
Issue #916

Add prevent destroy to firewall and cosmos resources to ensure are not accidently deleted by terraform.

### DIFF
--- a/templates/core/terraform/firewall/firewall.tf
+++ b/templates/core/terraform/firewall/firewall.tf
@@ -19,7 +19,10 @@ resource "azurerm_firewall" "fw" {
     public_ip_address_id = azurerm_public_ip.fwpip.id
   }
 
-  lifecycle { ignore_changes = [tags] }
+  lifecycle { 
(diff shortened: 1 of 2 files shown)


Lybecker:
Will the `make tre-destroy` still work?

marrobi:
@Lybecker good point. Could wrap them with a condition of based on an ENV var maybe? We need to block them being destroyed accidentally somehow. 

Could use Azure locks, but a destroy could destroy the lock too.

marrobi:
Can't do conditions - https://www.terraform.io/docs/language/meta-arguments/lifecycle.html#literal-values-only so you are correct would block a destroy. Should look at Azure management locks instead.

Lybecker:
Can the make tre-destroy modify the prevent_destroy before the actual TF destroy?

marrobi:
Nope, I'll give this another go with azure locks.

Issue #792, opened 2021-09-08
[BUG] Identity does not have permission to purge secrets from Key Vault


SvenAelterman:
I am using the identity as created by AzureTRE script. When running destroy.sh (`make tre-destroy`), I receive error messages that the identity is unable to purge secrets from Key Vault

**Steps to reproduce**

1. Create identities with the aad-app-reg.sh script.
1. Deploy the TRE with that identity.
1. Attempt to destroy the TRE with that identity.
1. Receive error:
Error: purging of Secret "mysql-aztrehack22-password" (Key Vault "https://kv-aztrehack22.vault.azure.net/") : keyvault.BaseClient#PurgeDeletedSecret: Failure responding to request: StatusCode=403 -- Original Error: autorest/azure: Service returned an error. Status=403 Code="Forbidden" Message="The user, group or application 'appid=04b07795-8ddb-461a-bbee-02f9e1bf7b46;oid=6e41d2a3-ec8f-417d-9fe4-e2f29db038c7;numgroups=11;iss=https://sts.windows.net/6c7dbaa5-c725-4e29-a340-123bdf8d0049/' does not have secrets purge permission on key vault 'kv-aztrehack22;location=eastus2'. For help resolving this issue, please see https://go.microsoft.com/fwlink/?linkid=2125287" InnerError={"code":"ForbiddenByPolicy"}

Note, this error occurs multiple times before Terraform exits with error code 1.

christoferlof:
Add purge permission to policy

PR #935, opened 2021-09-24
KV Purge permissions to deployer account

christoferlof:
## What is being addressed

`make tre-destroy` failed to remove keyvault secrets since deployment account was lacking purge permission

## How is this addressed

- Access policy for _deployer_  account updated to grant purge permission on secrets.
- Nexus and Gitea modules now having dependency on KeyVault module to ensure order of removal.

Closing #792


Lybecker (approved):
LGTM

PR #938, opened 2021-09-24
Update grep statement to improve log match on TF errors

### LINKED ISSUE #829: [BUG] Clean CI/CD run fails even if no actual error occured
**Describe the bug**

**terraform_wrapper.sh**

```bash

    TF_ERROR=$(cat tmp$$.log |  grep -c 'Error') || true; 
    if [[ $LOCKED_STATE > 0  ]];
    then
        RUN_COMMAND=1
        echo "Error acquiring the state lock"
        sleep 10
    elif [[ $TF_ERROR > 0  ]];
    then
        echo "Terraform Error"
        exit 1
    fi
```

log file has lines that contain word error such as
2021-09-13T09:59:54.6261442Z [0m[1mmodule.gitea[0].null_resource.webapp_vault_access_identity (local-exec):[0m [0m      "detailedErrorLoggingEnabled": null,

which are not related to an actual error

**Steps to reproduce**

1. Setup clean CI/CD pipeline / environment
2. Run workflow / make all
3. Run fails

christoferlof:
## What is being addressed

`terraform_wrapper.sh` reports error even though execution completed successfully. See #829 

## How is this addressed

- Update grep statement to match on `'Error '` instead, giving a better match, excluding the messages that previously generated the false positives.
- Closes #829

### DIFF
--- a/devops/scripts/terraform_wrapper.sh
+++ b/devops/scripts/terraform_wrapper.sh
@@ -54,27 +54,27 @@ while [ "$1" != "" ]; do
 done
 
 
-if [[ -z ${mgmt_resource_group_name+x} ]]; then  
+if [[ -z ${mgmt_resource_group_name+x} ]]; then
     echo -e "No terraform state resource group name provided\n"
     usage
 fi
 
-if [[ -z ${mgmt_storage_account_name+x} ]]; then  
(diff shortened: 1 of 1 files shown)


marrobi (approved):
Basic, but LGTM

Issue #132, opened 2021-05-31
[Story] Refactor InnerEye AML spike into Workspace bundle


christoferlof:
## Description

As a cloud administrator
I want to deploy an InnerEye workspace
So that I can provide that to our researchers for their demo and training purposes

## Acceptance critiera

- [x] Spike *[InnerEye deployment to Azure ML in closed vnet](https://github.com/microsoft/AzureTRE/tree/features/InnerEyeAMLPLDeployment)* refactored into a Workspace bundle #205 #207 
- [x] Workspace definition and Workspace services clearly separated in the bundle
- [x] Documentation on resources being provisioned. #249 
- [x] Workspace bundle deployed to demo environment using Porter directly.
- [x] #Refactor innereye and AML to single workspace service

christoferlof:
Also see #19 for additional context

PR #940, opened 2021-09-24
Create InnerEye workspace template and remove redundant templates (6)

marrobi:
# PR for issue #908 and story #132

## What is being addressed

Create an innereye workspace template that deploys the base workspace and the inner eye service template.

Deleted the old inner eye service tempaltes and redundant workspace temapltes as the service  deployment now works.

I have tested deployment via the API and all looks good. It might be decide to get rid of this template too and stick to workspace services.


deniscep (approved):
LGTM

PR #950, opened 2021-09-28
Add initial Azure Virtual Desktop with AAD Join Workspace Service

marrobi:
# PR for issue #1014

## What is being addressed

Add basic AVD service with AAD join using bicep templates.

## Work to do
- [ ] Uninstall action in bundle needs to work - might need all items tagging with the service_id?
- [ ] upload artifacts to accessible storage location ?
- [ ] add firewall rules - https://docs.microsoft.com/en-us/azure/firewall/protect-azure-virtual-desktop
- [ ] Readme moving to docs and having bundle specific documentation
- [ ] Testing(!)
- [ ] User Assignment - " desktop virtualization user" role assignment to application group and Virtual Machine User login to the VM,


SvenAelterman:
Additional work noted:

- [ ] Remove localAdminPassword param and generate a new unique value to be stored in the hub's KV.
- [ ] Output the needed hub firewall rules as `rules_collection`.
- [ ] Create unique ARM deployments by appending the service ID and the deployment timestamp.
- [ ] Update the `connection_uri` output from Bicep to a shortlink, or update to the April 2022 AVD endpoint URL.
- [ ] Fix the `customRdpProperty` on the hostpool for TRE purposes (disable drive redirection, disable clipboard)
- [ ] Attempt to configure the session hosts to disable drive redirection and clipboard with policy.
- [ ] Create a parameter for the friendly name of the workspace and the desktop application because those are user-facing. Provide reasonable defaults.

marrobi:
Superseded by #1865

PR #952, opened 2021-09-29
Add management locks and disable porter logs

marrobi:
# PR for issue #916 #926

## What is being addressed
- Upgrade porter version and disable logging from porter to prevent exceeding table storage limits
- Remove porter output folder
- Add management locks to firewall and cosmos db

If firewall tries to get be destroyed then not sure what will happen.


tompaana (approved):
Minor suggestion: Line 195 in `deploy_tre.yml` could be updated too from 0.38.6 to 0.38.7.
Otherwise, looks good!

PR #958, opened 2021-09-30
Update innereye_service string

deniscep:
E2E test was referencing a wrong string for innereye_service testing. Removed as quick fix, now correcting to the correct string.


marrobi (approved):
LGTM

PR #959, opened 2021-09-30
Fixes deployment workflow for merges and updates docs

### LINKED ISSUE #957: [BUG] Deployment workflow does not run when merged
The newly added conditions to `deploy_tre.yml` cause it to **not** run after merge. An additional condition is required or an alternative fix.

tompaana:
# Fixes #957


marrobi (approved):
LGTM

Issue #1003, opened 2021-10-07
[Task] Refine error messages being returned in aad_authentication.py


marrobi:
Story: XX

**Description**
The error returned here: https://github.com/microsoft/AzureTRE/blob/1e5edc698c3c48d88df8fd2a6a9c71b95566eeda/api_app/services/aad_authentication.py#L39

Can be caused by an issue with decoding the token:
 https://github.com/microsoft/AzureTRE/blob/1e5edc698c3c48d88df8fd2a6a9c71b95566eeda/api_app/services/aad_authentication.py#L35

Or an issue getting user roles from Graph:

https://github.com/microsoft/AzureTRE/blob/1e5edc698c3c48d88df8fd2a6a9c71b95566eeda/api_app/services/aad_authentication.py#L36

Can the detail of the 401 be different depending on which call fails.

PR #1005, opened 2021-10-07
add separate messages for if we can't decode the token or if we can't…

TessFerrandez:
… get role assignments

# PR for issue #1003 

## What is being addressed

more detailed error messages when we get a 401


marrobi (approved):
LGTM

LizaShak (approved):

Issue #1011, opened 2021-10-07
[BUG] - api-webapp is not fetching the docker image


eladiw:
Remember to label the issue with the appropriate area, and also assign it to yourself if you plan to work on the issue in the near future.

**Describe the bug**
the web app is failing to fetch the docker image. 
This is due to a bug (probably) that fails fetching the image using managed identity

**Steps to reproduce**

1. deploy tre
2. check the deployment center tab (Azure)
3. verify the image was not pulled

eladiw:
correct

PR #1012, opened 2021-10-07
api-webapp docker fetching is done using username & password

eladiw:
#1011 

## What is being addressed

docker image is not pulled

## How is this addressed

- switched to setting the username and password instead of managed identity


deniscep (changes requested):
looks good, and approved, but blocking merge until we test 2.80 Azure TF provider update

Issue #1008, opened 2021-10-07
[Task] Deployment creates AMPLS and connects the assets to it


tompaana:
Story: #742 

Azure Monitor Private Link Scope (AMPLS) needs to be created by deployment and assets (App Insights and Log Analytics) need to be associated with it.

This has to be taken into consideration: Deleting Azure Monitor resources requires that you first disconnect them from any AMPLS objects they are connected to. It's not possible to delete resources connected to an AMPLS.

Terraform does not support [AMPLS](https://docs.microsoft.com/en-us/azure/azure-monitor/logs/private-link-security) so a script is needed.

PR #1023, opened 2021-10-11
Deployment creates Azure Monitor Private Link Scope

tompaana:
# PR for issue #1008 

* Azure Monitor resources moved into its own Terraform module
* Azure Monitor module creates AMPLS


marrobi (approved):
LGTM

What is the plan within the workspaces?

PR #1027, opened 2021-10-11
Store and return template title

### LINKED ISSUE #708: [Story] Store and return template display name
## Description

As a UI developer
I want the API to return the friendly name of a workspace template 
So that I can display it to users

## Acceptance criteria

- [ ] Extract title from template schema json document and sent in registration payload as `display_name`
- [ ] `display_name` stored in state store 
- [ ] API operations for get template/s returns returns the title as `display_name` in addition tot he existing `name` field.

marrobi:
Fixes #708

## What is being addressed

Enable template titles to be stored and returned

## How is this addressed

- Add template title to resource_template to db function, domain models and schemas
- Added default of "" for templates with no title or description in the db
- Update tests


api_app/db/repositories/resource_templates.py
```diff
@@ -35,7 +35,7 @@ def get_templates_information(self, resource_type: ResourceType, parent_service_
         """
         Returns name/description for all current resource_type templates
```
TessFerrandez, line 36:
```suggestion
        Returns name/title/description for all current resource_type templates
```

Issue #1028, opened 2021-10-11
[BUG] Incorrect case in Core App Service Plan


tamirkamara:
**Describe the bug**
The case of the App Service plan size in incorrect which causes redundant updates each time terraform apply is called.

PR #851, opened 2021-09-14
Simple capitalization fix for App Svc size value

SvenAelterman:
## What is being addressed

No behavior modification. Fix `P1v3` to `P1V3` to be consistent with commonly used capitalization standards.

PR #1029, opened 2021-10-11
Fix app service plan case (revert #851)

tamirkamara:
# PR for issue #1028 

## What is being addressed

The size definition of the app service plan should be exact to remove redundant updates on terraform apply


marrobi (approved):
LGTM - odd as the docs have it in lower case.

PR #1032, opened 2021-10-12
Provide via API the current state of a user resource

### LINKED ISSUE #757: Provide via API the current state of a user resource
Feature #18 

## Description

As a Researcher
I want to see the current (as in Azure) state of a user resource
So that I know if my resource, e.g. VM is running or stopped, or maybe starting.

## Acceptance criteria

- [ ] Get user resource API returns current state

## Note
Maybe store the resourceId via a bundle output. Depending on the resourceType get the status from Azure rest API.

marrobi:
Fixes #757

## What is being addressed

Add azure resource status to user resources

## How is this addressed

- modify route to add status if azure_resource_id exists
- add azureStatus to user resource models
- update requirements.txt files with newer versions (required some due to old azure packages)
- added servcie, currently only retrieves status for Vms, can be extended later
- add azure_resource_id as an output to guacamole vm
- add tests

### DIFF
--- a/api_app/requirements.txt
+++ b/api_app/requirements.txt
@@ -1,6 +1,6 @@
 # API
-aiohttp==3.7.4.post0
 azure-core==1.19.0
+aiohttp==3.7.4
 azure-cosmos==4.2.0
 azure-identity==1.6.1
 azure-mgmt-cosmosdb==6.4.0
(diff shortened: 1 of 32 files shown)


api_app/requirements.txt
```diff
@@ -1,16 +1,17 @@
 # API
 aiohttp==3.7.4.post0
```
TessFerrandez, line 2: This needs to change back to 3.7.3 or 3.7.4 - dependabot is complaining -- note this also exists in the resource processor reqs

TessFerrandez (approved):
LGTM

PR #1053, opened 2021-10-20
Add APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL = True and Tracing Middleware class

marrobi:
## What is being addressed

Set APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL = True and tracing middleware class to resolve high CPU issues.

## How is this addressed
- Add APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL = True to terraform
- Add TracerMiddleware class
- Add additional metrics
- [x] Run for 24 hours to see if issue resolved.


marrobi:
Looks like CPU is still ramping up, this code change was deployed at the red line. Will monitor:
image

api_app/main.py
```diff
@@ -48,6 +43,11 @@ def get_application() -> FastAPI:
     application.add_event_handler("startup", create_start_app_handler(application))
     application.add_event_handler("shutdown", create_stop_app_handler(application))
 
+    try:
+        application.middleware("http")(RequestTracerMiddleware(application, exporter=AzureExporter(connection_string=f'InstrumentationKey={os.getenv("APPINSIGHTS_INSTRUMENTATIONKEY")}', sampler=ProbabilitySampler(1.0))))
```
martinpeck, line 47: What is happening on this line?

It looks like, instead of calling the `add_middleware` API, passing in a Middleware instance, you're calling the `middleware` decorator, getting back the decorator, and with this you're then injecting your Callable. That feels wrong.

This is what you're calling...

https://github.com/encode/starlette/blob/6c9bc55af7d9d202c2bc1299c9a65e844640c197/starlette/applications.py#L192

... might be better to call `add_middleware` (as you do elsewhere here).

Either way...the code looks strange to me (even though it may work)

marrobi: Thanks. I took most of this from a sample, but did try changing to use `add_middleware` myself as agree looked off, but could not see how to specify that it is `http` middleware, unless I can do something like:

```python
application.add_middleware(BaseHTTPMiddleware, dispatch=RequestTracerMiddleware(application, exporter=AzureExporter(connection_string=f'InstrumentationKey={os.getenv("APPINSIGHTS_INSTRUMENTATIONKEY")}', sampler=ProbabilitySampler(1.0)))) 
```

Taking a hint from https://github.com/encode/starlette/blob/6c9bc55af7d9d202c2bc1299c9a65e844640c197/starlette/applications.py#L198 ?

martinpeck: all the "http" does is add the BaseHTTPMiddleware, so yeah....that code you've got there looks right

martinpeck: well...actually....why do you want to use the BaseHTTPMiddleware? Why not just add the RequestTracerMiddleware directly?

marrobi: Thanks - changes in.

martinpeck (approved):
LGTM. With those changes the code in main.py is much cleaner/idiomatic

Issue #1052, opened 2021-10-19
[Task] Create private endpoint in Core VNET (SharedSubnet) for AMPLS to connect to


tompaana:
Story: #742

PR #1054, opened 2021-10-20
Configures network for AMPLS

tompaana:
# PR for issue #1052 

## Changes

* Creates private DNS zones required by Azure Monitor
* Creates a private endpoint for the AMPLS to `SharedSubnet`
* Creates a dedicated storage account for App Insights (required when AMP has Private Link enabled)
  * `azurerm_application_insights` does not support linking storage account so had to switch to ARM template deployment


marrobi (approved):
If it works, then looks good :)

Not an easy one. We need to verify we get requests into app insights once merged.

Issue #835, opened 2021-09-13
[Story] Services should use WS app registration for authentication


eladiw:
## Description

As a software engineer
I want to make sure that when accessing deployed ws services the token will be validated against the matching ws app registration
So that the auth flow will be according to best practice and won't require additional calls to the graph api.

Currently when a token is sent to a ws service, (e.g. guacamole vm), it is verified using the TRE API app registration. Then an additional call to AAD (graph api) is done in order to finish the flow.

## Acceptance critiera

- [x] services/ws API operations remain AS IS (still using the TRE API app reg)
- [ ] Deployed services should be injected with the auth details (possibly: app reg id, app secret, redirect url) so that the service will have this information for token validation
- [x] Guacamole service (which is a specific example of the general solution) will switch from the TRE API app reg to the given ws app reg (using the injected values from the prev acceptance criteria)
- [x] Amend TRE app registration creation script to work for workspaces

Technical notes:
1. This extension might be used for guacamole (https://guacamole.apache.org/doc/gug/openid-auth.html). It should be tested.
2. Second option for Guacamole will be to use the oauth proxy project (https://github.com/oauth2-proxy/oauth2-proxy). this option is the lower priority option and should be checked after the guacamole extension. for this option a guacamole image change is needed

PR #1058, opened 2021-10-24
Guacamole OpenID extension support

eladiw:
#835 

## What is being addressed

Guacamole currently accepts a header containing a token. This token is passed as-is to the api app to get the list of VMs.
This PR suggests to use the OpenID extension which is configured to get a token from the Azure AD provider (can be extended to other providers), and the validation part is done by this new extension.
This PR builds on top another feature, which is using the correct app registration in the api app.

## How is this addressed

- Installed the OpenID extension
- Removed the authorization part from TREs custom extension and instead got the token from the oid extension

## Notes
a new param was added to the porter schema : 'openid_client_id' which must be passed for Guacamole to operate correctly.


eladiw:
Changed to draft so it won't be merged before #1062

marrobi (approved):
LGTM - lets get this merged into the feature branch.

Issue #983, opened 2021-10-05
Support workspace tokens in API


LizaShak:
## Description

#985 is a pre-decessor

As a workspace owner
I want to be able to send API requests to edit/create workspace resources using different workspaces app registrations
So that I can access the management API and authorized based on the workspace app registration

## Technical notes

- When we create a workspace and appId param is passed on which represents the app registration used for this workspace
- Having a token issued for workspace
- Send  a request to create workspace service with the issued token to the management api
- Take the aud claim from the token
- Query the cosmos resources collection with the workspaceId==workspace_id on the request url and authorize the call only if appId param equals the aud claim. 

## Acceptance criteria

- [x] API requests, made on TRE level (creation of: shared services,  workspace... ) remain the current authentication logic(the access is made through SWAGGER UI app registration and the token aud claim is TRE API)
- [x] Create workspace service API with token issued for a different workspace gets 401 Unauthorized response.
- [x] Create workspace service API with token issued for the workspace but is expired gets 401 Unauthorized response.
- [x] Create workspace service API with a valid token issued for the workspace is authorized and the workspace service is created.

- [x] Update docs to match new experience
- [x] #1076
- [x] Ensure end to end tests work

marrobi:
@LizaShak end to ends passed so thing good to close this 👍

PR #1062, opened 2021-10-27
Support workspace tokens in API

LizaShak:
# PR for issue #983

## What is being addressed

This change allows the api to authenticate workspaces app registrations tokens.

## How is this addressed

- Apply on workspaces specific routes - meaning routes that require workspace_id as part of the url
- When token issued for workspace is received
- Take the aud claim from the token
- Query the cosmos resources collection with the workspaceId==workspace_id on the request url and authorize the call only if appId param equals the aud claim.


TessFerrandez (approved):
LGTM! Haven't tested the changes, but assuming they work fine, the code looks good

e2e_tests/conftest.py
```diff
@@ -33,3 +33,18 @@ async def token(verify) -> str:
...
+@pytest.fixture
+async def workspace_owner_token(verify) -> str:
+    async with AsyncClient(verify=verify) as client:
+
+        headers = {'Content-Type': "application/x-www-form-urlencoded"}
+        payload = f"grant_type=password&resource={config.AUTH_APP_CLIENT_ID}&username={config.USERNAME}&password={config.PASSWORD}&scope=open_id+profile+email&client_id={config.AUTH_APP_CLIENT_ID}"
+
+        url = f"https://login.microsoftonline.com/{config.AUTH_TENANT_ID}/oauth2/token"
+        response = await client.post(url, headers=headers, data=payload)
+        token = response.json()["access_token"]
```
marrobi, line 47: @LizaShak I'm getting "invalid key" for `access_token` here: https://github.com/microsoft/AzureTRE/runs/4301934706?check_suite_focus=true

Any ideas?

marrobi: Sorted, issue with `client_id` in above command - was set to the workspace app ID. Also app registrations needed scope adding and admin consent granting.

marrobi (approved):
Happy with this.

eladiw (approved):

PR #1108, opened 2022-01-07
Converted dockerfile to arch agnostic

jjgriff93:
# PR for issue

## What is being addressed

Currently the dev container will only run on x86/x64 architectures and will not work on arm32 or arm64 architectures (like the newer Mac M1 chipsets). This PR makes the dev container Dockerfile architecture agnostic so it can run on both.

## How is this addressed

- Changed the container base image from the older buster variant to the latest default (bullseye) which works cross-platform
- Converted Terraform and Powershell installs from apt repositories to zip/gz packages as their apt repositories don't currenctly have any arm images available
- Fixed some arm dependency issues with python wheel installs
- Installed az cli manually and pinned version 2.18.0 as this is the only version currently available supporting both Debian amd & arm64

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -2,7 +2,7 @@
 
 # [Choice] Python version: 3, 3.9, 3.8, 3.7, 3.6
 ARG VARIANT="3"
-FROM mcr.microsoft.com/vscode/devcontainers/python:dev-${VARIANT}-buster
+FROM mcr.microsoft.com/vscode/devcontainers/python:dev-${VARIANT}
 
 # [Option] Install Node.js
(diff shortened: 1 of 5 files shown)


Lybecker:
@jjgriff93 have your build and deployed the solution from scratch with bullseye? We had some issues with AZ CLI on bullseye and therefore downgraded to buster.

jjgriff93:
> @jjgriff93 have your build and deployed the solution from scratch with bullseye? We had some issues with AZ CLI on bullseye and therefore downgraded to buster.

@Lybecker I found that version 2.18.0-2 of the az cli worked at least initially; however later in the deployment it failed due to not having the az acr subcommands. Abandoning for now until the az cli supports ARM

Issue #1112, opened 2022-01-11
Powershell missing from devcontainer


daltskin:
When running make `tre-start` or `make tre-stop` an error is displayed.

**Steps to reproduce**

1. Build devcontainer
1. Run `make tre-start`
1. Script will error with: 
     The term 'Set-AzContext' is not recognized as a name of a cmdlet, function, script file, or executable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.

This is because the Powershell Az module is missing from devcontainer

PR #1113, opened 2022-01-11
#1112 Install az module for all users

daltskin:
# PR for issue #1112 

Install PowerShell Az module for all users

Resolves issue within devcontainer with az commands not being found


marrobi (approved):
LGTM

Issue #1115, opened 2022-01-11
Run `make tre-stop` at the end of the TRE Deploy github workflow


marrobi:
## Description

As a Azure Administrator
I want to shut down/deallocate resources at the end of a CD run
So that Azure spend is minimized

## Acceptance criteria

- [x] After e2e tests run - pass or fail - run `make tre-stop`
- [x] Ensure that all resources are started prior to e2e tests running.

marrobi:
I tried this, but need to handle az logon. Might be worth getting #1132 sorted first.
 
 ```
  tre_stop:
      name: "Deallocate Resources"
      runs-on: ubuntu-latest
      environment: Dev
      needs: [e2e_tests]
      if: |
        ${{
        always() &&
        github.event.name == 'schedule'
        || github.event_name == 'push'
        || github.event_name == 'workflow_dispatch'
        || contains(github.event.pull_request.labels.*.name, 'safe to test')
        }}
      steps:
        - name: Checkout code
          uses: actions/checkout@v2
        - name: Azure Login
          uses: azure/login@v1
          with:
            creds: ${{ secrets.AZURE_CREDENTIALS }}
        - name: Deallocate Resources
          shell: bash
          env:
            TRE_ID: "${{secrets.TRE_ID}}"
          run: |
            export ARM_SUBSCRIPTION_ID=$(echo "$AZURE_CREDENTIALS" | jq -r '.subscriptionId')
            pwsh -c Install-Module Az -Force -Scope CurrentUser
            make tre-stop
```

marrobi:
Fixed in #973

PR #1116, opened 2022-01-11
Fix Azure charges

ChaseKnowlden:
# PR for issue #1115 

## What is being addressed

GitHub Actions deploys, but doesn't stop and costing a lot of money.

## How is this addressed

- Changed Deploy Strategy to stop the process at the end of the run so it doesn't charge $1,000 easily.

### DIFF
--- a/.github/workflows/deploy_tre.yml
+++ b/.github/workflows/deploy_tre.yml
@@ -81,6 +81,7 @@ jobs:
           export TF_VAR_arm_subscription_id=$ARM_SUBSCRIPTION_ID
 
           make all
+          make tre-stop
 
       - name: Notify dedicated teams channel
         uses: sachinkundu/ms-teams-notification@1.4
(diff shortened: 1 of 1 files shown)


marrobi:
Hi @ChaseKnowlden, thanks for the PR, coincidentally we created an issue earlier today for this here:  https://github.com/microsoft/AzureTRE/issues/1115

We will get it tested and merged of all is good. Thanks.

marrobi:
@ChaseKnowlden Taking a closer look it seems you have added the command before the end to end tests run.

This step should be added at the end of the workflow, and run unconditionally, something like (not tested):

```
  tre_stop:
    name: "Deallocate Resources"
    runs-on: ubuntu-latest
    environment: Dev
    if: |
      github.event.name == 'schedule'
      || github.event_name == 'push'
      || github.event_name == 'workflow_dispatch'
      || contains(github.event.pull_request.labels.*.name, 'safe to test')
    steps:
      - name: Deallocate Resources
        if: always()
        shell: bash
        env:
          # required env vars here
        run: |
          make tre-stop
```

I also think we will need `make tre-start` after the `make all` command to ensure everything is running before the tests.

Issue #1122, opened 2022-01-13
When registering user resource template and workspace service template is not found exception not clear


marrobi:
**Describe the bug**
When try and register a user resource template, but the workspace service template does not exist the error message is not clear. 

`TEMPLATE_DOES_NOT_EXIST = "Could not retrieve the 'current' template with this name"`

need to add 

`WORKSPACE_SERVICE_TEMPLATE_DOES_NOT_EXIST = "Could not retrieve the workspace service template specified"`

gauravagrwal:
Hi @marrobi, here we need to replace the current error message with the new one. right?

PR #1124, opened 2022-01-13
updated the error message

gauravagrwal:
# PR for issue
#1122 

## What is being addressed

- Updating the error message when the workspace service template is not found

## How is this addressed

- Updated the `strings.py` file where all the error messages are listed,
- and updated the `workspace_service_templates.py` and `workspace_templates.py` which were referencing that error message.

### DIFF
--- a/api_app/api/routes/workspace_templates.py
+++ b/api_app/api/routes/workspace_templates.py
@@ -22,7 +22,7 @@ def get_current_template_by_name(template_name: str, template_repo: ResourceTemp
         template = template_repo.get_current_template(template_name, resource_type, parent_service_template_name)
         return template_repo.enrich_template(template)
     except EntityDoesNotExist:
-        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=strings.TEMPLATE_DOES_NOT_EXIST)
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=strings.WORKSPACE_SERVICE_TEMPLATE_DOES_NOT_EXIST)
     except DuplicateEntity:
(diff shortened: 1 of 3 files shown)


api_app/api/routes/workspace_templates.py
```diff
@@ -22,7 +22,7 @@ def get_current_template_by_name(template_name: str, template_repo: ResourceTemp
         template = template_repo.get_current_template(template_name, resource_type, parent_service_template_name)
         return template_repo.enrich_template(template)
     except EntityDoesNotExist:
-        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=strings.TEMPLATE_DOES_NOT_EXIST)
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=strings.WORKSPACE_SERVICE_TEMPLATE_DOES_NOT_EXIST)
```
marrobi, line 25: I believe this one needs to remain as `resource_type: ResourceType` might be of a different type.

marrobi (changes requested):

marrobi:
@gauravagrwal thank you for the PR. Believe a small change is required. Thanks.

api_app/api/routes/workspace_templates.py
gauravagrwal: okay, so both error messages(`TEMPLATE_DOES_NOT_EXIST` and `WORKSPACE_SERVICE_TEMPLATE_DOES_NOT_EXIST`) will be there. right?

marrobi: In this place just `TEMPLATE_DOES_NOT_EXIST` as we do not know the resourceType. Ideally the `resourceType` passed in would be substituted into the error but happy with either for now.

marrobi:
@gauravagrwal all looks good. We have an issue with our end to end tests, will get that sorted then rerun your PR test and merge. Thanks for the contribution.

martinpeck (approved):
LGTM. Thanks

Issue #1139, opened 2022-01-19
Extract ENV variables from TF output json


ross-p-smith:
Whilst debugging the letsencrypt and static files, it would be quicker if we output terraform to a json file and then convert them into an env rather than calling `terraform output` for each variable

PR #1140, opened 2022-01-19
Convert Tf output to env

ross-p-smith:
This is for #1139 

When we build the infrastructure through terraform, we will create a tre_output.json

When we need values from terraform, we will use this file to create a tre.env file and load in the variables.

If the tre_output.json is not present it will create it so that build jobs can be parallelised and not dependant on the file


marrobi (approved):
LGTM

Issue #1141, opened 2022-01-19
Documentation states to use oidc-redirect instead of oauth2-redirect


ross-p-smith:
When setting up your tenant for authentication it states to use a oidc-redirect

**Describe the bug**
This did not work and you have to use oauth2-redirect

PR #1142, opened 2022-01-19
Updating help/documentation with oauth2

ross-p-smith:
#1141 

## What is being addressed

Fixing the documentation and help text in script to use oauth2-redirect


marrobi (approved):
Good catch

Issue #700, opened 2021-08-23
When running make tre-deploy in succession terraform updates/replaces resources that have not changed.


marrobi:
**Describe the bug**
Each time run make tre-deploy terraform updates/replaces resources unnecessarily.

**Steps to reproduce**

1. Deploy TRE with `make`
2. Re run `'make tre-deploy' and monitor terraform output.

christoferlof:
Possibly caused by policies changing resources after deployment?

marrobi:
Step one might be to flatten the terraform to remove the number of dependencies in each module.

martinpeck:
Examples of this:

- firewall gets deleted during deployment
- VM scalesets get deleted
- private endpoints get deleted

Suspect this is due to data resources in terraform.

Next steps:
- refactor scripts so that a plan can be easily run and compared with previous plan
- inspect TF files for obvious issues
- run multiple deployments

marrobi:
Here's an occurrence:

```
module.resource_processor_vmss_porter[0].azurerm_role_assignment.subscription_owner must be replaced
-/+ resource "azurerm_role_assignment" "subscription_owner" ***
      ~ id                               = "/subscriptions/***/providers/Microsoft.Authorization/roleAssignments/deb370d9-e9c0-6ede-a44b-1f1ee3d2146d" -> (known after apply)
      ~ name                             = "deb370d9-e9c0-6ede-a44b-1f1ee3d2146d" -> (known after apply)
      ~ principal_type                   = "ServicePrincipal" -> (known after apply)
      ~ role_definition_id               = "/subscriptions/***/providers/Microsoft.Authorization/roleDefinitions/8e3af657-a8ff-443c-a75c-2fe8c4bcb635" -> (known after apply)
      ~ scope                            = "/subscriptions/***" -> (known after apply) # forces replacement
      + skip_service_principal_aad_check = (known after apply)
        # (2 unchanged attributes hidden)
    ***
```

marrobi:
If look at the multiple attempts on this run various things are going on. The firewall was deleted even though it should have a lock on it.

https://github.com/microsoft/AzureTRE/runs/4895836850?check_suite_focus=true

tanya-borisova:
> Here's an occurrence:
> 
> ```
> module.resource_processor_vmss_porter[0].azurerm_role_assignment.subscription_owner must be replaced
> -/+ resource "azurerm_role_assignment" "subscription_owner" ***
[...]

PR #1143, opened 2022-01-19
Remove unused references to subnets in modules

ross-p-smith:
#700 

When I was looking at the subnets - the app_gw subnet is passed into two modules and not used.


marrobi (approved):
LGTM

Issue #1132, opened 2022-01-17
Start/Stop TRE requires you to log into PowerShell when you have been using bash up to this point


ross-p-smith:
Using the make files assumes that you have used `az login` at the start. When you want to `make tre-stop`, you then have to log into PowerShell.

1. az login
2. make all
3. make tre-stop

daltskin:
Thinking is to remove the PowerShell script and run the commands in bash - removing the need for PowerShell altogether.   However the Azure Firewall PowerShell commands for allocating/deallocating are not available in az cli (this has been raised here: https://github.com/Azure/azure-cli-extensions/issues/4301)

Looking at the PowerShell $firewall.[Deallocate()](https://github.com/Azure/azure-powershell/blob/cd035673d8152c56c89fc91f4f77779918d8d3e4/src/Network/Network/Models/AzureFirewall/PSAzureFirewall.cs#L211) method - it's setting the IpConfigurations to an empty list: 

Which could be achieved by running the following az command: https://docs.microsoft.com/en-us/cli/azure/network/firewall/ip-config?view=azure-cli-latest#az-network-firewall-ip-config-delete

The $firewall.[Allocate()](https://github.com/Azure/azure-powershell/blob/cd035673d8152c56c89fc91f4f77779918d8d3e4/src/Network/Network/Models/AzureFirewall/PSAzureFirewall.cs#L144) method creates a new ManagementIPConfiguration which could be achieved using [az network firewall ip-config delete](https://docs.microsoft.com/en-us/cli/azure/network/firewall/ip-config?view=azure-cli-latest#az-network-firewall-ip-config-create)

az network application gateway start/stop is already supported in az cli: https://docs.microsoft.com/en-us/cli/azure/network/application-gateway?view=azure-cli-latest#az-network-application-gateway-start

@marrobi thoughts?

marrobi:
If we can do without the PowerShell and stops it costing, all good with me!

PR #1149, opened 2022-01-20
Remove PowerShell from container

daltskin:
# PR for issue #1132 

## What is being addressed

Removing need for PowerShell preventing need to login to azure multiple times:

- FW & AGW are now "stopped" from incurring costs via az cli in bash.

## How is this addressed

See proposal in #1132 discussion


devops/scripts/control_tre.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+set -e
+
+if [[ "$1" == *"start"* ]]; then
+  echo -e "Starting Firewall - creating ip-config"
+  az network firewall ip-config create -f "fw-$TRE_ID" -g "rg-$TRE_ID" -n "AzureFirewallIpConfiguration0" --public-ip-address "pip-fw-$TRE_ID" --vnet-name "vnet-$TRE_ID" > /dev/null
```
marrobi, line 6: Should it check if it already has an IP config before creating one? What is the behavior if the firewall is already "started"?

marrobi: I'm just trying this, any idea how long it should take? Is there any output to let the user know something is happening or if any errors? I see the output is sent to `/dev/null`.

marrobi (commented):
Couple of questions...

devops/scripts/control_tre.sh
daltskin: It can take >5 mins to run - check you've rebuilt your container :)

When I was debugging locally I was frequently starting/stop - but I'll add an initial check to see if there is one there already

daltskin: Added check for az cli azure-firewall extension within check_dependencies.
Added additional status checks to reduce execution time

marrobi (approved):
LGTM

PR #1152, opened 2022-01-20
Run `make tre-stop` after e2e tests and add `make tre-start` after deploy

marrobi:
# PR for issue #115

## What is being addressed
Stop Azure resoures after e2e tests run. Also ensure they are runnign after a deployemnt and before the tests.


marrobi:
Deployed and deallocate works here: https://github.com/microsoft/AzureTRE/runs/4889247794?check_suite_focus=true

.github/workflows/deploy_tre.yml
```diff
@@ -84,6 +84,9 @@ jobs:
 
           make all
 
+          az extension add --name azure-firewall
+          make tre-start
```
daltskin, line 88: You shouldn't need to explicitly run tre-start straight after a deployment - but no harm in running as it'll just check/report the current status.

It might be useful to separate out another make command that reports overall TRE status eg. `make tre-status`

Makefile
```diff
@@ -117,14 +117,14 @@ letsencrypt:
 
 tre-stop:
 	echo -e "\n\e[34m»»» 🧩 \e[96mStopping TRE\e[0m..." \
-	&& . ./devops/scripts/check_dependencies.sh \
+	&& . ./devops/scripts/check_dependencies.sh azfirewall \
```
daltskin, line 120: Initial thinking was that azure-firewall extension is a base dependency with any TRE development/deployment, just like az cli.  Since, if deploying it's essential you can stop/start it to manage costs.  Having explicit reference here is good idea though.

.github/workflows/deploy_tre.yml
marrobi: So I was thinking if App Gateway (or in future VMs) is deallocated and the TF hasn't touched it it may be stopped at this point.

devops/scripts/check_dependencies.sh
```diff
@@ -12,12 +12,13 @@ if [ $? -ne 0 ]; then
...
-echo -e "\n\e[96mChecking for Azure CLI extension(s)\e[0m..."
-az extension show -n azure-firewall > /dev/null 2>&1
-if [ $? -ne 0 ]; then
-  echo -e "\e[31m»»» ⚠️ Azure CLI azure-firewall extension is not installed! 😥 Please go to http://aka.ms/cli to set it up"
-  exit
+if [[ "$1" == *"azfirewall"* ]]; then
+  echo -e "\n\e[96mChecking for Azure CLI extension(s)\e[0m..."
+  az extension show -n azure-firewall > /dev/null 2>&1
+  if [ $? -ne 0 ]; then
+    echo -e "\e[31m»»» ⚠️ Azure CLI azure-firewall extension is not installed! 😥 Please go to http://aka.ms/cli to set it up"
```
daltskin, line 19: These messages could be more "devcontainer" friendly eg. "Please go to http://aka.ms/cli to set up or rebuild your devcontainer"

Makefile
marrobi: I was running another make command - not start/stop locally and it got me hence thought worth adding.

devops/scripts/check_dependencies.sh
marrobi: You wrote this bit? :D

daltskin (approved):

Issue #865, opened 2021-09-15
Provide operation history to TRE users via API


marrobi:
## Description

As an TRE Administrator I want to be able to view the operation history for all resources so that I can see who created/updated/deleted what, when and the outcome of the operation.

As an Researcher I want to be able to view the operation history for my resources so I can see when I created/updated/deleted resources and the outcome of the operation.

As an Workspace Owner I want to be able to view the operation history for all resources in a workspace so that I can see who created/updated/deleted what, when and the outcome of the operation.

damoodamoo:
Proposal:
- Update resource document in Cosmos to include `operations: []` (structure below)
- An operation is a create / update / delete of the resource
- API adds a new entry in list each time
- Each operation has a unique Id and newest ones are appended to the top of the list, so clients can query for the outcome of a specific operation if needed
- This replaces the current `deployment` object

```json
{
  "id": "guid....",
  ...,
  "operations": [
    { 
      "id": "guid-here",
      "status": "waiting",
      "message": "etc",
      "type": "update",
      "triggered_by": "...",
      "triggered_when": datetime,
      "completed_when": datetime
      ...
    },
    { 
      "id": "guid-blah",
      "status": "complete",
      "type": "create",
      ...
    }
  ]
```

marrobi:
@damoodamoo Are operations always associated with a resource?

Should these operations live under a resource or as a new collection in cosmos, referencing the resources.

Other than that, looks good.

damoodamoo:
ok good question - my thinking was that operations would always relate to a resource - an operation being 'an action that happens to a resource'. If 'Add User' was an operation, would that not also relate to a resource, or not?

damoodamoo:
also relevant is how do you see this data being surfaced - I was assuming a client would navigate to a resource and then view the operations history for that resource. Do you see there being an `/operations` API endpoint that would surface all operations across the TRE regardless of resource?

marrobi:
I think you are probably correct as far as there would always be a TRE resource mirrored. with any "deployed" or configurable resource. It's not the case for shared services at the moment but will be. How about "tre start" or "tre stop"...

As for surfacing the data - see the user stories above in this issue. Queries will not always be done at a resource level. For example in time might want to see what VMs were deployed in the TRE (or workspace) over a period of time.

damoodamoo:
ok to summarise:
- New cosmos collection for `operations`
[...]

Issue #1075, opened 2021-11-16
Split out workspace routers into separate files and move security dependencies per route


marrobi:

marrobi:
@martinpeck your opinion would be valued on this.

martinpeck:
Unassigning myself. I wonder if this issue should actually be closed. cc @marrobi

PR #1153, opened 2022-01-20
Operations History

damoodamoo:
# PR for issue #865 

- [X] New `Operations` cosmos collection and repository
- [X] `deployment:{}` removed from resource document in favour of an `operation` doc
- [X] New resources being created now create an `operation` document, linked to the resource. The `operation_id` is sent in the deployment message, and in the update message flowing back to the API
- [X] API updates the `operation` document based on the outcome of the runner
- [X] Moved 'is deleted' logic away from checking the `deployment.status` to checking an `isActive` flag on the resource object. When the resource marked as deleted `isActive` is set to 0. All GETs are modified to get `WHERE isActive != false`
- [X] `db_migrations` script updated to move all `deployment:{}` blocks to their own `operation` documents, and create+set the `isActive` flag on all existing documents.
- [X] New API routes added to `workspace`, `workspace_service` and `user_resource` - `/operations` and `/operations/{operation_id}` - to get all or a specific operation for that resource
- [X] Existing unit tests updated to include operations logic
- [x] ~New unit tests specifically for operations repository (unsure if these are really needed)~
- [x] E2E tests pass
- [x] Tested in Azure


api_app/api/routes/workspaces.py
```diff
@@ -133,6 +136,15 @@ async def delete_workspace(workspace=Depends(get_workspace_by_id_from_path), wor
 
     return WorkspaceIdInResponse(workspaceId=workspace.id)
 
+# workspace operations
+@workspaces_shared_router.get("/workspaces/{workspace_id}/operations", response_model=OperationInList, name=strings.API_GET_RESOURCE_OPERATIONS, dependencies=[Depends(get_current_workspace_owner_or_researcher_user_or_tre_admin)])
```
marrobi, line 140: Lets talk about what the appropriate user roles are for accessing these endpoints.

api_app/models/domain/resource.py
```diff
@@ -41,16 +23,17 @@ class Resource(AzureTREModel):
     templateName: str = Field(title="Resource template name", description="The resource template (bundle) to deploy")
     templateVersion: str = Field(title="Resource template version", description="The version of the resource template (bundle) to deploy")
     properties: dict = Field({}, title="Resource template parameters", description="Parameters for the deployment")
-    deployment: Deployment = Field(Deployment(status=Status.NotDeployed, message=""), title="Deployment", description="Fields related to deployment of this resource")
+    isActive: bool = Field(True, title="Is Active", description="Is the resource active? Will be False when deleted.")
```
marrobi, line 26: Have we still got the concept of disabled?

A resource had to be disabled before being deleted.

api_app/api/routes/workspaces.py
```diff
@@ -101,7 +104,7 @@ async def retrieve_workspace_by_workspace_id(workspace=Depends(get_workspace_by_
 
 
 @workspaces_core_router.post("/workspaces", status_code=status.HTTP_202_ACCEPTED, response_model=WorkspaceIdInResponse, name=strings.API_CREATE_WORKSPACE, dependencies=[Depends(get_current_admin_user)])
-async def create_workspace(workspace_create: WorkspaceInCreate, workspace_repo=Depends(get_repository(WorkspaceRepository))) -> WorkspaceIdInResponse:
+async def create_workspace(workspace_create: WorkspaceInCreate, workspace_repo=Depends(get_repository(WorkspaceRepository)), operations_repo=Depends(get_repository(OperationRepository))) -> WorkspaceIdInResponse:
```
Lybecker, line 107: I suggest returning the operation id instead of workspace id (also when doing updates).

When modifying a workspace, workspace service or a user ressource, the initiator wants to know what is the status of that change. That info can be retrieved via the operation. When the operation is done, then then workspace id becomes relevant.
If the workspace id is needed up front, then lookup the operation first, which will contain the workspace id

damoodamoo: Yes i've been wondering this, and if we're happy with this model I think it makes sense. Do we take it a step towards a HATEOAS structure and return something like: 

```json
HTTP STATUS 202 (Accepted)
{
	"operation": {
		"href": "/api/workspaces/<workspace-guid-here>/operations/<op-guid-here>",
		"operation_id": "guid-here"
                "resource_id": "resource-guid-here"
	}
}
```

?

api_app/api/routes/workspaces.py
damoodamoo: So far i've copied the roles from the 'parent' resource, thinking being if you have permission to modify a resource then you can see the operations associated with it?

marrobi: I don't think a researcher can modify a workspace, or they shouldn't be able to anyway.

api_app/models/domain/resource.py
damoodamoo: Disabled is in the json `properties` - here: https://github.com/microsoft/AzureTRE/blob/0300d32ef29616dc89a34c0362cf35c91f2a95b2/api_app/models/domain/resource.py#L47. The `IsActive` flag doesn't affect this logic.

marrobi: Thanks.

api_app/api/routes/workspaces.py
damoodamoo: you're right - a researcher user can _view_ (`GET`) a workspace. Do we want them to be able to see ops history too? If not then what should the access be for Workspace operations / WS Service Operations / User Resource Operations ?

marrobi: I think for now - the same people as can create/modify them - hence need to check the operations.

api_app/api/routes/workspaces.py
damoodamoo: @marrobi what do you think to this?

marrobi: It would be good to get both the operation ID and resource ID back. For example if want to get the calculated properties of a new resource immediately after create.

Although in the first instance operation id might be sufficient.

api_app/api/routes/workspaces.py
damoodamoo: this is done - only difference is the workspace operations, as the workspace PATCH was tre_admin only, but means that all the operations for workspaces would need to move to the core router which doesn't feel ideal - so i've gone with TRE Admin OR workspace owner who can see the ops.

api_app/api/routes/workspaces.py
damoodamoo: Here's what i've gone for - think it covers all the requirements. `resourcePath` was added to the resource object and to the operation as I need to know the path so I can query the right endpoint for the ops. Also added a `Location` header to the `POST` and `DELETE` endpoints which will refer you to the correct ops endpoint to poll.

```json
{
  "operation": {
    "id": "fc2e867d-c726-424d-8f64-13392b23bbef",
    "resourceId": "b374eb9b-7fd1-4448-8af5-7e02ef2749ee",
    "resourcePath": "/workspaces/a1bceefb-092a-4163-b67a-2a40e49d3ea3/workspace-services/343d3543-33cb-4cc8-9e35-1e38a0e0c6a5/user-resources/b374eb9b-7fd1-4448-8af5-7e02ef2749ee",
    "resourceVersion": 0,
    "status": "not_deployed",
    "message": "This resource has not yet been deployed",
    "createdWhen": 1643718851.996725,
    "updatedWhen": 1643718851.996725
  }
}
```

api_app/api/routes/workspaces.py
marrobi: Make sense. The dependencies could be organized in a better way - https://github.com/microsoft/AzureTRE/issues/1075 .

api_app/service_bus/deployment_status_update.py
```diff
@@ -55,39 +55,50 @@ async def receive_message():
...
-        [dict]: Dictionary representing a resource with the deployment sub doc updated
+        [Operation]: Updated Operation object to persist
     """
-    if resource["deployment"]["status"] in [Status.DeletingFailed, Status.Deleted]:
-        return resource  # cannot change terminal states
-    if resource["deployment"]["status"] in [Status.Failed, Status.Deployed, Status.Deleting]:
+
+    if operation.status in [Status.DeletingFailed, Status.Deleted]:
+        return operation  # cannot change terminal states
+    if operation.status in [Status.Failed, Status.Deployed, Status.Deleting]:
```
tanya-borisova, line 71:
```suggestion
    if operation.status in [Status.Failed, Status.Deployed, Status.Deleting] \
        and message.status not in [Status.Deleted, Status.DeletingFailed]:
```

e2e_tests/test_workspace_service_creation.py
```diff
@@ -57,9 +57,9 @@ async def test_create_guacamole_service_into_base_workspace(admin_token, workspa
             "openid_client_id": f"{config.TEST_WORKSPACE_APP_ID}"
         }
     }
-    workspace_service_id, install_service_status = await post_workspace_service_template(workspace_id, service_payload, workspace_owner_token, verify)
+    workspace_service_id, install_service_status = await post_workspace_service(workspace_id, service_payload, workspace_owner_token, verify)
 
-    await ping_guacamole_workspace_service(workspace_id, workspace_service_id, workspace_owner_token, verify)
+    #await ping_guacamole_workspace_service(workspace_id, workspace_service_id, workspace_owner_token, verify)
```
tanya-borisova, line 62: Is this a leftover from testing?

e2e_tests/test_workspace_creation.py
```diff
@@ -34,46 +34,4 @@ async def test_getting_templates(template_name, admin_token, verify) -> None:
         assert (response.status_code == status.HTTP_200_OK), f"GET Request for {template_name} creation failed"
 
 
-@pytest.mark.extended
-@pytest.mark.timeout(1800)
-async def test_create_devtestlabs_workspace(workspace_owner_token, admin_token, verify) -> None:
```
tanya-borisova, line 39: Why are these ones gone? Did you intend to replace them with tests for `post_workspace` (as you import that method)?

e2e_tests/helpers.py
```diff
@@ -219,10 +193,15 @@ async def disable_and_delete_workspace_service(workspace_id, workspace_service_i
         headers = {'Authorization': f'Bearer {token}'}
 
         await disable_workspace_service(workspace_id, workspace_service_id, token, verify)
-        await delete_workspace_service(workspace_id, workspace_service_id, token, verify)
+        operation_dict = await delete_workspace_service(workspace_id, workspace_service_id, token, verify)
+
+        print("OP DICT", operation_dict)
```
tanya-borisova: My favourite kind of testing ;)

tanya-borisova (commented):
Looks good, the generalisation from deployment status to operation makes sense to me 👍

e2e_tests/test_workspace_creation.py
damoodamoo: these workspace types don't exist any more, but the tests still hung around

e2e_tests/test_workspace_service_creation.py
damoodamoo: yup it's back in now :)

marrobi (commented):
As the safe to test label hasn't been doing what we expected, this branch should be tested via `workflow_dispatch`.

martinpeck (approved):

PR #1154, opened 2022-01-21
Convert data inputs into variable inputs for modules in Terraform

tanya-borisova:
# PR for issue https://github.com/microsoft/AzureTRE/issues/700

In several modules in Terraform, data inputs are used unnecessarily. Data inputs are intended for resources defined outside of Terraform, but in this case resources are defined within the same Terraform environment. This is also causing issues when Terraform thinks a resource has been updated when in reality it hasn't. 

## How is this addressed

- Remove `firewall/data.tf`, `gitea/terraform/data.tf`, `sonatype-nexus/terraform/data.tf`

**NOTE**: This is in draft for now because I'm resolving merge conflicts.


marrobi:
We will be moving Gitea, Nexus and Firewall out as shared services to enable them to be managed as part of #23 . To be deployed separately after the core infrastructure. Fine to do this, but when we do that the data pieces will need to go back in.

Maybe I should create an issue and split them out into porter bundles to be installed via Make now to avoid you having to do this.

tanya-borisova:
> Maybe I should create an issue and split them out into porter bundles to be installed via Make now to avoid you having to do this.

Sounds like a good idea to me

templates/shared_services/gitea/terraform/variables.tf
```diff
@@ -47,8 +62,33 @@ variable "gitea_storage_limit" {
...
 }
+
+variable "firewall_name" {
+  type        = string
+  description = "Name of the firewall to connect to"
+}
+
+variable "firewall_resource_group_name" {
+  type        = string
+  description = "Name of the firewall to connect to"
```
jjgriff93, line 93: Name of the resource group containing the firewall resource

jjgriff93 (approved):

templates/shared_services/gitea/terraform/variables.tf
tanya-borisova: Thanks for spotting this :)

Issue #1156, opened 2022-01-21
Automatically login to az if you are in CI


ross-p-smith:
When we are running commands in the makefile, we checkdependencies and check whether we are logged into azure.

Add a check to see if we are running in CI and then if we are; login to AZ with a Service Principal which is setup for the build. This will also mean that you do not have to login to az directly in the yml

PR #1157, opened 2022-01-21
Auto az login if in CI pipeline

ross-p-smith:
# PR for issue
#1156 

## What is being addressed
If we set the TF_IN_AUTOMATION then we will log in using the service principal


Lybecker (approved):
LGTM

Issue #1135, opened 2022-01-18
SUGGESTION: Move away from docker-compose.yml, and reintroduce mounts for local .azure folder


martinpeck:
It looks like a docker compose is contributing to a couple of issues:

- networking/DNS resolution issues and an impact on `az cli` for those using Docker Desktop
- issues when mounting the .azure folder from the host

In addition, it doesn't appear to be used for hosting multiple containers.

Suggest we move away and then add back the .azure folder mounts, as we know this will work on WSL, MacOS, and Linux, and we see no issue as to why it shouldn't work on Codespaces

When testing:

- [ ] test dev container on MacOS with Docker Desktop
- [ ] test dev container on WIndows (WSL) with Docker Desktop
- [ ] test dev container on WIndows (WSL) with Docker Engine
- [ ] test dev container in Codespaces

PR #1159, opened 2022-01-24
Remove docker compose

ross-p-smith:
# PR for issue #1135

## What is being addressed
Move to a devcontainer.json rather than a compose.yml

## How is this addressed
I've also added default values to the Dockerfile


ross-p-smith (commented):
@marrobi - We might have to work together and add some variables to the build to remove this

.github/workflows/deploy_tre.yml
```diff
@@ -51,6 +46,7 @@ jobs:
       - name: Deploy TRE
         shell: bash
         env:
+          TF_IN_AUTOMATION: 1
```
marrobi, line 49: Should this be part of this PR?

ross-p-smith: Yes, if you can add these as secrets in your build,       -e ARM_CLIENT_ID \
      -e ARM_CLIENT_SECRET \
      -e ARM_TENANT_ID \
      -e ARM_SUBSCRIPTION_ID \

marrobi: What's this got to do with the docker compose file? Thought this was a separate issue regarding amending the deployment scripts?

martinpeck: Hey @marrobi. I had a chat with @ross-p-smith and this change isn't part of the move away from Compose so he'll remove this file from the PR.

martinpeck (changes requested):
Once `deploy_tre.yml` is removed, I'm happy with the rest of the PR.

.github/workflows/deploy_tre.yml
ross-p-smith: I was trying to be nice, but I will fix up GH Actions as a separate task

ross-p-smith:
@marrobi and @martinpeck - I've added the capability to run the container as a non root user into this PR. (Also removed the GH Action change)

martinpeck (approved):

Issue #1165, opened 2022-01-24
Running `make tre-stop` causes the next `make tre-deploy` to fail


tanya-borisova:
**Describe the bug**

Running `make tre-stop` will cause the next `make tre-deploy` ran in the same environment to fail.

**Steps to reproduce**

1. Run `make tre-deploy tre-start tre-stop` 
2. Run `make tre-deploy` again

**Investigation**

This seems to be due to the fact that commands `make tre-start` and `make tre-stop` are [manipulating](https://github.com/microsoft/AzureTRE/blob/main/devops/scripts/control_tre.sh#L8) the resource `fw-ip-configuration` which is [managed by terraform](https://github.com/microsoft/AzureTRE/blob/main/templates/core/terraform/firewall/firewall.tf#L17). 

Because of this, Terraform wants to recreate it:

![MicrosoftTeams-image](https://user-images.githubusercontent.com/52954123/150855645-b9ffd677-90ff-4d0f-a50c-3ebfd7b28db7.png)

However there is also a [lock](https://github.com/microsoft/AzureTRE/blob/main/templates/core/terraform/firewall/firewall.tf#L25) on its parent resource. Because of this, Terraform fails to recreate it: 

![image](https://user-images.githubusercontent.com/52954123/150855989-e500fc6b-5c25-45b4-9903-da37f8578d44.png)

Given that the resource managed by Terraform is also modified elsewhere, I think we should add a `lifecycle` [attribute](https://www.terraform.io/language/meta-arguments/lifecycle#ignore_changes) there, to reflect the fact that the `fw-ip-configuration` resource is modified by processes outside Terraform:

```
lifecycle {
    ignore_changes = [
      ip_configuration,
    ]
  }
```

marrobi:
The lock is only a do not delete lock, so confused as to which it cannot create the IP - its trying to replace the entire firewall from what I can see.

I'm happy with ignoring changes on ip configuration if it fixes it. 

Good investigation work.

tanya-borisova:
> confused as to which it cannot create the IP - its trying to replace the entire firewall from what I can see

Confused about this as well, from what I understand this can happen if there's no way to update the firewall in the API (or it's not implemented in the provider). I'll look again into this today

tanya-borisova:
[...]

PR #1169, opened 2022-01-25
Ignore changes to ip_configuration in Terraform

tanya-borisova:
# PR for issue https://github.com/microsoft/AzureTRE/issues/1165

Ignore changes to ip_configuration in Terraform, as these are updated by tre-start and tre-stop commands. 
Details of the investigation are in the parent ticket 

## How is this addressed

- Added `lifecycle { ignore_changes }` attribute for `ip_configuration` 

- [ ] Local E2E test


tanya-borisova:
This approach won't work as the ignored resource is also a Terraform output. Terraform apparently can't work like that and needs to track a resource if it outputs it.

Issue #1171, opened 2022-01-25
[BUG] Resource processor image fails to build with Terraform install error


jjgriff93:
**Describe the bug**
When building the resource processor image as part of `make all` the install of Terraform fails with

```
E: Encountered a section with no Package: header
E: Problem with MergeList /var/lib/apt/lists/apt.releases.hashicorp.com_dists_buster_main_binary-amd64_Packages.lz4
E: The package lists or status file could not be parsed or opened.
```

**Steps to reproduce**

1. Make all

jjgriff93:
Same issue we were facing with dev container install of terraform - will centralise install script and call from both dev cont and resource processor image

PR #1172, opened 2022-01-25
Fix for broken tf install in resource processor

jjgriff93:
# PR for issue

## What is being addressed

Fix for a bug #1171  where TF installation in resource processor image fails due to corrupt repository.

## How is this addressed

- Consolidated install method used for dev container and reused in resource processor image


ross-p-smith (approved):
It would be lovely if we could find a way to have a single script that was used in 2 different places

stuartleeks (approved):
lgtm

Issue #1175, opened 2022-01-25
"Publish and register bundles" should be a make target


ross-p-smith:
Story: #1166 

**Description**
Within the current deploy-tre.yml workflow there is a lot of bash to publish and register bundles. Wrap this in a script that we can call from a Make Command as this will eventually move to the dev container anyway.

Install yq in the devcontainer

PR #1181, opened 2022-01-26
Build and publish porter automatically from a make target

ross-p-smith:
# PR for issue
#1175

## What is being addressed
Amending the Actions workflow to reduce inline script to make it easier to move to the dev container. There is now a new target `make build-and-register-bundle` which builds the porter bundles, retrieves the token from the api, posts the json to the endpoint and checks that it has been registered.

## How is this addressed

image
The bundles all successfully register

### DIFF
--- a/devops/scripts/build_and_register_bundle.sh
+++ b/devops/scripts/build_and_register_bundle.sh
@@ -47,7 +47,7 @@ while [ "$1" != "" ]; do
             echo "Bundle type must be workspace, workspace_service or user_resource, not $1"
             exit 1
         esac
-        bundle_type=$1
+        BUNDLE_TYPE=$1
         ;;
     -c| --current)
         current="true"
@@ -66,7 +66,7 @@ while [ "$1" != "" ]; do
     shift # remove the current value for `$1` and use the next
 done
 
-if [[ -z ${bundle_type+x} ]]; then
+if [[ -z ${BUNDLE_TYPE+x} ]]; then
     echo -e "No bundle type provided\n"
     usage
 fi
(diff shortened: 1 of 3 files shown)


devops/scripts/build_and_register_bundle.sh
```diff
@@ -0,0 +1,100 @@
...
+esac
+
+export TOKEN=$(curl -X POST -H "Content-Type: application/x-www-form-urlencoded" -d \
+  "grant_type=password&resource=${RESOURCE}&client_id=${CLIENT_ID}&username=${USERNAME}&password=${PASSWORD}&scope=default)" \
+  https://login.microsoftonline.com/${AUTH_TENANT_ID}/oauth2/token | jq -r '.access_token')
+
+# We now need to traverse back to our scripts directory
+# TODO: Fix this traversing
+cd ../../../
+make register-bundle DIR=${BUNDLE_DIR}
```
marrobi, line 91: Hmm, @ross-p-smith  is there a reason these stages need to be separate. Is the only difference one gets the token and one does not?

ross-p-smith: So I've left the existing development flow of 
- `make porter-build`
- `make register-bundle`

so that developers can authorise themselves

There is a new target `make build-and-register-bundle: porter-build` which does it all and manages the token

marrobi: Ok, makes sense, do we need to add this into the docs somewhere, can see it causing confusion. What each command does and when to use? I know this stuff and makes my head hurt.

Happy for you to create a separate issue, then can get this merged?

marrobi (commented):
LGTM

devops/scripts/build_and_register_bundle.sh
ross-p-smith: Deal - I've raised #1193 which I will do whilst I am waiting for builds

marrobi (approved):
LGTM

Issue #900, opened 2021-09-20
If deploy two app service plans into a workspace they cannot integrate with the same subnet


marrobi:
**Describe the bug**
A limitation of Azure Web Apps is each app service plan must have a dedicated vnet for vnet integration. This means if deploy two workspace services with their own app service plans vnet integration fails on the second.

**Steps to reproduce**

1. Deploy Guacamole into a workspace
2. Deploy InnerEye Inference into the workspace

Possible resolution - deploy app service plan into the workspace rather than with the workspace-service

marrobi:
As an initial fix will deploy the app service plan in to our base workspace template.

There may be a requirement down the line to install an app service environment, into a workspace, but this brings deployment overhead.

PR #1187, opened 2022-01-27
Split out app service plan into workspace template

marrobi:
# PR for issue #900

## What is being addressed

Move app service plan into base worksapce template and configure guacamole and inner eye bundles to reference this app service plan using a data element.


templates/workspace_services/guacamole/terraform/web_app.tf
```diff
@@ -1,21 +1,13 @@
-resource "azurerm_app_service_plan" "guacamole" {
-  name                = "plan-${local.webapp_name}"
-  location            = data.azurerm_resource_group.ws.location
+data "azurerm_app_service_plan" "workspace" {
+  name                = "plan-${var.workspace_id }"
```
tamirkamara, line 2: nit if it works

```suggestion
  name                = "plan-${var.workspace_id}"
```

templates/workspaces/base/terraform/appserviceplan.tf
```diff
@@ -0,0 +1,14 @@
+resource "azurerm_app_service_plan" "workspace" {
+  count = var.deploy_app_service_plan ? 1 : 0
```
tamirkamara, line 2: Just a thought, isn't this the first time we might not have the resource we expect to?
I mean, if we try to deploy guacamole, it depends on the workspace being ready for it...

tamirkamara (approved):
A couple of notes/questions.

templates/workspaces/base/terraform/appserviceplan.tf
marrobi: Yes, do ive defaulted it to true and not exposed it., so will always be there.

 Down the line after the shared service work we may look at how can update the workspace and enable the app service plan via an output of the workspace service.

Issue #1166, opened 2022-01-24
Use DevContainer within GitHub Action Workflows


martinpeck:
Within the dev container we install tools, and set up configuration. This dev container is then used by engineers working on the project.

However, when it comes to the github action workflows we don't make use of the container. Here, we run different scripts/steps to install porter, install powershell extensions etc etc.

If, instead, we built the dev container and then ran the `make all` *within* the DC we'd have much higher confidence that the build will work in a PR. Also, this makes migrating from github actions to AzDO, or Gitlab, or other build runners much easier because the build essentially becomes...

1. docker build "the dev container'
2. docker run "stuff within the dev container"

... which is a low bar, and achievable in most build runners

PR #1196, opened 2022-01-27
Use DevContainer for the build

ross-p-smith:
# PR for issue
#1166 

## What is being addressed
This PR pushes all of the tasks of the build into the DevContainer. This means that what we build in our local developer container is exactly the same when it comes to CI.

## How is this addressed
- [X] Build the docker devcontainer
- [X] Upload the devcontainer to ACR
- [X] Use devcontainer to `make mgmt-deploy`
- [X] Parallelise the build and push of the docker images
- [X] Register the bundles in the devcontainer
- [X] Add PyTest to the devcontainer
- [X] Re-usable workflow for devcontainer run

I've commented out 1 test for the moment
image
image


tamirkamara (commented):
I looked for a run that completed, especially for the deploy-tre one but could find it. Can you force run it and include a successful run in the PR description so we could look?

.github/actions/devcontainer_run_command/action.yml
```diff
@@ -0,0 +1,213 @@
...
+        TEST_WORKSPACE_APP_ID: "${{ inputs.TEST_WORKSPACE_APP_ID }}"
+        TRE_ID: "${{ inputs.TF_VAR_tre_id }}"
+        TF_VAR_tre_id: "${{ inputs.TF_VAR_tre_id }}"
+        TF_VAR_location: ${{ inputs.TF_VAR_location }}
+        TF_VAR_terraform_state_container_name:
+          ${{ inputs.TF_VAR_terraform_state_container_name }}
+        TF_VAR_mgmt_resource_group_name: ${{ inputs.TF_VAR_mgmt_resource_group_name }}
+        TF_VAR_mgmt_storage_account_name:
+          ${{ inputs.TF_VAR_mgmt_storage_account_name }}
+        TF_VAR_core_address_space: ${{ inputs.TF_VAR_core_address_space }}
```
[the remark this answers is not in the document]
ross-p-smith: I agree, that this is a mess. We need some consistent variable standards as we have mixed and matched it seems. I created #1215 to address this.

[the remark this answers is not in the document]
stuartleeks: I think it would be worth updating the env var handling, but that feels like a separate task/PR to me :-)

.github/actions/devcontainer_run_command/action.yml
```diff
@@ -0,0 +1,213 @@
...
+        TF_VAR_tre_address_space: ${{ inputs.TF_VAR_tre_address_space }}
+        TF_VAR_swagger_ui_client_id: "${{ inputs.TF_VAR_swagger_ui_client_id }}"
+        TF_VAR_aad_tenant_id: "${{ inputs.TF_VAR_aad_tenant_id }}"
+        TF_VAR_api_client_id: "${{ inputs.TF_VAR_api_client_id }}"
+        TF_VAR_api_client_secret: "${{ inputs.TF_VAR_api_client_secret }}"
+        TF_VAR_api_app_service_plan_sku_tier: "${{ inputs.TF_VAR_api_app_service_plan_sku_tier }}"
+        TF_VAR_api_app_service_plan_sku_size: "${{ inputs.TF_VAR_api_app_service_plan_sku_size }}"
+        TF_VAR_acr_name: ${{ inputs.TF_VAR_acr_name }}
```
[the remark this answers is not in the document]
ross-p-smith: Yes - I think that there is a further piece of work (maybe of #1215) to tackle this as it's really confusing what the variables are used for. I also agree that TF_VAR has bleed into the pipeline which maybe it shouldn't have

.github/actions/devcontainer_run_command/action.yml
```diff
@@ -0,0 +1,213 @@
...
+    TF_VAR_api_client_id:
+      description: "The API Client Id. (Same as Resource)"
+      required: true
+    TF_VAR_api_client_secret:
+      description: "The API Client Secret."
+      required: true
+    TF_VAR_api_app_service_plan_sku_tier:
+      description: "The SKU tier of the App Service for the API."
+      required: false
+      default: "PremiumV3"
```
tamirkamara, line 105: This is a duplicate of the default value defined in TF (also next var). Do you have to specify a value here? What would happen if you don't?

.github/workflows/deploy_tre.yml
```diff
@@ -44,53 +38,167 @@ jobs:
...
+        if: ${{ success() }}
+        shell: bash
         run: |
-          export TF_VAR_arm_subscription_id=$ARM_SUBSCRIPTION_ID
+          set -e
+          docker image tag tredev:latest ${{ secrets.ACTIONS_ACR_NAME }}.azurecr.io/tredev:latest
+          docker image push ${{ secrets.ACTIONS_ACR_NAME }}.azurecr.io/tredev:latest
 
-          make all
+  build_docker_images:
```
tamirkamara, line 117: This job outputs warnings I think we should try to avoid. Not sure exactly where it comes from...

> Build Docker Images (build-and-push-gitea)
error: Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get "http://%2Fvar%2Frun%2Fdocker.sock/v1.24/containers/buildx_buildkit_builder-19ce2ecb-5976-4003-a284-c3fac728017b0/json": dial unix /var/run/docker.sock: connect: permission denied

.github/actions/devcontainer_run_command/action.yml
ross-p-smith: Perfect - they were originally in the workflow.yml, so replicated. Never checked the default in Terraform - They have now gone as I can't see us changing them in the pipeline

.github/workflows/deploy_tre.yml
ross-p-smith: I've removed the installation of Buildkit using a Docker action and used it via ENV vars and the warning has gone away

.github/workflows/deploy_tre.yml
```diff
@@ -44,53 +38,160 @@ jobs:
...
-          terraform_version: 1.0.5
+          registry: ${{ secrets.ACTIONS_ACR_NAME }}.azurecr.io
+          username: ${{ secrets.ACTIONS_ACR_NAME }}
+          password: ${{ secrets.ACTIONS_ACR_PASSWORD }}
 
-      - name: Deploy TRE
+      - name: Set Variables
+        run: |
+          echo "USER_UID=$(id -u)" >> $GITHUB_ENV
+          echo "USER_GID=$(id -g)" >> $GITHUB_ENV
```
tamirkamara, line 54: Is this only used once for building the devcontainer? 
If yes, do you really need this here and insert it to the env of all steps?

tamirkamara (changes requested):

.github/workflows/deploy_tre.yml
ross-p-smith: I've pulled these 2 inline with the docker build. These were left over when I was refactoring.

stuartleeks (approved):
lgtm

Issue #1206, opened 2022-01-28
Bundles broken in the build


ross-p-smith:
**Describe the bug**
At the moment the build is not able to register the bundles

**Steps to reproduce**

1. Start a tre_deploy build and it fails in the Bundle job

PR #1207, opened 2022-01-28
Pass in credentials for Bundles to publish

ross-p-smith:
# PR for issue #1206 

## What is being addressed
Pass in the Azure Credentials so that we can log into Azure


marrobi (approved):
LGTM

PR #1210, opened 2022-01-31
Remove access token

ross-p-smith:
# PR for issue
#1206 

## What is being addressed
- When we are building porter from docker-from-docker we need to set the socket up correctly
- When we build-and-register the bundles we do not ask for the TOKEN, instead we get it from the auth service
- We no longer need to log into Azure in the pipeline as the CheckDependencies does this for us


marrobi (approved):
LGTM

Issue #1197, opened 2022-01-28
Cannot build AML Compute Instance User resource


marrobi:
Fails with error cannot cd to terraform directory.

Porter Mixins are being executed before code is being copied.

PR #1214, opened 2022-02-01
Fix indentation in AML compute instance bundle

marrobi:
# PR for issue #1197

## What is being addressed

As a result of some linting indentation needed fixing.


ross-p-smith (approved):
Sorry!

Issue #1213, opened 2022-02-01
.gitattributes does not enforce LF line endings for files without an extension


marrobi:
File names such as `run`, are not covered by the `.gitattributes` setting to enforce LF. When checking out on Windows often line endings are changed to CRLF.

PR #1216, opened 2022-02-01
Force LF line endings on checkout

marrobi:
# PR for issue #1213

## What is being addressed

Force all line endings to lf on checkout to avoid line endings being changed when checked out to Windows file systems.


marrobi:
PR from incorrect branch.

Issue #1177, opened 2022-01-26
Split out Azure Firewall, Gitea and Nexus from core TRE components


tanya-borisova:
## Description

As a customer of Azure TRE
I want to opt-in into having {Azure Firewall, Gitea, Nexus} deployed
So that I can decide whether it meets my needs

## Acceptance criteria

- [ ] Firewall deployment is decoupled from the deployment of the core TRE components
- [ ] Gitea deployment is decoupled from the deployment of the core TRE components
- [ ] Nexus deployment is decoupled from the deployment of the core TRE components

## Notes

This is related (and likely part of) issue https://github.com/microsoft/AzureTRE/issues/23 
This is also related to resolving bug https://github.com/microsoft/AzureTRE/issues/1165 (as implementing this will likely make the bug go away)

tanya-borisova:
It turns out that Azure Firewall currently can't be destroyed from Terraform (but can from Portal). 
Link to the issue in the azurerm Terraform provider: https://github.com/hashicorp/terraform-provider-azurerm/issues/14866 

To deploy this change, we would have to ask everybody to manually delete Firewall in their dev deployment, as well as delete it manually for the CI deployment.

tanya-borisova:
It also turns out that Gitea and Nexus won't work as part of core terraform if Firewall is extracted into the shared service, so all 3 components are extracted into Porter bundles as part of this ticket

tanya-borisova:
Talked to @marrobi  today and we agreed that in order to minimise disruption to existing users of TRE, I will attempt to run `terraform state rm` for the resources in core terraform, followed by `terraform import` in the Porter bundles terraform states.
This is to ensure continuity of existing deployments of Firewall, Gitea and Nexus.

PR #1226, opened 2022-02-02
Move shared services into Porter bundles

tanya-borisova:
# PR for issue https://github.com/microsoft/AzureTRE/issues/1177

## What is being addressed

Currently, shared services (Firewall, Gitea and Nexus) are deployed as part of `make tre-deploy` command and are part of core terraform module. 
That makes it impossible to opt out of Firewall, and cumbersome to configure whether Gitea or Nexus should be deployed. 

In this pull request, Porter bundles are made for these shared services. 

This PR is also ground work for automatic update of shared services in the future (https://github.com/microsoft/AzureTRE/issues/23). 

## How is this addressed

- [x] Moved firewall terraform module to shared services
- [x] Merged routetable and terraform modules, as their functionality is very similar
- [x] Initialised Porter bundles for Terraform, Gitea and Nexus 
- [x] **Testing strategy**: Checked terraform plans to make sure they match 

## Note: Bug in Terraform provider

It turns out that Azure Firewall currently can't be destroyed from Terraform (but can from Portal).
Link to the issue in the azurerm Terraform provider: hashicorp/terraform-provider-azurerm#14866

To deploy this change, we would have to ask everybody to manually delete Firewall in their dev deployment, as well as delete it manually for the CI deployment.

## Note: Creation of the Service Principal

Following merge of this pull requests, developers will have to create a Service Principal if needed to deploy Firewall, Gitea or Nexus. A doc will be created on Wiki / in Markdown on how to do this.


Makefile
```diff
@@ -120,19 +120,46 @@ letsencrypt:
...
-	echo -e "\n\e[34m»»» 🧩 \e[96mStarting TRE\e[0m..." \
+tre-stop:
+	echo -e "\n\e[34m»»» 🧩 \e[96mStopping TRE\e[0m..." \
 	&& . ./devops/scripts/check_dependencies.sh azfirewall \
 	&& . ./devops/scripts/load_env.sh ./templates/core/.env \
 	&& . ./devops/scripts/load_env.sh ./devops/.env \
-	&& ./devops/scripts/control_tre.sh start
+	&& ./devops/scripts/control_tre.sh stop
+
+firewall-install:
```
tanya-borisova, line 137: Discussion point: 
I haven't yet included these commands as part of the main deploy. Should we include these services by default?

templates/shared_services/firewall/terraform/firewall.tf
```diff
@@ -11,11 +11,11 @@ resource "azurerm_public_ip" "fwpip" {
...
   name                = "fw-${var.tre_id}"
-  resource_group_name = var.resource_group_name
+  resource_group_name = local.core_resource_group_name
   location            = var.location
   ip_configuration {
     name                 = "fw-ip-configuration"
-    subnet_id            = var.firewall_subnet.id
+    subnet_id            = data.azurerm_subnet.firewall.id
     public_ip_address_id = azurerm_public_ip.fwpip.id
   }
```
tanya-borisova, line 21: Discussion point: 
Do we still need lock on the Firewall (below)?  I wasn't sure if it was added as a workaround to the problem of the Firewall being recreated (see https://github.com/microsoft/AzureTRE/issues/1165) or for another reason

marrobi: I think a do not delete is still a good thing. Given it is stateful we need to prevent accidental deletion, but not update.

templates/shared_services/gitea/terraform/gitea-webapp.tf
```diff
@@ -267,6 +267,7 @@ resource "azurerm_role_assignment" "gitea_acrpull_role" {
 resource "null_resource" "webapp_vault_access_identity" {
```
marrobi, line 267: Look like this is now supported by the terraform provider: https://github.com/hashicorp/terraform-provider-azurerm/issues/13388

templates/shared_services/gitea/terraform/variables.tf
```diff
@@ -29,3 +29,18 @@ variable "acr_name" {
   type        = string
   description = "Name of Azure Container Registry"
 }
+
+variable "arm_tenant_id" {
```
marrobi, line 33: Are these needed if not running the rest command?

marrobi (changes requested):
I'll check out branch and test once done the `keyVaultReferenceIdentity` fix.

Good work, seem to be getting there.

Makefile
tanya-borisova: Yes, for local development we will need to create a Service Principal. As far as I understand, there is already one for running in CI.

templates/shared_services/gitea/terraform/variables.tf
tanya-borisova: Nope these won't be needed! I will update the keyVaultReferenceIdentity instead, thanks!

Makefile
damoodamoo: I vote we get on with adding shared services via the API rather than spend too much time smoothing out this process

damoodamoo (approved):

Makefile
marrobi: @damoodamoo  agree.

tanya-borisova: I was just going to add a little README on how to create a service principal for deployment for the local dev flow and be done with it for now

templates/shared_services/gitea/terraform/variables.tf
marrobi: @tanya-borisova can these go now?

tanya-borisova: @marrobi yes, sorry forgot about that -- updated now

Makefile
marrobi: @tanya-borisova where are these new bundles being deployed as part of the pipeline via `make all`? I can't see it in the PR. They do need to be deployed - especially when others pick up the repository given templates are dependent.

tanya-borisova: @marrobi  done! I added them as a step to tre-deploy, and there is a script that creates a service principal too, that hopefully will make it easier for people

templates/shared_services/firewall/terraform/firewall.tf
tamirkamara: @marrobi what state is saved in the firewall?
The aggregate of all the interactions with it?

tanya-borisova:
Remaining piece of work in this PR: after discussion with @marrobi , I'll try to move the Terraform resources from core state to the new state created in the Porter bundles, without destroying and recreating them. This should be possible to achieve via running `terraform state rm` for the resources in core terraform, followed by `terraform import` in the Porter bundles terraform states.

damoodamoo:
this has been continued + updated in #1365 in Tanya's absence

Issue #1221, opened 2022-02-02
Refactor TF templates to remove sub-modules


eladiw:
## Description

As a TRE developer
I want to consolidate the existing modules into a single 'module'
So that it will be easier to work and maintain the terraform templates

TF modules comes with an overhead of inputs and outputs files.
when using a lot of modules, the developer need to pass a lot of values in and out the different modules. Also, sometimes a module separation is needed and it involves a lot of refactoring.
The [recommendation](https://www.terraform.io/language/modules/develop#when-to-write-a-module), is to use modules with moderation.

This issue is about reducing the number of modules to the minimum.

## Acceptance criteria

- [ ] most of the modules are consolidated into the main script
- [ ] still separated to multiple files and still easy to find and maintain

marrobi:
@eladiw assigned to yourself to ensure nobody else picks it up.

PR #1234, opened 2022-02-03
Consolidated TF modules

eladiw:
#1221 

## What is being addressed

This PR consolidates the most simple modules into the root TF directory.
In order to simplify the dependencies

### What was not changed
The modules for:
appgateway
azure-monitor
network
and resource processor

were not consolidated as each held eith multiple files (e.g. app gateway) or had configurations files in the same directory which makes sense 'bundled' together.

## How is this addressed

- modules which are relatively simple, were moved to the root directory. all inputs/outputs have been adjusted


templates/core/terraform/firewall.tf
```diff
@@ -1,6 +1,6 @@
 resource "azurerm_public_ip" "fwpip" {
   name                = "pip-fw-${var.tre_id}"
-  resource_group_name = var.resource_group_name
```
tanya-borisova, line 3: I wouldn't touch firewall too much as it is extracted out in this PR: https://github.com/microsoft/AzureTRE/pull/1226

eladiw: Thanks, will remove once merged

marrobi:
@eladiw I'm good with this in principal. Removes a lot of unnecessary TF. However we need to ensure that resources that have state are preserved. 

Thinking about Cosmos and KeyVault particularly which if recreated are going to disrupt a number of existing deployments. The firewall will be moved.

@tanya-borisova is looking into this for the firewall and other shared services.

marrobi:
Another issue is when moving the private endpoint, such as the API app, the IP address changes. App Gateway does not refresh automatically, hence the Web App health status has been in a failed state since the test run. I updated the health probe and came to life. This is a known issue, and will discuss offline.

eladiw:
@marrobi  good points!

Please take a look at the latest revision. In which I've added the 'moved' blocks.
This will allow existing deployments to work with this new version of tf templates.

We can discuss in a few months if we want to delete this file...
https://www.terraform.io/language/modules/develop/refactoring#moved-block-syntax

eladiw:
> @marrobi good points!
> 
> Please take a look at the latest revision. In which I've added the 'moved' blocks. This will allow existing deployments to work with this new version of tf templates.
> 
> We can discuss in a few months if we want to delete this file... https://www.terraform.io/language/modules/develop/refactoring#moved-block-syntax

btw, it was tested using @tamirkamara env-in-pr branch. deployed using 'old' tf, and upgraded without any issues

eladiw:
image

These were not consolidated, as I thought its better to keep them in a separate directory (because they are not as simple). however I am open to discuss this

templates/core/terraform/firewall.tf
eladiw: reverted

stuartleeks:
> Please take a look at the latest revision. In which I've added the 'moved' blocks. This will allow existing deployments to work with this new version of tf templates.

Oooh, nice. I hadn't seen that `moved` blocks had been added - that's really useful!

templates/core/terraform/bastion.tf
```diff
@@ -1,6 +1,6 @@
 resource "azurerm_public_ip" "bastion" {
   name                = "pip-bas-${var.tre_id}"
-  resource_group_name = var.resource_group_name
+  resource_group_name = azurerm_resource_group.core.name
   location            = var.location
```
ross-p-smith, line 4: Shall we do location whilst we are here too? (Take it from the azurerm_resource_group.core.location)

ross-p-smith: Absolutely LOVE the "moved" - today I learned something new!

martinpeck (commented):
Looks good. A couple of questions: 

- Have you got any numbers/feel for how this improves/changes deployment times?
- Do you think this will cause any issues for those with existing environments? i.e. should we add the `breaking-change` tag to this PR, and then add a note into the docs if there are fixup/warnings?

eladiw:
> Looks good. A couple of questions:
> 
>     * Have you got any numbers/feel for how this improves/changes deployment times?
> 
>     * Do you think this will cause any issues for those with existing environments? i.e. should we add the `breaking-change` tag to this PR, and then add a note into the docs if there are fixup/warnings?

1. unfortunately, I didn't see any improvement regarding the deployment time..
2. The added 'moved' blocks should eliminate a breaking change

martinpeck (approved):
Thanks @eladiw ! LGTM.

martinpeck:
Please coordinate with @tanya-borisova so that we can merge #1226 first, if possible.

ross-p-smith (approved):

PR #1235, opened 2022-02-03
Add config to save dev container bash history to the mounted volume

stuartleeks:
This change persists bash history across dev container rebuilds.

The `commandhistory` volume was mounted in `devcontainer.json`, but the container wasn't configured to use it for bash history.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -68,3 +68,9 @@ RUN wget -O /usr/local/bin/yq https://github.com/mikefarah/yq/releases/download/
   && chmod +x /usr/local/bin/yq
 
 USER $USERNAME
+
+# Save command line history 
+RUN echo "export HISTFILE=$HOME/commandhistory/.bash_history" >> "$HOME/.bashrc" \
+    && echo "export PROMPT_COMMAND='history -a'" >> "$HOME/.bashrc" \
+    && mkdir -p $HOME/commandhistory \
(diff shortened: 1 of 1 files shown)


.devcontainer/Dockerfile
```diff
@@ -68,3 +68,9 @@ RUN wget -O /usr/local/bin/yq https://github.com/mikefarah/yq/releases/download/
   && chmod +x /usr/local/bin/yq
 
 USER $USERNAME
+
+# Save command line history 
+RUN echo "export HISTFILE=$HOME/commandhistory/.bash_history" >> "$HOME/.bashrc" \
+    && echo "export PROMPT_COMMAND='history -a'" >> "$HOME/.bashrc" \
```
ross-p-smith, line 74: Could this overwrite other bash history we may have from other devcontainers - do we need "tre" in there?

tanya-borisova (approved):

.devcontainer/Dockerfile
ross-p-smith: Oh sorry - that is sorted out in the devcontainer.json - ignore me!

stuartleeks: The mounted volume has `tre` in it, so would only conflict if another dev container uses the same volume name :-)

```
		"type=volume,source=tre-bashhistory,target=/home/vscode/commandhistory",
```

ross-p-smith (approved):

Issue #1233, opened 2022-02-03
'make letsencrypt' fails while configuring network storage rules


dram1964:
'make letsencrypt' target fails to run because of an error in the ./core/terraform/scripts/letsencrypt.sh

**Error Message**
When running 'make letsencrypt', the command errors with: the following arguments are required: --resource-group/-g

**Steps to reproduce**

1. Following TRE instructions at [TRE docs](https://microsoft.github.io/AzureTRE/tre-admins/setup-instructions/deploying-azure-tre/)
2. Run make all to build TRE
3. Run 'make letsencrypt'

**Suggested Fix**
Added '--resource-group ${RESOURCE_GROUP} to the `az storage account network-rule` commands on lines 16 and 106 of the letsencrypt.sh file. 'make letsencrypt' then completes without error.

ross-p-smith:
Just out of interest do you use default azure config? Do you have anything set if you type `az config get`?

PR #1236, opened 2022-02-03
Fixing missing Resource Group

ross-p-smith:
# PR for issue 
#1233 

## What is being addressed
We were not explicitly setting the RESOURCE_GROUP_NAME in the letsencrypt and uploadstaticweb.sh

## How is this addressed
- I have also renamed the variable so that it can be re-used in other parts of the developer experience.
- There are also further ServiceBus outputs that are needed in scripts


marrobi (approved):
LGTM

Issue #1224, opened 2022-02-02
PR validation should test deployment and E2Es


tamirkamara:
## Description

As a TRE Developer
I want a stronger PR validation that tests deployment and some of the E2E tests
So that I'll be less likely to break the main branch.

* Need to find a good balance between this goal and the time we wait for validation to run.

## Acceptance criteria

- [ ] PR validation tests deployment of a TRE
- [ ] PR validation tests core E2E scenarios
- [ ] Validation runs or different PRs don't interfere with each other 
- [ ] Resources provisioned for validation are deleted automatically to reduce cost

ross-p-smith:
Are we saying that a PR build should build a fresh copy of TRE and run the E2E tests in order to merge. Then the build of main deploys over the top of a persistent environment?

tamirkamara:
Yes @ross-p-smith. 
It remains to be defined what is the difference between the E2E that run in the PR and the one after merge to main. If there's no difference probably we don't need both.
As to the persistent environment - I wonder if anyone actually use it but that's a different discussion.

martinpeck:
What's the time budget for these builds?
Are we holding ourselves to (for example) 20 mins for a PR build to complete?
Is there a different budget for the main build?

tamirkamara:
@martinpeck I really wanted to meet that 20 mins mark... However the deploy itself runs for ~20 mins, add to that building images, running tests, pulling devcontainers and we get a higher number.

PR #1242, opened 2022-02-03
Azure testing in PR validation

tamirkamara:
# PR for issue #1224 

## What is being addressed

We currently deploy to Azure and run E2E tests only after a PR is merged to main. This has caused issues as many times we have broken the main branch.

## How is this addressed

- The deploy workflow will now create a new environment on Azure for each **branch** when a PR is triggered (or run with the other configured triggers).
- The new environment will be deleted after a PR closes or the source branch is deleted (see the added workflow). Costs will get reduced when tre-stop will be operational again (add issue number)
Note that the delete action is commented out since some of the logic could be verified only after the workflow will be on the main branch. We should turn it on after a few real test drives.

An example run can be seen [here](https://github.com/microsoft/AzureTRE/actions/runs/1790815831). Run time is ~45 mins.


marrobi:
@tamirkamara we've discovered that PRs triggered via `pull_request_target` - i.e. adding the label, checkout the target branch (as the name suggests!), i.e. `main`, not the actual PR. So I think what you have is good for environments, but does not do what we need as will not deploy the PR. This is an issue in the workflow at main at present.

So we need this plus another PR that fixes the incorrect branch being run.

tamirkamara:
I understand the challenge @marrobi . I only found out this week that we run deploy from PRs and not just on merge to main. 
Anyway, I think this is not directly related to this pr in which each git ref runs in its own environment.
> The branch or tag ref that triggered the workflow run.
Sorting out the pr triggers and security of external contributions can be done in another pr.

.github/workflows/clean_validation_envs.yml
```diff
@@ -0,0 +1,71 @@
...
+            # az group delete --resource-group "${tre_rg}-mgmt" --yes --no-wait
+          }
+
+          echo "Refs:"
+          git show-ref
+
+          open_prs=$(gh pr list --state open --json number,title)
+
+          # Resource groups that start with a specific string and have the ci_git_ref tag whose value starts with "ref"
+          az group list --query "[?starts_with(name, 'rg-tr') && tags.ci_git_ref != null && starts_with(tags.ci_git_ref, 'refs')].[name, tags.ci_git_ref]" -o tsv |
```
eladiw, line 51: isnt the prefix condition too restrictive?
is it possible to have a rg with other prefix?
if so maybe another tag such as 'tre deployment' might be better (instead of checking a prefix)

Makefile
```diff
@@ -12,96 +18,96 @@ build-and-push-gitea: build-gitea-image push-gitea-image
 build-and-push-guacamole: build-guacamole-image push-guacamole-image
 
 bootstrap:
-	echo -e "\n\e[34m»»» 🧩 \e[96mBootstrap Terraform\e[0m..." \
+	$(call target_title, "Bootstrap Terraform") \
```
eladiw, line 21: whats this change

eladiw (commented):
looks good. few questions

.github/workflows/clean_validation_envs.yml
tamirkamara: The prefix should be the same one we set in the other workflow where the env gets created. Honestly I could just use the tag but wanted another condition to add some assurence.
The issue with just using tags is that the "main" env is also created in the same process and will get all tags.

Makefile
tamirkamara: I started to use recipes (macros) to write code once and reuse as needed. I'm working on other cases but it's not ready yet.

tamirkamara: Line 10

> target_title = @echo -e "\n\e[34m»»» 🧩 \e[96m$(1)\e[0m..."

.github/workflows/clean_validation_envs.yml
```diff
@@ -0,0 +1,71 @@
...
+      - name: Azure Login
+        uses: azure/login@v1
+        with:
+          creds: ${{ secrets.AZURE_CREDENTIALS }}
+
+      - name: clean
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -o errexit
```
ross-p-smith, line 30: I think this is great, I could see this having a benefit in AzDO builds too. If this was in a script, behind something like `make deleteEnvironment` it would allow us to call it locally and also means we  have less logic in GH Actions

ross-p-smith (changes requested):
Would be nice to see the part that deletes unwanted environments in a re-usuable script

.github/workflows/clean_validation_envs.yml
tamirkamara: I understand moving the logic to a script file, but not sure about putting it inside the make file - it doesn't really have anything to do with the tre solution but more as a tiding step of the ci.
Will it make sense to do it without the make target and call the script directly in the ci step?

tamirkamara: Done @ross-p-smith

daltskin (commented):
For PR builds can we re-instate the tre stop to prevent incurring firewall/apg charges asap (https://github.com/microsoft/AzureTRE/issues/1165 shouldn't be an issue for new deployments)

tamirkamara:
@daltskin This isn't exactly a new deployment for each PR **run** and you could use the same environment multiple times on the same PR which will save you time (but will cost more). 
I thought this is a good first step and perhaps #1165 will be fixed soon and we can come back to this.

ross-p-smith (approved):

eladiw (approved):

Issue #1237, opened 2022-02-03
build broken in main due to keyvalue secret change


ross-p-smith:
Run the build on "main" and it fails in tre-deploy stage

**Steps to reproduce**

│ Error: keyvault.BaseClient#RecoverDeletedSecret: Failure responding to request: StatusCode=403 -- Original Error: autorest/azure: Service returned an error. Status=403 Code="Forbidden" Message="The user, group or application 'appid=***;oid=fb152351-3675-4a9f-8a51-0582e6157fda;iss=https://sts.windows.net/***/' does not have secrets recover permission on key vault 'kv-***;location=***'. For help resolving this issue, please see https://go.microsoft.com/fwlink/?linkid=2125287" InnerError={"code":"ForbiddenByPolicy"}
│ 
│   with module.gitea[0].azurerm_key_vault_secret.gitea_password,
│   on ../../shared_services/gitea/terraform/gitea-webapp.tf line 244, in resource "azurerm_key_vault_secret" "gitea_password":
│  244: resource "azurerm_key_vault_secret" "gitea_password" {
│ 
╵
╷
│ Error: keyvault.BaseClient#RecoverDeletedSecret: Failure responding to request: StatusCode=403 -- Original Error: autorest/azure: Service returned an error. Status=403 Code="Forbidden" Message="The user, group or application 'appid=***;oid=fb152351-3675-4a9f-8a51-0582e6157fda;iss=https://sts.windows.net/***/' does not have secrets recover permission on key vault 'kv-***;location=***'. For help resolving this issue, please see https://go.microsoft.com/fwlink/?linkid=2125287" InnerError={"code":"ForbiddenByPolicy"}
│ 
│   with module.gitea[0].azurerm_key_vault_secret.db_password,
│   on ../../shared_services/gitea/terraform/mysql.tf line 58, in resource "azurerm_key_vault_secret" "db_password":
│   58: resource "azurerm_key_vault_secret" "db_password" {
│ 
╵

PR #1247, opened 2022-02-04
Import TF Resources

marrobi:
# PR for issue #1237 

WIP


.github/workflows/import_resources_to_tf.yml
```diff
@@ -0,0 +1,44 @@
...
+  deploy_management:
+    name: Import TF Resources
+    runs-on: ubuntu-latest
+    environment: Dev
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Import Resources
+        uses: ./.github/actions/devcontainer_run_command
```
eladiw, line 17: why is it using a devcontainer script?
If it is not related to devcontainer maybe this code should be extracted?

PR #1257, opened 2022-02-07
Register AAD workspace - developer experience

ross-p-smith:
## What is being addressed

- New `make register-aad-workspace` which will log you into the AAD_TENANT_ID and register an AAD app for your workspace and then log you back into the tenant you were on
- Updated the developer experience so that it copies the `templates\core\.env` file to your api_app when launching debug so that you have to copy less keys/values around. This has also meant I have to write less to templates/core/tre.env as this should be the values that have come from your deployment.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,12 @@
+{
+  // See https://go.microsoft.com/fwlink/?LinkId=733558
+  // for the documentation about the tasks.json format
+  "version": "2.0.0",
+  "tasks": [
+    {
+      "label": "Copy_env_file_for_debug",
+      "command": "cp --force /workspaces/AzureTRE/templates/core/.env /workspaces/AzureTRE/api_app/.env",
+      "type": "shell"
+    }
+  ]
+}
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -5,7 +5,7 @@
   "tasks": [
     {
       "label": "Copy_env_file_for_debug",
-      "command": "cp --force /workspaces/AzureTRE/templates/core/.env /workspaces/AzureTRE/api_app/.env",
+      "command": "cp --force ${workspaceFolder}/templates/core/.env ${workspaceFolder}/api_app/.env",
       "type": "shell"
     }
   ]
--- a/devops/scripts/register-aad-workspace.sh
+++ b/devops/scripts/register-aad-workspace.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -e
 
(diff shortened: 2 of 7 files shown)


.vscode/tasks.json
```diff
@@ -0,0 +1,12 @@
+{
+  // See https://go.microsoft.com/fwlink/?LinkId=733558
+  // for the documentation about the tasks.json format
+  "version": "2.0.0",
+  "tasks": [
+    {
+      "label": "Copy_env_file_for_debug",
+      "command": "cp --force /workspaces/AzureTRE/templates/core/.env /workspaces/AzureTRE/api_app/.env",
```
martinpeck, line 8: Can `/worksplaces/AzureTRE` be replaced with `${workspaceFolder}`?

devops/scripts/register-aad-workspace.sh
```diff
@@ -0,0 +1,28 @@
...
+
+: ${AAD_TENANT_ID?"You have not set you AAD_TENANT_ID in ./templates/core/.env"}
+
+LOGGED_IN_TENANT_ID=$(az account show --query tenantId -o tsv)
+CHANGED_TENANT=0
+
+if [ "${LOGGED_IN_TENANT_ID}" != "${AAD_TENANT_ID}" ]; then
+  echo "Attempting to sign you onto ${AAD_TENANT_ID} to add an App Registration."
+
+  # First we need to login to the AAD tenant (as it could be different to the subscription tenant)
```
martinpeck, line 12: This comment is repeated (here and line 26). I suspect here it should say `as it is different` rather than `as it could be different`

devops/scripts/register-aad-workspace.sh
```diff
@@ -0,0 +1,28 @@
...
+# Then register an App
+./scripts/aad-app-reg.sh \
+-n "${TRE_ID} - Workspace One" \
+-r "https://${TRE_ID}.${RESOURCE_LOCATION}.cloudapp.azure.com/api/docs/oauth2-redirect" \
+-w
+
+if [ "${CHANGED_TENANT}" -ne 0 ]; then
+  echo "Attempting to sign you back into ${LOGGED_IN_TENANT_ID}."
+
+  # First we need to login to the AAD tenant (as it could be different to the subscription tenant)
```
martinpeck, line 26: repeated comment

Should be replaced with `switch back to the subscription tenant (as it was different to the AAD tenant)`

martinpeck (changes requested):

martinpeck (approved):
LGTM

PR #1261, opened 2022-02-07
Network size

ross-p-smith:
# PR for issue
#1254 

## What is being addressed
- Allow vscode to debug tests inline
- vscode default test task now runs the unit tests (Run task)
- address_space_size can be passed in (small, medium, large or custom)
- Added unit tests for tshirt sized address_space


api_app/db/repositories/workspaces.py
```diff
@@ -70,6 +67,24 @@ def create_workspace_item(self, workspace_input: WorkspaceInCreate, auth_info: d
...
+        address_space_size = workspace_properties.get("address_space_size", "small")
+
+        #773 allow custom sized networks to be requested
+        if (address_space_size == "custom"):
+            # if address_space isn't provided in the input, generate a new one.
+            # TODO: #772 check that the provided address_space is available in the network.
+            return workspace_properties.get("address_space") or self.get_new_address_space()
+
+        # We allow the users some predefined TShirt sizes
+        predefined_address_spaces = { "small": 24, "medium": 22, "large": 16 }
```
martinpeck, line 82: FYI (no change required) but Python 3.10 has `match select` which would have make this a little easier.

I wonder if it makes more sense to push the 'predefined_address_spaces` variable up into a constant, defined outside of this method. It's not a big deal, but at present it's being defined per call.

TBH, this might be better as an `if/elif/else` statement given that there are only two options (medium and large) that result in anything other than the `else`.

martinpeck (approved):

api_app/db/repositories/workspaces.py
```diff
@@ -70,6 +67,24 @@ def create_workspace_item(self, workspace_input: WorkspaceInCreate, auth_info: d
...
+    def get_address_space_based_on_size(self, workspace_properties: dict):
+
+        # Default the address space to 'small'. This was the pre-existing default
+        address_space_size = workspace_properties.get("address_space_size", "small")
+
+        #773 allow custom sized networks to be requested
+        if (address_space_size == "custom"):
+            # if address_space isn't provided in the input, generate a new one.
+            # TODO: #772 check that the provided address_space is available in the network.
+            return workspace_properties.get("address_space") or self.get_new_address_space()
```
tamirkamara, line 79: So this means that if the user requested a workspace with a custom address_space (e.g. 1.2.3.4/24) then they also need to supply address_space_size=custom?
If yes, wouldn't it make more sense to switch it around and not care about the address_space_size if an address_space was provided?

stuartleeks: I think I quite like the explicitness of setting `address_space_size=custom` 🤷

tamirkamara: The confusion I saw is that we say the size is custom but we expect the user to supply the full space and not the custom size. But it's not that important... we can leave it as is.

tamirkamara (approved):
Regarding my comment - fine with it either way you choose.

api_app/db/repositories/workspaces.py
marrobi: I like the fact its explicit too. Very clear you are "going off piste".

api_app/db/repositories/workspaces.py
ross-p-smith: My next ticket is validating the custom address range, so some of this will change. Personally I don't like if/elif and would prefer to use the dictionary as it's extensible and reads better

Issue #1240, opened 2022-02-03
Update Resource Properties via API


damoodamoo:
Feature #23 

The API needs to accept `PATCH` operations for all resources which update the state of the resource in the data store. To avoid data loss, the existing `properties:{}` block should be copied into a `history:[]` array within the resource document. The `history` item should include audit details about who / what  and when the change was made.

As part of allowing updates to properties, eTag checking should be implemented across _all_ `PATCH` endpoints, to ensure each submission contains the most up-to-date information and one change is not overwriting another. 

eg: 
```json
{
    "id": "95fa5b15-e0ec-4277-aee4-964a5bcb0c52",
    "templateName": "tre-workspace-base",
    "templateVersion": "0.1.5",
    "resourceVersion": 1,
    "properties": {
        "display_name": "Updated Title",
        "description": "I've made an update",
        "app_id": "6d8aabae-0318-4a4b-a433-848607dd0db5",
        "address_space": "10.1.5.0/24",
    },
    "history": [
         {
            "resourceVersion": 0,
            "updatedBy": "user@provider.com",
            "timestamp": 123456765.1234,
            "properties": {
               "display_name": "My new workspace",
               "description": "Initial desc",
               "app_id": "6d8aabae-0318-4a4b-a433-848607dd0db2",
               "address_space": "10.1.5.0/24"
           },
          { ... },
          { ... },
          { ... },
    ]
   ...
 }
```

Acceptance:
- [x] Ability to send a `PATCH` request to each resource type
- [x] All previous properties moved to history
- [x] Appropriate unit testing
- ~[ ] E2E test added for `PATCH`~ 

> NOTE: Currently each resource accepts a `PATCH` which _only_ sets the `enabled` property. These will be updated to cover all properties, including the `enabled` flag, if sent.

TO DO:
- [x] `enabled` flag moved to top level of resource document
- [x] Allow update of `isEnabled`, and `properties`.
- [x] Unit tests for property history
- [x] custom annotation added to schema docs to indicate the property is updateable
- [x] translate `updateable: false` custom annotation into `readOnly` when validating `PATCH`ed properties
- [x] validate properties from resource template
- [x] return `_etag` in all resources
- [x] require `etag` in all `PATCH`es

tanya-borisova:
I had a brief thought about only using the `history` block and treating the top history item as a current property block, but actually it will make things more complicated without a clear benefit. So I think your design is good as it is written up here!

damoodamoo:
[...]

PR #1265, opened 2022-02-08
ETag enforcement for PATCH operations

damoodamoo:
PR to partially address #1240 

It:
- Returns the `_etag` property in all resources
- Enforces that the user send the `etag` header in the `PATCH` request
- Uses the cosmos precondition to ensure that the etag matches the document being updated
- Moves the `enabled` flag to the top level of the document so it applies uniformly to all resources, and is called `isEnabled`.
- e2e tests updated to send the * as etag header to patch + disable the resource


api_app/models/domain/operation.py
```diff
@@ -29,7 +29,7 @@ class Operation(AzureTREModel):
     id: str = Field(title="Id", description="GUID identifying the operation")
     resourceId: str = Field(title="resourceId", description="GUID identifying the resource")
     resourcePath: str = Field(title="resourcePath", description="Path of the resource undergoing change, ie '/workspaces/guid/workspace-services/guid/'")
-    resourceVersion: int = Field(0, title="resourceVersion", description="Version if the resource this operation relates to")
+    resourceVersion: int = Field(0, title="resourceVersion", description="Version of the resource this operation relates to")
```
martinpeck, line 32: nice description

damoodamoo, line 32: this is genius!

api_app/db/repositories/workspaces.py
```diff
@@ -91,9 +96,12 @@ def get_new_address_space(self, cidr_netmask: int = 24):
         new_address_space = generate_new_cidr(networks, cidr_netmask)
         return new_address_space
 
-    def patch_workspace(self, workspace: Workspace, workspace_patch: WorkspacePatchEnabled):
-        workspace.properties["enabled"] = workspace_patch.enabled
-        self.update_item(workspace)
+    def patch_workspace(self, workspace: Workspace, workspace_patch: WorkspacePatch, etag: str) -> Workspace:
+        workspace.isEnabled = workspace_patch.isEnabled
+
+        # TODO - validate update workspace props here
```
ross-p-smith, line 102: Is it worth linking to a task here?

damoodamoo:
Abandoning this as the underlying work has moved on and builds on this

PR #1262, opened 2022-02-08
Add automation app support

stuartleeks:
Closes #1231

- [X] update `aad-app-reg.sh` to allow an additional App to be created (with `TREAdmin` role)
- [X] update the bundle registration called from CI/CD to use this App as the identity when registering bundles if configured
- [x] update docs for `aad-app-reg.sh` with new argument

As part of the `aad-app-reg.sh` changes, it was hard to find a meaningful single-letter switch, so this PR also includes adding long-forms of switches (and updates the doc to use the long-form for readability).

Whilst updating `publish_register_bundles.sh` the `eval` sections of the script have also been removed to simplify the script (and calling code)

Documenting the steps to add create the automation account manually will be done for #1232

PR #1269, opened 2022-02-09
Fix 'make register-bundle' interactive flow

stuartleeks:
The changes in #1262 broke the flow for running `make register-bundle` without auth env vars set. This PR addresses that.

- Allow command to run without auth
- Fix arg passing


devops/scripts/publish_register_bundle.sh
```diff
@@ -60,15 +62,19 @@ while [ "$1" != "" ]; do
         access_token=$1
         ;;
     *)
+      echo "Hello 2: $1"
```
ross-p-smith, line 65: is this debug?

stuartleeks: 🤦

stuartleeks: gone

ross-p-smith (approved):

devops/scripts/publish_register_bundle.sh
```diff
@@ -63,12 +64,15 @@ while [ "$1" != "" ]; do
...
+    fi
     shift # remove the current value for `$1` and use the next
 done
 
 if [[ -z ${access_token} ]]; then
     echo -e "WARNING!!! No Azure access token provided. Automatic bundle registration not possible. Use the script output to self-register. See documentation for more details.\n"
-    usage
 fi
 
 if [[ -z ${tre_url} ]]; then
```
oliver7598, line 78: line 80  "usage" in check on tre_url is irrelevant if an access_token is not provided and causes the script to fail.

Suggest moving if statement on like 78 down to after line 101 adding in the following at line 76:
 elif [[ -n  ${access_token} ]]; then
echo -e "No TRE URL provided\n"
      usage
    fi

oliver7598 (changes requested):
line 80 "usage" in check on tre_url is irrelevant if an access_token is not provided and causes the script to fail.

Suggest moving if statement on like 78 down to after line 101 adding in the following at line 76:
elif [[ -n ${access_token} ]]; then
echo -e "No TRE URL provided\n"
usage
fi

devops/scripts/publish_register_bundle.sh
stuartleeks: I've updated to allow no url if there's no access_token

Thanks!

oliver7598: Looks good to me and runs locally as expected

oliver7598 (approved):

PR #1272, opened 2022-02-09
Devcontainer cache for main and PR

ross-p-smith:
# PR for issue

## What is being addressed
- Add GH CLI to the devcontainer
- ACTIONS_DEVCONTAINER_TAG is either latest or REF_ID depending on whether we are main or not
- We --cache-from our ACTIONS_DEVCONTAINER_TAG as well as known "latest" which means we get a cache hit all of the time
- We push our latest container on all builds. It just so happens that only main can push to "latest"

TODO - have another build that cleans up ACR on branch deletion/PR completion


.github/workflows/deploy_tre.yml
```diff
@@ -39,6 +39,7 @@ jobs:
       RUN_ACR_NAME: ${{ github.ref == 'refs/heads/main' && secrets.ACR_NAME || format('tre{0}', steps.run-id.outputs.refid) }}
       RUN_STATE_STORAGE_ACCOUNT_NAME: ${{ github.ref == 'refs/heads/main' && secrets.STATE_STORAGE_ACCOUNT_NAME || format('tre{0}mgmt', steps.run-id.outputs.refid) }}
       RUN_MGMT_RESOURCE_GROUP: ${{ github.ref == 'refs/heads/main' && secrets.MGMT_RESOURCE_GROUP || format('rg-tre{0}-mgmt', steps.run-id.outputs.refid) }}
+      DEVCONTAINER_LABEL: ${{ steps.run-id.outputs.dev_container_label }}
```
tamirkamara, line 42: Label or tag?

.github/workflows/deploy_tre.yml
```diff
@@ -51,6 +52,12 @@ jobs:
           echo "using id of: ${REFID} for GitHub Ref: ${GITHUB_REF}"
           echo "::set-output name=refid::${REFID}"
 
+          if [ ${GITHUB_REF} == 'refs/heads/main' ]; then
+            echo "::set-output name=dev_container_label::latest"
+          else
+            echo "::set-output name=dev_container_label::${REFID}"
+          fi
```
tamirkamara, line 59: If this in the script because you wanted to be more explicit compared with the conditional expressions I used in the other outputs of this job?

ross-p-smith: I wanted it to be readable, but will put it inline if you prefer

tamirkamara (approved):

PR #1280, opened 2022-02-10
Use Docker BuildKit cache in builds when running via make

### LINKED ISSUE #1271: Buildkit setting doesn't get into the devcontainer in CI
**Describe the bug**
When running github actions using the devcontainer, the value of docker buildkit isn't passed to the docker run command. This prevents optimizations when building images (like cache)

**Steps to reproduce**
1. Look into the custom action file: .github/actions/devcontainer_run_command/action.yml
1. The input includes DOCKER_BUILDKIT with a default of 1 (which is good)
1. The run command step doesn't make use of this input which means buildkit is off

tamirkamara:
# Fixes #1271 

## What is being addressed

Docker buildkit isn't used when the CI builds images through the devcontainer and caching isn't possible.

## How is this addressed

- Pass the DOCKER_BUILDKIT value from the action to the docker run command so that it will be available inside the devcontainer for all other docker operations
- Create a make recipe for building and publishing images
- Create & use buildkit cache layers when building images via make
- Move Guacamole image build to a later stage as core deployments doesn't need to wait for it

### DIFF
--- a/.github/workflows/deploy_tre_reusable.yml
+++ b/.github/workflows/deploy_tre_reusable.yml
@@ -140,8 +140,9 @@ jobs:
           docker image push \
             ${{ secrets.ACTIONS_ACR_NAME }}.azurecr.io/tredev:${{ secrets.ACTIONS_DEVCONTAINER_TAG }}
 
-  build_docker_images:
-    name: Build Docker Images
+  build_core_images:
+    # used to build images used by core infrastructure
+    name: Build Core Docker Images
     runs-on: ubuntu-latest
     needs: [deploy_management]
     environment: Dev
@@ -151,8 +152,7 @@ jobs:
         target: [
           build-and-push-api,
           build-and-push-resource-processor,
-          build-and-push-gitea,
-          build-and-push-guacamole]
+          build-and-push-gitea]
 
     steps:
       - name: Checkout
@@ -177,14 +177,14 @@ jobs:
         with:
           github-token: ${{ github.token }}
           ms-teams-webhook-uri: ${{ secrets.MS_TEAMS_WEBHOOK_URI }}
-          notification-summary: "Make Images Failed"
+          notification-summary: "Make Core Images Failed"
           notification-color: dc3545
           timezone: Europe/Zurich
 
   deploy_tre:
     name: Deploy TRE
     runs-on: ubuntu-latest
-    needs: [build_docker_images]
+    needs: [build_core_images]
     environment: Dev
     steps:
       - name: Checkout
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy_tre_reusable.yml
```diff
@@ -278,11 +278,49 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Docker build
+        uses: ./.github/actions/devcontainer_run_command
+        with:
+          COMMAND: "make ${{ matrix.target }}"
+          ACTIONS_ACR_NAME: ${{ secrets.ACTIONS_ACR_NAME }}
+          ACTIONS_ACR_PASSWORD: ${{ secrets.ACTIONS_ACR_PASSWORD }}
+          ARM_TENANT_ID: "${{ secrets.ARM_TENANT_ID }}"
```
ross-p-smith, line 302: You are missing ACTIONS_DEVCONTAINER_TAG

Makefile
```diff
@@ -37,74 +37,61 @@ mgmt-destroy:
 	&& . ./devops/scripts/load_terraform_env.sh ./devops/.env \
 	&& cd ./devops/terraform && ./destroy.sh
 
+# A recipe for building images. Parameters:
+# 1. Image name suffix
+# 2. Version file path
+# 3. Docker file path
+# 4. Docoer context path
+# Example: $(call build_image,"api","./api_app/_version.py","api_app/Dockerfile","./api_app/")
+define build_image
```
ross-p-smith, line 46: Don't get me wrong, I ❤️ this, the repeating code in the makefile were making me shudder. However, I wonder whether this would be better in a script with some proper checking of the existence of the variables. As when it goes wrong, debugging a makefile is not fun! Happy to leave it too

ross-p-smith (changes requested):
I think you are missing ACTIONS_DEVCONTAINER_TAG

Makefile
```diff
@@ -37,74 +37,61 @@ mgmt-destroy:
 	&& . ./devops/scripts/load_terraform_env.sh ./devops/.env \
 	&& cd ./devops/terraform && ./destroy.sh
 
+# A recipe for building images. Parameters:
+# 1. Image name suffix
+# 2. Version file path
+# 3. Docker file path
+# 4. Docoer context path
```
stuartleeks, line 44: nit:

```suggestion
# 4. Docker context path
```

.github/workflows/deploy_tre_reusable.yml
tamirkamara: fixed.

Makefile
tamirkamara: fixed

Makefile
tamirkamara: Not sure what will be in the script and how will that play nicely with the whole way env is loaded and used.
Anyway, I didn't much of the logic here, just centralized it with a few params.
Feels to me that revamping make might be a separate thing but willing to discuss...

ross-p-smith: Agreed - I think this makefile will get changed incrementally anyway. It's a step in the right direction

PR #1281, opened 2022-02-10
Preserve Resource Property History

damoodamoo:
PR to address part of #1240 :

- eTags used to prevent competing write (patch) operations overwriting each other in cosmos
- `isEnabled` moved to top level of document
- Resource document/object updated to contain a `history: []` array, which a new entry is added to each time a `PATCH` is made
- `PATCH` operations updated to use the same underlying method in the higher level Resource Repository (rather than each containing update logic)
- `<ResourceType>Patch` objects combined into single `ResourcePatch` object

Still to do in a future PR - Check which properties it is 'legal' to update and which to error on.


tanya-borisova (approved):
LGTM

ross-p-smith (approved):

Issue #772, opened 2021-09-05
Check that a custom address_space for a workspace is free


tamirkamara:
Story: #52 

**Description**
When an api call is made with a pre-defined address_space (i.e. a custom one), we should make sure that it's available for allocation. Otherwise the bundle deployment will probably fail.

PR #1283, opened 2022-02-10
Validate the custom address space

ross-p-smith:
# PR for issue #772 

## What is being addressed
When you are creating a workspace and you define an "address_space_size" of `custom`, then this PR will validate that the address space is within the TRE_ADDRESS_SPACE when TRE was originally created.

### DIFF
--- a/api_app/services/cidr_service.py
+++ b/api_app/services/cidr_service.py
@@ -1,5 +1,6 @@
 from typing import List
 from ipaddress import IPv4Network, NetmaskValueError, collapse_addresses
+from xmlrpc.client import boolean
 
 from core import config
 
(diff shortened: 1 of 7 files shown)


api_app/services/cidr_service.py
```diff
@@ -1,5 +1,6 @@
 from typing import List
 from ipaddress import IPv4Network, NetmaskValueError, collapse_addresses
+from xmlrpc.client import boolean
```
martinpeck, line 3: I don't think you want this (see other comment)

api_app/services/cidr_service.py
```diff
@@ -29,6 +24,28 @@ def generate_new_cidr(allocated_subnets: List[str], required_cidr_block_type: in
...
+
+    free_subnets = remove_subnet([allocation_network], core_network)
+
+    for subnet_string in allocated_subnets:
+        free_subnets = remove_subnet(free_subnets, IPv4Network(subnet_string))
+
+    return free_subnets
+
+
+def is_network_available(allocated_subnets: List[str], requested_CIDR) -> boolean:
```
martinpeck, line 39: return type should be `bool`

martinpeck (changes requested):
LGTM. Just one small change (use of `bool` instead of the imported type `boolean`)

api_app/services/cidr_service.py
ross-p-smith: Do you know - I dont even know where that came from!

martinpeck (approved):
LGTM

PR #1286, opened 2022-02-10
Add unit tests into build validation

ross-p-smith:
Adding the unit tests into the build validation. 

I could not find the unit tests, so added them. @tamirkamara - it looks like the upload/publish of tests results were in the DockerBuild workflow (thanks, didn't spot that), but the tests were never actually run. So I've cleaned this up

### DIFF
--- a/.github/workflows/build_docker_images.yml
+++ b/.github/workflows/build_docker_images.yml
@@ -112,16 +112,3 @@ jobs:
           file: ./templates/shared_services/gitea/Dockerfile
           cache-from: type=gha
           cache-to: type=gha,mode=max
-
-      - name: Publish Unit Test Results
-        uses: EnricoMi/publish-unit-test-result-action@v1
-        if: always()
-        with:
-          files: test-results/**/*.xml
-
-      - name: Upload Unit Test Results
-        if: always()
-        uses: actions/upload-artifact@v2
-        with:
-          name: test-results
-          path: test-results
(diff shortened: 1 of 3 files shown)


tamirkamara:
IIRC, the docker files include running the tests, so how is this different than the workflow that build images?

jjgriff93 (approved):

tamirkamara:
> Adding the unit tests into the build validation.
> 
> I could not find the unit tests, so added them. @tamirkamara - it looks like the upload/publish of tests results were in the DockerBuild workflow (thanks, didn't spot that), but the tests were never actually run. So I've cleaned this up

@ross-p-smith I'm used to always running rests as part of the docker build process so if something is wrong there it means that an image could be built inspite of failing tests.
I'd prefer to see that working there rather than externalizing it. But that my opinion...

ross-p-smith:
> > Adding the unit tests into the build validation.
> > I could not find the unit tests, so added them. @tamirkamara - it looks like the upload/publish of tests results were in the DockerBuild workflow (thanks, didn't spot that), but the tests were never actually run. So I've cleaned this up
> 
> @ross-p-smith I'm used to always running rests as part of the docker build process so if something is wrong there it means that an image could be built inspite of failing tests. I'd prefer to see that working there rather than externalizing it. But that my opinion...

Just saw you comment @tamirkamara - I'll move the tests back into the Docker build as understand now - the api app is containerised there

PR #1302, opened 2022-02-14
Simpler bundle registration

### LINKED ISSUE #1278: Simplify automatic bundle registration
## Description

I want to register bundles in a simple "atomic" way
So that I could perform the action manually and automatically through CICD

We currently have multiple make targets involved / look similar, with multiple script files that in turn call other make targets. This makes it hard to split the build/public/register activities in different stages in the ci that could reduce the total run time.

## Acceptance criteria

- [ ] A single make target (and script) that is able to register the bundle without build/publish (will reference a remote-published bundle)
- [ ] The target/script supports getting an access token for the TRE API
- [ ] Existing logic in make build-and-register-bundle, register-bundle-payload is removed
- [ ] Existing logic in scripts: devops/scripts/publish_register_bundle.sh, devops/scripts/build_and_register_bundle.sh is removed

tamirkamara:
# Fixes #1278 

## What is being addressed

Simplifying the steps needs for a person and the ci to build, publish and register a bundle with the api making it available for use.

## How is this addressed

- Have distinct targets not depended on each other (main change in the register)
- Register target always references a remote bundle, meaning you don't need to publish in order to register
- As a result, it is now possible for the ci to publish (and build) bundles in parallel to the deploy-tre job and later handle the registration. This saves us ~5 mins.
- Rename all relevant targets to be bundle-something


devops/scripts/register_bundle_with_api.sh
```diff
@@ -0,0 +1,154 @@
...
+
+    if [[ -z "$2" ]]; then
+      # if no more args then stop processing
+      break
+    fi
+
+    shift # remove the current value for `$1` and use the next
+done
+
+explain_json=$(porter explain --reference ${acr_name}.azurecr.io/$(yq eval '.name' porter.yaml):v$(yq eval '.version' porter.yaml) -o json)
```
ross-p-smith, line 87: is it worth having something like 
${acr_name:?} 
so that we check the variables are actually set?

ross-p-smith (approved):

devops/scripts/register_bundle_with_api.sh
stuartleeks: In other scripts we have checks that required variables are set. Adding that here would be good IMO - it feels easier to track down an error when the message is "--acr-name must be set" rather than a cryptic error from porter that the acr couldn't be found.

E.g. https://github.com/microsoft/AzureTRE/blob/main/devops/scripts/publish_register_bundle.sh#L75

stuartleeks (approved):
lgtm aside from the suggestion from @ross-p-smith

devops/scripts/register_bundle_with_api.sh
tamirkamara: done.

PR #1303, opened 2022-02-14
Enable PyPI proxy and consume via Linux VMs

jjgriff93:
# PR for issue

## What is being addressed

Nexus will now proxy PyPI as a source repository for pip packages, and Linux VMs deployed as Guacamole user resources are configured to consume from this proxy source without the need for users to manually specify alternate package sources/enable manual trust of remote repositories.

## How is this addressed

- Added firewall FQDN exception for files.pythonhosted.org
- Added configuration for Nexus pypi proxy
- Created clouding script to configure `pip.conf` on Linux VMs when provisioned to consume from proxy source


ross-p-smith (approved):
From my expert pypi repo knowledge - LGTM :-)

Issue #1301, opened 2022-02-14
Linux VM scale set password requirements incorrect


vjmanda:
**Describe the bug**
`make all` may fail due to the Linux VM Scale Set password not meeting the requirements

```
Error: creating Linux Virtual Machine Scale Set "vmss-rp-porter-vmtre1" (Resource Group "rg-vmtre1"): 
compute.VirtualMachineScaleSetsClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: 
Code="InvalidParameter" Message="The supplied password must be between 6-72 characters long and must satisfy at least 3 of 
password complexity requirements from the following:\r\n1) Contains an uppercase character\r\n2) Contains a lowercase 
character\r\n3) Contains a numeric digit\r\n4) Contains a special character\r\n5) Control characters are not allowed" 
Target="adminPassword"

with module.resource_processor_vmss_porter[0].azurerm_linux_virtual_machine_scale_set.vm_linux,
on resource_processor/vmss_porter/main.tf line 53, in resource "azurerm_linux_virtual_machine_scale_set" "vm_linux":
53: resource "azurerm_linux_virtual_machine_scale_set" "vm_linux" {
```

**Steps to reproduce**

1. run `make all`

**Fix**
Add additional properties to random_password in /templates/core/terraform/resource_processor/vmss_porter/main.tf

replace
```
resource "random_password" "password" {
  length           = 16
  special          = true
  override_special = "_%@"
}
```

with
```
resource "random_password" "password" {
  length           = 16
  lower            = true
  min_lower        = 1
  upper            = true
  min_upper        = 1
  number           = true
  min_numeric      = 1
  special          = true
  min_special      = 1
  override_special = "_%@"
}
```

marrobi:
Good catch, very odd hasn't come up before. Are you OK to PR the change? Thanks.

vjmanda:
👍🏽

PR #1311, opened 2022-02-15
#1301-Fix-LinuxVMScaleSetPasswordRequirements

vjmanda:
# PR for issue
[#1301 - Linux VM scale set password requirements incorrect](https://github.com/microsoft/AzureTRE/issues/1301)

## What is being addressed
TRE deployment may fail due to random_password generated for Linux VM scale set not meeting the password requirements. It currently does not have the required password requirements in the TF resource.

Currently only has 
```
resource "random_password" "password" {
  length           = 16
  special          = true
  override_special = "_%@"
}
```

## How is this addressed
The fix applies additional password requirements to random_password
```
resource "random_password" "password" {
  length           = 16
  lower            = true
  min_lower        = 1
  upper            = true
  min_upper        = 1
  number           = true
  min_numeric      = 1
  special          = true
  min_special      = 1
  override_special = "_%@"
}
```


ross-p-smith (approved):
Great spot!

PR #1313, opened 2022-02-16
Enable tre stop/start

tamirkamara:
## What is being addressed

This PR is partially resolves the problems described in #1165 and enables the system to stop and start the TRE to conserve costs, but **only** when deployed in debug mode.
This [test](https://github.com/microsoft/AzureTRE/runs/5214057060?check_suite_focus=true#step:3:326) shows a TRE that was previously stopped and brought up within the run.

## How is this addressed

- All locks are disabled in debug mode
- make tre-deploy calls tre-start before executing Terraform
- control_tre scripts checks if a TRE exists before doing anything. This is to a new tre deployment which can happen since tre-start is called before tre-deploy
- Change the type of firewall diagnostics settings to save Terraform redundant changes.

### DIFF
--- a/.github/actions/devcontainer_run_command/action.yml
+++ b/.github/actions/devcontainer_run_command/action.yml
@@ -103,7 +103,10 @@ inputs:
     description: "Indicates if the API endpoint has valid TLS certificate and if we validate it during E2E."
     required: false
     default: "true"
-
+  TF_VAR_debug:
+    description: "A value indicating if the system should be deployed in debug mode."
+    required: false
+    default: "false"
 runs:
   using: composite
   steps:
@@ -190,5 +193,6 @@ runs:
           -e IS_API_SECURED \
           -e TF_VAR_ci_git_ref="${{ github.ref }}" \
           -e DOCKER_BUILDKIT \
+          -e TF_VAR_debug=${{ inputs.TF_VAR_debug }} \
           '${{ inputs.ACTIONS_ACR_URI }}tredev:${{ inputs.ACTIONS_DEVCONTAINER_TAG }}' \
         bash -c "${{ inputs.COMMAND }}"
--- a/Makefile
+++ b/Makefile
@@ -93,7 +93,7 @@ push-gitea-image:
 push-guacamole-image:
 	$(call push_image,"guac-server","./templates/workspace_services/guacamole/version.txt")
 
-tre-deploy:
+tre-deploy: tre-start
 	$(call target_title, "Deploying TRE") \
 	&& . ./devops/scripts/check_dependencies.sh nodocker \
 	&& . ./devops/scripts/load_env.sh ./templates/core/.env \
--- a/templates/core/terraform/firewall/firewall.tf
+++ b/templates/core/terraform/firewall/firewall.tf
@@ -23,6 +23,7 @@ resource "azurerm_firewall" "fw" {
 }
 
 resource "azurerm_management_lock" "fw" {
+  count      = var.debug ? 0 : 1
   name       = azurerm_firewall.fw.name
   scope      = azurerm_firewall.fw.id
   lock_level = "CanNotDelete"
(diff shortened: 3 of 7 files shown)


ross-p-smith (approved):
LGTM

templates/core/terraform/firewall/firewall.tf
```diff
@@ -23,16 +23,19 @@ resource "azurerm_firewall" "fw" {
...
 }
 
 resource "azurerm_monitor_diagnostic_setting" "firewall" {
-  name                       = "diagnostics-firewall-${var.tre_id}"
-  target_resource_id         = azurerm_firewall.fw.id
-  log_analytics_workspace_id = var.log_analytics_workspace_id
+  name                           = "diagnostics-firewall-${var.tre_id}"
+  target_resource_id             = azurerm_firewall.fw.id
+  log_analytics_workspace_id     = var.log_analytics_workspace_id
+  log_analytics_destination_type = "Dedicated"
```
martinpeck, line 37: It wasn't clear how this impacts start/stop stuff.
Is this part of that change, or unrelated?

Makefile
```diff
@@ -93,7 +93,7 @@ push-gitea-image:
 push-guacamole-image:
 	$(call push_image,"guac-server","./templates/workspace_services/guacamole/version.txt")
 
-tre-deploy:
+tre-deploy: tre-start
```
martinpeck, line 96: So, `tre-deploy` will always first try and start the TRE now?

templates/core/terraform/main.tf
```diff
@@ -99,6 +99,7 @@ module "firewall" {
   location                   = var.location
   resource_group_name        = azurerm_resource_group.core.name
   log_analytics_workspace_id = module.azure_monitor.log_analytics_workspace_id
+  debug                      = var.debug
```
martinpeck, line 102: I'm not sure I'd have picked "debug" as it as an implication that we're also setting this deployment up to have additional logging, and be easier to debug. I'm not sure that's true.

I'd have done with "development" rather than "debug" but this is fine. Naming stuff is hard :)

martinpeck (approved):

templates/core/terraform/main.tf
tamirkamara: I understand what you mean... this wasn't my choice and it's being using in a few other places to have TF implement other behaviors - like in the KV IIRC.

Makefile
tamirkamara: Yes, that' right. The reason is that if the TRE is "stopped" then TF will fail with weird errors

templates/core/terraform/firewall/firewall.tf
tamirkamara: Kind of related:
If _log_analytics_destination_type_ isn't set then TF uses a default value of _AzureDiagnostics_. After the deployment that value somehow changes into _null_ (probably something on Azure).
On some runs, TF tries to recreate this resource and fails (something about the resource already there). I saw that a few times while testing the stop/start. 
The easy solution was to make sure TF doesn't try to change it at all. A value of _Dedicated_ is the way forward anyway so it made sense.

templates/core/terraform/firewall/variables.tf
```diff
@@ -33,3 +33,8 @@ variable "web_app_subnet" {
   })
   description = "The ID of the web app subnet"
 }
+
+variable "debug" {
```
stuartleeks, line 37: Is this variable still needed?

.github/actions/devcontainer_run_command/action.yml
```diff
@@ -190,5 +193,6 @@ runs:
           -e IS_API_SECURED \
           -e TF_VAR_ci_git_ref="${{ github.ref }}" \
           -e DOCKER_BUILDKIT \
+          -e TF_VAR_debug=${{ inputs.TF_VAR_debug }} \
```
stuartleeks, line 196: Do we still need TF_VAR_debug in this file?

templates/core/terraform/firewall/variables.tf
tamirkamara: fixed

.github/actions/devcontainer_run_command/action.yml
tamirkamara: fixed

Issue #1202, opened 2022-01-28
Documentation: URL to InnerEye is incorrect


sjmercer65:
on the page:
https://microsoft.github.io/AzureTRE/tre-templates/workspaces/inner-eye/

The first two links on this page are to InnerEye and AzureML - both are incorrect.

InnerEye:
https://microsoft.github.io/AzureTRE/tre-templates/workspace_services/innereye
Should be (I THINK)
https://microsoft.github.io/AzureTRE/tre-templates/workspaces/inner-eye/

The URL for AzureML is also incorrect (leads to a page with huge icons and no formatting) - I don't know what the right one would be.

marrobi:
@sjmercer65 do you think you might be able to raise a PR to fix this?

The links should go back to the workspace service pages, such as: `https://microsoft.github.io/AzureTRE/tre-templates/workspace-services/azure-ml/`

Could do with a link checker as part of the mkdocs pipeline.

GowthamChowdary:
i'd like to work on this if it's still open

GowthamChowdary:
i'd appreciate some guidance on this, everything looks fine except for the template_schema.json which has 
`"https://github.com/microsoft/AzureTRE/templates/workspaces/base/template_schema.json"`
but should be 
`"https://github.com/microsoft/AzureTRE/tre-templates/workspaces/base/template_schema.json"` ?

marrobi:
@GowthamChowdary Thanks for the offer of assistance. I believe @sjmercer65 is referring to the links that are at the top of the docs page next to the two bullets.

<img width="553" alt="image" src="https://user-images.githubusercontent.com/17089773/154234246-44e27dd6-0d5d-4fef-8dee-50e0130c206b.png">

GowthamChowdary:
Alright, thanks for clarifying that, i thought that was generated from
mkdocs. I’ll do a PR to fix the links.

sjmercer65:
Thanks Gowtham!

PR #1315, opened 2022-02-16
fix urls

GowthamChowdary:
# PR for issue
#1202 

## What is being addressed
the urls on the website are wrong

Describe the current behavior you are modifying

## How is this addressed
modified urls


ross-p-smith (approved):
Many thanks!

PR #1322, opened 2022-02-17
Adding a job to run extended tests (only on main)

eladiw:
# PR for issue

## What is being addressed

We would like to have all the extended test running after merge to main.

## How is this addressed

- added the command to the makefile
- added a workflow job to run this command on main only

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -6,19 +6,49 @@ on:
   workflow_dispatch:
 
 jobs:
-  e2e_tests:
-    name: "Run E2E Tests"
+  e2e_tests-smoke:
+    name: "Run E2E Tests (Smoke)"
     runs-on: ubuntu-latest
     environment: Dev
     steps:
       - name: Checkout
         uses: actions/checkout@v2
 
-      - name: Run e2e Tests
+      - name: Run E2E Tests
         uses: ./.github/actions/devcontainer_run_command
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/e2e.yml
```diff
@@ -6,19 +6,49 @@ on:
...
-      - name: Run e2e Tests
+      - name: Run E2E Tests
         uses: ./.github/actions/devcontainer_run_command
         with:
-          DISPLAY_NAME: "Run e2e Tests"
-          COMMAND: "make test-e2e"
+          DISPLAY_NAME: "Run E2E Tests (Smoke)"
+          COMMAND: "make test-e2e-smoke"
+          ACTIONS_ACR_NAME: ${{ secrets.ACTIONS_ACR_NAME }}
+          ACTIONS_ACR_PASSWORD: ${{ secrets.ACTIONS_ACR_PASSWORD }}
```
ross-p-smith, line 23:
```suggestion
          ACTIONS_ACR_PASSWORD: ${{ secrets.ACTIONS_ACR_PASSWORD }}
```

ross-p-smith (changes requested):
Without this tag, I don't think it will work

eladiw:
> > Without this tag, I don't think it will work
> 
> How is it working with the smoke tests without the tag?

Added

ross-p-smith (approved):

PR #1323, opened 2022-02-17
Add secondary docker cache

### LINKED ISSUE #1319: Docker cache for core images in new CI environments
## Description

As a TRE Developer
I want the CI to use docker caching when building core images (api, resource processor) for **new** environments
So that workflow will complete sooner.

## Acceptance criteria

- [ ] When building core images in a new environment, docker is still able to use cache from the "main" ACR.

tamirkamara:
Fixes #1319 

## What is being addressed

When the CI create a new environment, the ACR is empty and caching isn't possible which results in a full build for the core images. That takes time, especially the resource processor which is 5-6 minutes.

## How is this addressed

- Have docker use a secondary cache-from directive and point it to the ACR used for the main branch.


eladiw (approved):

.github/workflows/deploy_tre.yml
```diff
@@ -108,3 +108,4 @@ jobs:
       TF_STATE_CONTAINER: ${{ secrets.TF_STATE_CONTAINER }}
       TRE_ADDRESS_SPACE: ${{ secrets.TRE_ADDRESS_SPACE }}
       TRE_ID: ${{ format('tre{0}', needs.prepare-not-main.outputs.refid) }}
+      SECONDARY_CACHE_ACR_NAME: $${{ secrets.ACR_NAME }}
```
ross-p-smith, line 111: Is this the same value as line 87? I'm assuming that this is the tre_acr that main uses? If that's the case, would it be safer to use ACTIONS_ACR_NAME? Just thinking out loud that someone could write some Terraform that dropped the repo

.github/workflows/deploy_tre_reusable.yml
```diff
@@ -54,6 +54,8 @@ on:
         required: true
       TRE_ID:
         required: true
+      SECONDARY_CACHE_ACR_NAME:
```
ross-p-smith, line 57: Do you think it would be better labelled CI_CACHE_ACR_NAME

ross-p-smith (approved):

.github/workflows/deploy_tre.yml
tamirkamara: No it's not the save value as in 87. 
@marrobi and I had a discussion before this whole env in pr validation started - it would have been ideal if we could use the same ACR across all environments, but there're challenges there (especially because of porter).
If we use the ACTIONS_ACR then we'll have to push those images there too which we currently don't.
I think this is a good compromise that will only reference the main ACR as a cache but wouldn't interact with it which is something the run from main is responsible of doing.
I think that "cache" wise it will be virtually be the same since the previous version of an image in the new branch is likely be the latest from main (which is where the new branch came from).

ross-p-smith: I'm happy with that - just going to be hard for someone in a year to work out all the different acr's! But yes I agree with the compromise

Issue #1326, opened 2022-02-17
Workspace service and user resource documentation bug


daltskin:
**Describe the bug**
Following the guide here: https://microsoft.github.io/AzureTRE/tre-admins/setup-instructions/installing-workspace-service-and-user-resource results in the below error as the make targets have changed:

Error: unable to pull bundle before invoking explain command: unable to pull bundle [trename].azurecr.io/tre-service-guacamole:v0.1.9: unable to pull remote bundle: [trename].azurecr.io/tre-service-guacamole:v0.1.9: not found

**Steps to reproduce**

1.  Follow https://microsoft.github.io/AzureTRE/tre-admins/setup-instructions/installing-workspace-service-and-user-resource
1.  make bundle-register DIR=./templates/workspace_services/guacamole BUNDLE_TYPE=workspace_service

PR #1327, opened 2022-02-17
Update docs with new make target for publish & register of bundles

daltskin:
# PR for issue #1326

Fix documentation following make target changes in #1302


tamirkamara (approved):
Good catch. Can you just give this PR a nicer title?

ross-p-smith (approved):

Issue #1330, opened 2022-02-17
Standardise on vscode files


ross-p-smith:
We have two places where we define vscode settings/launch and tasks.

Define them in 1 place so it is consistent: - devcontainer.json

PR #1331, opened 2022-02-17
Removed .vscode files

ross-p-smith:
# PR for issue #1330 

## What is being addressed
- Moved launch.json and tasks.json into devcontainer.json
- Also fixed up a missing key/value pair (address_space_size) from the Swagger example


daltskin (approved):
LGTM

PR #1337, opened 2022-02-18
Fix up typo in docs with bundle-register

ross-p-smith:
make bundle-register DIR=./templates/workspace_services/guacamole BUNDLE_TYPE=workspace_service

previous command had wrong path


daltskin (approved):
LGTM

PR #1339, opened 2022-02-18
Shared workspace storage

### LINKED ISSUE #1266: Shared storage in the workspace
As a researcher I want to access storage shared between users in my workspace across my Windows and Linux virtual machines without additional configuration.

# Acceptance Criteria
- [ ] A User can log onto a Windows VM and access shared storage without running any scripts or manual configuration
- [ ] A User can log onto a Linux VM and access shared storage without running any scripts or manual configuration

oliver7598:
Fixes #1266 

## What is being addressed
User resources deployed within a workspace should have access to a shared file storage.

## How is this addressed
A file share has been added to the storage account deployed as part of a base workspace.  This is then used as part of the user resource templates to mount the drive on deployment.

For Linux, the drive is mounted and can be accessed through the file shares name at root.
This was achieved by:
- Creating the required folders on disk.
- Storing the connection credentials to the SMB
- installing autofs 
- Configuring autofs to mount the SMB
- Symlink the created folder into root as autofs only loads the mount on access request.
- Run the above as part of the existing vm_config.sh script which runs via cloud-init

For Windows the drive is mounted each time on user login.
- Command line command to mount the SMB
- Command stored as a .cmd file to run as a startup script
- Script runs on user login and mounts the drive
- Above is triggered via custom script extension.

Share name and a True/False flag on whether to provide shared storage to the VM can be provided as input arguments via the swagger api.  These currently default to "vm-shared-storage" and "true".

If a different shared storage name is provided the base workspace storage.tf file must be amended to reflect this.


templates/workspace_services/guacamole/user_resources/guacamole-azure-linuxvm/template_schema.json
```diff
@@ -16,6 +16,22 @@
...
+        "$id": "#/properties/shared_storage_access",
+        "type": "string",
+        "title": "Shared Storage Access",
+        "description": "true/false on whether to set-up shared storage access",
+        "enum": [
+          "true",
+          "false"
+        ]
+      },
+      "shared_storage_name": {
```
marrobi, line 30: Can we set a default? Not sure if we should be exposing this at all. But doesn't hurt.

templates/workspace_services/guacamole/user_resources/guacamole-azure-windowsvm/template_schema.json
```diff
@@ -16,6 +16,22 @@
...
+        "$id": "#/properties/shared_storage_access",
+        "type": "string",
+        "title": "Shared Storage Access",
+        "description": "true/false on whether to set-up shared storage access",
+        "enum": [
+          "true",
+          "false"
+        ]
+      },
+      "shared_storage_name": {
```
marrobi, line 30: Same as above, can we have a default?

ross-p-smith:
This branch is now in the repo and currently building - https://github.com/microsoft/AzureTRE/actions/runs/1864697404

templates/workspace_services/guacamole/user_resources/guacamole-azure-linuxvm/template_schema.json
oliver7598: Default value of "vm-shared-storage" for name and "True" for shared storage access is set in porter.  The name used is the name of the file share created as part of the workspace storage deployment.

marrobi: The json schema is used to create a UI. So if is blank the UI will provide a blank form input, and send a blank value overriding porter.

marrobi: Hmm, you might be right, images choice are loading with nothing selected. So they would definitely be better with a default. 

Thinking more I'm not sure we want/need end users to be able to specify the shared storage name as leaves a lot of potential for providing invalid values that can't be validated client side, or via the API prior to accepting the request.

oliver7598: I'll omit the name from the template and just have the option to add/remove the shared storage access instead

templates/workspace_services/guacamole/user_resources/guacamole-azure-windowsvm/template_schema.json
oliver7598: Resolving as linked to review on guacamole-azure-linuxvm/template_schema.json

templates/workspace_services/guacamole/user_resources/guacamole-azure-linuxvm/template_schema.json
oliver7598: This has now been actioned.  Happy for me to resolve this?

oliver7598: Only argument I can think pro having the end user be able to enable/disable the access would be how granular the end users would like the IG control.  Currently all shared storage is read/write so data could be moved round between VMs in a workspace.

oliver7598: No problem will action that now and commit the changes.

templates/workspaces/base/terraform/storage.tf
```diff
@@ -8,6 +8,12 @@ resource "azurerm_storage_account" "stg" {
   lifecycle { ignore_changes = [tags] }
 }
 
+resource "azurerm_storage_share" "shared_storage" {
+  name                 = "vm-shared-storage"
+  storage_account_name = azurerm_storage_account.stg.name
+  quota                = 50
+}
+
```
tamirkamara, line 16: This is probably not the perfect place to comment about this, but, this storage account will now be "compromised", as in, can't be used for anything else since its key is visible from each vm.
1. Are we using it for anything else?
2. Perhaps we should make it more visible to a developer to be careful when using it? Names, comments, maybe delete the blob private endpoint since that probably not a valid use for this account?

oliver7598: The storage account referenced is the one created as part of a base workspace deployment.  Prior to the changes in this PR this storage account was not in use.  Part of this PR creates a File Share in the SA which is then made available for access to the VMs within Guacamole.

I had originally tried alternative methods of mounting the drives on both Linux and Windows although all failed to persist the drive or execute correctly.

marrobi: As the workspace is seen as the security boundary I think this is acceptable, and not sure is any other way to avoid this from a technical perspective.

The requirement I see from customers is generally when data is within the workspace any researcher should be free to access that data.

We should have some documentation calling this out. Part of the base workspace template maybe?

ross-p-smith (approved):
LGTM

PR #1353, opened 2022-02-21
Adding ID token (hybrid flow) to the app registration script

### LINKED ISSUE #1137: App Registration missing ID token needed for Guacamole
To get Guacamole to work ID token needs to be enabled in App Registration

eladiw:
Fixes #1137 

## What is being addressed

Currently when creating a new app registration (or updating) for WS, the required id token is not being set.
This is fixed in this pr

## How is this addressed

- Added missing properties in the json payload
- needed properties taken from : https://github.com/Azure/azure-cli/issues/10579#issuecomment-601643464

Screenshot:
image


scripts/aad-app-reg.sh
```diff
@@ -404,16 +428,30 @@ if [[ -n ${apiAppObjectId} ]]; then
...
     wait_for_new_app_registration $apiAppId
 
     # Update to set the identifier URI.
+    echo "Updating identifier URI"
     az ad app update --id ${apiAppId} --identifier-uris "api://${apiAppId}"
 fi
 
+echo "Setting API permissions (email / profile / ipaddr)"
+# unfortunatly we need to assign these permissions using the non descriptive guids...(taken from the manifest)
+az ad app permission add --id ${apiAppId} --api 00000003-0000-0000-c000-000000000000 --api-permissions 64a6cdd6-aab1-4aaf-94b8-3cc8405e90d0=Scope 37f7f235-527c-4136-accd-4a02d197296e=Scope 14dad69e-099b-42c9-810b-d002981feec1=Scope
```
stuartleeks, line 446: nit - could we create variables in the script for these GUIDs? At least that way we'd have something that describes them in the script 😄

tamirkamara (approved):

Issue #478, opened 2021-07-19
[Story] Configure Github workflows to allow PRs from forks to be deployed and validated


marrobi:
## Description

As a  project maintainer
I want to validate an incoming PR from a fork deploys successfully and tests pass.
So that I can merge PRs from external contributors

## Acceptance criteria

- [x] PRs that have been "reviewed" and labelled/commented on by maintainers can deploy via CD and access secrets
- [x] External contributors cannot access secrets or bypass the labelling mechanism.

See https://securitylab.github.com/research/github-actions-preventing-pwn-requests/ and suggest a solution

christoferlof:
Not sure it's a good idea to deploy from forks. We need to run validation on PRs from forks. Lets discuss how to potentially split this into multiple checks/workflows.

marrobi:
I believe we need the ability to spin up a fresh deployment based on a PR from a fork on demand.

tompaana:
Tested a possible solution (from the link in the description):

In a workflow:

```yaml
on:
  pull_request_target:
    types: [labeled]

jobs:
  build:
    ...
    if: github.event.pull_request.head.repo.full_name == github.repository || contains(github.event.pull_request.labels.*.name, 'safe to test')
```

The result: PRs created from a branch of the source repository run the workflow as usual. PRs coming from forks require the custom label, "safe to test" in this case, to run the workflow - it will be in "skipped" state until the label is added.

daltskin:
Currently E2E tests are running on the main branch, not on the fork.

E2E tests should be verifying code from the fork location merge with main.

tamirkamara:
The link in the description a great overview but also should be read carefully - at the very bottom it states:

> We have also noticed a pattern that has a vulnerable intent of use, however due to misunderstanding of pull_request_target ends up being a broken, but not vulnerable, workflow.

which is exactly what we have now.

stuartleeks:
Closing as base implementation is done. Follow-on work is tracked in #1538

PR #1356, opened 2022-02-21
Add pr_comment_bot workflow

stuartleeks:
Allow running tests by commenting `/test` on a PR as OWNER/COLLABORATOR

deploy_tre_reusable now takes two optional parameters to identify the repo and commit to checkout. If these are set then it uses that commit, otherwise retains the previous checkout behaviour

This addresses #478. Not marking as closes that issue since we can only test this once the `pr_comment_bot` workflow is in `main`

- [X] add workflow triggered by PR comments to run test workflow
- [X] move non-sensitive secrets to inputs to prevent masking in logs (will still be masked in places where the value is a repo secret)
- [x] skip build when only `.md` files have changed and add explicit check status

### DIFF
--- a/.github/workflows/deploy_tre.yml
+++ b/.github/workflows/deploy_tre.yml
@@ -22,6 +22,8 @@ jobs:
     name: "Deploy main"
     if: github.ref == 'refs/heads/main'
     uses: ./.github/workflows/deploy_tre_reusable.yml
+    with:
+      ciGitRef: ${{ github.ref }}
     secrets:
       AAD_TENANT_ID: ${{ secrets.AAD_TENANT_ID }}
       ACR_NAME: ${{ secrets.ACR_NAME }}
@@ -110,6 +112,8 @@ jobs:
     if: ${{ github.ref != 'refs/heads/main' && needs.prepare-not-main.outputs.not-md == 'true' }}
     needs: [prepare-not-main]
     uses: ./.github/workflows/deploy_tre_reusable.yml
+    with:
+      ciGitRef: ${{ github.ref }}
     secrets:
       AAD_TENANT_ID: ${{ secrets.AAD_TENANT_ID }}
       ACR_NAME: ${{ format('tre{0}', needs.prepare-not-main.outputs.refid) }}
--- a/.github/workflows/deploy_tre_reusable.yml
+++ b/.github/workflows/deploy_tre_reusable.yml
@@ -81,11 +81,11 @@ jobs:
     environment: Dev
     steps:
       - name: Checkout main
-        if: ${{ inputs.prRepo == "" }} # if not running for a PR, checkout main
+        if: ${{ inputs.prRepo == '' }} # if not running for a PR, checkout main
         uses: actions/checkout@v2
 
       - name: Checkout PR
-        if: ${{ inputs.prRepo != "" }} # if running for a PR, checkout the PR commit
+        if: ${{ inputs.prRepo != '' }} # if running for a PR, checkout the PR commit
         uses: actions/checkout@v2
         with:
           repository: ${{ inputs.prRepo }}
--- a/.github/workflows/deploy_tre_reusable.yml
+++ b/.github/workflows/deploy_tre_reusable.yml
@@ -12,6 +12,10 @@ on:
         description: The git ref to checkout
         type: string
         required: false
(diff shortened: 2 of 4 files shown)


.github/workflows/pr_comment_bot.yml
```diff
@@ -0,0 +1,151 @@
...
+          echo -e "\tLatest commit ref: $ref"
+          echo
+
+          echo "Setting outputs"
+          echo "::set-output name=prRef::${ref}"
+          echo "::set-output name=prRepo::${pr_owner}/${pr_repo}"
+          echo "Done"
+
+          REFID=$(echo ${GITHUB_REF} | shasum | cut -c1-8)
+          echo "using id of: ${REFID} for GitHub Ref: ${GITHUB_REF}"
```
tamirkamara, line 113: What will be the GITHUB_REF for a PR from a fork in the context for this workflow?

.github/workflows/deploy_tre.yml
```diff
@@ -37,17 +41,14 @@ jobs:
...
       SWAGGER_UI_CLIENT_ID: ${{ secrets.SWAGGER_UI_CLIENT_ID }}
       TEST_APP_ID: ${{ secrets.TEST_APP_ID }}
       TEST_USER_NAME: ${{ secrets.TEST_USER_NAME }}
       TEST_USER_PASSWORD: ${{ secrets.TEST_USER_PASSWORD }}
       TEST_WORKSPACE_APP_ID: ${{ secrets.TEST_WORKSPACE_APP_ID }}
       TF_STATE_CONTAINER: ${{ secrets.TF_STATE_CONTAINER }}
       TRE_ADDRESS_SPACE: ${{ secrets.TRE_ADDRESS_SPACE }}
-      TRE_ID: ${{ secrets.TRE_ID }}
 
   prepare-not-main:
```
tamirkamara, line 53: The plan is to remove the "not main" parts is the next PR?

stuartleeks: yes - I wanted to add the new way of triggering and then remove the labels approach once we confirm comments are working

.github/workflows/pr_comment_bot.yml
stuartleeks: Ah, good spot... for the PR comment trigger this will be `main`. I'll update this to use the same PR ref that will be used for the CI annotation on the deployment

.github/workflows/deploy_tre_reusable.yml
```diff
@@ -59,17 +72,25 @@ on:
...
 
 jobs:
   deploy_management:
     name: Deploy Management
     runs-on: ubuntu-latest
     environment: Dev
     steps:
-      - name: Checkout
+      - name: Checkout main
+        if: ${{ inputs.prRepo == '' }} # if not running for a PR, checkout main
```
tamirkamara, line 84: This won't always checkout main, for example if deploy_tre is "dispatched" (I'm doing it myself sometimes). So maybe "default" is more appropriate here.

.github/workflows/pr_comment_bot.yml
```diff
@@ -0,0 +1,156 @@
...
+
+          REFID=$(echo ${github_pr_ref} | shasum | cut -c1-8)
+          echo "using id of: ${REFID} for GitHub Ref: ${github_pr_ref}"
+          echo "::set-output name=refid::${REFID}"
+
+  run_test:
+    # Run the tests with the re-usable workflow
+    needs: [pr_comment]
+    if: ${{ needs.pr_comment.outputs.command == 'run-tests' && steps.filter.outputs.not-md == 'true' }}
+    name: Run tests
```
tamirkamara, line 124: I believe this name should be identical to the step above bypassing the tests (i.e. Deploy PR). But let's see what happens and fix in the second PR you're planning.

tamirkamara (approved):
Approved with a couple of small comments.

.github/workflows/deploy_tre_reusable.yml
stuartleeks: Ah yes - I have updated the name & comment

.github/workflows/pr_comment_bot.yml
stuartleeks: I've updated the name - 🤞

PR #1360, opened 2022-02-22
CI checks core image version update

### LINKED ISSUE #1359: Core images will not be used unless the version is updated
**Describe the bug**
If one makes a change to a core image (like the API) but without updating the version file, then the image won't be used by a preexisting TRE environment.
This is especially problematic while developing and running CICD as it can cause failures in the environment representing the main branch (which is preexisting).

**Steps to reproduce**

1. Deploy a new TRE environment `make all`
1. Make a change in the API that will be visible to you without changing _version.py file.
1. Update the TRE environment by running `make all` again
1. Notice that the API hasn't changed

tamirkamara:
Fixes #1359 

## What is being addressed

We tag core images with a value from a version file. If the application is updated but the version file isn't, then it might cause failures after the merge to main. 
The reason is that in PR validation we deploy a new environment, but on main we use a preexisting one. If the version file isn't change then the existing container running in the environment won't get replaced and that can produce errors in the E2E tests.

## How is this addressed

- Add a check in the CI that will enforce updating the version file when the respective application is updated. This only tests that the file was updated (git method) and not the actual number.
You can see a run result here: https://github.com/microsoft/AzureTRE/runs/5286484296?check_suite_focus=true
- An additional change is to remove the job conditions as we can run this workflow for any PR and not just native (non-fork) or with the label.

### DIFF
--- a/.github/workflows/build_docker_images.yml
+++ b/.github/workflows/build_docker_images.yml
@@ -16,11 +16,6 @@ jobs:
   docker_build:
     name: Build images
     runs-on: ubuntu-latest
-    if: |
-      github.event.pull_request.head.repo.full_name == github.repository
-      || github.event_name == 'workflow_dispatch'
-      || contains(github.event.pull_request.labels.*.name, 'safe to test')
-
     steps:
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


tamirkamara:
@damoodamoo, @eladiw - you're tagged as reviewers as you were on different ends of this issue - one made a change in the api (with no version update) and the second had tests fail after merge to main.

damoodamoo (approved):
Makes sense - a common approach might also be to use the commit sha as the version id? But this allows more deliberate use and not updating if you don't want to.

tamirkamara:
@damoodamoo There might be a reason why they chose not to go in the sha route before and I didn't want to make too many changes.
It might be something to do with the API displaying a "nice" version so that will remain.

PR #1365, opened 2022-02-22
Split out Firewall / Nexus / Gitea

damoodamoo:
Continuation of @tanya-borisova 's work, with some simplifications:

- Deployment of shared services are done via terraform only, not porter running locally. To run via porter required a new service principal and this added complexity. Since this is a stop-gap solution I'm proposing we just run the TF deployments for each one (via `make` targets)
- State for the firewall is 'moved' by being removed from the core deployment and imported into the firewall deployment. This is done via a custom make target. NOTE that the default behaviour will be to recreate the firewall. To move firewall state, run `make migrate-firewall-state`
- Gitea and Nexus will be recreated [BREAKING CHANGE]
- Gitea keyvault secrets have been gently renamed to allow for deletion/recreation 

`make tre-deploy` should be seamless to the user and no other configuration changes are required.


templates/core/terraform/deploy.sh
```diff
@@ -18,7 +18,7 @@ LOG_FILE="tmp$$.log"
   -n $TF_VAR_terraform_state_container_name \
   -k ${TRE_ID} \
   -l ${LOG_FILE} \
-  -c "terraform plan -out ${PLAN_FILE} && \
+  -c "terraform plan -out ${PLAN_FILE} && terraform state list"
   terraform apply -input=false -auto-approve ${PLAN_FILE} && \
```
tamirkamara, line 22: Just for my understanding - so apply will now run without the terraform_wrapper script. What does it mean?

templates/core/terraform/deploy.sh
```diff
@@ -18,7 +18,7 @@ LOG_FILE="tmp$$.log"
...
-  terraform output -json > ../tre_output.json"
+  terraform output -json > ../tre_output.json
```
tamirkamara, line 23: Since this also runs outside the wrapper script, does it need to be "concatenated" the the apply command?

templates/shared_services/check_sp.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
tamirkamara, line 1: Can you add generic bash options to better handle issues?
```bash
set -o errexit
set -o pipefail
set -o nounset
```

templates/shared_services/firewall/porter.yaml
```diff
@@ -0,0 +1,71 @@
...
+    description: "The name of the Terraform state storage account"
+  - name: tfstate_container_name
+    type: string
+    default: "tfstate"
+    description: "The name of the Terraform state storage container"
+
+mixins:
+  - exec
+  - terraform:
+      clientVersion: 1.0.4
```
tamirkamara, line 38: Perhaps a good opportunity to also update this to 1.1.5 (or 6 even)?
(same comment for the other bundles as well)

templates/shared_services/firewall/terraform/import_state.sh
```diff
@@ -0,0 +1,96 @@
+#!/bin/bash
```
tamirkamara, line 1: Bash options...

templates/shared_services/firewall/terraform/remove_state.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+# TODO: maybe rename script
+# TODO: add a comment explaining wtf is going on
```
tamirkamara, line 3: Bash options...
Do you plan to keeps those comments?

templates/shared_services/sonatype-nexus/porter.yaml
```diff
@@ -0,0 +1,65 @@
...
+      description: "Deploy shared service"
+      input: false
+      vars:
+        tre_id: "{{ bundle.parameters.tre_id }}"
+        location: "{{ bundle.parameters.azure_location }}"
+      backendConfig:
+        resource_group_name: "{{ bundle.parameters.tfstate_resource_group_name }}"
+        storage_account_name: "{{ bundle.parameters.tfstate_storage_account_name }}"
+        container_name: "{{ bundle.parameters.tfstate_container_name }}"
+        key: "tre-shared-service-nexus"
```
tamirkamara, line 47: I think you need the tre_id in the key as well

templates/shared_services/gitea/porter.yaml
```diff
@@ -0,0 +1,74 @@
...
+      vars:
+        tre_id: "{{ bundle.parameters.tre_id }}"
+        location: "{{ bundle.parameters.azure_location }}"
+        mgmt_resource_group_name: "{{ bundle.parameters.tfstate_resource_group_name }}"
+        acr_name: "{{ bundle.parameters.acr_name }}"
+      backendConfig:
+        resource_group_name: "{{ bundle.parameters.tfstate_resource_group_name }}"
+        storage_account_name: "{{ bundle.parameters.tfstate_storage_account_name }}"
+        container_name: "{{ bundle.parameters.tfstate_container_name }}"
+        key: "tre-shared-service-gitea"
```
tamirkamara, line 54: I think you need the tre_id in the key, like you had done for the firewall

templates/shared_services/sonatype-nexus/terraform/main.tf
```diff
@@ -0,0 +1,15 @@
+# Azure Provider source and version being used
+terraform {
+  required_providers {
+    azurerm = {
+      source  = "hashicorp/azurerm"
+      version = "=2.86.0"
```
tamirkamara, line 6:
```suggestion
      version = "=2.97.0"
```

tamirkamara (dismissed):

templates/core/terraform/deploy.sh
damoodamoo: right - this had been changed for debugging before me, and I didn't realise it was all supposed to feed into the wrapper script - i've reverted now so it's as-was.

templates/shared_services/check_sp.sh
damoodamoo: i've gone with `set -e` as that's what the majority of the other scripts use

templates/shared_services/firewall/porter.yaml
damoodamoo: 1.1.5 as it's the same version as in the DC

templates/shared_services/sonatype-nexus/porter.yaml
damoodamoo: good catch, thanks

templates/shared_services/gitea/Dockerfile.tmpl
```diff
@@ -0,0 +1,29 @@
...
+
+# Install Azure CLI (neeeded for import_state.sh)
+RUN apt-get update \
+    && apt-get install -y ca-certificates jq curl apt-transport-https lsb-release gnupg \
+    && curl -sL https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor | tee /etc/apt/trusted.gpg.d/microsoft.gpg > /dev/null \
+    && AZ_REPO=$(lsb_release -cs) \
+    && echo "deb [arch=amd64] https://packages.microsoft.com/repos/azure-cli/ $AZ_REPO main" | tee /etc/apt/sources.list.d/azure-cli.list \
+    && apt-get update && apt-get -y install azure-cli
+
+# This is a template Dockerfile for the bundle's invocation image
```
martinpeck, line 15: do we need these comments?

templates/shared_services/sonatype-nexus/Dockerfile.tmpl
```diff
@@ -0,0 +1,21 @@
+FROM debian:stretch-slim
+
+ARG BUNDLE_DIR
+
+RUN apt-get update && apt-get install -y ca-certificates
+
+# This is a template Dockerfile for the bundle's invocation image
```
martinpeck, line 7: comments needed?

templates/shared_services/sonatype-nexus/azure.json
```diff
@@ -0,0 +1,32 @@
...
+      }
+    },
+    {
+      "name": "azure_tenant_id",
+      "source": {
+        "env": "ARM_TENANT_ID"
+      }
+    }
+  ]
+}
```
martinpeck, line 32: nit: missing new line

martinpeck (changes requested):

templates/shared_services/sonatype-nexus/Dockerfile.tmpl
damoodamoo: don't know - it's the boilerplate comments, I don't seen any reason to remove it?

damoodamoo: I think they do no harm and might provide some instruction to the next person who wants to work with the dockerfile?

templates/shared_services/firewall/parameters.json
```diff
@@ -0,0 +1,38 @@
+{
+  "schemaVersion": "1.0.0-DRAFT+TODO",
```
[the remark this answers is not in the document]
damoodamoo: they're made up - i've removed the TODO from here but they're just identifiers for the version of this bundle

martinpeck (approved):

PR #1369, opened 2022-02-23
Add Alternative Method of running letsencrypt.sh

### LINKED ISSUE #1233: 'make letsencrypt' fails while configuring network storage rules

### LINKED ISSUE #1368: letsencrypt.sh fails when running from AzDO in same region as the deployment
When `letsencrypt.sh` is run is attempts to write a file into the App Gateway storage account (the static web app) to create an `index.html`. In order to do this it needs to...

1. add the *public* IP address of the machine running the script into the Allow list
2. write the file
3. remove the *public* IP address of the machine

This works fine *apart from* a situation where the machine running the script has access to internal networking routes to the storage account. So, in a situation where...

- An Azure Host Build Agent runs this script from (for example) the `UK South` region
- The deployment, and hence storage account, is in that same region

... then an internal route is taken and public IP is never used.

I've tried the following...

1. Tried to add the private/internal IP to the allow-list (won't work: you can only add public IPs)
2. Tried turning on routing preferences (no difference...still fails)
3. Tried using internetrouted endpoints (no difference...still fails)

Using logging on the storage account, in all cases I see a 10.*.*.* address being used to access the storage account, and consequently rejected due to firewall rules.

Work arounds:

1. Turn the fire wall rules off during this step
2. Ensure that your build agent is in a different data centre
3. Use a custom build agent with NAT and public IP.

martinpeck:
This PR fixes #1368 by providing an alternative strategy to working with the storage account network rules

### DIFF
--- a/templates/core/terraform/scripts/letsencrypt.sh
+++ b/templates/core/terraform/scripts/letsencrypt.sh
@@ -8,15 +8,42 @@ if [[ -z ${STORAGE_ACCOUNT} ]]; then
   exit 1
 fi
 
-IPADDR=$(curl ipecho.net/plain; echo)
-
 # The storage account is protected by network rules
-# The rules need to be temporarily lifted so that the index.html file, if required, and certificate can be uploaded
-echo "Creating network rule on storage account ${STORAGE_ACCOUNT} for $IPADDR"
-az storage account network-rule add \
-  --account-name "${STORAGE_ACCOUNT}" \
-  --resource-group "${RESOURCE_GROUP_NAME}" \
-  --ip-address $IPADDR
+#
+# The rules need to be temporarily lifted so that the script can determine if the index.html file
+# already exists and, if not, create it. The firewall rules also need lifting so that the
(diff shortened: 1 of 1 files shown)


stuartleeks (approved):
LGTM - having this as opt-in feels like a good approach

PR #1372, opened 2022-02-24
Purge keyvaults to allow reuse of the same TRE ID

tamirkamara:
## What is being addressed

Reusing the same TRE ID is blocked because the keyvault is deleted but not purged

## How is this addressed

- Delete and purge a keyvault if protection is off
- Move the "destroy" logic into a separate script to allow it to be reused later (as a make target)
- Increase the timeout allowed for the ci workflow since purging a keyvault can take time


stuartleeks (approved):
LGTM!

PR #1379, opened 2022-02-25
Revert to author_association check

stuartleeks:
Continued work on #478 

A previous PR moved from the `author_association` check to attempting to test for write access in the repo using the `/repos/{owner}/{repo}/collaborators/{username}` api ([docs](https://docs.github.com/en/rest/reference/collaborators#check-if-a-user-is-a-repository-collaborator)). While this worked well in a test repo, it failed here. The API always returned a 404, even for endpoints that were tested on a developer laptop with a PAT (the PAT required authorisation for the microsoft org).

After further investigation/testing, we tried adding my user directly to the collaborators (rather than as part of the group that is added) with the same permissions as it has in the group, and the `author_association` value suddenly started to identify me as a `COLLABORATOR`. During this test, the collaborators API also identified me as a collaborator.

This PR reverts to the `author_association` check as both options appear to have the same requirement and it is the simpler of the two options. The `author_association` check also allows the test to be part of the `if` condition on the workflow step, avoiding running the workflow when the condition isn't met, which saves workflow execution time.

### DIFF
--- a/.github/workflows/pr_comment_bot.yml
+++ b/.github/workflows/pr_comment_bot.yml
@@ -28,7 +28,7 @@ jobs:
     # only allow commands where:
     # - the comment is on a PR
     # - the commenting user has write permissions (i.e. is OWNER or COLLABORATOR)
-    if: ${{ github.event.issue.pull_request }}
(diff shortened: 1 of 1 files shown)


martinpeck (approved):
Based on discussions yesterday, lgtm

PR #1383, opened 2022-02-25
Fix ordering. Update to use PR merge ref

stuartleeks:
Updates for #478 

- Update ordering of steps and set the ref explicitly on the `dorny/paths-filter` step to ensure the diff is correct
- Update to use the `potentialMergeCommit` SHA instead of the latest PR commit as this seems to be what the checks are expecting (it also more closely mirrors the `pull_request` trigger behaviour)


jjgriff93 (approved):
‘Nit’ some of the variables are camelcased whereas GitHub actions bars tend to be snakecase, just a consistency thing

.github/workflows/pr_comment_bot.yml
```diff
@@ -142,7 +148,7 @@ jobs:
     uses: ./.github/workflows/deploy_tre_reusable.yml
     with:
       prRef: ${{ needs.pr_comment.outputs.prRef }}
-      prRepo: ${{ needs.pr_comment.outputs.prRepo }}
+      # prRepo: ${{ needs.pr_comment.outputs.prRepo }}
```
daltskin, line 151: Maybe remove this?

PR #1400, opened 2022-03-01
Add header to RP prune script

### LINKED ISSUE #1399: Resource Processor cron job fails
**Describe the bug**
The cron job in the Resource Processor responsible of pruning docker and freeing up disk space fails to run

**Steps to reproduce**

1. Run via cron tab with the following command: `cd / && run-parts --verbose --report /etc/cron.weekly`
1. See it fails with: `exec format error`1.

tamirkamara:
Fixes #1399 

Crontab failed to execute the script and disk space run out.

## How is this addressed

- Not sure what shell crontab (run-parts command) uses when it tries to execute the script, so specifying it solved the problem.

A small additional change is to debug the destroy script results when run in the CI


ross-p-smith (approved):

templates/core/terraform/main.tf
```diff
@@ -14,7 +14,7 @@ provider "azurerm" {
   features {
     key_vault {
       purge_soft_delete_on_destroy    = var.keyvault_purge_protection_enabled ? false : true
-      recover_soft_deleted_key_vaults = true
+      recover_soft_deleted_key_vaults = false
```
stuartleeks, line 17: Was this intended to be included in this PR? It's not in the PR description (and probably warrants discussion?)

tamirkamara: It was an attempt to fix an env problem. Reverted now.

Issue #1420, opened 2022-03-02
Ability to be able to debug E2E tests


ross-p-smith:
As a TRE Developer I want the variables defined in various .ENV files to be the same across the Unit Tests / E2E Tests / the api_app and the resource Provider so that I can have a central .ENV file that is generated for me.

## Acceptance criteria

- [x] USERNAME= TEST_USER_NAME #1499
- [x] PASSWORD=TEST_USER_PASSWORD #1499
- [x] CLIENT_ID=TEST_APP_ID #1499
- [X] ~~TEST_WORKSPACE_APP_ID=??~~
- [X] RESOURCE=API_CLIENT_ID #1424 
- [X] SCOPE="api://${RESOURCE}/user_impersonation" #1424
- [X] WORKSPACE_SCOPE="api://${TEST_WORKSPACE_APP_ID}/user_impersonation" #1424
- [X] Amend the aad reg script to tell users to copy the values into /workspaces/AzureTRE/templates/core/.env #1437 
- [x] Amend the workspace registration script so that the automation Admin App is given the workspace Owner permission #1468

PR #1424, opened 2022-03-02
Ability to locally debug E2E tests

ross-p-smith:
# PR for issue #1420 

## What is being addressed

Debug the E2E tests without setting any .ENV files

## How is this addressed

- Streamline some ENV vars
- Add correct launch settings
- Amend Smoke Tests to use AUTOMATION ACCOUNT


.github/actions/devcontainer_run_command/action.yml
```diff
@@ -45,10 +45,10 @@ inputs:
   ARM_SUBSCRIPTION_ID:
     description: "Azure Subscription Id."
     required: true
-  RESOURCE:
+  API_CLIENT_ID:
     description: "The API Client Id."
     required: true
-  AUTH_TENANT_ID:
+  AAD_TENANT_ID:
     description: "The Tenant Id where the App is registered."
```
tamirkamara, line 52: Isn't this the tenant of the TEST_USER?
If yes, maybe a better name and description is in order?

tamirkamara (changes requested):

.github/actions/devcontainer_run_command/action.yml
ross-p-smith: Yes, but they are the same thing. It's confusing because we were copying AAD_TENANT_ID into AUTH_TENANT_ID. This means that we can set one value in /workspaces/AzureTRE/templates/core/.env as per the docs and the tests just work

PR #1437, opened 2022-03-03
Docs fixes

ross-p-smith:
## What is being addressed
The docs are not building on main, I believe because of a missing reference to the guacamole linux vm


stuartleeks (approved):
LGTM

PR #1439, opened 2022-03-03
Sort RGs before delete

### LINKED ISSUE #1438: Destroy might fail when workspace are involved
**Describe the bug**
A destroy action might fail when a workspace has certain reference on the core-rg routetable.

**Steps to reproduce**

Not easy to reproduce, but this is an example:
https://github.com/microsoft/AzureTRE/runs/5409819486?check_suite_focus=true

tamirkamara:
Fixes #1438 

## What is being addressed

Workspace group might have references to the routetable in the core resource group. The destroy script will first try to delete the core RG and that will fail in these situations.

## How is this addressed

- Reverse sort the resource groups so workspace ones (e.g. rg-treid-ws-12ab) will be deleted first

### DIFF
--- a/devops/scripts/destroy_env_no_terraform.sh
+++ b/devops/scripts/destroy_env_no_terraform.sh
@@ -109,7 +109,9 @@ else
 fi
 
 # this will find the mgmt, core resource groups as well as any workspace ones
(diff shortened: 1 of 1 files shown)


tamirkamara:
Forced approved since this script isn't used by the deployment workflow.

stuartleeks (approved):

Issue #1267, opened 2022-02-08
Specify the storage quota for workspace shared storage


marrobi:
As a TRE Administrator when deploying a workspace I want to specify a storage  quota for shared storage when deploying a workspace

oliver7598:
I can pick this up as a continuation of my shared storage work.  Would have addressed in recent PR but only just seeing this.

PR #1441, opened 2022-03-04
Exposed shared storage quota for workspace

oliver7598:
# PR for issue
#1267 

## What is being addressed
Currently when shared storage is created as part of the base workspace it is set to have a quota for 50.

## How is this addressed
- Base workspace version number iterated 
- Optional property added to `template_schema.json` called `shared_storage_quota` with a default value of 50.
- Property mapped to workspace storage account file share vm-shared-storage.


jjgriff93 (approved):
Looks good!

templates/workspaces/base/terraform/variables.tf
```diff
@@ -8,6 +8,11 @@ variable "tre_resource_id" {
   description = "Resource ID"
 }
 
+variable "shared_storage_quota" {
```
martinpeck, line 11: can you add a description to this variable please (other variables have descriptions).
I would suggest...

```Quota (in GB) to set for the VM Shared Storage.```

File: templates/workspaces/base/template_schema.json [hunk 2]
```diff
@@ -7,5 +7,11 @@
     "required": [
     ],
     "properties": {
+      "shared_storage_quota": {
+        "$id": "#/properties/shared_storage_quota",
+        "type": "integer",
+        "title": "Shared Storage Quota'",
+        "description": "Quote (in GB) to set for the VM Shared Storage."
```
martinpeck, line 14: Typo `Quote` should be `Quota`

martinpeck (changes requested):

templates/workspaces/base/terraform/variables.tf
oliver7598: Changes implemented

martinpeck (approved):
LGTM. Thanks!

martinpeck:
Tests have been previously run (3 times)

Issue #1306, opened 2022-02-15
Remove manual steps from each set of template docs


marrobi:
## Description

At present template docs include manual steps which confuse Administrators as they follow these steps rather than using the API to deploy the workspace.

We should move the method around manually deploying the templates into the developer docs with general documentation.

## Acceptance criteria

- [x] Manual steps moved from each template docs to developer docs

PR #1443, opened 2022-03-04
Move templates section under dev docs

daltskin:
# PR for issue #1306 

## What is being addressed

Reduce confusion within the docs by moving out the template manual steps.

## How is this addressed

- Moved out the entire Templates section under the TRE Developers section to keep it all under one area


marrobi:
I think we still need a list of the services, and each one should ideally have a diagram and overview of what purpose it serves. From a Workspace Owner/Researcher perspective

I don't think we need the manual steps for individual service,  but a description of how to deploy a template manually as part of the authoring process.

Issue #1414, opened 2022-03-02
Anaconda Repo Nexus Proxy


oliver7598:
DSVMs within Guacamole are currently deployed with anaconda although have no access to the anaconda repo for package management,

Currently `conda install` commands fail as they are trying to get the packages via https://repo.anaconda.com/ rather than a configured proxy.

To amend this:
- [x]  Nexus needs configuring to proxy https://repo.anaconda.com/
- [x] Guacamole user resources need to configure the proxy on initial boot.

oliver7598:
Can this be assigned to me

oliver7598:
Branch merged - issue can be closed

PR #1454, opened 2022-03-07
Nexus conda config

oliver7598:
# PR for issue
#1414 

## What is being addressed
DSVMs within Guacamole are currently deployed with anaconda although have no access to the anaconda repo for package management,

Currently conda install commands fail as they are trying to get the packages via https://repo.anaconda.com/ rather than a configured proxy.

## How is this addressed
- `repo.anaconda.com` added to `nexus_allowed_fqdns` variable to allow the external repo to be reached through the firewall
- `configure_nexus.sh` script amended to configure conda proxy endpoint
- Linux and Windows Guacamole user resources amended to configure conda proxy on User Resource deployment


marrobi (commented):
@oliver7598 nice way to handle the requirement for conda binaries to be there.

There will likely be times where users need to configure conda/pip etc post deployment - for example in a container that hasn't had the cloud init scripts run.

Not necessarily part of this issue, but are the scripts available/can they be made available to be run interactively to configure the repositories?

oliver7598:
@marrobi At current these script probably wont be available post deployment but it should be easy enough to extract them out and make them available.

In addition to this I'm assuming exposing the download link for Anaconda/PIP via Nexus would also be required to enable this?  If so I could pick this up after addressing another issue I am looking at and assuming no other higher priority issues arise.

marrobi:
@oliver7598 oli
> In addition to this I'm assuming exposing the download link for Anaconda/PIP via Nexus would also be required to enable this?

To install the binaries? Good point, PIP can be done via apt which is mirrored, but might be need to make other binaries not covered by a package manager available via nexus in raw format.

scripts/configure_nexus.sh
```diff
@@ -102,7 +102,7 @@ if [[ ${STATUS_CODE} == 404 ]]
 fi
 
 #Check if the repo already exists
-export STATUS_CODE=$(curl -iu admin:$NEXUS_PASS -X "GET" "${NEXUS_URL}/service/rest/v1/repositories/apt/proxy/pypi" -H "accept: application/json" -k -s -w "%{http_code}" -o /dev/null)
+export STATUS_CODE=$(curl -iu admin:$NEXUS_PASS -X "GET" "${NEXUS_URL}/service/rest/v1/repositories/proxy/pypi" -H "accept: application/json" -k -s -w "%{http_code}" -o /dev/null)
```
martinpeck, line 105: I'm probably missing something, but why remove the `apt` from the URL?

martinpeck (changes requested):
Looks good, but I have one question

scripts/configure_nexus.sh
oliver7598: Did this as a error correct as there is both a pypi_proxy_conf.json and apt-pypi_proxy_conf.json.
One is an apt repo and one is not. The curl request checks to see if the repo already exists before creating it so the standard proxy will never be created at the moment due to this error.

martinpeck: Ah, right. Got it.

martinpeck (approved):

martinpeck:
Thanks @oliver7598 !

PR #1461, opened 2022-03-08
Debug Logs

damoodamoo:
WIP


.github/workflows/deploy_tre.yml
```diff
@@ -146,3 +146,4 @@ jobs:
       TRE_ADDRESS_SPACE: ${{ secrets.TRE_ADDRESS_SPACE }}
       TRE_ID: ${{ format('tre{0}', needs.prepare-not-main.outputs.refid) }}
       CI_CACHE_ACR_NAME: ${{ secrets.ACR_NAME }}
+      TF_LOG: ${{ secrets.TF_LOG }}
```
ross-p-smith, line 149: is the value a secret? Would it be better as a normal input and then you could see its value in the logs? secrets end up being ***

stuartleeks: If it's coming from a GH secret in the repo, doesn't that get masked regardless of how you pass it in to other actions etc?

PR #1464, opened 2022-03-08
Fix: update `/test` builds to ensure PR code is tested

stuartleeks:
Remove the `checkout` step in devcontainer_run_command action as the parent jobs already include `checkout` steps (and target the ref for the PR)

### DIFF
--- a/.github/actions/devcontainer_run_command/action.yml
+++ b/.github/actions/devcontainer_run_command/action.yml
@@ -119,11 +119,6 @@ inputs:
 runs:
   using: composite
   steps:
-    - name: Checkout
-      uses: actions/checkout@v2
-      with:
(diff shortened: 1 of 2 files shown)


stuartleeks:
This line shows the PR code being exercised: https://github.com/microsoft/AzureTRE/runs/5466574832?check_suite_focus=true#step:7:120

This line from the previous run (which still had the duplicated checkouts) didn't output the test output: https://github.com/microsoft/AzureTRE/runs/5466411136?check_suite_focus=true#step:7:589

damoodamoo (approved):

stuartleeks:
(force-approve to avoid re-testing after reverting the test code)

PR #1465, opened 2022-03-08
API health endpoint returns service bus and cosmos connectivity

anatbal:
# PR for issue #1361

## What is being addressed
 
 - API health endpoint returns service bus and cosmos connectivity.
- This PR removes the old health check which is just checking the api webapp is up and running, and replacing it with pinging the service bus and cosmos. 
- This includes removing the /status api route since it's improved version is now /health


api_app/services/health_checker.py
```diff
@@ -1,24 +1,59 @@
...
 from models.schemas.status import StatusEnum
 from resources import strings
 
 
+@asynccontextmanager
+async def default_credentials():
+    """
+    Context manager which yields the default credentials.
+    """
+    credential = DefaultAzureCredential(managed_identity_client_id=config.MANAGED_IDENTITY_CLIENT_ID)
```
ross-p-smith, line 21: Will this work when we debug this locally? (Running not as a MSI?)

anatbal: Seems like it works. I debugged locally

ross-p-smith:
image - Positive test works!

anatbal:
> I configured my cosmos endpoint incorrectly and I got this. image Should it still return a json message with that error in the "message" element of the cosmos check and still return the Service Bus status .

Fixed it

templates/core/terraform/appgateway/appgateway.tf
```diff
@@ -107,7 +107,7 @@ resource "azurerm_application_gateway" "agw" {
     protocol                                  = "Https"
     path                                      = "/api/health"
     timeout                                   = "30"
-    unhealthy_threshold                       = "3"
+    unhealthy_threshold                       = "10"
```
martinpeck, line 110: I'm curious why we've increased this threshold from 3 to 10.

Previously we'd wait 45 seconds (15 x 3) before declaring the API unhealthy
Now we'll wait 150 seconds.

Have we picked this number based upon something related to the addition of Service Bus to the health endpoint?

api_app/api/routes/api.py
```diff
@@ -7,16 +7,15 @@
 
 from api.dependencies.database import get_repository
 from db.repositories.workspaces import WorkspaceRepository
-from api.routes import health, status, workspaces, workspace_templates, workspace_service_templates, user_resource_templates, shared_service_templates
+from api.routes import health, workspaces, workspace_templates, workspace_service_templates, user_resource_templates, shared_service_templates
 from core import config
 
 core_tags_metadata = [
-    {"name": "health", "description": "Verify that the API is up and running"},
+    {"name": "health", "description": "Verify that the system is up and running"},
```
martinpeck, line 14: If we're changing this, I suspect `TRE` is better than `system`.

templates/core/terraform/appgateway/appgateway.tf
anatbal: You are right. My bad. For some reason I only looked at the 3 and didn't do the 15*3 calculation. 45 seconds should be enough even with the Service Bus addition. Returned it to the original value

ross-p-smith (approved):

PR #1466, opened 2022-03-08
Fix Typo in Base Template

martinpeck:
This fixes the build issue in `main`

A single quote breaks the way that CURL is called in `register_bundle_with_api.sh`. The same quote does not break the Swagger docs endpoint.

Because the automation route isn't tested by developers this wasn't seen locally. In fact, the changes to the makefile included in this PR are needed to allow testing locally.

The PR which allowed this to slip in fell foul of the fact that `/test` was testing `main`, not the PR branch.


ross-p-smith (approved):

martinpeck:
closes #1467

Issue #1436, opened 2022-03-03
Use a Service Principal for the E2E and have a scripted way to set it all up


ross-p-smith:
## Description

As a TRE Administrator
I want to use a SPN as a test user
So that I can use a Tenant that has MFA enabled

## Acceptance criteria

- [x] Move the Authentication Tenant used for the build to be the same as the subscription 
- [x] Use a Service Principal rather than a user in the E2E tests
- [x] Amend aad setup script to add the automation user to the workspace application

PR #1468, opened 2022-03-09
Automation SPN has permissions on Workspace App

ross-p-smith:
# PR for issue #1436

## What is being addressed

When we create a Workspace App Registration in AAD, we can now give permission to an Automation Account to be able to access this, this enables the client credential flow in the E2E tests and means the tests do not need to have a User. Some tenants have MFA enabled which means the user cannot login in headless mode.

## How is this addressed

- Add a new --automation-clientid switch to ./scripts/aad-app-reg.sh which takes an existing Application(ClientId)
- Retrieve the existing requiredResourceAccess for the AutomationApp
- Remove the permission from the workspace (if present) so that it is idempotent and we dont add duplicates
- Add the permission for the workspace to the Automation Account
- Grant Admin Consent


stuartleeks (approved):
LGTM

Issue #1274, opened 2022-02-09
Add Docker apt repository to default nexus configuration


marrobi:
## Description

As a Researcher
I want to use Docker on my Linux VM
So that I can run containers

See: https://docs.docker.com/engine/install/ubuntu/

## Acceptance criteria

- [x] Docker apt repository part of Nexus default configuration
- [x] I can install docker on a Linux VM as per the instructions at https://docs.docker.com/engine/install/ubuntu/

oliver7598:
This will be achieved as part of the work for issue #1089

oliver7598:
This can be closed

Issue #1089, opened 2021-12-01
Problem using Nexus to proxy docker hub


dvoet:
**Describe the bug**
I am trying to setup nexus in the AzureTRE environment to proxy docker hub. I believe I have nexus setup correctly and appropriate docker hosts added the the firewall rules. When I try to run `docker login` from a VM running in a workspace shared service subnet I see error similar to what is described [here](https://stackoverflow.com/questions/49127414/docker-login-on-private-nexus-error-response-from-daemon-login-attempt-to-ht). According to [Sonatype docs](https://help.sonatype.com/repomanager3/formats/docker-registry/ssl-and-repository-connector-configuration) docker needs to be exposed on a different port than the standard ssl port nexus is already running on. However nexus is running in an Azure App Service and I don't see a way to expose custom ports.
**Steps to reproduce**

1. setup Nexus to [proxy docker hub](https://help.sonatype.com/repomanager3/formats/docker-registry/proxy-repository-for-docker)
1. add `*.docker.io` and `*.docker.com` to firewall rules to allow nexus to access them
1. create workspace and start vm in shared services subnet
1. connect to vm using bastion and run `docker login nexus-[tre_id].azurewebsites.net:[port]`
1. see timeout error

Acceptance criteria

- [x] #1479 
- [x] #1480 
- [x] #1481 
- [ ] Configure Docker Hub proxy in Nexus using a standalone port

oliver7598:
Can this be assigned to me

tamirkamara:
@oliver7598 do you have an idea on how to address this?

oliver7598:
@tamirkamara myself and @jjgriff93 are going to pair on this as a priority over the next sprint

tamirkamara:
Sounds good. Can you please share more details on the plan once you have it?
IIRC, I was the one who added this to begin with, so let me know if there's something I can help with...

marrobi:
This doc will be useful - might be we add a second app service just for docker?

jjgriff93:
Current challenge is exposing a second port in App Service to allow docker hub connections, which is prevented by this limitation: https://docs.microsoft.com/en-us/archive/blogs/waws/things-you-should-know-web-apps-and-linux#you-can-only-expose-one-port-to-the-outside-worldapplies-to-web-app-for-containers

jjgriff93:
Potential solution is moving nexus to container instances which does support multiple ports, Oliver and I will do some investigation into this

[...]

PR #1469, opened 2022-03-09
Nexus Docker Apt Config

oliver7598:
# PR for issue #1274

## What is being addressed
Docker should be able to be installed on a Linux VM as detailed [here](https://docs.docker.com/engine/install/ubuntu/)

## How is this addressed
- Additional json config file added to configure Nexus to allow the download of the public gpg key at :
https://nexus-<tre-id>.azurewebsites.net/repository/docker-public-key/gpg
- Additional json config file added to configure Nexus to provide a dock proxy apt repo at https://nexus-<tre-id>.azurewebsites.net/repository/docker
- VM Deploy script had additional step to download gpg key and store at `/etc/apt/trusted.gpg.d/docker-archive-keyring.gpg`
- Apt source added signed by gpg key for https://nexus-<tre-id>.azurewebsites.net/repository/docker

## Additional Changes to Nexus Configuration
There were also a few additional changes to streamline the addition of future Nexus configurations.
- All json config files moved from `.\scripts\` to `.\scripts\nexus_config\`
- Two additional properties added to all json config files. `baseType` and `repoType`.  `baseType` is the type of proxy being configured (e,g, apt, docker, r, nuget, raw, pypi) and  `repoType` is the type of repo being set up (proxy, hosted, group)
- configure_nexus.sh now reads the `.\scripts\nexus_config\` folder and for each json file in the folder the repository will be created and configure.
- Addition of future nexus repositories will now only require a new json file to be added.

## Additional Changes to Linux `vm_config.sh` file
- Content of `/etc/apt/sources.list.d/` folder has been cleared.  On running apt-get update apt sources in this folder were used to update existing packages resulting in a series of errors being displayed in the console (See example below)
image

NOTE: I have also left in the creation of the Nexus Docker Hub proxy and the opening of the firewall to allow communication to `*.docker.com` and `*.docker.io` as a basis for addressing #1089.  Although, as mentioned in the comments in the issue, further port configuration is required to fully achieve this and there are ongoing discussions on how this will be achieved.


marrobi (approved):
LGTM

scripts/configure_nexus.sh
```diff
@@ -62,67 +62,26 @@ if [ -z "$NEXUS_PASS" ]; then
     export NEXUS_PASS=$NEW_PASSWORD
```
tamirkamara, line 62: Not the perfect place to put this comment in:
Seems like we have a lot of files directly related to Nexus at the root of the project folder structure. What do you think about moving all of them into the nexus template folder @oliver7598 ?

tamirkamara (changes requested):

scripts/configure_nexus.sh
marrobi: I tend to agree. Although might be that's taken as a separate issue that includes building these steps into the nexus bundle?

tamirkamara: I'm fine with waiting as long as we open an issue and reference it here for the comment to be resolved. 
Although, I don't see any knock-on affect with regards to the nexus things (aad app reg is outside the scope). If this is going to impact github actions and/or AzDo pipelines something else is wrong as we shouldn't directly use any of this...

To clarify, we're talking about moving the configure_nexus.sh script and the nexus_config folder into the template folder.

oliver7598: Nexus config scripts and related files have been moved into the nexus shared-service folder.  

Docs also updated to reflect new script path in configure shared services section

tamirkamara (approved):
LGTM. Thank you for contributing.

Issue #1470, opened 2022-03-09
ENV Files Need Consolidating and Refactoring


martinpeck:
We have too many .env files, and pretty much everything in the `Makefile` needs values from both/all.

We need to revisit this in order to streamline dev setup.

A full review of the two .env files, plus what's actually needed for debugging locally, should take place.

marrobi:
The reason for a separate .env file for the core TRE deployment, to the devops/bootstrap, was that will likely have multiple TRE environments, hence TRE_IDs, sharing the same ACR etc.

From a GitHub actions perspective its repository secrets vs environment secrets.

Agree needs some thought.

LizaShak:
Tasks to complete the story:

1. Have one config file at root (remove the previous ones)
2. Consider what should be  the config file format:  yml/json...
3. Auth setup updates the new config file
4. Making sure all make commands are working
5. Decide how should we handle the private.env?  Make sure the local debugging works.
6. Consider all the .env.sample files, how are we handling them? for example: /workspaces/AzureTRE/templates/workspaces/base/.env.sample
7. Update the Deployment Repo  
8. Update documentation

@martinpeck @marrobi @tamirkamara WDYT? Anything you would like add?

tamirkamara:
Another item we might want to include is how to migrate existing settings to the new file.

martinpeck:
Out of JSON and YAML, I'd probably suggest that YAML is the better one given that these files will be edited by humans.

That said, have we considered TOML as the format?

It's closer to what we have (name/value pairs), while allowing us to have a little more structure. 

Also, it's very easily loaded in our Python code (via https://pypi.org/project/tomli/)

Just a thought. Please ignore it if you feel

martinpeck:
Out of interest, why are we considering YAML/JSON etc? Is that because it's easier to parse programatically, or because it allows us to add structure?

At the end of the day, we still "flattern" these files into environment variables. I assume any additional structure/hierarchy will, presumably, get lost once we export the name/value pairs as environment variables.

So, if the change of format is...

1. to make things more readable
2. to better group/structure the values

... but, we're still expecting them to be edited by hand, then we want to make reading/editing as easy as possible and as error free as possible, even if that makes parsing/processing the files a little harder for us. For me, that makes TOML a slightly better choice as it's easy to read, easy to edit, and probably quite easy to migrate from .env files.

LizaShak:
@martinpec

[...]

PR #1471, opened 2022-03-09
Add TRE_URL to .env.example file

martinpeck:
Previous PR (#1441) added additional environment variables into the mix for `make bundle-register`

The result is that without the `TRE_URL` env var set you may see an error when you run `make bundle-register` in the situation where you have these two defined env vars...

```
AUTOMATION_ADMIN_ACCOUNT_CLIENT_ID
AUTOMATION_ADMIN_ACCOUNT_CLIENT_SECRET
```

You may see this error:

image

This PR adds an addition environment variable into the `.env.example` file that the user needs to copy to the .env file within the `templates/core/` folder.

THIS IS FAR FROM PERFECT. We absolutely should create this env var ourselves, but after reviewing this for a while I realised that the .env file set up in the project is a bit messy (tracked in #1470)

For now, consider this PR as some sign-posting/documentation.

### DIFF
--- a/templates/core/.env.sample
+++ b/templates/core/.env.sample
@@ -12,10 +12,10 @@ API_CLIENT_ID=__CHANGE_ME__
 API_CLIENT_SECRET=__CHANGE_ME__
 SWAGGER_UI_CLIENT_ID=__CHANGE_ME__
 
-# Automation accounts (e.g. bundle registration)
-AUTOMATION_ADMIN_ACCOUNT_CLIENT_ID=__CHANGE_ME__
-AUTOMATION_ADMIN_ACCOUNT_CLIENT_SECRET=__CHANGE_ME__
+# The following 3 environment variables are only required
+# if you want to automated bundle registration.
+# Change TRE_URL's location (e.g. westeurope) as appropriate
 
-# Bundle Registration via Automation needs to know TRE_URL.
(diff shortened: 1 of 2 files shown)


templates/core/.env.sample
```diff
@@ -15,3 +15,7 @@ SWAGGER_UI_CLIENT_ID=__CHANGE_ME__
 # Automation accounts (e.g. bundle registration)
 AUTOMATION_ADMIN_ACCOUNT_CLIENT_ID=__CHANGE_ME__
 AUTOMATION_ADMIN_ACCOUNT_CLIENT_SECRET=__CHANGE_ME__
+
+# Bundle Registration via Automation needs to know TRE_URL.
```
marrobi, line 19: Thinking....

Can we put these "optional" or rather CI/CD type variables under a separate header saying "Automated deployments such as deployments which are part of CI/CD pipelines will need the following variables:"

For users wanting to kick the tires, they do not need to set these, or the above automation env vars, so is slightly confusing.

Happy to discuss.

martinpeck: so, comment all three of these out in the example?

marrobi: Great. Thanks.

marrobi (approved):
LGTM

PR #1483, opened 2022-03-10
[IGNORE] debug repro

damoodamoo:
repro PR, will be abandoned


stuartleeks:
/test-destroy-env

recovered the KV, so this destroy should find and purge it

PR #1487, opened 2022-03-10
Remove template manual steps from docs and generalise

daltskin:
# PR for issue #1306 

## What is being addressed

Remove the repetitive manual steps and generalise under the registration section

## How is this addressed

- Updated env files with comments from the previous env tables for each service


ross-p-smith:
By removing the steps taht a developer has to go through - are we not making it harder to deploy as a developer? I know have to go and work out the paths and the names of the bundles. I think we are missing something that allows me to `make guacamole-windows-vm` and it interacts with the API. By removing this, I dont know how I would go about automating it - happy for you to close this!

daltskin:
The current docs are organised by role.  As the admins are the ones who deploy/create workspaces they will use the api/scripts.  The manual steps were removed to avoid causing them confusion of trying to follow them.

martinpeck (approved):
LGTM

martinpeck:
I somewhat agree with @ross-p-smith. While this resolves the repetition in the docs, it means you have to have read all of the docs to know about manual registration. Couldn't we have left a link to Manual Deployment in each of these templates?

PR #1493, opened 2022-03-13
fix delete e2e workspace

tamirkamara:
## What is being addressed

Cleanup script always skips deleting old e2e workspaces.

## How is this addressed

- Adjust the running workflow condition to not check the cleanup workflow itself before deciding to delete workspaces.

### DIFF
--- a/devops/scripts/clean_ci_validation_envs.sh
+++ b/devops/scripts/clean_ci_validation_envs.sh
@@ -65,9 +65,10 @@ while read -r rg_name rg_ref_name; do
   fi
 done
 
-# check if any workflows run on the main branch (except the cleanup one)
+# check if any workflows run on the main branch (except the cleanup=current one)
 # to prevent us deleting a workspace for which an E2E (on main) is currently running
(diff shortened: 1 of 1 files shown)


tamirkamara:
/test-force-approve
This change isn't checked via the tests.

devops/scripts/clean_ci_validation_envs.sh
```diff
@@ -65,14 +65,16 @@ while read -r rg_name rg_ref_name; do
   fi
 done
 
-# check if any workflows run on the main branch
+# check if any workflows run on the main branch (except the cleanup one)
 # to prevent us deleting a workspace for which an E2E (on main) is currently running
-if [ $(gh api "https://api.github.com/repos/microsoft/AzureTRE/actions/runs?branch=main&status=in_progress" --jq ".total_count") == 0 ]
+if [[ -z $(gh api "https://api.github.com/repos/microsoft/AzureTRE/actions/runs?branch=main&status=in_progress" | jq '.workflow_runs | select(.[].name != "Clean Validation Environments")') ]]
```
eladiw, line 70: is there a possibility to use id of some kind instead of name? so it won't break if someone rename

tamirkamara: I used the name from the github context which is even better...

eladiw (approved):

PR #1495, opened 2022-03-13
Fix delete e2e workspaces (2)

tamirkamara:
## What is being addressed

Following #1493 another fix had to be made to correct the way jq uses value from the environment.


eladiw (approved):

tamirkamara:
/test-force-approve
This workflow/script isn't tested.

PR #1498, opened 2022-03-14
Update getting-started.md

chboudry:
doc update : 

Github codespace is in preview and can only be used by Microsoft Github organization member. 
Customers won't be able to use it for this project.


martinpeck:
> Customers won't be able to use it for this project

Customers have to fork the repo to work on it anyway, so any customer forking this into a GitHub org should be able to use it.

I don't think we want to remove this tip. The link takes the reader to the relevant information and can help them determine If Codespaces is an option, or not.

PR #1499, opened 2022-03-14
Better variable names for Test account

ross-p-smith:
# PR for issue #1420 

## What is being addressed
This was the last tidy up task of renaming USER, PASSWORD, and CLIENT_ID to be TEST_USER_NAME, TEST_USER_PASSWORD, and TEST_APP_ID so that they matched what was setup in the GH secrets and to be more explicit in the code what they are used for.


martinpeck (approved):
LGTM

Issue #1480, opened 2022-03-10
(Nexus Docker Hub) [Spike] Generate Letsencrypt SSL cert


jjgriff93:
As part of solution to #1089, we need to generate an SSL certificate for consumption by Nexus (as it will no longer be hosted on App Service due to #1479 and thus won't have an included cert) so that it can proxy Docker Hub over https.

An approach outlined by @marrobi can potentially address this:

```
 ┌──────────────────────────────────────────┐    ┌────────────────────────────────┐
 │                                          │    │                                │
 │ Public IP: mytre123.uksouth.cloudapp.net │    │  Storage static site           │
 │                                          ├───►│                                │
 │                                          │    │                                │
 └──────────────────────────────────────────┘    └────────────────────────────────┘
                                                        ▲
                                                        │
                                                        │
                                                        │
                                                        │
                                                        │
 ┌──────────────────────────────────┐                   │
 │                                  ├───────────────────┘
 │  Letsencrypt cert                │
 │  processor (run on schedule)     ├────────────────────┐
 │                                  │                    │
 └──────────────────────────────────┘                    │
                                                         │       ▼
                                                         │
                                                         │
                                                         │
                                                 ┌───────▼────────────────────────┐
                                                 │                                │
                                                 │  Key Vault                     │
                                                 │                                │
                                                 └────────────────────────────────┘
                                                         ▲
                                                         │
                                                         │
                                                         │
                          ┌──────────────────────────────┼────────────────────────────────────┐
                          │                              │                                    │
                          │   VNET                       │                                    │
                          │                              │                                    │
                          │                              │                                    │
                          │                      ┌───────┴─────────────────────────────┐      │
                          │                      │                                     │      │
                          │                      │ VM                                  │      │
                          │                      │                                     │      │
                          │                      │                                     │      │
                          │                      └─────────────────────────────────────┘      │
                          │                                                                   │
                          │                                                                   │
                          │                                                                   │
                          │                      ┌─────────────────────────────────────┐      │
                          │                      │                                     │      │
                          │                      │  Private DNS Zone:                  │      │
                          │                      │  map mytre123.uksouth.cloudapp.net  │      │
                          │                      │         to VM IP address            │      │
                          │                      │                                     │      │
                          │                      │                                     │      │
                          │                      │                                     │      │
                          │                      └─────────────────────────────────────┘      │
                          │                                                                   │
                          │                                                                   │
                          │                                                                   │
[...]

Issue #1479, opened 2022-03-10
(Nexus Docker Hub) Change Nexus hosting to VM


jjgriff93:
As part of a solution to #1089, we need to host Nexus on a service that allows for exposing multiple custom ports. Currently App Service only allows one which is already consumed by Nexus' standard traffic over port 8081.

We need to move the Nexus container to be hosted by a VM:

- [x] Update the Terraform for the Nexus shared service to deploy a Linux VM
- [x] Use a cloudinit script to install necessary dependencies (i.e. Docker engine)
- [x] Also use this script to start the Nexus container and expose the port
- [x] Modify any related networking components to allow traffic to and from Nexus

marrobi:
You are still going to require a SSL cert and private endpoint to enable
 apt and pypi to work.

jjgriff93:
Yep good point, I've removed the test at the end as we don't expect this to work until the follow on task is completed

PR #1514, opened 2022-03-15
Nexus Letsencrypt Certificate

oliver7598:
# PR for issue #1480

Related to #1089 and #1479

## What is being addressed

Lets Encrypt certificate to be generated for `nexus-<tre_id>.<location>.cloudapp.azure.com` to allow use in newly hosted nexus service in a VM (#1479).

## How is this addressed

- Additional storage account created hosting a static site
- Public IP created with domain `nexus-<tre_id>`
- Application gateway used to route IP to static site.
- Letsencrypt used to generate certificate based on above resources, triggered by a `make` command
- New make command `nexus-cert-install` deploys required azure resources
- New make command `nexus-letsencrypt` runs letsencrypt script and stored cert to base TRE keyvault (kv-<tre_id>)

## Future Changes

The certificate produced from the changes in this PR need testing with #1479 to confirm whether it successfully addresses #1089.  

Once this is confirmed to be working in its current state there may be a few possible optimizations:
 - Make the nexus-cert shared service more generic/parameterized so it can be used to create additional certificates for other components
 - Additional support for a wildcard cert if a custom domain is used
 - Letsencrypt extracted out of templates/core/terraform/scripts/ and out of the new nexus-cert shared service to reduce repeated code.


templates/core/terraform/scripts/letsencrypt.sh
```diff
@@ -3,6 +3,8 @@ set -e
 
 script_dir=$(realpath $(dirname "${BASH_SOURCE[0]}"))
 
+echo $FQDN
```
ross-p-smith, line 6: Is this debug?

oliver7598: Yes it is there should have been no changes to this file, Will remove now

templates/shared_services/nexus-cert/terraform/certificate.tf
```diff
@@ -0,0 +1,49 @@
...
+      exportable = true
+      key_type   = "RSA"
+      reuse_key  = false
+    }
+    secret_properties {
+      content_type = "application/x-pkcs12"
+    }
+  }
+
+  # The certificate will get replaced with a real one, so we don't want Terrafomr to try and revert it.
```
stuartleeks, line 40: nit: typo
```suggestion
  # The certificate will get replaced with a real one, so we don't want Terraform to try and revert it.
```

templates/shared_services/nexus-cert/scripts/letsencrypt.sh
```diff
@@ -0,0 +1,92 @@
...
+        --only-show-errors
+
+    # Wait a bit for the App Gateway health probe to notice
+    echo "Waiting 30s for health probe"
+    sleep 30s
+else
+    echo "index.html already present"
+fi
+
+ledir=$(cd ${script_dir} && cd ../ && pwd)/letsencrypt
```
stuartleeks, line 44: nit: could this be simplified?

```suggestion
ledir="${script_dir}/../letsencrypt"
```

templates/shared_services/nexus-cert/terraform/staticweb.tf
```diff
@@ -0,0 +1,30 @@
+data "azurerm_client_config" "deployer" {}
+
+resource "azurerm_storage_account" "staticweb" {
```
stuartleeks, line 3: Do we have anything in the repo that explains why we are deploying the storage static site, and how it is used?

templates/shared_services/nexus-cert/terraform/certificate.tf
oliver7598: Will correct but typo probably exists in terraform/core/appgateway also

templates/shared_services/nexus-cert/terraform/staticweb.tf
oliver7598: I've not added anything into the documentation about how it is used for this use case.  Although its very similar to how the cert is acquired for the API using letsencrypt.  Looking at the docs, the only reference I can find to any of this is references to letsencrypt in `deploying-azure-tre.md`.

If this is required to be documented for both letsencrypt use cases, I can raise an additional issue for this to prevent any delays in resolving #1089 by doing the documentation as part of this PR.

stuartleeks: Good point, I think that there are two potential cases of documentation here:

1. User docs. Is there anything that the user needs to do to trigger the lets-encrypt cert process for nexus? If so, that should be documented in user-facing docs
2. Developer/code docs. I think that the process we're using here is quite clever. But "clever" doesn't mean that it's obvious, so I was suggesting that some docs/notes are added to the code

In my initial comment, I was referring to the code docs, which I think would be good to include as part of the code in this PR so that future developers working on this code can quickly get an overview of how the letsencrypt process works and what the various components are used for. (potentially including a diagram similar to [this](https://github.com/microsoft/AzureTRE/issues/1089#issuecomment-1063928777)?)

oliver7598: I'll get something together relating to the letsencrypt and how it sits together/is used this afternoon for both the additions in this PR and the existing structure for the other Letsencrypt use case.

As for the user docs I imagine 1/2 would be related to changes in this PR, 1/2 to the changes in #1479.  

Ideally I imagine a single `make` command would be ideal to deploy the resources required to get the cert, generate the cert and deploying nexus (replacing the existing `make nexus-install`? which is currently run as part of `make deploy-shared-services`, with each part still being able to be run independently.

docs/tre-developers/letsencrypt.md
```diff
@@ -0,0 +1,52 @@
...
+                F[Web App]
+            end
+            G[Private DNS Zone < web-app-name >.< location >.cloudapp.net]
+        end
+    end
+
+    A --> |1. Request to            | B
+    B --> |2. Attempts to hit       | C
+    C --> |3. App Gateway routes    | D
+    D --> |4. Respondes             | C
```
stuartleeks, line 46:
```suggestion
    D --> |4. Responds             | C
```

docs/tre-developers/letsencrypt.md
```diff
@@ -0,0 +1,52 @@
...
+    C --> |5. Respondes             | B
```
stuartleeks, line 47:
```suggestion
    C --> |5. Responds             | B
```

marrobi (commented):
Couple of questions:
1. Looks like this deploys its own app gateway - is a cost associated with this - does it need its own app gw? Can the public IP be associated with the existing app gateway maybe?
2. How do renewals work?

I like the idea of a lets encrypt shared service to handle requesting and renewing certs.

daltskin:
I've create a story to track management of the letsencrypt certs #1540

Makefile
```diff
@@ -143,9 +143,23 @@ gitea-install:
...
+
+nexus-letsencrypt:
+	$(call target_title, "Requesting LetsEncrypt SSL certificate for Nexus") \
+	&& . ./devops/scripts/check_dependencies.sh nodocker,certbot \
+	&& . ./devops/scripts/load_env.sh ./templates/core/.env \
+	&& . ./devops/scripts/load_env.sh ./devops/.env \
+	&& . ./devops/scripts/load_terraform_env.sh ./devops/.env \
+	&& . ./devops/scripts/load_terraform_env.sh ./templates/core/.env \
+	&& pushd ./templates/shared_services/nexus-cert/scripts/ > /dev/null && . ./outputs.sh && popd > /dev/null \
+	&& . ./devops/scripts/load_env.sh ./templates/shared_services/nexus-cert/private.env \
```
ross-p-smith, line 160: nit: This may be better as just `.env` as we are leaking a bit of a hack that I had to do for the debugging.

templates/shared_services/nexus-cert/terraform/appgateway.tf
```diff
@@ -0,0 +1,171 @@
+resource "azurerm_public_ip" "appgwpip" {
+  name                = "pip-nexus-${var.tre_id}"
+  resource_group_name = local.core_resource_group_name
+  location            = var.location
+  allocation_method   = "Static"
+  sku                 = "Standard"
+  domain_name_label   = "nexus-${var.tre_id}"
```
ross-p-smith, line 7: If you made the whole domain_name_label a var, you are nearly 90% complete on making this re-usuable

templates/shared_services/nexus-cert/terraform/appgateway.tf
```diff
@@ -0,0 +1,171 @@
...
+
+resource "azurerm_user_assigned_identity" "agw_id" {
+  resource_group_name = local.core_resource_group_name
+  location            = var.location
+  name                = "id-agw-nexuscert-${var.tre_id}"
+
+  lifecycle { ignore_changes = [tags] }
+}
+
+resource "azurerm_application_gateway" "agw" {
```
ross-p-smith, line 20: Are we creating a new AppGateway?

templates/shared_services/nexus-cert/terraform/appgateway.tf
oliver7598: As is, its a 1-1 relationship, in the future this could achieve the fully dynamic shared service 
that uses a  dynamic tf file to achieve the goal. But as an is this is needed  
Asia k was focusing on getting the Nexus/Docker solution configured for the current needs and ammended dependant on future needs while the wider issues is addressed

templates/shared_services/nexus-cert/terraform/appgateway.tf
oliver7598: Currently, yes. But I imagine a single shared gateway may be adaquate.  But I will test this fully on Monday as I ran out of time today.

I had designed it as a separate component for now specifically for letsencrypt but if integrated I`m happy to ammended.

oliver7598: @ross-p-smith After a bit of digging I don't think we'd be able to amend the existing app gateway on deployment of this and a new one would be required.  It wouldn't be needed outside of acquiring the certificate however so could be stopped/started when required which would avoid the additional cost?

oliver7598: I have also tested the uses of a container instance although this is resolving to the wrong FQDN.  The IP seems to be generated by the CI so ends in azurecontainer.io.  This would maybe work if Nexus was also hosted in a CI although would then need to look at writing the auth hook to the container rather than a storage account blob.

jjgriff93:
Closing as addressed in #1584

PR #1515, opened 2022-03-15
Remove build trigger on label

stuartleeks:
As part of the clean-up for #478, remove the previous, label-triggered PR builds


.github/workflows/deploy_tre.yml
```diff
@@ -50,100 +46,3 @@ jobs:
       TF_STATE_CONTAINER: ${{ secrets.TF_STATE_CONTAINER }}
       TRE_ADDRESS_SPACE: ${{ secrets.TRE_ADDRESS_SPACE }}
       TRE_ID: ${{ secrets.TRE_ID }}
-
-  prepare-not-main:
```
tamirkamara, line 54: Deleting this means there's no way to deploy without a pr and also that dispatch is only targeting the main branch. Is that what you want?

stuartleeks: Ah, good point. I forgot this workflow is also sometimes used to build a branch without a PR.

I was hoping that this change would allow us to remove this and simplify the workflow graph:

image

Do you have a sense for what the scenarios are where people would want to trigger a build for a non-`main` branch where creating a PR or draft PR wouldn't make sense?

tamirkamara: I have used it many times to have the ci test something for me prior to opening the pr. But that was for branches in this repo and not Forks (which is how we should do things from now).

stuartleeks: Ok, since the code I removed would only be invoked via `workflow_dispatch`, I'll add a new workflow that contains the removed steps. As you mention, this is unlikely to be used in the normal flow (as we'll be working in forks and so secrets wouldn't be passed to the workflow), but having it will allow us to push a branch to the main repo and test workflow changes there before merging to `main`.

stuartleeks: @tamirkamara  - I've added `deploy_tre_branch.yml` with the non-`main` build flow. Let me know if that will handle your requirements. Thanks!

stuartleeks:
/test-force-approve

Skipping tests as the only relevant test would be `deploy_tre_branch.yml` but that can't be triggered until it has been pushed to `main` at least once

Issue #1243, opened 2022-02-03
Update Resource in Azure following API Update


damoodamoo:
Feature #23 

When an update is made via the API (#1240), it needs to be reflected in Azure. The new change needs to be sent via the service bus to the runner, which will run an update operation via the porter bundle.

For instance: A user wants to resize a VM they have created as a `user-resource`. They submit a `PATCH` via the API, and the new properties are fed to the runner, which will resize the machine. 

Acceptance:
- [x] Submit the new properties to the runner
- [x] If output properties are returned from the runner, merge with props
- [x] Demonstrate updating of a single resource
- [x] E2E test confirming successful operation

PR #1517, opened 2022-03-15
Resource Updates

damoodamoo:
# PR for issue #1243 

## What it does
Feeds the newly `PATCH`ed properties through to the runner, along with the `upgrade` command. A large majority of the existing deployment / status update code remains as-is, with just a few tweaks to better handle upgrades. The `PATCH` endpoints now return an operation doc + location header which can be used to track the upgrade.


Makefile
```diff
@@ -284,14 +284,18 @@ static-web-upload:
 	&& ./templates/core/terraform/scripts/upload_static_web.sh
 
 test-e2e-smoke:
-	$(call target_title, "Running E2E smoke tests") && \
-	cd e2e_tests && \
-	python -m pytest -m smoke --verify $${IS_API_SECURED:-true} --junit-xml pytest_e2e_smoke.xml
+	$(call target_title, "Running E2E smoke tests") \
+	&& cp --force ./templates/core/.env ./e2e_tests/.env \
+	&& . ./devops/scripts/load_env.sh ./templates/core/private.env \
```
ross-p-smith, line 289: private.env will not be there when run in the GitHub Build

damoodamoo:
bad build + deletion stuck, re-pr-ing

PR #1518, opened 2022-03-15
Debug the Smoke Tests

ross-p-smith:
Add an easy way for developers to be able to debug the E2E tests. Currently we have `make setup-local-debugging` which works, but everytime your redeploy your infrastructure, the firewall settings are replaced. This now does this at a Terraform level, so is persisted. It is triggered by the following setting in /workspaces/AzureTRE/templates/core/.env
```
ENABLE_LOCAL_DEBUGGING=true
```

Also added a vscode launch configuration to be able to debug the smoke tests

### DIFF
--- a/templates/core/terraform/variables.tf
+++ b/templates/core/terraform/variables.tf
@@ -135,11 +135,16 @@ variable "stateful_resources_locked" {
   description = "Used to add locks on resources with state"
 }
 
-
 variable "ci_git_ref" {
   default     = ""
   description = "The git ref used by the ci to deploy this TRE"
   type        = string
+}
+
+variable "enable_local_debugging" {
+  default     = false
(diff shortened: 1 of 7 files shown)


ross-p-smith:
\test

templates/core/terraform/variables.tf
```diff
@@ -135,13 +135,18 @@ variable "stateful_resources_locked" {
...
-
 variable "ci_git_ref" {
   default     = ""
   description = "The git ref used by the ci to deploy this TRE"
   type        = string
 }
 
+variable "enable_local_debugging" {
+  default     = false
+  description = "This will allow Cosmos to be accesible from your local IP address and add some extra role permissions."
```
martinpeck, line 146: nit: `accesible` should be `accessible`. If you're fixing this, suggest change `Cosmos` to `CosmosDB`

PR #1519, opened 2022-03-15
Include TRE_ID in debugging SP name

stuartleeks:
If multiple developers are using the same AAD tenant then it looks like the first developer is able to run `make setup-local-debugging` but other developers using that tenant will get an error similar to the error below:

```
WARNING: Found an existing application instance of "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxx". We will patch it
ERROR: Insufficient privileges to complete the operation.
```

Adding the `TRE_ID` into the SP name will ensure uniqueness and allow multiple developers to run in a shared AAD tenant.

### DIFF
--- a/scripts/setup_local_debugging.sh
+++ b/scripts/setup_local_debugging.sh
@@ -45,7 +45,7 @@ az role assignment create \
 
 # Configure SP for local resource processor debugging (Porter can't use local creds)
 echo "Configuring Service Principal for Resource Processor debugging..."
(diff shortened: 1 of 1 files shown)


damoodamoo:
hiya - i mentioned this in the chat - i have this covered in my PR already - #1517

stuartleeks:
@damoodamoo - yeah, @ross-p-smith  mentioned this, but I wanted to pull it out into a small PR that can be pushed through quickly 😄

damoodamoo:
i'll unwind my change then

ross-p-smith (approved):

PR #1524, opened 2022-03-15
Fix for Workspace AAD

ross-p-smith:
When registering the AAD for your workspace, it does not let you assign Users to the "Workspace Owner" "Workspace Researcher" app roles.

## How is this addressed

- There was some missing "


stuartleeks (approved):
LGTM

PR #1541, opened 2022-03-18
Test for access_token and output response text if absent

stuartleeks:
Rather than a KeyError when auth fails, check for the access_token key
and output the response (with error details) if not found


stuartleeks:
/test-force-approve

Previous run validated that the change hadn't broken the success path

daltskin (approved):

Issue #1538, opened 2022-03-18
Tidy-up/polish PR comment bot implementation


stuartleeks:
Following on from the implementation work for #478, there are some aspects of the implementation that could be tidied up and polished. This includes:

- [x] Move GitHub script code to external js file for better editing experience in VSCode (#1616 )
- [x] Add `/test-extended` command to also run the extended tests (#1576 )
- [x] Add tests for the GitHub script code (#1660)
- [x] When running `/test-destroy-env`, run the destroy twice: once generating the environment ID for the PR and once for the branch. This way, we can clean up  both environments. (note: the destroy needs to gracefully handle the environment not existing) (#1663 )
- [x] Consider splitting the cleanup step to a separate job. This would be clearer when looking at the job run history, as well as limiting the scope of the workflow requiring credentials (implemented in #1663 )
- [x] Change command test logic: Trim first line and check for leading `/`. If found, assume a command is intended and continue to match. If not match then output help message `'/foo' not recognised. Available commands are ...` (implemented in #1727)
- [x] Add a comment when a user without permissions attempts to run commands (implemented in #1727)
- [x] Show who left the comment with the command that the bot is responding to in any comments the bot leaves in response (implemented in #1727)
- [x] Have the bot add a comment in response to `/test-force-approve` commands indicating the SHA being approved and the original commenter (implemented in #1727)
- [x] Include the resource group name (or at least the pr ref SHA) in bot comments for test runs to make it easier when troubleshooting runs (implemented in #1727)
- [ ] require head SHA on external PRs for `/test` (#1731)
- [ ] Ignore `/test` etc on closed PRs? (Possibly still respond to `/test-destroy-env`?). (#1731)<br>Update: running `/test` on closed PRs fails as there's no `potentialMergeCommit` generated by GitHub. <br>Update 2: running `/test` on open PRs with merge conflicts fails as there's no `potentialMergeCommit` generated

Other aspects that could be considered for tidy-up:
- Update 3rd party actions to use specific (validated) SHAs
- addressing linter warnings for the workflow YAML files
- ~~In `deploy_tre_reusable.yml`, move non-sensitive fields (e.g. `TRE_ID` to `input` rather than `secret`) to improve log output by removing unnecessary masking (Initially attempted in #1544 but later reverted)~~

PR #1544, opened 2022-03-18
Move non-sensitive values from secrets to inputs

stuartleeks:
Improve log output for PR builds by moving non-sensitive values to inputs
(Note: any value still derived from repo secrets will remain masked e.g. for builds on main)

part of #1538

### DIFF
--- a/.github/workflows/deploy_tre.yml
+++ b/.github/workflows/deploy_tre.yml
@@ -22,13 +22,15 @@ jobs:
     uses: ./.github/workflows/deploy_tre_reusable.yml
     with:
       ciGitRef: ${{ github.ref }}
+      ACR_NAME: ${{ secrets.ACR_NAME }}
+      ACTIONS_DEVCONTAINER_TAG: 'latest'
+      MGMT_RESOURCE_GROUP: ${{ secrets.MGMT_RESOURCE_GROUP }}
+      TRE_ID: ${{ secrets.TRE_ID }}
     secrets:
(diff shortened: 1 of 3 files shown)


stuartleeks:
Previous successful run via `deploy_tre_branch.yml`

.github/workflows/deploy_tre.yml
```diff
@@ -22,13 +22,15 @@ jobs:
     uses: ./.github/workflows/deploy_tre_reusable.yml
     with:
       ciGitRef: ${{ github.ref }}
+      ACR_NAME: ${{ secrets.ACR_NAME }}
```
ross-p-smith, line 25: This is a globally unique acr - do we really want that displayed in our logs? I toyed with this in the past, but I think it's better that the public dont see it.

ross-p-smith: So if it's still masked, what benefit does this change have?

stuartleeks: For builds on `main`, none. But the `deploy_tre_reusable.yml` is also used from the PR builds, so making it an input there reduces masking on PR builds (where `ACR_NAME` points to the specific instance for the PR deployment)

ross-p-smith (approved):

stuartleeks:
Redeployed using `deploy_tre_branch.yml` after resolving merge conflicts: https://github.com/microsoft/AzureTRE/runs/5624442302?check_suite_focus=true

PR #1548, opened 2022-03-20
Fix Resource Processor logging

### LINKED ISSUE #1547: Logs are missing in Resource Processor
**Describe the bug**

The resource processor application doesn't output some of the logs. Especially, when a deployment was successful.

**Steps to reproduce**

1. Ask the API to deploy a workspace 
1. Wait for it to complete successfully
1. Review the logs in the RP or AppInsights, see that entries after the "stdout" one don't really include logs of the deployment.

tamirkamara:
Fixes #1547 

## What is being addressed

Resource processor should log the right output (stdout vs stderr).

## How is this addressed

- Fix the shell logging to output stdout on success
- Refactor our logging config/setup to better cope with local debugging and Application Insights
- Use a message scoped logger when one is already available
- Given the amount of testing done manually, upgrade all packages used to their latest version
- Increase the frequency of the RP pruning job as once a week isn't enough


eladiw (approved):

templates/core/terraform/resource_processor/vmss_porter/cloud-config.yaml
```diff
@@ -39,12 +39,18 @@ write_files:
...
+  - path: /etc/cron.hourly/docker-prune
+    # An hourly cron job to have docker free disk space. Running this frquently
+    # since disk might get full fast, but we prune only when free space is low.
     content: |
       #!/bin/bash
       set -o errexit
-      docker system prune -f
+      used_percent=$(df / --output=pcent | tail -1 | sed 's/[^0-9]//g')
+      echo "Used disk space percent: $${used_percent}"
+      if (( used_percent > 60 )); then
```
eladiw, line 50: 60 seems a bit low, why not 80 or so?

tamirkamara: The machine comes online with a volume of 29GB, 20% free.
80% means less than 6GB free.
Since we have images 1.5 GB or even 3+ (Guacamoles VMs) I thought it will be better to have 60% set for pruning.
If we continue to see space errors we'll increase the disk since (and the threshold too).

resource_processor/shared/logging.py
```diff
@@ -96,9 +106,13 @@ def shell_output_logger(console_output: str, prefix_item: str, logger: logging.L
     """
     logger.log(logging_level, prefix_item)
 
-    if console_output is None:
+    if (console_output is None or len(console_output) == 0):
```
martinpeck, line 109: can this be collapsed to just...

```
if console_output:
```

given that both None and "" are falsey ?

tamirkamara: done

Issue #1526, opened 2022-03-16
Storage share fails to delete from base workspace


tamirkamara:
**Describe the bug**
When trying to delete a (base) workspace instance, the action might fail due to a problem removing the storage share:

shares.Client#Delete: Failure sending request: StatusCode=0 -- Original Error: Delete "https://stgws84a2.file.core.windows.net/vm-shared-storage?restype=share": EOF

> True deleting_failed aad1d169-e124-43ed-859a-07b0cdbb84a2: Error context message = ╷ │ Error: deleting File Share "vm-shared-storage" (Storage Account "stgws84a2" / Resource Group "rg-***-ws-84a2"): shares.Client#Delete: Failure sending request: StatusCode=0 -- Original Error: Delete "https://stgws84a2.file.core.windows.net/vm-shared-storage?restype=share": EOF │  │  ╵ Error: error running command /cnab/app/terraform terraform destroy -auto-approve -input=false -var address_space=10.0.68.0/24 -var location=*** -var shared_storage_quota=50 -var tre_id=*** -var tre_resource_id=aad1d169-e124-43ed-859a-07b0cdbb84a2: exit status 1 Error: mixin execution failed: exit status 1 Error: 1 error occurred: 	* container exit code: 1, message: <nil>   az login --identity -u 4c1a5d14-72a0-433f-b96b-4d96d3a6e6df && az acr login --name *** && porter uninstall "aad1d169-e124-43ed-859a-07b0cdbb84a2"  --reference ***.azurecr.io/tre-workspace-base:v0.1.10  --param address_space="10.0.68.0/24" --param arm_use_msi="true" --param azure_location="***" --param id="aad1d169-e124-43ed-859a-07b0cdbb84a2" --param tfstate_container_name="tfstate" --param tfstate_resource_group_name="***mgmt" --param tfstate_storage_account_name="***state" --param tre_id="***" --cred ./vmss_porter/azure.json --allow-docker-host-access --force && porter show aad1d169-e124-43ed-859a-07b0cdbb84a2

**Steps to reproduce**

1. More likely to happen after we merge to main, but not every time

PR #1550, opened 2022-03-20
Base workspace storage share issue - updates/debug

tamirkamara:
## What is being addressed

Deleting a base workspace sometimes fails due to an error on the storage share as described on #1526.
This PR is a step in trying to pinpoint the problem.

## How is this addressed

- Turn up TF logs on destroy. Due to their size they will be accessible from the Resource Processor machine.
- Upgrade all TF components - core and provider
- Add a param that will allow the base workspace to be deployed from a local machine (=not inside the vnet) to allow easier debugging
- Add missing parameters to the parameters.json file


templates/workspaces/base/terraform/variables.tf
```diff
@@ -35,3 +35,9 @@ variable "app_service_plan_sku" {
   default     = "P1v3"
   description = "App Service Plan SKU"
 }
+
+variable "storage_is_exposed_externally" {
+  type        = bool
+  default     = false
```
ross-p-smith, line 41: Could you wire this up to "enable_local_debugging" and have one place where you turn this on?

ross-p-smith (approved):

eladiw (approved):

PR #1558, opened 2022-03-21
Rename automation account

LizaShak:
# PR for issue

## What is being addressed

Rename automation account

## How is this addressed

- Rename automation account var name


tamirkamara (changes requested):
Can we not say Automation Account? That has a special context on Azure...
How about just TEST_ACCOUNT_CLIENT_ID?

ross-p-smith (approved):

tamirkamara (approved):

Issue #1559, opened 2022-03-21
MLflow - Shared storage container missing


vjmanda:
**MLflow - Shared storage container missing**
MLflow errors on the host machine when it tries to track artefacts. It requires a shared storage container called `vm-shared-storage` to be created.

![image](https://user-images.githubusercontent.com/44143177/159260176-0f8ccd5f-69eb-4e60-8e3d-ea69f059480c.png)

PR #1563, opened 2022-03-21
Add Missing Shared Storage Container

vjmanda:
# PR for issue #1559 

## What is being addressed

MLflow hosts need a storage container on the workspace storage account. 

## How is this addressed

Added storage container `mlartefacts` to MLflow terraform deployment.


templates/workspaces/base/terraform/storage.tf
```diff
@@ -18,6 +18,12 @@ resource "azurerm_storage_share" "shared_storage" {
   ]
 }
 
+resource "azurerm_storage_container" "shared_storage" {
```
marrobi, line 21: Can this container be created inside the ML flow workspace service, and named with an ML flow specific name? We should avoid adding things to the base workspace unless they are actually needed by multiple services and/or by all deployments.

PR #1564, opened 2022-03-21
Fix secret values on main build without triggering masking on PR builds

stuartleeks:
Enable passing secret values from main build, but allow PR builds to pass values as inputs so that non-secret values aren't turned it to secrets and masked in the output (making debugging PR build failures harder)

### DIFF
--- a/.github/workflows/deploy_tre.yml
+++ b/.github/workflows/deploy_tre.yml
@@ -22,15 +22,13 @@ jobs:
     uses: ./.github/workflows/deploy_tre_reusable.yml
     with:
       ciGitRef: ${{ github.ref }}
-      ACR_NAME: ${{ secrets.ACR_NAME }}
-      ACTIONS_DEVCONTAINER_TAG: 'latest'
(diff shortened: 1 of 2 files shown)


damoodamoo (approved):

tamirkamara:
@stuartleeks 
This makes the workflows even more complex. Is it really that important we don't mask the ACR for example? Maybe we should just let it be?

stuartleeks:
@tamirkamara  - yeah, it's frustrating that there isn't a simpler way to achieve this. However, since the `ACR_NAME` has the same value as the `TRE_ID` for PR builds, treating `ACR_NAME` has a secret means that we mask the TRE_ID.

I'm planning to have another look at this tomorrow and see if there's a decent way forwards or whether to revert to it all being secrets again.

stuartleeks:
It turns out that this gets complicated quickly as we don't seem to be able to reference secrets when passing inputs to `deploy_tre_reusable.yml`

Reverted in #1568

PR #1565, opened 2022-03-21
Use JSON output for az commands

stuartleeks:
Script was failing if `az` default output isn't JSON. Setting `AZURE_CORE_OUTPUT` at the top of the script to force the output of all `az` commands in the script to be in JSON format (without affecting the user's default)


tamirkamara (approved):

stuartleeks:
/test-force-approve

(I don't believe `aad-app-reg.sh` is exercised anywhere in CI)

PR #1568, opened 2022-03-22
Revert to masked input

stuartleeks:
We have a challenge: we want to keep masking for the ACR_NAME etc in main builds,
but we want to remove the masking for PR builds to improve debugging.

GH workflows don't seem allow us to reference secrets when
passing inputs via workflow_call, so this becomes tricky to manage as
the reusable workflow needs masked values in one case and
unmasked in the other


stuartleeks:
Successful build with `deploy_tre_branch`: https://github.com/microsoft/AzureTRE/actions/runs/2020623071

tamirkamara (approved):

stuartleeks:
(this reverted the changes for #1544 )

PR #1570, opened 2022-03-22
Use Local IP Address

ross-p-smith:
## What is being addressed

When building TRE from behind a network where ipecho.net/plain is blocked. Allow the setup to allow your local IP address to be injected in.

## How is this addressed

- Set the PUBLIC_DEPLOYMENT_IP_ADDRESS to your  local IP address in the .env file to circumvent the dynamic nature of getting the IP address from a web page


marrobi:
Local IP to me means LAN IP. Maybe we should call it DEBUG_PUBLIC_IP or similar?

ross-p-smith:
> 

Bit nervous about the word DEBUG as it's not for debugging, this is the machine that is deploying TRE. There is a wider question that we are deploying a Trusted Research Environment, but we are dependent on a 3rd party URL to determine what IP address to grant access to some of the services. Thoughts @marrobi ?

marrobi:
Agree, the external URL dependency isn't ideal. How about `CLIENT_PUBLIC_IP`?

ross-p-smith:
> Agree, the external URL dependency isn't ideal. How about `CLIENT_PUBLIC_IP`?

Sorry - must have missed this comment whilst I was changing the name - do you prefer CLIENT_PUBLIC_IP or the one I just changed it to DEPLOYMENT_IP_ADDRESS?

marrobi:
I think important to say PUBLIC. CLIENT or DEPLOYMENT is fine.

templates/core/.env.sample
```diff
@@ -20,7 +20,10 @@ SWAGGER_UI_CLIENT_ID=__CHANGE_ME__
 # TEST_ACCOUNT_CLIENT_SECRET=__CHANGE_ME__
 # TRE_URL=https://${TRE_ID}.westeurope.cloudapp.azure.com
 
-
+# If you are behind a firewall and you cannot browse
+# to https://ipecho.net/plain as its blocked then set
+# your local IP address here
```
martinpeck, line 25: Should we be more explicit that "this is the IP address of the machine performing the deployment. e.g. your build agent, or your local machine"?

martinpeck (approved):

templates/core/.env.sample
martinpeck: i.e. spell it out why you might want to do this, and what this IP address represents

daltskin:
Can we add this to the docs too: https://microsoft.github.io/AzureTRE/tre-admins/environment-variables/#for-azure-tre-instance-in-templatescoreenv

Issue #701, opened 2021-08-23
[Story] Add application health monitoring to VMSS resource processor


marrobi:
## Description

As an Azure Administrator
I want to have health monitoring enabled of the VMSS
So that operating system updates are carried out automatically

https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-automatic-upgrade#requirements-for-configuring-automatic-os-image-upgrade

https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-health-extension

## Acceptance criteria

- [ ] Add internal HTTP health endpoint to runner.py to verify it is running
- [ ] Add health extension to VMSS ( https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-health-extension
 )
- [ ] Enable automatic OS updates and verify status in the portal ( https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-automatic-upgrade#configure-automatic-os-image-upgrade )

PR #1572, opened 2022-03-23
Automatic OS upgrades to VMSS resource processor

anatbal:
# PR for issue #701 

## What is being addressed
Add automatic OS upgrades to VMSS resource processor


eladiw (approved):

marrobi:
Looks good. We should also look at maintenance windows, and gracefully shutting down - as what happens if a TRE resource deployment is in progress when the machine is updated.

Lower priority, but can you create issues for future as you feel might be appropriate, I expect a combination of the below:

https://docs.microsoft.com/en-us/azure/virtual-machines/virtual-machine-scale-sets-maintenance-control
https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-terminate-notification

Issue #1573, opened 2022-03-23
"State Store endpoint is not responding" in PR build


stuartleeks:
This build run failed: https://github.com/microsoft/AzureTRE/runs/5639936288?check_suite_focus=true#step:3:346

```
State Store endpoint is not responding

make: *** [Makefile:274: bundle-register] Error 1
Warning: Template API check for workspace_service tre-service-guacamole returned http status: 404
Error: Process completed with exit code 2.
```

Re-running succeeded, so it may be a timing/dependency issue.

stuartleeks:
The theory is that the error in the description was because the API wasn't up and running. #1574 does nothing to address the cause of the API not running, but adds a check earlier in the process to make it easier to detect. 

Since this will change the error, we could close this and open an issue if we encounter problems with the API not being ready?

tamirkamara:
Closing as this hasn't been seen for a while.

PR #1574, opened 2022-03-23
Add API healthcheck test

stuartleeks:
- Add `make api-healthcheck`
- incorporate healthcheck in CI after TRE Core deployment

Following #1573, add a healthcheck to the build to catch deployments with unhealthy API earlier in the build

### DIFF
--- a/.github/workflows/deploy_tre_reusable.yml
+++ b/.github/workflows/deploy_tre_reusable.yml
@@ -223,7 +223,7 @@ jobs:
       - name: Deploy TRE Core
         uses: ./.github/actions/devcontainer_run_command
         with:
-          COMMAND: "TF_VAR_ci_git_ref=${{ inputs.ciGitRef }} TF_LOG=${{ secrets.TF_LOG }} make deploy-core"
+          COMMAND: "TF_VAR_ci_git_ref=${{ inputs.ciGitRef }} TF_LOG=${{ secrets.TF_LOG }} make deploy-core && make api-healthcheck"
           ACTIONS_ACR_NAME: ${{ secrets.ACTIONS_ACR_NAME }}
           ACTIONS_ACR_URI: ${{ secrets.ACTIONS_ACR_URI }}
           ACTIONS_ACR_PASSWORD: ${{ secrets.ACTIONS_ACR_PASSWORD }}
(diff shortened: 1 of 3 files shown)


stuartleeks:
Successful run with `deploy_tre_branch.yml`: https://github.com/microsoft/AzureTRE/runs/5659256061?check_suite_focus=true#step:3:1518

marrobi:
Do we need to wait for things to get to a healthy stage? Or is expected all should be healthy when the deployment finished? @anatbal is just about to add the resource processor health to the health endpoint, and cloud init takes a while to finish, but VMSS deployment will have completed.

.github/workflows/deploy_tre_reusable.yml
```diff
@@ -218,7 +218,7 @@ jobs:
       - name: Deploy TRE Core
         uses: ./.github/actions/devcontainer_run_command
         with:
-          COMMAND: "TF_VAR_ci_git_ref=${{ inputs.ciGitRef }} TF_LOG=${{ secrets.TF_LOG }} make deploy-core"
+          COMMAND: "TF_VAR_ci_git_ref=${{ inputs.ciGitRef }} TF_LOG=${{ secrets.TF_LOG }} make deploy-core && make api-healthcheck"
```
tamirkamara, line 221: What about splitting the extra make command into a separate step?

stuartleeks:
At the point where the TRE Core deployment step completes, I think that the resources should be deployed and ready. Otherwise, the project's CI process needs to include a wait, anyone else implementing their own CI to deploy AzureTRE also needs to add that step, and so do people creating a dev environment.

.github/workflows/deploy_tre_reusable.yml
stuartleeks: Yeah, I was on the fence about whether to split or not. Happy to split it out!

stuartleeks: Done

tamirkamara (approved):

stuartleeks:
/test-force-approve

Skipping pr-bot tests as it won't load new workflow. Have tested via `deploy_tre_branch` as per previous comment.

PR #1575, opened 2022-03-23
Fix cleanup bug that deleted main env

tamirkamara:
## What is being addressed

The cleanup script catch the "mgmt" resource group and deleted that despite not supposed to.

## How is this addressed

- Include the "ws" string when searching for workspace resource group that should be deleted.
- Reduce the time we wait before stopping an environment.


stuartleeks (approved):
lgtm

tamirkamara:
/test-force-approve
This workflow/script isn't relevant for running tests.

PR #1579, opened 2022-03-23
Check forks & native branches activity on clean up

tamirkamara:
## What is being addressed

The cleanup scripts doesn't check activity on forks or native branches, so it doesn't _stop_ their environments.

## How is this addressed

- Add the PR's _updatedAt_ date in case we don't find a commit (happens when the ref is from a fork)
- Add activity check to native branches as well


stuartleeks (approved):
LGTM. 

(I thought that there was a potential alternative approach with the `potentialMergeCommit`/`mergeCommit` on PRs, but some PRs seem to have neither set!)

tamirkamara:
/test-force-approve
tests don't cover this change

stuartleeks:
/test-force-approve

as per https://github.com/microsoft/AzureTRE/pull/1579#issuecomment-1076545871

PR #1582, opened 2022-03-23
Sync guacamole tags

### LINKED ISSUE #848: Guacamole does not get an image tag from resource processor
**Describe the bug**
When creating a new guacamole service, the tag is hard coded in the porter.yaml rather than dynamically transferred from the resource processor.

### LINKED ISSUE #1581: Guacamole app fails to start due to wrong image tag
**Describe the bug**
Guacamole image tag is out of sync and causing the app to fail, and also the E2E.

**Steps to reproduce**

1. Create a workspace
1. Create a Guacamole workspace service
1. Browse to the app url, get 500 / internal server error
1. View the app service logs and see that it doesn't find the image

tamirkamara:
Fixes #1581, Fixes #848 

## What is being addressed

The Guacamole workspace service bundle referred to an image tag that didn't exist.

## How is this addressed

- Bundle image tag set to an empty string as a default which will instruct terraform to take the value from the version.txt file. If the bundle will get a value, that will be used for the image tag.
- Add folders to the bundle dockerignore file that aren't required for the bundle installer. This saves ~300 MB in its size.
- Upgrade Terraform version used in the bundle.
- E2E tests are using the logging lib (vs. the plain print)
- Turn live_cli logs in pytest so that we'll see what is going on while the tests are running rather than waiting for it to end. Also include the timestamp of each entry.

### DIFF
--- a/templates/workspaces/base/porter.yaml
+++ b/templates/workspaces/base/porter.yaml
@@ -82,7 +82,7 @@ uninstall:
   - terraform:
       description: "Tear down workspace"
       # https://github.com/microsoft/AzureTRE/issues/1526
-      logLevel: Debug
+      # logLevel: Debug
       vars:
         tre_id: "{{ bundle.parameters.tre_id }}"
         tre_resource_id: "{{ bundle.parameters.id }}"
--- a/e2e_tests/helpers.py
+++ b/e2e_tests/helpers.py
@@ -4,8 +4,12 @@ from contextlib import asynccontextmanager
 from httpx import AsyncClient
 from starlette import status
 
+import logging
+
 import config
 from resources import strings
+
+LOGGER = logging.getLogger(__name__)
 
 
 class InstallFailedException(Exception):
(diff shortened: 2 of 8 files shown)


templates/workspaces/base/porter.yaml
```diff
@@ -82,7 +82,7 @@ uninstall:
   - terraform:
       description: "Tear down workspace"
       # https://github.com/microsoft/AzureTRE/issues/1526
-      logLevel: Debug
+      # logLevel: Debug
```
stuartleeks, line 85: nit: is this intentionally left commented out?

stuartleeks (approved):
LGTM

templates/workspaces/base/porter.yaml
tamirkamara: Yes, I will enable it in my other pr that deals with resource processor logging

marrobi (approved):
LGTM

PR #1589, opened 2022-03-24
Aid debugging e2e test failures

stuartleeks:
To aid debugging of e2e test failures due to auth, output the response content when the response fails to be parsed as JSON


martinpeck (approved):

stuartleeks:
/test

Previous run hit `Status=400 Code="HashConflictOnDifferentRoleAssignmentIds" Message="A hash conflict was encountered for the role Assignment ID. Please use a new Guid."`

PR #1590, opened 2022-03-24
Fix "waiting for update of instances for Linux Virtual Machine Scale Set"

anatbal:
# PR for issue #1591 

## What is being addressed

The error that occurs is:
│ Error: waiting for update of instances for Linux Virtual Machine Scale Set "vmss-rp-porter-trejd1" (Resource Group "rg-trejd1"): Code="MaxUnhealthyUpgradedInstancePercentExceededInRollingUpgrade" Message="Rolling Upgrade failed after exceeding the MaxUnhealthyUpgradedInstancePercent value defined in the RollingUpgradePolicy. 100% of instances are in an unhealthy state after being upgraded - more than the threshold of 10%. The most impactful error is: Instance found to be unhealthy or unreachable.. First failed instance ID is: 0."

## How is this addressed

- Changing the threshold to MaxUnhealthyUpgradedInstancePercent from 10 to 100 precent.


daltskin:
Fix for #1591

daltskin (approved):

anatbal:
> @anatbal have we done a deployment, made the VMSS instance unhealthy and then check it upgrades?

I am trying to do one now

Issue #1556, opened 2022-03-21
Deploy shared services via API in build


tanya-borisova:
Currently, there is a temporary solution in place for deploying shared services.
In this task, change this to build and deploy shared services as Porter bundles

- [x] Register shared service templates via `register_bundle_via_api.sh` 
- [x] Build bundles for Firewall, Gitea and Nexus shared service as part of the build
- [x] Fixes in bundles for Firewall, Gitea and Nexus to make them deployable
- [x] Deploy shared services enabled by default as part of the deploy
- [x] Manual e2e test in an environment

Issue #1533, opened 2022-03-16
E2E test for shared service deployment


tanya-borisova:
Add an end-to-end test for a successful shared service deployment

damoodamoo:
i'm wondering if we actually need this as the shared service deployments will become part of the build, so will be implicitly tested... but to discuss

PR #1531, opened 2022-03-16
API for shared services

tanya-borisova:
# PR for issue

Closes issue #1239 

## What is being addressed

Add API for shared services: creation, deletion, update, list, and operations

## How is this addressed

- [x] Models and repository for shared services, that inherits from base resource.py repository
- [x] Tests for shared services repository
- [x] API endpoints for shared services and shared service operations - same as exist for workspace operations
- [x] Tests for API endpoints for shared services
- [x] Code reuse with workspace services (some methods are extracted into `api_app/routes/resource_helpers.py` -- I will leave comments where code has been moved 
- [x] Add some fixes for `firewall` shared service that were needed to achieve a successful deploy (mostly just remove `azure_location` parameter as it isn't necessary) 

## How is it tested

I tested that locally I can build a firewall bundle and then successfully deploy it via the API (screenshot is logs of resource processor). No code changes in resource processor were necessary

MicrosoftTeams-image (1)

Issue #1244, opened 2022-02-03
Update a secondary resource from a primary


damoodamoo:
Feature #23 

### Status: In Progress

- [x] #1677
- [x] #1678
- [x] #1736
- [x] #1679
- [x] #2069

## Requirements
- A resource should be able to update properties on another resource before or after its installation process 
- The steps need to happen in serial. All 'pre' steps must happen in order, before the deployment starts.
- The user should be able to have a view on the overall status of the operation, including the pre + post steps as part of the overall deployment
- The pre + post steps should be able to reference properties from the resource object

## Example
A user submits a `POST` request for a VM user resource. In the resource template, a `pipeline: {}` block is defined, and contains steps which describe the updating of a firewall. The API reads this array, and creates an `Operation` document that contains the steps for the pipeline. Each step is executed, using the existing porter runner method, one after the other, until they are all done and the API marks the `Operation` as complete.

## Sequence Diagram
The API layer centrally orchestrates the whole process. When a user submits a request for a resource which has a `pipeline: []` array within the template, the API will enqueue the first step, handing the `operation_id` and `step_id` to the Resource Processor. When the resource processor returns the status update of a deployment, it will reference the `operation_id` and a `step_id`. The API will get the `Operation` document, check if there are more steps to be carried out, and enqueue the next. When all steps are complete, the overall status will be updated. 

```mermaid
sequenceDiagram
    User->>API: POST user_resource {...}
    API->>API: Template has pipeline {}? 
    loop Steps
        API ->> API:Create operations document <br/>containing multiple steps 
    end
    API ->> Cosmos:Operations doc
    API ->> API:Substitute property values for <br>resource defined in step[0]
    API ->> Cosmos:Update resource defined in step[0]
    API ->> Resource Processor: Enqueue upgrade/install action
    API ->> User: Operations Doc
    Note left of User: {steps:[<br>{status:in_progress}, <br>{status:waiting}<br>]}
    Resource Processor --> User: 
    Resource Processor --> User: 
    Resource Processor --> User: 
    Resource Processor ->> Resource Processor: Run bundle <br>upgrade
    Resource Processor ->> API:Enqueue status update message
    API ->> Cosmos: Get Operation doc
    Cosmos ->> API: Operation doc
    loop While Operation has more Steps
        API ->> Cosmos:Get resource template
        Cosmos ->> API: Resource template
        API ->> API:Substitute property values for <br>resource defined in step[N]
[...]

PR #1597, opened 2022-03-25
Build and deploy shared services as part of make tre-deploy

tanya-borisova:
# PR for issue

Closes https://github.com/microsoft/AzureTRE/issues/1556 

## How is this addressed

- [x] Made changes to the script [register_bundle_with_api.sh](https://github.com/microsoft/AzureTRE/compare/tborisova/1556-build-shared-services-as-part-of-the-build?expand=1#diff-f1a4a282ce8199af220677f924428ac8e1c97e589452ebd383f9d1c9bf601e01) to register shared services and also optionally deploy shared services
- [x] Add `make` targets for building, registering and deploying shared services
- [x] Made changes to Porter bundles to make them deployable (e.g. template_schema.json was missing in some)
- [x] Pass location from resource processor to porter bundles as several bundles depend on it 
- [x] Manually test e2e in an environment


marrobi (commented):
Is the plan for the share service deployments to become part of the e2e tests? Feels like they should sit there given that is where we test the deployments of other resources.

tanya-borisova:
@marrobi we also have this task on adding shared services to E2E tests: https://github.com/microsoft/AzureTRE/issues/1533

Although @damoodamoo raised a point that this might not be necessary if they are already part of the build, as is done here.

I think they need to be in the build to preserve the behaviour (as in, shared services get deployed as part of `make tre-deploy`). I can also see the argument of adding some tests for consistency. What do you guys think is best?

devops/scripts/register_bundle_with_api.sh
```diff
@@ -14,24 +15,26 @@ function usage() {
 
     Options:
         -r, --acr-name                Azure Container Registry Name
-        -t, --bundle-type             Bundle type, workspace or workspace_service
-        -w, --workspace-service-name  The template name of the user resource
+        -t, --bundle-type             Bundle type: workspace, workspace_service, user_resource or shared_service
+        -w, --workspace-service-name  The template name of the user resource (if registering a user_resource), or shared service (if registering a shared_service)
```
ross-p-smith, line 19: I don't think you are using the --workspace-service-name for your shared services. Is the conditional on -d now and not this?

tanya-borisova: Oh yes, a good spot -- forgot to update this message

templates/shared_services/firewall/porter.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 name: tre-shared-service-firewall
-version: 0.0.5
+version: 0.0.13
```
ross-p-smith, line 3: Should this go to 6?

templates/shared_services/gitea/porter.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 name: tre-shared-service-gitea
-version: 0.0.1
+version: 0.0.7
```
ross-p-smith, line 3: nit: 0.0.2

templates/shared_services/sonatype-nexus/porter.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 name: tre-shared-service-nexus
-version: 0.0.1
+version: 0.0.8
```
ross-p-smith, line 3: 0.0.2

ross-p-smith (approved):
LGTM

resource_processor/shared/config.py
```diff
@@ -4,6 +4,7 @@
 def get_config() -> dict:
     config = {}
 
+    config["azure_location"] = os.environ["RESOURCE_LOCATION"]
```
tamirkamara, line 7: The RP is location agnostic and the API is the one aware of location and sends it in most payloads automatically. 
1. Can you check if this is also true for a shared service template?
2. Is this an intentional change in the pattern?

Makefile
```diff
@@ -279,6 +272,16 @@ bundle-register:
 	&& cd ${DIR} \
 	&& ${ROOTPATH}/devops/scripts/register_bundle_with_api.sh --acr-name "$${ACR_NAME}" --bundle-type "$${BUNDLE_TYPE}" --current --insecure --tre_url "$${TRE_URL}" --verify --workspace-service-name "$${WORKSPACE_SERVICE_NAME}"
 
+shared-service-register-and-deploy:
+	@# NOTE: ACR_NAME below comes from the env files, so needs the double '$$'. Others are set on command execution and don't
+	$(call target_title, "Registering ${DIR} bundle") \
```
tamirkamara, line 277: Title is wrong

devops/scripts/register_bundle_with_api.sh
```diff
@@ -97,19 +105,24 @@ if [[ -z ${acr_name:-} ]]; then
     usage
 fi
 
-if [[ -z ${BUNDLE_TYPE:-} ]]; then
+if [[ -z ${bundle_type:-} ]]; then
```
tamirkamara, line 108: Why the casting change? This is an environment variable ad I think it supposed to be upper case per [Bash best practices](https://google.github.io/styleguide/shellguide.html#s7-naming-conventions).
(anyway, you left a bunch with their upper case casting)

tamirkamara (changes requested):
I didn't look at everything, but had to request changes due to the RP becoming location aware which is a change into how the system works and probably not consistent with the other template types.

resource_processor/shared/config.py
tanya-borisova: @tamirkamara  No, I didn't realise that the RP is location agnostic and the API is location aware. Don't recall it being mentioned in the docs. I will change this, thanks!

devops/scripts/register_bundle_with_api.sh
tamirkamara: my bad. Hard to do this over the phone...

tamirkamara:
After reviewing this on a proper screen I think we should chat about several aspects:
1. Including the "deploy" action in the flow (through the make file and register-... script) only asks the API to deploy it. There's no feedback loop that checks if the shared service has actually been deployed.
This can create issues when (for example) you call the "install" for the firewall and on the next step, try to install nexus as the firewall data terraform object might fail.
2. It doesn't work for me - I tried to deploy a new TRE and then run make firewall-install, the deployment eventually failed and rendered the system unreachable as the app gateway started to return http 502. 
Looking at the docker logs, looks like TF has crashed without completing the deployment.
3. I haven't gone through the API code, but what will happen when another make firewall-install is issued? Will it save another resource?
4. A small thing - why didn't you take the location from the core resource group as it was before this PR? You anyway depend on its name to be right, so using a data object looks fine...

templates/shared_services/firewall/terraform/variables.tf
```diff
@@ -1,6 +1,15 @@
-variable "tre_id" {}
+variable "tre_id" {
```
damoodamoo, line 1: think if we're passing in the `tre_id` we should be able to get the RG via a data block and use that location, to avoid having to pass the location as well?

Makefile
```diff
@@ -16,9 +16,10 @@ build-and-push-api: build-api-image push-api-image
 build-and-push-resource-processor: build-resource-processor-vm-porter-image push-resource-processor-vm-porter-image
 build-and-push-gitea: build-gitea-image push-gitea-image
 build-and-push-guacamole: build-guacamole-image push-guacamole-image
+build-and-push-firewall: build-firewall-bundle push-firewall-bundle
```
damoodamoo, line 19: can't see where these targets actually are - did they get removed and this left behind?

devops/scripts/register_bundle_with_api.sh
```diff
@@ -159,26 +172,39 @@ else
...
 
-    if [[ ${status_code} != 200 ]]
-    then
+    if [[ ${status_code} != 200 ]]; then
       echo "::warning ::Template API check for ${bundle_type} ${template_name} returned http status: ${status_code}"
       exit 1
     fi
   fi
+
+  if [[ "${deploy_shared_service}" = "true" ]]; then
```
damoodamoo, line 197: as @tamirkamara said - this wants to have a polling loop in it to block until the shared service is deployed. it's in the e2e test code, in python - would just need to replicated here in bash

damoodamoo:
This is great progress on a really thorny piece of work! 

What's the expected behaviour for TRE's with a firewall already deployed - will this just post to the API and then essentially no-op? We'll also need to think about passing the firewall rules in via properties - but that's perhaps left to a new task.

Makefile
tanya-borisova: Good catch, yes these aren't used anymore

templates/shared_services/firewall/terraform/variables.tf
tanya-borisova: Either way works. I added the location back as it was before my PR  https://github.com/microsoft/AzureTRE/pull/1531/files#diff-244ec695609d33e2b1de334c5e0591a000b4ca09b6552d15559e22e5713b3957 for consitency with two other shared services. 

cc @tamirkamara  as he asked the same question

devops/scripts/register_bundle_with_api.sh
tanya-borisova: That's a good point. I'll look into implementing that now

templates/shared_services/firewall/terraform/variables.tf
tamirkamara: I think to align on the approach regarding the location. My initial comment was that getting the RP to provide location is something we need to think carefully about, but as I also noted in my comment from yesterday, you have the core RG name anyway so you can use the RG data object which is what we use in all bundles (except workspace ones)

tanya-borisova:
@damoodamoo  

> What's the expected behaviour for TRE's with a firewall already deployed - will this just post to the API and then essentially no-op?

Right now yes, it will no-op at Terraform level. 
Perhaps I should change the logic so that API fails when trying to deploy the same shared service of the same version. 

> We'll also need to think about passing the firewall rules in via properties - but that's perhaps left to a new task.

Yes, let's make that a separate task (perhaps the "Demonstrate updating firewall rule" line you have in item https://github.com/microsoft/AzureTRE/issues/1244 can be extracted into a task for this).

tanya-borisova:
@tamirkamara 

> It doesn't work for me - I tried to deploy a new TRE and then run make firewall-install, the deployment eventually failed and rendered the system unreachable as the app gateway started to return http 502.
Looking at the docker logs, looks like TF has crashed without completing the deployment.

Thanks for the bug report. I can't reproduce it in my environment -- any chance you have the docker logs saved so I can use it for debugging? 

The rest of the questions are the same as David asked, I replied to them above.

damoodamoo:
@tanya-borisova sounds good - let's try and keep the scope of this work contained and we can revisit edge cases of deploying twice to the API in later work. I'm planning to pull this and test it locally today so will report back...

Makefile
```diff
@@ -84,6 +84,21 @@ build-guacamole-image:
 build-mlflow-image:
 	$(call build_image,"mlflow-server","templates/workspace_services/mlflow/mlflow-server/version.txt","templates/workspace_services/mlflow/mlflow-server/docker/Dockerfile","templates/workspace_services/mlflow/mlflow-server")
 
+firewall-install:
+	make bundle-build DIR=./templates/shared_services/firewall/ \
```
tamirkamara, line 88: please convert all internal make calls to $(MAKE)

devops/scripts/register_bundle_with_api.sh
```diff
@@ -159,26 +172,39 @@ else
...
+    ("shared_service") tre_get_path="/api/shared-service-templates";;
   esac
 
   echo -e "Server Response:\n"
   eval "curl -X 'POST' ${tre_url}/${tre_get_path} -H 'accept: application/json' -H 'Content-Type: application/json' -H 'Authorization: Bearer ${access_token}' -d '${payload}' ${options}"
   echo -e "\n"
 
-  if ${verify}
-  then
+  if [[ "${verify}" = "true" ]]; then
```
tamirkamara, line 186: Why do you need "true"? Isn't Boolean somethin that works without the string notation?

devops/scripts/register_bundle_with_api.sh
tamirkamara: This becomes too complicated for a single script. I meant to write this even before but probably forgot... IMO we should mix the register work with the deploy one and prefer to split this into another script.

tamirkamara (changes requested):

devops/scripts/register_bundle_with_api.sh
tanya-borisova: What do you mean by a boolean? There are no booleans in Bash. You can declare a string variable, or a number, and compare it.
In this case, the "verify" was already declared as a string variable, and all I did was to make the if statements in the script more consistent, and do an explicit comparison (as per recommended, for example, here: https://stackoverflow.com/a/21210966/3488743)

devops/scripts/register_bundle_with_api.sh
tanya-borisova: I agree with that but splitting won't be trivial (e.g. I will need to extract retrieving of the access token into a separate script, and make the new script work with any type of bundle). 
I will leave a comment here with a link to a new issue instead.

tanya-borisova:
There is a bug in this version of PR: when resource processor is deploying the firewall shared service, after the deployment of firewall it blocks itself all traffic (as all traffic by default is denied). 

I think it should be possible to solve this via creating a Firewall policy first and then providing said policy when creating the firewall. But while I work on this, I'll move the PR back to draft

Makefile
tanya-borisova: Sorry, I don't understand what do you mean by that. Do you mean `$(call)`? If so, then I don't think I can set the values of env, such as `DIR` here, in the same statement

templates/core/terraform/resource_processor/vmss_porter/main.tf
```diff
@@ -163,3 +163,11 @@ resource "azurerm_role_assignment" "subscription_owner" {
   role_definition_name = "Owner"
   principal_id         = azurerm_user_assigned_identity.vmss_msi.principal_id
 }
+
+resource "azurerm_key_vault_access_policy" "resource_processor" {
```
tamirkamara, line 167: What is this used for?

tamirkamara (changes requested):

tamirkamara (approved):

templates/core/terraform/resource_processor/vmss_porter/main.tf
tanya-borisova: Gitea deployment doesn't work otherwise -- the deploying entity needs access to a secret (I can't remember which one it was but probably the administrator password)

api_app/db/repositories/shared_services.py
```diff
@@ -49,7 +50,11 @@ def get_deployed_shared_service_by_id(self, shared_service_id: str, operations_r
         return shared_service
 
     def get_shared_service_spec_params(self):
-        return self.get_resource_base_spec_params()
+        params = self.get_resource_base_spec_params()
+        params.update({
+            "azure_location": config.RESOURCE_LOCATION,
+        })
+        return params
```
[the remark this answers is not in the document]
tanya-borisova: So the thing is, when I was building the shared service API, my intention was to pass through azure_location (however for it to work this change was missing). 
Now in this PR we seem to want to not do that for shared services, and instead use the core resource group. 
I think azure_location is still used in shared service template, so if I remove it, I need to remove it there as well. In my mind it would be better to keep things consistent in this PR and remove references to azure_location in a separate PR.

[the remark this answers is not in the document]
tanya-borisova: For the sake of getting this PR through, I have removed this change. I will remove the remaining references in a follow-up PR.

tanya-borisova:
I don't really know what should be the way forward with this PR. The last 4 runs failed with the same error (permission denied when trying to connect to Docker socket from the build agent): 
image
This doesn't seem to be related to the PR, but doesn't look like a flake either. 
I wouldn't like to merge it without succeeding in CI first, since the changes to deployment are significant, would like to make sure it doesn't break CI for anybody else.

tamirkamara:
> I don't really know what should be the way forward with this PR. The last 4 runs failed with the same error (permission denied when trying to connect to Docker socket from the build agent): image This doesn't seem to be related to the PR, but doesn't look like a flake either. I wouldn't like to merge it without succeeding in CI first, since the changes to deployment are significant, would like to make sure it doesn't break CI for anybody else.

I don't know exactly why it happens but might be related to running porter and then trying to run another docker based action. I tested it on a side branch and was able to replicate it easily so merging it like so would mean that all CI validation would be fail going forward.
On the positive side, I suggest to split the shared services deployment activities into a publish step and a register/deploy step much like we do with the other bundles. This accomplishes 2 things:
1. It makes the pipeline complete successfully. An example run: https://github.com/microsoft/AzureTRE/actions/runs/2092453109
2. It reduces the total run time, as publishing (and building) the bundles runs in parallel to the deploy-core steps
The change is here: https://github.com/microsoft/AzureTRE/commit/215af780502ea6689da224a82ab902deeee498a0

tanya-borisova:
@tamirkamara I think your proposal is a good idea, at the very least it achieves consistency with the other bundles, as well as reducing the build time, so I will merge these changes in. Thanks!

> might be related to running porter and then trying to run another docker based action

Porter runs on resource processor, where the error that I linked is from running `docker login` on the build agent. So I don't think this would explain what's going on. 
@stuartleeks  and I will try to take a closer look at why the build agent error is happening, but that shouldn't prevent this PR from getting merged.

Makefile
tanya-borisova: With everything else I forgot to fix it, but I'll add a note to do it in my follow-up PR

tanya-borisova:
Successful CI run (I needed to run it with deploy_tre_branch.yml because this PR changes the workflows too)

Issue #1352, opened 2022-02-21
App registration's redirect url should be set automatically when creating a workspace


eladiw:
## Description

As a TRE Workspace Owner/Azure TRE Developer
I want the redirect url to be set automatically in the app registration
So that I won't need this manual step before I can use services such as Guacamole

## Acceptance criteria

- [ ] When creating a new WS, the app registration for guacamole can be determined and set, automatically
- [ ] Either in the script or using the RP 
- [ ] Design is needed as part of this user story
- [ ] ~~#220 will add the AAD Application automatically when the workspace is created.~~

joalmeid:
Duplicate of #1096

martinpeck:
@ross-p-smith  to add a few dependencies and details

PR #1600, opened 2022-03-26
Dev Flow for creating AAD assets

### LINKED ISSUE #1513: Granting Admin Consent to Automation Account fails first time
When you run 
```
./aad-app-reg.sh --name 'TRE' --swaggerui-redirecturl https://mytre.region.cloudapp.azure.com/api/docs/oauth2-redirect --admin-consent --automation-account
```
the first time, it can fail with saying that the automation account has not been created. This is just an aad propogation bug and just needs a wait around it.
This is around Line 764 of the script

**Steps to reproduce**

1. Run the script outlined above to create a new App Registration

ross-p-smith:
# PR for issue #1352
Fixes #1513

## What is being addressed

- new `make auth` which sets up TRE API. SwaggerUI, Automation Account and Workspace.
- Moved some of the AAD functions as a start to slim down the aad-app-reg.sh script.
- SwaggerUI app now has correct user_impersonation for api and workspace
- Add the Application.ReadWrite.OwnedBy permission and grant admin permission to it. This lays the work for the guacamole redirect to be updated by using the Workspace client/secret.


tamirkamara:
@ross-p-smith Can you explain how the entire follow is going to look like?

scripts/aad-app-reg.sh
```diff
@@ -379,6 +337,7 @@ JSON
 
 declare roleUserReadAll=$(get_msgraph_role "User.Read.All" )
 declare roleDirectoryReadAll=$(get_msgraph_role "Directory.Read.All" )
+declare roleApplicationReadWriteAll=$(get_msgraph_role "Application.ReadWrite.All" )
```
damoodamoo, line 340: Do we need `ReadWrite.All` ? If it's so that this app can create / manage others - I think we'd get away with `Application.ReadWrite.OwnedBy` - bit more tightly scoped to the apps for TRE? Definitions here - https://docs.microsoft.com/en-us/graph/permissions-reference#application-permissions-4

ross-p-smith: With OwnedBy
{"error":{"code":"Authorization_RequestDenied","message":"Insufficient privileges to complete the operation.","innerError":{"date":"2022-03-28T08:33:23","request-id":"4bd4caa4-bc35-4b5e-90ff-ee17efdcfb67","client-request-id":"4bd4caa4-bc35-4b5e-90ff-ee17efdcfb67"}}}

ross-p-smith: Put it back to "All" and it works again. :-( I am unsure why - happy to walk through it as may have missed something - could put a note in the code?

damoodamoo: yea it's not a massive deal - the app is definitely an owner of the one it's trying to modify? but if not let's use `.All` then.

stuartleeks: I agree that it would be preferable to use `Application.ReadWrite.OwnedBy`.
Sounds like the app isn't the owner of the object it's trying to modify? If so, is that something that we expect will change as this work progresses? (i.e. add a comment in the code and an issue to track re-visiting this once the work is further on)

stuartleeks (approved):
LGTM

PR #1606, opened 2022-03-28
stop/start mysql

### LINKED ISSUE #1605: TRE cost can be reduced by stopping MySQL
## Description

As a TRE Developer I want to better control the cost of my environment by stop/start the MySQL in the core RG. 
This resource is number 4 in the cost breakdown and the next one in the list we can stop (after the firewall and appgw)
This is around 150$/month

## Acceptance criteria

- Control script is able to stop and start the MySQL

tamirkamara:
Fixes #1605 

## What is being addressed

Reduce costs when the TRE isn't in use by stopping Azure MySQL as well

## How is this addressed

- As described.

### DIFF
--- a/devops/scripts/control_tre.sh
+++ b/devops/scripts/control_tre.sh
@@ -35,6 +35,13 @@ if [[ "$1" == *"start"* ]]; then
   else
     echo "Application Gateway already running"
   fi
+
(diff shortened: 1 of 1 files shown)


stuartleeks (approved):
lgtm

Issue #1413, opened 2022-03-02
All operation updates should include the user who made the change


daltskin:
## Description

As a TRE Administrator
I want to see who made changes to shared services

## Acceptance criteria

- [x] The user that made the change is returned within the operation log
- [x] The user that made the change is returned within the operation history

PR #1612, opened 2022-03-28
Log requesting user in API

damoodamoo:
# PR for issue #1413 

## What is being addressed

The user is retrieved from the token in the API already. This user object is now submitted to cosmos and tracked in the resource, as well as resource history and each operation. 

Also noticed the timestamp wasn't being set properly when creating / updating a resource, so set that and added to resource history.


ross-p-smith:
You'll need to bump the api version number

PR #1620, opened 2022-03-29
Move workspace_service_name check further down

### LINKED ISSUE #1587: make register-bundle fails if not providing workspace_service_name
register_bundle_with_api.sh now has a check that `if [ ${BUNDLE_TYPE} == "user_resource" ] && [ -z ${workspace_service_name:-} ];` line 105 which is only needed if you are using the automation account to register with the API automatically.  

If you want the payload to be returned to console and don't supply `workspace_service_name` the script fails before the payload is output to the console.

I suggest this if statement be moved down to line ~152 to allow the json to be output to the console if no access token is acquired which is still before `workspace_service_name` is used.

ross-p-smith:
# Fixes #1587 

## What is being addressed

Should still output the bundle payload if you are not trying to automatically register it

## How is this addressed

- Moved the check further down in the script


stuartleeks (approved):
lgtm

PR #1626, opened 2022-03-29
Add bash linting & Fix the TF one

tamirkamara:
## What is being addressed

We have many bash files with substantial logic, but we don't lint them. We also don't lint Terraform files for a few months ever since SuperLinter changed the flag name...

## How is this addressed

- Update the TF flag name
- Add more linters: bash, bash-exec (.sh files need to be executable), github actions
- Update line length for yaml files to 120 and have it apply for the linter as well (unfortunately it has to be in 2 places)
- Remove RUN_LOCAL flag which makes this lint ONLY changed files (compared to main)
- Use the slim version of SuperLinter
- Have a make target that runs SuperLinter locally.

A child PR is opened for linting fixes to make it easier to understand the changes: https://github.com/microsoft/AzureTRE/pull/1627


stuartleeks (approved):
LGTM

PR #1627, opened 2022-03-29
Bash & Workflow linting issues

tamirkamara:
Bash changes:
- double escape special chars - `\n` -> `\\n`
- add shell type header `#!/bin/bash`

Workflow changes:
- run container command action - most inputs need to be optional since they are not always passed in
- disable yamllint rule for the "on" key as it's not valid
- double quotes for all vars in run steps (this need to happen for regular bash script in *.sh files but it was too much for me - need proper testing)


stuartleeks (approved):
LGTM

.github/actions/devcontainer_run_command/action.yml
```diff
@@ -47,55 +47,55 @@ inputs:
     required: true
```
ross-p-smith, line 47: Just out of interest - why turn these all to optional?

ross-p-smith (approved):

.github/actions/devcontainer_run_command/action.yml
tamirkamara: Because we don't send them on every call to this action. 
When I added the Github Actions linter it started to complain that a required input wasn't being passed...

ross-p-smith: Makes sense!

tamirkamara:
Closing for now as the bash fixes are way to much for me to handle. Will reopen small ones.

Issue #1615, opened 2022-03-29
Docs and Code Still Refer to the Removed Endpoint /api/status


martinpeck:
We have removed the /api/status endpoint, but the docs still refer to it:

https://microsoft.github.io/AzureTRE/tre-admins/setup-instructions/deploying-azure-tre/#validate-the-deployment

martinpeck:
`strings.py` also refers to this endpoint
`test_provisioned_health_api.py` still has a test for this endpoint

anatbal:
@martinpeck I can't seem to find a reference for it in strings.py. Do you mind pointing out where?

PR #1629, opened 2022-03-30
Fix Docs and Code References of Removed Endpoint /api/status 

anatbal:
# PR for issue #1615 

## What is being addressed

We have removed the /api/status endpoint, but the docs still refer to it. In addition, a fix for a test in the e2e tests that still refers to it.


martinpeck (approved):
LGTM

Issue #1635, opened 2022-03-30
Make Firewall deployable from Resource Processor


tanya-borisova:
There is a bug in this version of PR: when resource processor is deploying the firewall shared service, after the deployment of firewall it blocks itself all traffic (as all traffic by default is denied). 

I think it should be possible to solve this via creating a Firewall policy first and then providing said policy when creating the firewall. But while I work on this, I'll move the PR back to draft

_Originally posted by @tanya-borisova in https://github.com/microsoft/AzureTRE/issues/1597#issuecomment-1081920933_

PR #1636, opened 2022-03-30
Make firewall deployable from resource processor

tanya-borisova:
# PR for issue

Closes https://github.com/microsoft/AzureTRE/issues/1635

## What is being addressed

This PR has been split from https://github.com/microsoft/AzureTRE/pull/1597  and addresses the problem of deploying Firewall from Resource Processor.

Currently, Firewall is being deployed separately from the rules. Firewall [denies all traffic by default](https://docs.microsoft.com/en-us/azure/firewall/rule-processing). This means that when Firewall shared service is deployed from the Resource Processor as a Porter bundle, immediately after deployment Resource Processor loses all access to internet (including Terraform servers) and cannot finish the deployment of Firewall rules (which would have opened up this traffic. 
To solve this problem, a Firewall policy containing the rules is created first. Then, when deploying Firewall, this policy is referenced which means that Resource Processor never loses access to Terraform. 

### Note on pricing

[No Azure Firewall Manager policy charges will be done for policies that are associated to a single firewall.](https://azure.microsoft.com/en-gb/pricing/details/firewall-manager/)

## How is this addressed

- Firewall policy is created in Firewall module
- Rule priorities have been changed: priority of Application rules [has to be higher](https://docs.microsoft.com/en-us/azure/firewall/rule-processing#network-rules-and-applications-rules) in the policy than priority of Network rules, otherwise Terraform complains. As such, rules are ordered so that network rules go first. Network rules also start from 200 and application from 300, as per [convention](https://docs.microsoft.com/en-us/azure/firewall/policy-rule-sets#rule-collection-groups)


tamirkamara:
Nice progress resolving this problem. A couple of questions:
1. I recall other bundles also making changes to the firewall. Are you going to handle those as well in this pr?
2. Have you considered multiple policies like described here? https://docs.microsoft.com/en-us/azure/firewall-manager/policy-overview#hierarchical-policies

marrobi:
I think policies are good, but is there a reason why could not just leave the deployment of the routes table to after the rules gave been added?

As until the route table is added traffic will not touch the firewall.

tanya-borisova:
> I think policies are good, but is there a reason why could not just leave the deployment of the routes table to after the rules gave been added?

That looks promising, after 1 test this has worked. I will run on a couple more TREs to be sure.
Since it's only like a 2 line change in code of the firewall, I will make it in my original PR https://github.com/microsoft/AzureTRE/pull/1597

The change in policies we may or may need to look into once we implement the update (and Tamir's suggestion on hierarchical policies might come in handy), but for now I will postpone this until https://github.com/microsoft/AzureTRE/pull/1597 is merged.

PR #1639, opened 2022-03-31
Service Bus Sessions + Parallel / Serial Processing

### LINKED ISSUE #1245: Use Sessions in Service Bus to enforce ordering
Service Bus supports Sessions if required. We should set the SessionId of each Service Bus message to be the Resource ID of the resource undergoing change. 

This would ensure that any consumers of the service bus messages would only process one change against a resource at a time, even if there were multiple concurrent consumers. 

Acceptance:
- [ ] Post a number of updates to a resource via the API
- [ ] Inspect that one follows another, even when multiple runners are consuming messages

damoodamoo:
# Closes #1245 

## What is being addressed

We need to support parallel execution and deployments, without introducing conflicts of multiple updates being attempted against a resource at once.

## How is this addressed

- Service bus queue for deployments has sessions enabled
- Session ID set to resource ID - so multiple updates to the same resource will exist in the same session in SB, and be queued behind each other to be processed in serial. Other deployments will be executed by spare processors in the RP, in parallel.
- Message receiver logic refactored in the runner:
  - `yield` logic removed, message processing is awaited in the loop
  - better exception handling for service bus transient errors, and clearer logging to describe what each process is doing
  - some `uamqp` logging levels raised to avoid polluting the logs with messages that look like errors that require action, when in fact they are expected (timeouts etc) 
- Introduction of basic, manual performance tests to validate behaviour
  - Not for the build, but a new `E2E Performance` debug task has been added, with 2x perf tests which test creating separate resources in parallel, and then test applying multiple concurrent updates to user resources to ensure those updates queue and are not acted upon in parallel.

In my testing I have run 10x processors in the RP, and via the perf tests have created 2x VMs, applying 5x updates to each. Without sessions these would all run together and 4 out of 5 would fail. With sessions, I have confirmed that those updates are queued and are picked up either by the same process - or a new process _after_ the previous one has finished and disconnected from the session.

In the resource processor logs you'll see entries such as:
```bash
# Process 17 attempting to get a new service bus session. This will eventually time out if no 'available' messages are there
runner  root    17      2022-03-31 16:04:19,589 INFO    Looking for new session...

# Process 17 timed out looking for messages on a new session
runner  root    17      2022-03-31 16:04:13,719 INFO    No sessions for this process. Sleeping 30s then will look again...

# Process 23 has a session that contains messages, and will hold on to that whilst there are more messages with the same 
# session ID
runner  root    23      2022-03-31 15:43:21,123 INFO    Get a session containing messages
runner  root    23      2022-03-31 15:44:21,572 INFO    Message received for resource_id=c8063edb-c9ef-4b1f-9627-bc3ccfff0968,
```


e2e_tests/config.py
```diff
@@ -14,3 +14,8 @@
 TEST_ACCOUNT_CLIENT_ID: str = config("TEST_ACCOUNT_CLIENT_ID", default="")
 TEST_ACCOUNT_CLIENT_SECRET: str = config("TEST_ACCOUNT_CLIENT_SECRET", default="")
 TEST_WORKSPACE_APP_ID: str = config("TEST_WORKSPACE_APP_ID", default="")
+
+# Perf test env vars - set these in private.env if you want to run perf tests and use an existing
```
ross-p-smith, line 18: Should say /templates/core/.env. private.env will be autogenerated and will wipe your settings. The launch target you created copies the /templates/core/.env into e2e folder and also then pulls in private.env. (Effectively tests read 2 .env files)

resource_processor/version.txt
```diff
@@ -1 +1 @@
-__version__ = "0.1.8"
+__version__ = "0.1.10"
```
ross-p-smith, line 1: nit: should be 9 :-)

ross-p-smith (approved):
LGTM

damoodamoo:
/test-force-approve

have run PR build + smoke tests already. e2e extended + performance run locally

damoodamoo:
/test-force-approve

have run PR build + smoke tests already. e2e extended + performance run locally

PR #1641, opened 2022-04-03
Require TRE ID to be lowercase.

### LINKED ISSUE #1631: CosmosDB and Application Gateway creation fails when tre_id contains capital letters
**Describe the bug**

When tre id contains capital letters the tre deployment fails on CosmosDb/Application Gateway  creation with the following error:
"name must be 3 - 50 characters long, contain only LOWERCASE letters, numbers and hyphens"

**Steps to reproduce**

1. Set TRE_ID environment variable to contain capital letters
1. Deploy TRE

LizaShak:
## What is being addressed

Fixes #1631. Some resources name require to be lowercase and fail during the deployment otherwise. To avoid failures during deployment and for easy maintenance I have added a requirement on the tre id to be lowercase.

## How is this addressed

- Add on TF var tre_id a condition which requires it to be lowercase


marrobi (approved):
LGTM

eladiw (approved):

PR #1642, opened 2022-04-03
Limit locations.

### LINKED ISSUE #1632: Deploy fails on creating Public IPs in a location which is not an availability zone
**Describe the bug**

When deploying TRE to a  location which is not an availability zone the deployment fails on the public ip creation. We could avoid getting this error and having all the resource deployed beforehand by pre checking the location is in availability zones. 

**Steps to reproduce**

1. Define LOCATION env variable to be a non [availability zone](https://docs.microsoft.com/en-us/azure/availability-zones/az-overview)
1. Deploy TRE

LizaShak:
# PR for issue #1632 

## What is being addressed

Fixes #1632 
Limit deployment locations. 

Context: public IP cannot be deployed within a location which is not within an availability zones. To avoid this error during the deployment in this PR a condition on the location was added.

## How is this addressed

- Add a condition on a terraform variable.


tamirkamara (commented):
I thought we'll enable deploying into a region without availability zones by using a flag...

PR #1647, opened 2022-04-04
Ability to select multiple VM sizes for the Guacamole Linux user resource 

### LINKED ISSUE #1200: Ability to select multiple VM sizes for the Guacamole user resource
When deploying a Guacamole VM user resource, you don't have the option of selecting any other size than the hardcoded value in the bundle (DS1_v2). This could be too small for some customers to work with large datasets etc. which require more performance.

I would like to be able to select from multiple VM sizes when passing the user resource json / having a dropdown in a UI.

anatbal:
# PR fixes #1200 

## What is being addressed

-  Add vm_size for the Linux vm user resource as already exists in the windows vm
- Upgrade from v2 to v5 in both
- Fixed uninstall bug in the windows resource
- Add upgrade to the Linux user resource


damoodamoo (approved):
Looks good!

PR #1649, opened 2022-04-05
Nexus config fix

oliver7598:
## What is being addressed

File path referenced in `nexus_config.sh` hadn't be changed since it was moved from the `/scripts/` folder to `/templates/shared_services/sonatype-nexus/scripts/` causing the script to fail when run.

Minor change to reflect the new script location.


jjgriff93:
Closing as addressed in #1584

PR #1653, opened 2022-04-06
Basic Health Logs

### LINKED ISSUE #1652: Health check endpoint should log all the service status as it queries
Currently the `/health` endpoint queries Cosmos / Service Bus / the RP - and returns the statuses. If any are not ok, the response is a 503.

There is currently no way to query that endpoint when the gateway has blocked access - so we at least need it to log the results so we can track back and see what service was down, when.

damoodamoo:
# Closes #1652 

Basic logging added when a service is not ok.


tanya-borisova (approved):

stuartleeks (approved):
LGTM

martinpeck (approved):

ross-p-smith (approved):

PR #1656, opened 2022-04-06
Session Termination

### LINKED ISSUE #1655: Service Bus Sessions never terminate
After a processor receives a message on a session, it hangs onto that session indefinitely, blocking the thread and meaning other messages cannot be processed. 

We need to terminate the session after each message has been processed / errored out.

damoodamoo:
# Fixes #1655 

`max_wait_time` set to `1` second in the receiver. This is the number of seconds the session will be held open after it's started, waiting for new messages. It defaults to infinite.

Means after each message the session is over and a new one is started, but the overall effect of stacking messages up per resource is the same.

### DIFF
--- a/resource_processor/vmss_porter/runner.py
+++ b/resource_processor/vmss_porter/runner.py
@@ -52,10 +52,11 @@ async def receive_message(service_bus_client):
     while True:
         try:
             logger_adapter.info("Looking for new session...")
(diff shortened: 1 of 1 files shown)


tanya-borisova (approved):
LGTM but I wonder if there's any test we can add for this case?

damoodamoo:
/test-force-approve

extended tests ran

Issue #1040, opened 2021-10-13
Add service bus connectivity and resource processor health to status endpoint


marrobi:
## Description

As a TRE Administrator
I want to know if the system is healthy
So that I can report issues quickly

## Acceptance criteria

- [x] #1361
- [x] Status endpoint returns resource processor health as per cosmos. Requires #701

anatbal:
@marrobi, @tamirkamara   After taking a look at [#119](https://github.com/microsoft/AzureTRE/pull/119) it seems like this task means bringing /api/status and /api/health back together, if I understand correctly?

PR #1658, opened 2022-04-06
Stop the /health endpoint returning 503 when a downstream service fails

damoodamoo:
`/health` should return a 200, unless the API is having an issue in which case another status will be returned and picked up in the App Gateway health probe. 

This change should prevent app gateway blocking calls to the API when the resource processor is being updated, for instance, and hopefully make the builds a bit more consistent. 

Also allows us to stop the RP in Azure whilst debugging locally, and still have the Azure API running to test against.


anatbal:
@marrobi  - when I first implemented this I followed this story #1040. I think since implemented this there were issues about tracking the system health and maybe we should think to reshape it a little. What do you think?

tamirkamara (approved):
Approving per our offline discussion to validate if this has any impact on the instability around deployments in our CI.

marrobi:
@damoodamoo @tamirkamara @anatbal do we have an issue to track this?

I think we actually need a "system health" - which provides full system health as at present. It is important this does not always return 200 so can be used by external monitoring systems. However I do think we need to auth on this endpoint as is there not a risk of DDoS on downstream services?

Ahen for the app gateway API health probe we should probe a basic "api health" endpoint. Be that a simple ping, or actually includes cosmos health too is open to discussion.

Thoughts?

damoodamoo:
I agree - we should have an api-health endpoint, returning a ping, always 200, and a system-health endpoint which returns the system status and a 503.
________________________________

tamirkamara:
IMO we have several different requirements and use-cases here...
We should move the discussion to a dedicate issue: https://github.com/microsoft/AzureTRE/issues/1673.

PR #1662, opened 2022-04-07
Specify porter mixins

### LINKED ISSUE #1661: Install only Porter mixins we use
## Description

As a TRE Developer I want the devcontainer to include the porter mixins we need for this project so that the image size will be smaller.

## Acceptance criteria

- [ ] Mixins are limited to what we need.

tamirkamara:
Fixes #1661

## What is being addressed

Porter default install script adds things we don't need which increases the devcontainer size.
All items below results in almost 300MB reduction in size:
image

## How is this addressed

- "manually" install porter and the mixins we need
- Use a symlink to avoid copying the porter binary twice (as the automated script does)
- All `apt-get install` run commands end with a cleanup procedure
- Added an INTERACTIVE argument to control installing additional tools that are needed for a developer but not in CI
- All `pip install` commands use `--no-cache-dir` to save space
- Remove azfirewall extension as a dependency as it's not needed in most cases and scripts install it on demand
- Make sure the e2e python dependencies are being installed as well


.devcontainer/Dockerfile
```diff
@@ -77,4 +96,4 @@ RUN echo "export HISTFILE=$HOME/commandhistory/.bash_history" >> "$HOME/.bashrc"
 
 # Install github-cli
 COPY ./.devcontainer/scripts/gh.sh /tmp/
-RUN /tmp/gh.sh
+RUN if [ "${INTERACTIVE}" = "true" ]; then /tmp/gh.sh; fi
```
eladiw, line 99:
```suggestion
RUN if [ "${INTERACTIVE}" = "false" ]; then /tmp/gh.sh; fi
```

?

tamirkamara: We use GH cli in non-interactive mode when we destroy environments. But since that happens outside the devcontainer it means that this is not required for non-interactive hence my original change is correct.

.devcontainer/Dockerfile
```diff
@@ -30,26 +33,40 @@ COPY .devcontainer/scripts/docker-client.sh /tmp/
...
 
 # Install Docker
-RUN apt-get update && sudo apt-get install -y  apt-transport-https ca-certificates curl gnupg  lsb-release \
+RUN apt-get update && sudo apt-get install -y apt-transport-https ca-certificates curl gnupg lsb-release \
     && curl -fsSL https://download.docker.com/linux/debian/gpg | gpg --dearmor -o /usr/share/keyrings/docker-archive-keyring.gpg \
     && echo "deb [arch=amd64 signed-by=/usr/share/keyrings/docker-archive-keyring.gpg] https://download.docker.com/linux/debian $(lsb_release -cs) stable" \
     | tee /etc/apt/sources.list.d/docker.list > /dev/null \
-    && apt-get update && apt-get install -y docker-ce docker-ce-cli containerd.io
+    && apt-get update && apt-get install -y docker-ce docker-ce-cli containerd.io \
+    && apt-get clean -y && rm -rf /var/lib/apt/lists/*
```
ross-p-smith, line 41: You clean on line 51 - wouldn't it be better to clean once at the end?

tamirkamara: No :-)
Each layer is independent, so if a run command doesn't end with the cleanup, the data is persistent in the layer. Cleaning in the end will be a separate layer that will only mask / mark as deleted but the container will still contain that data and its size will be larger.

.devcontainer/Dockerfile
```diff
@@ -30,26 +33,40 @@ COPY .devcontainer/scripts/docker-client.sh /tmp/
...
+    && /opt/certbot/bin/pip install --no-cache-dir certbot \
+    && apt-get clean -y && rm -rf /var/lib/apt/lists/*
 
 # Install Porter
-RUN export PORTER_HOME=/home/$USERNAME/.porter \
-    && curl -L https://cdn.porter.sh/latest/install-linux.sh | bash \
-    && ${PORTER_HOME}/porter mixin install docker \
-    && chown -R $USERNAME ${PORTER_HOME}
+# Not using the script from https://cdn.porter.sh/latest/install-linux.sh
+# as it installes things we don't need and duplicates the binary.
```
ross-p-smith, line 55:
```suggestion
# as it installs things we don't need and duplicates the binary.
```

.devcontainer/Dockerfile
```diff
@@ -30,26 +33,40 @@ COPY .devcontainer/scripts/docker-client.sh /tmp/
...
+ARG PORTER_MIRROR=https://cdn.porter.sh
+ARG PORTER_PERMALINK=v0.38.10
```
ross-p-smith, line 57: Real petty and feel free to ignore - do you think this would be better in a scripts\porter.sh

(just a thought!)

.devcontainer/Dockerfile
tamirkamara: done

.devcontainer/Dockerfile
tamirkamara: done

PR #1680, opened 2022-04-11
Allow Azure Trusted Services and add note in the docs

### LINKED ISSUE #1654: azureml service wrong Keyvault Configuration
With the default AML workspace is not able to access keyvault which will result in a "forbidden" error while trying to list datastore or create new ones.

To fix this : 
Key vault is missing the "Allow trusted Microsoft services to bypass this firewall" in the network blade of the Keyvault.

in Terraform it translates by adding a bypass property with a value of "AzureServices" on the network_acls block
https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/key_vault

marrobi:
# Allow Azure Trusted Services to Key Vault in base workspace

Closes #1654 

## What is being addressed

As per #1654 this setting is needed to enable certain Azure PaaS services to interact with the Key Vault. As it is set on the storage account, makes sense to have it set on the Key Vault.

As some users may not be comfortable with this added a note to the docs to explain.

### DIFF
--- a/docs/tre-templates/workspaces/base.md
+++ b/docs/tre-templates/workspaces/base.md
@@ -9,3 +9,12 @@ The base workspace template contains the following resources:
 - Key Vault
 - VNet Peer to Core VNet
 - Network Security Group
+- App Service Plan
+
+## Azure Trusted Services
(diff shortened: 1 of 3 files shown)


ross-p-smith (approved):
LGTM

PR #1681, opened 2022-04-11
Fix the E2E tests and add the new attributes for the new guacamole oauth proxy

anatbal:
# PR for issue

## What is being addressed
In order to create a new guacamole workspace service, we now should pass on the request payload 3 parameters: ws_client_id (former openid_client_id), ws_client_secret and oauth2_proxy_email_domain. In order for the e2e to work, we need to add those 3 to all required pipelines.


.github/actions/devcontainer_run_command/action.yml
```diff
@@ -59,6 +59,12 @@ inputs:
   TEST_WORKSPACE_APP_ID:
     description: "The Test Workspace Id used to interact with the API."
```
eladiw, line 60:
```suggestion
    description: "The Test Workspace application Id used to interact with the API."
```

.github/actions/devcontainer_run_command/action.yml
```diff
@@ -59,6 +59,12 @@ inputs:
...
     required: false
+  TEST_WORKSPACE_APP_SECRET:
+    description: "The Test Workspace secret used to interact with the API."
```
eladiw, line 63:
```suggestion
    description: "The Test Workspace application secret used to interact with the API."
```

.github/actions/devcontainer_run_command/action.yml
anatbal: Fixed

.github/actions/devcontainer_run_command/action.yml
anatbal: Fixed

PR #1683, opened 2022-04-11
Get rid of TRE_URL as this introduces bugs

tanya-borisova:
# PR for issue

## What is being addressed

- Get rid of TRE_URL as this introduces bugs (e.g. URL becomes this `https://${TRE_ID}.northeurope.cloudapp.azure.com/api/shared-services` without adding extra `eval` statements)
- In the `register_bundle_with_api.sh` script, explicitly check the HTTP response when registering template.
- Opt in to Gitea and Nexus as it used to be before

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -285,7 +285,7 @@ bundle-register:
 	&& . ./devops/scripts/load_env.sh ./templates/core/.env \
 	&& az acr login --name $${ACR_NAME}	\
 	&& cd ${DIR} \
-	&& ${ROOTPATH}/devops/scripts/register_bundle_with_api.sh --acr-name "$${ACR_NAME}" --bundle-type "$${BUNDLE_TYPE}" --current --insecure --tre_url "$${TRE_URL}" --verify --workspace-service-name "$${WORKSPACE_SERVICE_NAME}"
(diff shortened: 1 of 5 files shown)


jjgriff93 (approved):

tanya-borisova:
@tamirkamara  no, I think @ross-p-smith  mentioned that the TRE_URL was introduced as a fix for something else, and in it's current form it appears broken anyway, as it doesn't substitute the location either. 
Do you have a suggestion for how to make a fix for this so that it honors local flow?

ross-p-smith:
> This makes the assumption we always run in the cloud and would make it harder to run/debug the system locally. Was this taken into account @tanya-borisova?

I think what Tamir is suggesting; is that if he wanted to register bundles to his local deployment, he just amends the TRE_URL to localhost/blah and things work locally - is that correct @tamirkamara ?

ross-p-smith:
Just a random thought - would it help if you made /workspaces/AzureTRE/devops/scripts/register_bundle_with_api.sh line
```
tre_url=eval($1)
```

And then pass in the TRE_URL as before. Obvs haven't tested it! :-)

tamirkamara (approved):
lgtm

PR #1685, opened 2022-04-11
`pipeline: {}` init

### LINKED ISSUE #1677: Create basic, no-op `pipeline: {}` block in new `dev-vm` resource
This task is to enable future work on the `pipeline: {}` without interfering with existing functionality.

- [x] Copy the `guacamole-azure-linuxvm` to `guacamole-dev-vm`
- [x] Add a basic 3-step `pipeline: {}` to the schema to update the `display_name` of the firewall shared service, run main, then update the firewall name again.
- [x] Make sure the `pipeline: {}` block is ignored/removed during schema validation.
- [x] Start a doc to describe what the `pipeline:{}` is for. This will be updated as features are added in subsequent tasks.

damoodamoo:
# Closes #1677

This PR lays the groundwork for the pipeline work to be built upon.

- A new `dev-vm` has been added, to use in dev + testing without polluting the existing user resources
- Early docs have been added that will be updated at each stage as features are added
- `pipeline` property added as optional to the resource template model


docs/tre-templates/pipeline-templates/overview.md
```diff
@@ -0,0 +1,56 @@
+# Pipeline Templates
```
tanya-borisova, line 1: Nice work on updating docs right away 👍 
I think maybe make it super clear that it's a work in progress for now, so that people don't try to update things and expect it to work?

templates/workspace_services/guacamole/user_resources/guacamole-dev-vm/terraform/deploy.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
tanya-borisova, line 1: Do you need this script?

damoodamoo: I think so? It's a direct copy of the linux-vm where I assume it's needed?

tanya-borisova (approved):

damoodamoo:
/test-force-approve

have run tests already

PR #1686, opened 2022-04-11
Provided example workspace id for the SwaggerUI

### LINKED ISSUE #1549: WS id can be taken from url
**Describe the bug**
When using the 'second' swagger api (https://<azure_tre_fqdn>/api/workspaces/<workspace_id>/docs), the one intended for the ws owner. The url itself contains the ws id and then it is still required for each api call

If possible to reduce this parameter it will improve usability
**Steps to reproduce**

image

ross-p-smith:
# Fixes #1549

## What is being addressed

When we render the SwaggerUI, place the WorkspaceId as an 'example' in the OpenAPI spec.

(Would love some nitpick on the python as I'm a newbie!!) Shame @martinpeck is not lurking around.

image


ross-p-smith:
/test-force-approve

(tests previously ran successfully)

Issue #1697, opened 2022-04-14
Default value for the windows vm_size is incorrect


anatbal:
**Describe the bug**
The default value for the windows vm_size  is incorrect, as a result cannot create a windows vm without the "vm_size" attribute

PR #1698, opened 2022-04-14
Fix default value for windows vm_size

anatbal:
# PR for issue #1697

## What is being addressed

The default value for the windows vm_size is incorrect, as a result cannot create a windows vm without the "vm_size" attribute


marrobi (approved):
LGTM - the template schema is ok, which is why I have not hit this before.

anatbal:
Not testing this since there aren't any tests for user_resource in the e2e, which is described in a different issue

Issue #220, opened 2021-06-08
AD app registration created when new workspace is created


christoferlof:
## Description

As a TRE administrator
I want to create workspaces without providing identity provider parameters such as application id
So that I don't have to bother our cloud admin asking her to create the application registration for me first.

## Acceptance criteria

- [x] #1706
- [x] #211 
- [x] #1797
- [x] #1928
- [x] App registration deleted when the workspace is deleted (Terraform is a good option for creation/deletion)
- [ ] Documentation updated with environment variable and permissions for automatic app registration
- [x] #831
- [x] #1716

## Design 

Workspace templates must only change resources it itself creates. Changes done to management/core services such as Azure Firewall, Shared Service such as source mirror, Azure Active Directory etc. must be delegated.

Support for alternative identity providers other than Azure Active Directory for application registration must also be supported. It should be configurable which identity provider is used at TRE creation time (not part of this Story). Meaning the same identity provider will be used in all Workspaces within the same TRE instance.

Therefore, the responsible for creating application registrations is implemented into the TRE management API.

### Permissions

To be able to automatically create app registration the [Application.ReadWrite.All](https://docs.microsoft.com/en-us/graph/permissions-reference#delegated-permissions-4) permission must be granted to the TRE management API Service Principal

Azure AD is managed via the Microsoft Graph API.

### Changes to setup 

A new environment variable `auto_app_registration = [TRUE|FALSE] (default=FALSE)` indication if automatic app registration is enabled when setting up a TRE instance.
If the environment variable is set to true, then the Application.ReadWrite.All permission must be granted.

### Changes to the TRE API

Currently the following parameters are required in the REST API to create a Workspace

```JSON
{
  "workspaceType": "tre-workspace-vanilla",
  "properties": {
    "display_name": "the workspace display name",
    "description": "workspace description",
    "client_id": "9d52b04f-89cf-47b4-868a-e12be7133b36"
  }
}
```

The `app_id` should be made optional, but only if the `auto_app_registration` variable is set to TRUE.

The abstraction over identity provider is implemented in the [AccessService](https://github.com/microsoft/AzureTRE/blob/develop/management_api_app/services/access_service.py) class. This class must be extended to include an application registration method.

Incorporate and reuse as much as possible from the [manual app registration script](https://github.com/microsoft/AzureTRE/blob/develop/scripts/workspace-app-reg.py).

[...]

PR #1701, opened 2022-04-14
3. AAD Managed by Workspace creation (Turned off at the moment)

### LINKED ISSUE #1201: Debug and Purge Protection Core and Workspace

**Describe the bug**
Setting debug to true disables Purge Protection in the core
https://microsoft.github.io/AzureTRE/tre-admins/environment-variables/#for-shared-management-resources-in-devopsenv

Workspace are still created with Purge Protection enabled (and it cannot be disbaled)

Suggest that debug should consistently set Purge Protection for both core and workspace

This could be done by replicating the setting in
https://github.com/microsoft/AzureTRE/blob/main/templates/core/terraform/keyvault/keyvault.tf
to
https://github.com/microsoft/AzureTRE/blob/main/templates/workspaces/base/terraform/keyvault.tf

**Steps to reproduce**

1. Follow setup instructions https://microsoft.github.io/AzureTRE/tre-admins/setup-instructions/getting-started/
2. In Step 2 set Debug="false"
3. Create Workspace, observe that in the core KeyVault Purge protection is disabled and in the Workspace KeyVault purge protection is enabled

ross-p-smith:
# PR for issue #220 

## What is being addressed

This workspace bundle will create the workspace and create an AAD Application, then put the client id/secret in the keyvault that it has just created. This is only the Terraform at the moment, but thought it warranted a review. It also works from inside porter. But there are other pieces necessary for the full end-to-end to work. This will be featured by a feature flag `register_aad_application`

## How is this addressed

- [X] New workspace bundle
- [X] Terraform debug script model so that it can be tested independently from porter.
- [X] Fixes #1201 - Workspace Key Vault now has an optional purge protection for aiding debugging.
- [x] Pass new aad_auth.json file to porter from Resource Processor (#1702)
- [x] Add API client/Secret into core key vault (#1702)
- [x] Wire up Porter/KeyVault plugin to read credentials (#1702)
- [x] Add Application.ReadWrite.All to Api app (conditionally) (#1704)
- [X] Amend the API so that `auto_create` will trigger an automatic registration of the bundle #1705


api_app/schemas/workspace.json
```diff
@@ -22,6 +23,12 @@
           "description": "Description of the workspace",
           "updateable": true
       },
+      "app_id": {
```
marrobi, line 26: Not clear why this is being added. This means every workspace will need an app id passing into the API and the UI will require it. Not sure if this is the intention?

templates/core/terraform/main.tf
```diff
@@ -14,7 +14,7 @@ provider "azurerm" {
   features {
     key_vault {
       purge_soft_delete_on_destroy    = var.keyvault_purge_protection_enabled ? false : true
-      recover_soft_deleted_key_vaults = true
+      recover_soft_deleted_key_vaults = false
```
marrobi, line 17: Why are we changing this? I believe it was set to true so could redeploy with the same tre_id?

templates/workspaces/base/debug-porter.sh
```diff
@@ -0,0 +1,27 @@
...
+read -p "Would you like to uninstall this porter bundle? (y/N) " -n 1 -r
+echo
+if [[ ! $REPLY =~ ^[Yy]$ ]]
+then
+    exit 0
+fi
+
+porter uninstall -p ./parameters.json \
+  -c ./debug-aad-auth.json \
+  -c /workspaces/AzureTRE/resource_processor/vmss_porter/azure.json
```
marrobi, line 27: Is this specific to workspaces or could it be for any resource, and hence stored with scripts and called via the make file? We already have a `make porter-install` command.

templates/workspaces/base/porter.yaml
```diff
@@ -1,10 +1,18 @@
...
 
 credentials:
+  # Credentials for interacting with the AAD Auth tenant
+  - name: auth_client_id
+    env: AUTH_CLIENT_ID
+  - name: auth_client_secret
+    env: AUTH_CLIENT_SECRET
+  - name: auth_tenant_id
+    env: AUTH_TENANT_ID
+  # Credentials for interacting with Azure
```
marrobi, line 15: If these values are not present can we still deploy?

templates/workspaces/base/porter.yaml
```diff
@@ -53,6 +60,25 @@ parameters:
   - name: register_aad_application
     type: boolean
     default: false
+
+outputs:
+  - name: WorkspaceOwner
```
marrobi, line 65: WorkspaceOwner what?

templates/workspaces/base/porter.yaml
```diff
@@ -53,6 +60,25 @@ parameters:
...
+    type: string
+    applyTo:
+      - install
+  - name: WorkspaceResearcher
```
marrobi, line 69: As above.

templates/workspaces/base/terraform/aad/aad.tf
```diff
@@ -0,0 +1,114 @@
...
+      id   = "37f7f235-527c-4136-accd-4a02d197296e" # Openid
+      type = "Scope"                                # Delegated
+    }
+    resource_access {
+      id   = "14dad69e-099b-42c9-810b-d002981feec1" # Profile
+      type = "Scope"                                # Delegated
+    }
+  }
+
+  # Add this when Guacamole is added
```
marrobi, line 79: Not sure this needs to be in this PR?

templates/workspaces/base/terraform/outputs.tf
```diff
@@ -1,3 +1,19 @@
 output "workspace_resource_name_suffix" {
   value = local.workspace_resource_name_suffix
-}
\ No newline at end of file
+}
+
+output "WorkspaceOwner" {
```
marrobi, line 5: Can we be more explicit? WorkspaceOwnerAppRoleId?

templates/workspaces/base/terraform/outputs.tf
```diff
@@ -1,3 +1,19 @@
...
+  value = var.register_aad_application ? module.aad[0].app_role_workspace_owner_id : ""
+}
+
+output "WorkspaceResearcher" {
```
marrobi, line 9: As above.

marrobi (commented):
A quick look few, couple of comments/questions, happy to talk through if easier.

templates/core/terraform/main.tf
ross-p-smith: So I copied it from core which is actually false, and then had problems so added it to true in a previous PR and then am aligning it to core. What do you think?

marrobi: I actually think we should have it to true - have you had better luck with it set to true?

templates/workspaces/base/terraform/outputs.tf
ross-p-smith: Sure - this is because the json stored in cosmos is just "WorkspaceOwner". I'm making a breaking change here anyway, so I'll be more explicit.

templates/core/terraform/main.tf
stuartleeks: IIRC, this used to be true and @damoodamoo set to `false` when looking into build issues.
I think that the issue with recovery is that when you recover a KeyVault, secrets etc also get brought back with it. So then Terraform fails because secrets already exist and need importing. So having as `false` and failing early ended up a better experience.

I _think_ that the underlying issues here were around the PR build destroy step failing to purge the KV, but I think that's all working now?

ross-p-smith: Both set to false

templates/workspaces/base/debug-porter.sh
ross-p-smith: Agree - I've reworked this to use the `make bundle-install` - I didn't know they were there!! Thanks - retested using these and works

templates/workspaces/base/porter.yaml
ross-p-smith: These values are passed in by the RP and locally they are passed in by `make bundle-install`

marrobi: Will they always exist? Even if people don't want to automatically create app reg? I guess they will, just might not have the correct permissions.

marrobi (approved):
LGTM providing doesn't break any of the existing functionality.

ross-p-smith:
/test-force-approve

image

PR #1702, opened 2022-04-15
1. Add API creds into KeyVault so that RP can read it

ross-p-smith:
# PR for issue #220

## What is being addressed

This PR places the credentials for the API into the core KeyVault so that the Resource Provider can read them so that it can communicate with the Auth tenant. This is part 1 of multiple PR's

## How is this addressed

- Add 3 new secrets into KeyVault
- Retrofitted the api-app app config so that it uses these key vault secrets rather than clear text values.
- Add the keyvault name as an environment variable to the VMSS instance.
- Utilise a porter plugin that automatically reads from KeyVault (from the name above) 
- Use a new credential set in Porter to retrieve the necessary secrets from the Vault
- #1701 will continue this work


tamirkamara:
Why does the RP need the API app credentials?

ross-p-smith:
> Why does the RP need the API app credentials?

So that when `auto_app_registration` is set to true, it is able to create the App Registration for the workspace.

resource_processor/resources/commands.py
```diff
@@ -59,7 +59,9 @@ async def build_porter_command(config, logger, msg_body, custom_action=False):
                     f"{'invoke --action ' if custom_action else ''}"
                     f"{msg_body['action']} \"{installation_id}\" "
                     f" --reference {config['registry_server']}/{msg_body['name']}:v{msg_body['version']}"
-                    f" {porter_parameters} --cred ./vmss_porter/azure.json --allow-docker-host-access --force"
+                    f" {porter_parameters} --allow-docker-host-access --force"
+                    f" --cred ./vmss_porter/azure.json"
+                    f" --cred ./vmss_porter/api-creds.json"
```
tamirkamara, line 64: workspace_creation... instead of api-creds?

resource_processor/run.sh
```diff
@@ -9,8 +9,14 @@ if [[ -z "${MGMT_STORAGE_ACCOUNT_NAME}" ]]; then
...
 fi
 
+if [[ -z "${KEY_VAULT_NAME}" ]]; then
+  >&2 echo "Environment variable for Key Vault name missing"
+fi
+
+# Documentation here: - https://github.com/vdice/porter-bundles/tree/master/azure-keyvault
 cat > /root/.porter/config.toml << EOF
 default-storage = "azurestorage"
+default-secrets = "api"
```
tamirkamara, line 19: still need to be api?

resource_processor/run.sh
```diff
@@ -20,6 +26,13 @@ plugin = "azure.table"
 [storage.config]
 account="${MGMT_STORAGE_ACCOUNT_NAME}"
 resource-group="${MGMT_RESOURCE_GROUP_NAME}"
+
+[[secrets]]
+name = "api"
```
tamirkamara, line 31: still need to be api?

resource_processor/vmss_porter/api-creds.json
```diff
@@ -0,0 +1,26 @@
+{
```
tamirkamara, line 1: Should the file be renamed?

tamirkamara (approved):

ross-p-smith:
/test
(manually deleted environment)

ross-p-smith:
Merged this branch into the later branch and the tests pass on that branch.
/test-force-approve

ross-p-smith:
/test-force-approve

(Merged this branch into another and ran tests there)

PR #1704, opened 2022-04-19
2. Allow API App to have Application.ReadWrite.All

ross-p-smith:
# PR for issue #220 

## What is being addressed

In order to allow Workspaces to be automatically registered within AAD, we need an application that has the necessary AAD permissions to be able to create new AAD Applications and this is Application.ReadWrite.All. This permission is very high and some organisations may not permit this, so have this disallowed by default. If you set AUTO_WORKSPACE_APP_REGISTRATION to true, then when you register your AAD API Application, it has the correct permissions.

## How is this addressed

- Use AUTO_WORKSPACE_APP_REGISTRATION (default false) to turn on this behaviour.
- Re-run your `make auth`

## Other approaches
image
I did try only granting `Application.ReadWrite.OwnedBy` and Terraform fails with the message above


ross-p-smith:
/test-force-approve

There are no tests for the areas of code I have changed.

docs/tre-admins/auth.md
```diff
@@ -9,7 +9,8 @@ The following values are needed to be in place to run the script. (`/templates/c
 | Key | Description |
 | ----------- | ----------- |
 |TRE_ID|This is needed to build up the redirect URI for the Swagger App|
-|AAD_TENANT_ID|The tenant id of where your AAD identities will be placed. This can be different to the tenant where your Azure resources are created.
+|AAD_TENANT_ID|The tenant id of where your AAD identities will be placed. This can be different to the tenant where your Azure resources are created.|
```
marrobi, line 12: Can we use the work AUTH in here somewhere to distinguish between the ARM tenant?

docs/tre-admins/auth.md
```diff
@@ -9,7 +9,8 @@ The following values are needed to be in place to run the script. (`/templates/c
...
+|AUTO_WORKSPACE_APP_REGISTRATION| Default of `false`. Setting this to true grants the `Application.ReadWrite.All` permission to the API AAD Application. This allows it to create other AAD applications, e.g. Workspaces.
```
marrobi, line 13: IT doesn't actually grant it does it? Does it enable the script to grant it or the TF to use it?

marrobi (approved):
Couple of comments, but looks good.

docs/tre-admins/auth.md
ross-p-smith: Just to get this right, we used to have duplicates and have AAD_TENANT_ID and AUTH_TENANT_ID using the same value - now you want AAD_AUTH_TENANT_ID?? This isn't a change I'm making in this PR as it needs many files updating. I think that needs to be a separate story - a valid one I'll grant you!

docs/tre-admins/auth.md
ross-p-smith: This value is a top level flag that will be used by multiple pieces so that this feature can work. In this PR, the flag will be checked by the aad-app-reg.sh script and just enable a permission on the aad app. (as the text says)

It doesn't create workspace apps, it just says "the api app is able to create workspace apps" when required.

docs/tre-admins/auth.md
ross-p-smith: Created #1715

PR #1712, opened 2022-04-20
Rename auth file

ross-p-smith:
## What is being addressed

Renaming the credential file that porter will use to authenticate to AAD

Also standardised on auth_client_* names as per openid


templates/core/terraform/api-webapp.tf
```diff
@@ -49,8 +49,8 @@ resource "azurerm_app_service" "api" {
     "RESOURCE_LOCATION"                          = azurerm_resource_group.core.location
     "SWAGGER_UI_CLIENT_ID"                       = var.swagger_ui_client_id
     "AAD_TENANT_ID"                              = "@Microsoft.KeyVault(SecretUri=${azurerm_key_vault_secret.auth_tenant_id.id})"
-    "API_CLIENT_ID"                              = "@Microsoft.KeyVault(SecretUri=${azurerm_key_vault_secret.api_app_id.id})"
-    "API_CLIENT_SECRET"                          = "@Microsoft.KeyVault(SecretUri=${azurerm_key_vault_secret.api_app_secret.id})"
+    "API_CLIENT_ID"                              = "@Microsoft.KeyVault(SecretUri=${azurerm_key_vault_secret.api_client_id.id})"
+    "API_CLIENT_SECRET"                          = "@Microsoft.KeyVault(SecretUri=${azurerm_key_vault_secret.api_client_secret.id})"
```
marrobi, line 53: Should these be auth_ ?

ross-p-smith: This would mean a whole heap of changes in the API python code (which would need retesting - these have always been called that)

marrobi (approved):
LGTM thanks for creating the additional issue.

PR #1713, opened 2022-04-20
Add GitHub script to capture PR info

stuartleeks:
- Capture info and dump to compare with existing method of generation
- Update tests to create new mock per test for more reliable assertions on calls to mocks (i.e. avoid interference across tests)

Part of #1538 

Future PR(s) will update the pipeline to start using the outputs from the GitHub script and remove some of the bash steps in the workflow.

### DIFF
--- a/.github/scripts/build.js
+++ b/.github/scripts/build.js
@@ -15,25 +15,25 @@ async function getCommandFromComment({ core, context, github }) {
 
   // Determine PR SHA etc
   const prNumber = context.payload.issue.number;
-	const pr = (await github.rest.pulls.get({ owner: repoOwner, repo: repoName, pull_number: prNumber })).data;
-	console.log("==========================================================================================");
-	console.log(pr);
-	console.log("==========================================================================================");
+  const pr = (await github.rest.pulls.get({ owner: repoOwner, repo: repoName, pull_number: prNumber })).data;
+  console.log("==========================================================================================");
+  console.log(pr);
+  console.log("==========================================================================================");
 
-	const prRefId = getRefIdForPr(prNumber);
-	console.log(`prRefId: ${prRefId}`);
-	core.setOutput("prRefId", prRefId);
+  const prRefId = getRefIdForPr(prNumber);
+  console.log(`prRefId: ${prRefId}`);
+  core.setOutput("prRefId", prRefId);
 
-	console.log(`Using head ref: ${pr.head.ref}`)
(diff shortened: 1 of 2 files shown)


.github/scripts/build.js
```diff
@@ -11,6 +13,29 @@ async function getCommandFromComment({ context, github }) {
     return "none";
   }
 
+  // Determine PR SHA etc
+  const prNumber = context.payload.issue.number;
+	const pr = (await github.rest.pulls.get({ owner: repoOwner, repo: repoName, pull_number: prNumber })).data;
```
martinpeck: what's the reason to indent these lines like this?

martinpeck: is this a "tabs vs spaces" issue?

.github/scripts/build.js
```diff
@@ -81,6 +106,19 @@ async function userHasWriteAccessToRepo({ github }, username, repoOwner, repoNam
   return userHasWriteAccess
 }
 
+function getRefIdForPr(prNumber) {
+	// Trailing newline is for compatibility with previous bash SHA calculation
+	return createShortHash(`refs/pull/${prNumber}/merge\n`);
+}
+function getRefIdForBranch(branchName) {
+	// Trailing newline is for compatibility with previous bash SHA calculation
```
martinpeck: where are these functions used other than above (in new code)?

Wouldn't it be better to add the `\n` at the point where it's printed, rather than in this function? I makes this function less useful in the future

martinpeck (approved):
LGTM....just one comment on tabs, and one question

.github/scripts/build.js
stuartleeks: I just pushed a commit that standardises the indentation, which I think addresses this

.github/scripts/build.js
stuartleeks: These functions are used to generate the ID for the deployment resource group etc. In the current workflows they are generated via bash (`REFID=$(echo "${github_pr_ref}" | shasum | cut -c1-8)`).

These IDs are currently calculated based on the PR/branch and need to be consistent across the environment creation and destruction, so the newline is included to ensure consistency with the bash script approach so that when I add changes to use the JavaScript-generated IDs for creation, the destroy step still finds the environment

martinpeck: Ah, right. Makes sense.

stuartleeks:
/test-force-approve

(changes not covered by `/test`)

PR #1714, opened 2022-04-20
Upgrade workspace AzureRM provider

ross-p-smith:
## What is being addressed

The workspace was using version 2.9 of the AzureRM Terraform provider. This upgrade moves to 3.1 which was part of my wider auto-aad-workpsace work, but have broken it out into smaller chunks.


stuartleeks (approved):
LGTM

Issue #1703, opened 2022-04-19
Shared service object is created multiple times in Cosmos


tanya-borisova:
**Describe the bug**
Shared service is created multiple time in Cosmos when deploying shared service (e.g. Firewall) multiple times.
Since shared service is a singleton, only one active object should be present at one time.

**Steps to reproduce**

1. Run `make firewall-install` several times
1. See several objects in Cosmos created with `isActive` set to `true`

Issue #141, opened 2021-05-31
Update existing TRE and Azure resources


TessFerrandez:
As an Azure Administrator 
I would like to be able to update the deployed infrastructure and code within an Azure TRE instance, either manually or using CI/CD.
So that I can keep the TRE infrastructure and code up to date.
 
As a TRE adminstrator
I want to update workspaces
So that I can rename them

As a workspace owner
I want to update workspace services
So that I can make changes such as increase the amount of compute or change security settings

As a researcher
I want to update user resources
So that I can resize VMs

## Acceptance Criteria
- [x] Explore and prepare proposal for updates that can be done on a TRE instance.
- [x] #1744
- [x] Each template schema needs define which properties can (or cannot) be updated. These are returned as readonly properties by the API on get template (for an update operation) and validated by the API on PUT/PATCH that they haven't changed.
- [x] On workspace update the bundles updates with new parameters via an update message - this mechanism could also be used to retry a failed bundle deployment.

## Tasks
- [x] #1240

marrobi:
@damoodamoo Most of this is done I believe? Can we update the template version of a resource and hence bundle?

damoodamoo:
Yes 3/4 of the tasks above are done as part of #1240 . You currently cannot update the template of a resource - you can update the properties and the `isEnabled` flag.

tamirkamara:
@marrobi I think we can close this as complete. do you agree?

marrobi:
Happy to close it, we need to make sure we reinstate the workspace upgrade flow.

PR #1717, opened 2022-04-20
Fix shared service object created multiple times in cosmos

tanya-borisova:
# PR for issue https://github.com/microsoft/AzureTRE/issues/1703

## What is being addressed

Currently, there is a bug when a shared service object is created multiple times in Cosmos (all these objects are "active", i.e. isActive is set to true).
Only one shared service of a given template type can exist at the time. This PR implements this constraint.

## How is this addressed

- Fixed API and database repository for shared services to return 409 Conflict when there already is an active shared service of a given name.
- DB migration script that deletes all duplicated items

## NOTE

After this PR goes in, everybody who has TRE deployed *needs* to run the DB migration script, otherwise consecutive operations to shared services will fail.

The script can be run as follows:

```
STATE_STORE_ENDPOINT=https://cosmos-${TRE_ID}.documents.azure.com:443/ STATE_STORE_KEY=your_secret_here COSMOSDB_DATABASE_NAME=AzureTRE python3 scripts/db_migrations.py
```


api_app/db/repositories/shared_services.py
```diff
@@ -26,6 +27,10 @@ def shared_service_query(shared_service_id: str):
     def active_shared_services_query():
         return f'SELECT * FROM c WHERE {IS_ACTIVE_CLAUSE} AND c.resourceType = "{ResourceType.SharedService}"'
 
+    @staticmethod
+    def active_shared_service_with_template_name_query(template_name: str):
+        return f'SELECT * FROM c WHERE {IS_ACTIVE_CLAUSE} AND c.resourceType = "{ResourceType.SharedService}" AND c.templateName = "{template_name}"'
```
ross-p-smith, line 32: Do we need the template version in here too?

ross-p-smith (approved):
Looks great, just one question about whether we need to check the version too?

api_app/db/repositories/shared_services.py
tanya-borisova: This is a great question that I should've clarified explicitly :) 

If we had a template version here, that would mean that we allow several active shared services of the same resource type, but of different version.

Currently, I think we don't want to allow that (cc @damoodamoo  @marrobi), and instead we want to implement an upgrade action, which would upgrade an already existing shared service to a new version (see https://github.com/microsoft/AzureTRE/issues/141). 

So, currently, once a shared service is deployed, it cannot then be upgraded to a new version. This limitation will be solved as part of that issue I linked above

tanya-borisova:
I think https://github.com/microsoft/AzureTRE/pull/1726 should go in first, and I would propose to add running db-migrations as part of `make all`. Otherwise, as it currently stands, if people don't run migrations first, then consecutive deployments of shared services including firewall will fail with 500 Internal server error (as the code checks for invariant that only one shared service is active at a time).

Issue #1611, opened 2022-03-28
Tech debt: Refactor register_bundle_with_api.sh


tanya-borisova:
## Description

The script register_bundle_with_api.sh does several things at once and is becoming unwieldy. 
In this task, separate into own scripts:
- getting a token 
- registering a template
-  deploying of a service (done as part of https://github.com/microsoft/AzureTRE/pull/1597 )

PR #1718, opened 2022-04-20
Split register_bundle_with_api.sh into three scripts

tanya-borisova:
# PR for issue https://github.com/microsoft/AzureTRE/issues/1611

## What is being addressed

`register_bundle_with_api.sh` became unwieldy during some recent changes. Split this script into three: 
- getting an access token
- registering a bundle
- deploying a shared service 

## How is this addressed

- Split the script into three
- Updated the script to deploy a shared service to prepare it implementation of https://github.com/microsoft/AzureTRE/issues/141 (upgrade of shared services)
- Updated Makefile to reflect changes
- Updated deploy_tre_reusable Github workflow to register shared services bundles together with the rest of the bundles, and deploy shared services before the e2e tests


.github/workflows/deploy_tre_reusable.yml
```diff
@@ -450,18 +456,51 @@ jobs:
...
-          API_CLIENT_ID: "${{ secrets.API_CLIENT_ID }}"
-          AAD_TENANT_ID: "${{ secrets.AAD_TENANT_ID }}"
-          TEST_APP_ID: "${{ secrets.TEST_APP_ID }}"
-          TEST_ACCOUNT_CLIENT_ID: "${{ secrets.TEST_ACCOUNT_CLIENT_ID }}"
-          TEST_ACCOUNT_CLIENT_SECRET: "${{ secrets.TEST_ACCOUNT_CLIENT_SECRET }}"
-          TRE_ID: "${{ secrets.TRE_ID }}"
-          LOCATION: "${{ secrets.LOCATION }}"
+          CI_CACHE_ACR_NAME: ${{ secrets.CI_CACHE_ACR_NAME}}
 
-      - name: Register/deploy nexus
```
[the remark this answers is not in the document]
marrobi: Do the docs need updating to include the install steps as part of this PR?

damoodamoo (approved):

.github/workflows/deploy_tre_reusable.yml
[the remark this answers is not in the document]
tanya-borisova: @marrobi  good point, I will create a task to update shared service docs 👍

Makefile
```diff
@@ -291,18 +294,19 @@ bundle-register:
 	&& . ./devops/scripts/load_env.sh ./devops/.env \
 	&& . ./devops/scripts/load_env.sh ./templates/core/.env \
 	&& az acr login --name $${ACR_NAME}	\
+	&& . ./devops/scripts/get_access_token.sh \
 	&& cd ${DIR} \
 	&& ${ROOTPATH}/devops/scripts/register_bundle_with_api.sh --acr-name "$${ACR_NAME}" --bundle-type "$${BUNDLE_TYPE}" --current --insecure --tre_url "$${TRE_URL:-https://$${TRE_ID}.$${LOCATION}.cloudapp.azure.com}" --verify --workspace-service-name "$${WORKSPACE_SERVICE_NAME}"
 
-shared-service-register-and-deploy:
+deploy-shared-service:
```
stuartleeks, line 301: Is this a make target that consumers of AzureTRE are likely to use in their CI/CD?

If so, changing this will require them to update. (Will be less of an issue once we have releases where we can add breaking changes comments as part of the release notes)

tanya-borisova: @stuartleeks  to be honest, I don't know how likely they are going to be using it in their CI/CD. This target has been introduced on 6 of April, so if the customers' CI/CD pipelines have been created after that, then it's likely to be present there. 

Would it be better if I left the name of the target to be the same, just to be on a safe side?

stuartleeks: Possibly one for @marrobi ?

tanya-borisova: @marrobi  do you have an opinion on this? Have any customers set up CI in the past few weeks that you know of?

tanya-borisova: Talked offline and @marrobi  said he's not too worried so I'm going to go ahead and merge this

tanya-borisova:
/test-force-approve

Successful run: https://github.com/microsoft/AzureTRE/actions/runs/2218742300

Makefile
martinpeck, line 301: Would it be better if we...

- leave the `shared-service-register-and-deploy` rule in place
- have it call the new rule
- have it output a message saying "this rule is deprecated...please use deploy-shared-service instead"

... and then raise an issue to remove this rule later? It allows us to have a gap between people pulling this change, and then fixing their AzDO builds

PR #1719, opened 2022-04-20
E2E tests for shared services

tanya-borisova:
# PR for issue

## What is being addressed

Added E2E tests for shared services.
Moved deploying Nexus and Gitea into extended tests.

## How is this addressed

- Move deploying Nexus and Gitea into their own extended tests
- Some minor refactoring of the existing tests: moved test_get_workspace_service_templates into the corresponding file
- Renamed test files for (hopefully) more clarity
- Other fixes related to making the test work (returning Operation object in PATCH and a correct status code, fixes to Gitea shared service, fix to permissions on resource processor keyvault role)
- Added a Makefile target `prepare-for-e2e` that registers templates required for E2E tests to run in the environment


tanya-borisova:
Successful test run: https://github.com/microsoft/AzureTRE/actions/runs/2220688355

e2e_tests/helpers.py
```diff
@@ -83,6 +75,24 @@ async def post_resource(payload, endpoint, resource_type, token, admin_token, ve
         return resource_path, resource_id
 
 
+async def get_shared_service_id_by_name(template_name: str, verify, token):
+    async with AsyncClient(verify=verify, timeout=30.0) as client:
```
damoodamoo, line 79: use the same `timeout` value here as the others?

damoodamoo (approved):
looks good - some discussions offline to answer questions too, thanks! :)

e2e_tests/helpers.py
tanya-borisova: Good catch, missed that!

PR #1721, opened 2022-04-21
PR bot updates

stuartleeks:
Continuing work for #1538 

- Add tests to validate outputs are set correctly
- Update workflow to use outputs from GitHub script step
- Move help output to GitHub script and update command handling to treat all commands starting with `/` as commands and output message via comment if command isn't recognised
- Use core for logging. This gives more options for controlling the output, and cleans up the test run output.
- Simplify test context setup
- Remove redundant `az login` step
- Disable GitHub Actions linting due to `actionlint` bug (tracking re-enabling in #1723)
- Update build test script to install and run latest `actionlint` which has the above bug fixed (also provides a quicker way to run the workflow linting when working on the workflows)

### DIFF
--- a/.github/scripts/build.js
+++ b/.github/scripts/build.js
@@ -6,16 +6,21 @@ async function getCommandFromComment({ core, context, github }) {
   const repoParts = repoFullName.split("/");
   const repoOwner = repoParts[0];
   const repoName = repoParts[1];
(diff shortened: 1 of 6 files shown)


stuartleeks:
superceded by #1727

PR #1733, opened 2022-04-25
Fix makefile: Add missing back slashes

tanya-borisova:
# PR for issue

## What is being addressed

Fixing a bug in Makefile where \ is missing for a multi line command


stuartleeks (approved):
lgtm

PR #1738, opened 2022-04-26
Remove outputs from base workspace

ross-p-smith:
## What is being addressed
There is a bug in the porter installation of the base workspace stating that outputs are not present.

## How is this addressed

- remove outputs for now


templates/core/terraform/.terraform.lock.hcl
```diff
@@ -58,21 +58,21 @@ provider "registry.terraform.io/hashicorp/local" {
 }
 
 provider "registry.terraform.io/hashicorp/random" {
-  version = "3.1.2"
```
tanya-borisova, line 61: Does this need to change?

tanya-borisova (approved):
Just one question

templates/core/terraform/.terraform.lock.hcl
ross-p-smith: Removed - must have been updated by mistake - thanks! :-)

PR #1747, opened 2022-04-27
Add diagnostic settings for core resources

### LINKED ISSUE #1746: Add diagnostic settings for core resources
## Description

As a TRE Administrator / Developer
I want to be able to diagnose issues with different Azure resources comprising the core TRE solution
So that I can solve them

* This is also a recommended security practice

## Acceptance criteria

- Application gateway has diagnostic settings defined
- Service bus has diagnostic settings defined
- Keyvault has diagnostic settings defined

tamirkamara:
Fixes #1746 

## What is being addressed

When something gets wrong in some azure resources it's hard to figure out the problem without logs. I came across this with Application Gateway, but we have other resources as well that don't save logs.
It is also valuable information for security audits and trackback scenarios. 

## How is this addressed

- Add diagnostic settings definition for resource in the core tre infra: application gateway, service bus, keyvault
- Small: remove appgw ip data object as it's not really needed for static IPs.

### DIFF
--- a/templates/core/terraform/appgateway/appgateway.tf
+++ b/templates/core/terraform/appgateway/appgateway.tf
@@ -195,9 +195,9 @@ data "azurerm_log_analytics_workspace" "tre" {
 }
 
 resource "azurerm_monitor_diagnostic_setting" "agw" {
-  name                           = "diagnostics-agw-${var.tre_id}"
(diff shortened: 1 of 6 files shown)


ross-p-smith (approved):
One step closer to logs!

PR #1755, opened 2022-04-28
Add Guacamole redirection index page and update connection URIs

### LINKED ISSUE #1724: On first connection to a VM often get sent to the Guacamole connection listings, rather than a specific VM
Could this be as the connection list is only loaded on a visit to /guacamole, rather than to the VM. So if connect to the VM directly is no list?

When is the VM list populated - at login or visit to /guacamole? Has there been a change to ensure the VM list is refreshed?

marrobi:
# Closes #1724 

## What is being addressed

Added index page to redirect to Guacamole without the # which gets lost as part of the OIDC login flow. This means the TRE can link to this page rather than guacamole to connect VMs.


anatbal (approved):
LGTM

eladiw (approved):

PR #1757, opened 2022-04-28
Only Migrate database if ENABLE_LOCAL_DEBUGGING is true

### LINKED ISSUE #1759: make all/tre-deploy fails due to db-migrate
**Describe the bug**
Running make `all` (or `tre-deploy` for that matter) fails because it runs `db-migrate` that needs direct access to cosmos.

**Steps to reproduce**

1. Make sure you don't have `ENABLE_LOCAL_DEBUGGING=true` in your .env files
2. Deploy a new environment by running `make all` or `make tre-deploy`
3. See that the activity fail due to lack of access to cosmos

ross-p-smith:
# Fixes #1759 

## What is being addressed

Only migrate the database if you are able to connect to it
## How is this addressed

- Check the ENABLE_LOCAL_DEBUGGING flag and only execute the migrations if this is set.
- This is so local `make all` works. Yet to find a way to execute this from within CI/CD


tamirkamara (dismissed):
This can't be running in CI as it stands because:
1. The CI doesn't have access to cosmos
2. Some other error as shown here: https://github.com/microsoft/AzureTRE/runs/6244187047?check_suite_focus=true
3. Also connected to #1759 

BTW, using the bot to test a workflow change isn't useful as it always run from the main branch

damoodamoo (approved):

Issue #1241, opened 2022-02-03
Guacamole: Connection to user resource fails if allocated IP address of old resource which has failed to delete


oliver7598:
As mentioned in issue #1201 and #1163 User Resources currently fail to delete due to Key Vault purge protect.
This happens after the Virtual Machine is deleted and the IP used is freed.
If a new resource is then created and the old IP allocated, the connection uri created will match that of the previous resource.
On attempting to connect to the virtual machine a connection error is presented.

**Steps to reproduce**
1. Create a user resource in a Guacamole service
2. Delete the user resource
3. Wait for the delete to fail so the IP is free
4. Create a new user resource
5. Wait for it to be ready to use
6. Go to connection uri provided
7. Connection error

NOTE:  I have only tested this on a Ubuntu deployment.  Unsure if the same would happen for a Windows VM although I would expect it to.

marrobi:
I think the issue here is IP addresses are being used for connection name, we should be using the VM name which should always be unique for the workspace service.

marrobi:
I think his will be fixed if #1729 is completed.

PR #1758, opened 2022-04-30
Change VM name to display name and identifier to hostname

### LINKED ISSUE #1729: VM name should be the name, not the ip address

marrobi:
#  Closes #1729 #1241 

## What is being addressed

The IP address was used for the connection identifier, which is reused if a VM is deleted and then created. The connection name was the IP address which is confusing for users.

## How is this addressed

- Change connection name to display name
- Change connection identifier to VM hostname
- Connection still made using IP address


templates/workspace_services/guacamole/guacamole-server/guacamole-auth-azure/src/main/java/org/apache/guacamole/auth/azuretre/connection/ConnectionService.java
```diff
@@ -71,9 +71,10 @@ private static Map<String, GuacamoleConfiguration> getConfigurations(final Azure
                     if (templateParameters.has("hostname") && templateParameters.has("ip")) {
                         final String azureResourceId = templateParameters.getString("hostname");
                         final String ip = templateParameters.getString("ip");
-                        setConfig(config, azureResourceId, ip);
+                        final String display_name = templateParameters.getString("display_name");
+                        setConfig(config, azureResourceId, ip, display_name);
                         LOGGER.info("Adding a VM: {}", ip);
```
eladiw, line 76: you can also add the display name to the log line

templates/workspace_services/guacamole/user_resources/guacamole-dev-vm/terraform/main.tf
```diff
@@ -60,5 +60,5 @@ output "azure_resource_id" {
 }
 
 output "connection_uri" {
-  value = "https://${data.azurerm_app_service.guacamole.default_site_hostname}/?/client/${textencodebase64("${azurerm_network_interface.internal.private_ip_address}\u0000c\u0000azuretre", "UTF-8")}"
+  value = "https://${data.azurerm_app_service.guacamole.default_site_hostname}/?/client/${textencodebase64("${azurerm_linux_virtual_machine.linuxvm.name}\u0000c\u0000azuretre", "UTF-8")}"
```
eladiw, line 63: Does this change means, that I can use EITHER the old connection uri and the new one?

I mean, with the ip or the vm name?
Or, to rephrase the question, will the change still be correct if you reverted the change in this file?

marrobi: No, as the ID of the connection is now the hostname to prevent duplicates. So not "backwards compatible". Is a breaking change. Thoughts?

templates/workspace_services/guacamole/guacamole-server/guacamole-auth-azure/src/main/java/org/apache/guacamole/auth/azuretre/connection/ConnectionService.java
marrobi: Done

templates/workspace_services/guacamole/user_resources/guacamole-dev-vm/terraform/main.tf
eladiw: I see you added the displayname property. so I'm not 100% sure would the ip  stop working...

eladiw (approved):

templates/workspace_services/guacamole/user_resources/guacamole-dev-vm/terraform/main.tf
marrobi: I added the dispay_name property to the configuration.

The connection however has a single identifier that I changed to be hostname.

PR #1763, opened 2022-05-03
Don't use Owner

### LINKED ISSUE #495: Can the scope of the VMSS MSI be reduced
**Description**
Currently this is subscription owner, can it be reduced to a custom role with lower privileges.

ross-p-smith:
## Fixes #495

The User assigned Identity does not need to be Owner, but rather "User Access Administrator".


ross-p-smith:
/test-force-approve

(Previously succeeded)

damoodamoo (approved):

PR #1764, opened 2022-05-03
Set firewall props via the API 

### LINKED ISSUE #1736: Change firewall bundle to read rules from resource properties
In order for the firewall to be updated by the API / other resources as part of pipelines, it must read the 'configurable' properties (parts of the firewall we want to allow a user/process to update) from the resource properties in cosmos.

damoodamoo:
# Closes #1736 

This PR adds the capability of creating dynamic firewall rule collections from `properties` in the resource object. Whilst the work is focussed on the firewall, it also demonstrates the model for any resource. To make this happen, some more fundamental changes were needed:
- `properties` in the resource can now support complex objects / arrays rather than basic types. When these are fed to porter, they are base64 encoded and passed in the cmd line. Terraform decodes the string, then decodes the json object into HCL which can be mapped to terraform properties.
- The `resource_template` model was tweaked (and some tests added) to support nested properties in the resource template, and ensure that nested properties were validated. 
- Updated porter to latest stable `0.38.11`

The next step will be to use this in the substitution mechanism, so dynamic values from the primary resource can be set in the payload, and items can be added/removed from lists.

To demonstrate - the following `PATCH` payload can now be sent to the firewall shared service, to create a rule collection:
```json
{
    "rule_collections": [
        {
            "name": "banana",
            "priority": 103,
            "action": "Allow",
            "rules": [
                {
                    "name": "banana rule 1",
                    "description": "desc here",
                    "protocols": [
                        {
                            "port": "6969",
                            "type": "Http"
                        }
                    ],
                    "target_fqdns": [
                        "one.two.three.com",
                        "two.three.com",
                        "three.two.com"
                    ],
                    "source_addresses":[
                        "172.196.0.0"
                    ]
                }
            ]
        }
    ]
}
```


api_app/models/domain/resource_template.py
```diff
@@ -23,6 +23,7 @@ class Property(AzureTREModel):
     pattern: Optional[str] = Field(None, title="Pattern")
     updateable: Optional[bool] = Field(None, title="Indicates that the field can be updated")
     readOnly: Optional[bool] = Field(None, title="Indicates the field is read-only")
+    items: Optional[dict] = None
```
tanya-borisova, line 26: Why is this a `None`, and not a `Field(None, title="...")`? 
Also, if it has to be `None`, could you add a comment indicating what this is (since other fields have a `title` field that serves this purpose)

tanya-borisova (approved):
LGTM! Just one small nit

damoodamoo:
/test-force-approve
extended tests have passed previously

Issue #563, opened 2021-08-06
Research and implement database migrations


TessFerrandez:
## Description

As a TRE API developer
I want to be able to write database migrations
So that I can update my database models without manually having to update all the documents in cosmos

## Acceptance critiera

- [x] Understand and document how to create atomic migrations for cosmos
- [ ] #1824
- [ ] Don't run migrations that have already been executed.
- [ ] Add db migration step to the build so that we can guarantee that they are run every time

ross-p-smith:
Fixed as #1726

PR #1776, opened 2022-05-04
Add Migrations to API

ross-p-smith:
# PR for issue #563

## What is being addressed

Need a mechanism to be able to upgrade documents in the Cosmos database so that breaking changes are handled. 

## How is this addressed

- [x] New route called migrate with a single (no body) post
- [x] Update Unit Tests

## Design
The migrate route within the API is designed to be called whenever you want to migrate the Cosmos database. All migrations are written in an idempotent manner; therefore you should be able to migrate the same database more than once. The end intention is that this will be called during the CI/CD of TRE. Future migration authors can simple add a method into the Xmigration class if a migration already exists for the resource type. If one does not exist, you can simply derive from an existing repository and add a new class.


api_app/db/migrations/shared_services.py
```diff
@@ -0,0 +1,16 @@
...
+    def __init__(self, client: CosmosClient):
+        super().__init__(client)
+
+    async def deleteDuplicatedSharedServices(self):
+        template_names = ['tre-shared-service-firewall', 'tre-shared-service-nexus', 'tre-shared-service-gitea']
+
+        for template_name in template_names:
+            for item in self.query(query=f'SELECT * FROM c WHERE c.resourceType = "shared-service" AND c.templateName = "{template_name}" \
+                                                                ORDER BY c.updatedWhen ASC OFFSET 1 LIMIT 10000', enable_cross_partition_query=True):
+                print(f"Deleting element {item}")
```
martinpeck, line 15: print where?
should this be logging instead?

api_app/db/migrations/workspaces.py
```diff
@@ -0,0 +1,28 @@
...
+
+class WorkspaceMigration(WorkspaceRepository):
+    def __init__(self, client: CosmosClient):
+        super().__init__(client)
+
+    async def moveAuthInformationToProperties(self):
+        for item in self.query(query=WorkspaceRepository.workspaces_query_string()):
+            template_version = semantic_version.Version(item["templateVersion"])
+            if (template_version > semantic_version.Version('0.3.0') and "authInformation" in item):
+                print(f'Found workspace {item["id"]} that needs migrating')
```
martinpeck, line 14: should this be a logging statement?

api_app/db/migrations/workspaces.py
```diff
@@ -0,0 +1,28 @@
...
+
+                # Copy authInformation into properties
+                item["properties"]["sp_id"] = item["authInformation"]["sp_id"]
+                item["properties"]["app_role_id_workspace_researcher"] = item["authInformation"]["roles"]["WorkspaceResearcher"]
+                item["properties"]["app_role_id_workspace_owner"] = item["authInformation"]["roles"]["WorkspaceOwner"]
+
+                # cleanup
+                del item["authInformation"]
+                self.update_item_dict(item)
+                print(f'Upgraded authentication info for workspace id {item["id"]}')
```
martinpeck, line 28: should this be a logging statement?

martinpeck:
I guess my comment is about the design, rather than the code (which looks fine).
The migrations API calls a set of idempotent functions that migrate things.

However, we always call all migrations, for each invocation of the API. We don't have any way to know that a given migration is "done". For example, we don't have a migrations collection in the DB that we write to when a given migration has completed.

From an efficiency point of view it might be better to log successful migrations and then not run them again. More importantly, I'm worried that we might get into a situation where a future migration breaks a historic migration. 

e.g Migration 1 can no longer run because Migration 101 renamed a collection, and now we can never run Migration 102 without an exception being thrown by the API

ross-p-smith:
> I guess my comment is about the design, rather than the code (which looks fine). The migrations API calls a set of idempotent functions that migrate things.
> 
> However, we always call all migrations, for each invocation of the API. We don't have any way to know that a given migration is "done". For example, we don't have a migrations collection in the DB that we write to when a given migration has completed.
> 
> From an efficiency point of view it might be better to log successful migrations and then not run them again. More importantly, I'm worried that we might get into a situation where a future migration breaks a historic migration.
> 
> e.g Migration 1 can no longer run because Migration 101 renamed a collection, and now we can never run Migration 102 without an exception being thrown by the API

@martinpeck - are you able to add those comments to #563?

martinpeck (approved):

Issue #1688, opened 2022-04-12
Resource Processor logs not appearing in Log Analytics


marrobi:
Resource processor logs should appear in Log Analytics. I believe this was under Traces. They are no longer there.

This related is likely a side effect of logging not working. #1628

tamirkamara:
A theory for now, but I think this is where logs are lost:
https://github.com/microsoft/AzureTRE/blob/main/resource_processor/vmss_porter/runner.py#L227

A new python Process doesn't share the same logging config done on the class.

From App Insights:
<img width="321" alt="image" src="https://user-images.githubusercontent.com/26870601/163937202-7e3a4fe2-1343-4ac4-b077-be4e4a59d3b5.png">

From the console:
<img width="582" alt="image" src="https://user-images.githubusercontent.com/26870601/163937385-64ee2394-14ae-43ef-9ff8-a86d592488cc.png">

Those "Looking for new session..." lines are generated from the new process context and are missing in the AI logs.

@damoodamoo WDYT?

damoodamoo:
yes makes sense to me - a bit surprising though as the other settings in the `initialize_logging` method are used by the sub processes - such as the output format. perhaps for AI we need to initialize the logging for each process?

tamirkamara:
> yes makes sense to me - a bit surprising though as the other settings in the `initialize_logging` method are used by the sub processes - such as the output format. perhaps for AI we need to initialize the logging for each process?

or pass in the handler. At any case, I'm not a python expert so there might be a more elegant way to do this...

martinpeck:
Currently the logging is set up at the top of `runner.py` when the script is loaded.

I feel we should wrap the logger init code into a function, and then call it from within `start_runner_process`.
We can also call it from the `__main__` code (i.e. bottom of the script) as well, to log the initial startup of the runner.

BTW...I wonder if we'll need to add the PID into the logger (if it's not already present) so that we can tell the difference between each processes logs.

PR #1784, opened 2022-05-06
Resource Processor: Configure logging handler per process

tanya-borisova:
# PR for issue https://github.com/microsoft/AzureTRE/issues/1688

## What is being addressed

Currently, logs from child processes of Resource Procesor runner aren't appearing in LogAnalytics. This PR fixes that 

## How is this addressed

- Refactor logging initialisation into a separate method
- Call method in each thread so that each process is configured with the`AzureLogHandler` handler separately 
- Pass logger_adapter into methods instead of having a method-level variable
- Only configure console logger in the parent thread

Screenshots:
image

image


martinpeck (approved):
Nice! LGTM!

ross-p-smith (approved):

PR #1788, opened 2022-05-09
E2E tests in parallel 

### LINKED ISSUE #1787: Extended tests take too long to run
The extended tests all run in sequence, and take > 50m, which means the task gets cancelled.

We should try to parallelize these tests.

damoodamoo:
# Fixes #1787 

Uses `pytest-xdist` to allow tests to be run in parallel rather than sequentially.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -88,7 +88,9 @@
             "-m",
             "extended",
             "--verify",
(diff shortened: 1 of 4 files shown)


damoodamoo:
will abandon this as we're splitting out the tests to a matrix instead of running using parallel workers

PR #1797, opened 2022-05-10
E2E work with scope_id from workspace properties

ross-p-smith:
## What is being addressed

E2E Extended Tests work with using the new scope_id.

image
In a normal App Reg, you can see that the Application ID URI is made up of the Client ID. I need to break this connection for when we automatically create AAD Apps from Terraform. This change is about reading the Application ID URI (called Identifier URI in the Graph) from AAD when we create the workspace, rather than assuming it will be "api://client_id".

## How is this addressed

- New python function to get full address
- Remove quotes around AAD creation that shell check enforce
- Retrieve the Scope Id from AAD when we create a workspace
- Use this new scope when we create a workspace token for the E2E tests (Rather than assuming it is the same as the client_id)

image


api_app/api/routes/api.py
```diff
@@ -129,15 +135,15 @@ async def get_openapi_json(workspace_id: str, request: Request, workspace_repo=D
 async def get_workspace_swagger(workspace_id, request: Request, workspace_repo=Depends(get_repository(WorkspaceRepository))):
 
     workspace = workspace_repo.get_workspace_by_id(workspace_id)
-    ws_app_reg_id = workspace.properties['client_id']
+    scopes = get_scope(workspace)
```
tanya-borisova, line 138: Nit: name this `scope` as the function returns only one?

api_app/services/aad_authentication.py
```diff
@@ -193,7 +193,7 @@ def _get_app_auth_info(self, client_id: str) -> dict:
             raise AuthConfigValidationError(f"{strings.ACCESS_UNABLE_TO_GET_INFO_FOR_APP} {client_id}")
 
         app_info = graph_data['value'][0]
-        authInfo = {'sp_id': app_info['id']}
+        authInfo = {'sp_id': app_info['id'], 'scope_id': app_info['servicePrincipalNames'][0]}
```
tanya-borisova, line 196: Is this the branch when we don't get scope_id from the bundle?

docs/tre-admins/auth.md
```diff
@@ -192,7 +192,9 @@ Same as [TRE API](#authentication-tre-api).
 | API/permission name | Type | Description | Admin consent required |
 | ------------------- | ---- | ----------- | ---------------------- |
 | Microsoft Graph/User.Read (`https://graph.microsoft.com/User.Read`) | Delegated | Allows users to sign-in to the app, and allows the app to read the profile of signed-in users. It also allows the app to read basic company information of signed-in users. | No |
-| Workspace API/user_impersonation (`api://<Workspace API Application (client) ID>/user_impersonation`) | Delegated* | Allows the app to access the workspace API on behalf of the user | No | Granted for *[directory name]* |
+| Workspace API/user_impersonation (`api://<Workspace Scope ID>/user_impersonation`) | Delegated* | Allows the app to access the workspace API on behalf of the user | No | Granted for *[directory name]* |
+
+When the Workpsace AAD app is registered by the aad-app-reg.sh, the `Workspace Scope Id` is the same as the Client Id.
```
tanya-borisova, line 197: Could you add a sentence saying something like "Otherwise, when the Workspace AAD app is registered by the bundle, the `Workspace Scope Id` is defined as ....."

e2e_tests/helpers.py
```diff
@@ -33,28 +33,31 @@ def get_auth_header(token: str) -> dict:
     return {'Authorization': f'Bearer {token}'}
 
 
+def get_full_endpoint(endpoint: str) -> str:
+    full_endpoint = ""
+    if (config.TRE_URL != ""):
+        full_endpoint = f"{config.TRE_URL}{endpoint}"
+    else:
+        full_endpoint = f"https://{config.TRE_ID}.{config.RESOURCE_LOCATION}.cloudapp.azure.com{endpoint}"
+    LOGGER.info(f'POSTING RESOURCE TO: {full_endpoint}')
```
tanya-borisova, line 42: Remove this log line? This method can be called not only for posting a resource. Then you won't need to have the variable `full_endpoint` at all

e2e_tests/helpers.py
```diff
@@ -33,28 +33,31 @@ def get_auth_header(token: str) -> dict:
     return {'Authorization': f'Bearer {token}'}
 
 
+def get_full_endpoint(endpoint: str) -> str:
+    full_endpoint = ""
+    if (config.TRE_URL != ""):
+        full_endpoint = f"{config.TRE_URL}{endpoint}"
```
tanya-borisova, line 39:
```suggestion
        return f"{config.TRE_URL}{endpoint}"
```

e2e_tests/helpers.py
```diff
@@ -33,28 +33,31 @@ def get_auth_header(token: str) -> dict:
...
+    else:
+        full_endpoint = f"https://{config.TRE_ID}.{config.RESOURCE_LOCATION}.cloudapp.azure.com{endpoint}"
```
tanya-borisova, line 41:
```suggestion
        return f"https://{config.TRE_ID}.{config.RESOURCE_LOCATION}.cloudapp.azure.com{endpoint}"
```

e2e_tests/helpers.py
```diff
@@ -201,3 +196,60 @@ async def check_deployment(client, operation_endpoint, headers):
...
+async def get_identifier_uri(client, workspace_id: str, auth_headers) -> str:
+    workspace = await get_workspace(client, workspace_id, auth_headers)
+
+    if ("properties" not in workspace):
+        raise Exception("Properties not found in workspace.")
+
+    if ("scope_id" not in workspace["properties"]):
+        raise Exception("Scope Id not found in workspace properties.")
+
+    # Cope with the fact that scope id can have api:// at the front.
```
tanya-borisova, line 222: Can use the same incantation as in the other file if you want: 

`return f"api://{workspace['properties']['scope_id'].lstrip('api://')}"`

e2e_tests/helpers.py
```diff
@@ -201,3 +196,60 @@ async def check_deployment(client, operation_endpoint, headers):
...
+        return f"api://{workspace['properties']['scope_id']}"
+
+
+async def get_workspace_owner_token(admin_token, workspace_id, verify) -> str:
+    async with AsyncClient(verify=verify) as client:
+        auth_headers = get_auth_header(admin_token)
+        scope_uri = await get_identifier_uri(client, workspace_id, auth_headers)
+
+        if config.TEST_ACCOUNT_CLIENT_ID != "" and config.TEST_ACCOUNT_CLIENT_SECRET != "":
+            # Use Client Credentials flow
```
tanya-borisova, line 235:
```suggestion
            # Logging in as Enterprise Application: use Client Credentials flow
```

e2e_tests/helpers.py
```diff
@@ -201,3 +196,60 @@ async def check_deployment(client, operation_endpoint, headers):
...
+            payload = f"grant_type=client_credentials&client_id={config.TEST_ACCOUNT_CLIENT_ID}&client_secret={config.TEST_ACCOUNT_CLIENT_SECRET}&scope={scope_uri}/.default"
+            url = f"https://login.microsoftonline.com/{config.AAD_TENANT_ID}/oauth2/v2.0/token"
+
+        else:
+            # Use Resource Owner Password Credentials flow
```
tanya-borisova, line 240:
```suggestion
            # Logging in as a User: use Resource Owner Password Credentials flow
```

e2e_tests/helpers.py
```diff
@@ -201,3 +196,60 @@ async def check_deployment(client, operation_endpoint, headers):
...
+        raise Exception("Scope Id not found in workspace properties.")
+
+    # Cope with the fact that scope id can have api:// at the front.
+    if workspace["properties"]["scope_id"].startswith("api://"):
+        return workspace["properties"]["scope_id"]
+    else:
+        return f"api://{workspace['properties']['scope_id']}"
+
+
+async def get_workspace_owner_token(admin_token, workspace_id, verify) -> str:
```
tanya-borisova, line 229: Nit: since you are returning either a `str` or `None` depending on HTTP status code, use `Optional[str]` as a return type (you need to do `from typing import Optional`)

scripts/create_aad_assets.sh
```diff
@@ -23,7 +23,7 @@ fi
 ./scripts/aad/aad-app-reg.sh \
   --name "${TRE_ID}" \
   --swaggerui-redirecturl "https://${TRE_ID}.${LOCATION}.cloudapp.azure.com/api/docs/oauth2-redirect" \
-  --admin-consent --automation-account "${api_app_can_create_other_applications}"
+  --admin-consent --automation-account ${api_app_can_create_other_applications}
```
tanya-borisova, line 26: Does the linter not complain on this? Do you need to ignore the linter explicitly here?

e2e_tests/test_workspace_services.py
```diff
@@ -47,8 +48,55 @@ async def test_create_guacamole_service_into_base_workspace(admin_token, workspa
         }
     }
 
-    await post_resource(patch_payload, f'/api{workspace_service_path}', 'workspace_service', workspace_owner_token, None, verify, method="PATCH")
+    await post_resource(patch_payload, f'/api{workspace_service_path}', workspace_owner_token, verify, method="PATCH")
```
tanya-borisova, line 51: Looks much nicer with simplified parameters!

e2e_tests/test_workspace_services.py
```diff
@@ -47,8 +48,55 @@ async def test_create_guacamole_service_into_base_workspace(admin_token, workspa
...
-    await disable_and_delete_resource(f'/api{workspace_service_path}', 'workspace_service', workspace_owner_token, None, verify)
+    await disable_and_delete_resource(f'/api{workspace_service_path}', workspace_owner_token, verify)
 
-    await disable_and_delete_resource(f'/api{workspace_path}', 'workspace', workspace_owner_token, admin_token, verify)
+    await disable_and_delete_resource(f'/api{workspace_path}', admin_token, verify)
+
+
+@pytest.mark.extended_aad
+@pytest.mark.timeout(3000)
+async def test_create_guacamole_service_into_aad_workspace(admin_token, workspace_owner_token, verify) -> None:
```
tanya-borisova, line 60: Nit: add a short comment explaining the main difference with a previous test?

templates/core/terraform/variables.tf
```diff
@@ -35,12 +35,12 @@ variable "api_image_repository" {
 
 variable "api_app_service_plan_sku_tier" {
   type    = string
-  default = "PremiumV3"
+  default = "PremiumV2"
```
tanya-borisova, line 38: As discussed, let's extract these changes out

scripts/db_migrations.py
```diff
@@ -92,19 +92,33 @@ def moveAuthInformationToProperties(self, resources_container_name):
...
+                    del item["properties"]["app_id"]
+                    updated = True
 
-                # merge authInformation into properties
-                item["properties"] = {**item["authInformation"], **item["properties"]}
-                del item["authInformation"]
-                resources_container.upsert_item(item)
-                print(f'Upgraded authentication info for workspace id {item["id"]}')
+                if "scope_id" not in item["properties"]:
+                    updated = True
```
tanya-borisova, line 105: Very nit: you put `updated = True` as a last statement in other conditions, could you do the same here for consistency

templates/workspaces/base/.env.sample
```diff
@@ -1,26 +1,13 @@
...
-export TF_VAR_register_aad_application="true"
-export TF_VAR_client_id=$WORKSPACE_APP_CLIENT_ID
-
-export TF_VAR_mgmt_resource_group_name=__CHANGE_ME__
-export TF_VAR_mgmt_storage_account_name=__CHANGE_ME__
-export TF_VAR_terraform_state_container_name=tfstate
-export TF_VAR_auth_tenant_id="__CHANGE_ME__"
-export TF_VAR_auth_client_id="__CHANGE_ME__"
-export TF_VAR_auth_client_secret="__CHANGE_ME__"
+TRE_RESOURCE_ID="__CHANGE_ME__"
```
tanya-borisova, line 1: Nice 👍

templates/workspaces/base/terraform/.terraform.lock.hcl
```diff
@@ -22,21 +22,41 @@ provider "registry.terraform.io/hashicorp/azuread" {
 }
```
tanya-borisova, line 22: I do wonder if we even need these lock files in the repo?

templates/workspaces/base/terraform/aad/outputs.tf
```diff
@@ -10,6 +10,10 @@ output "client_id" {
   value = azuread_application.workspace.application_id
 }
 
+output "scope_id" {
```
tanya-borisova, line 13: Nit: could you add a comment here saying something like "This is where the scope_id is defined in case of automatic Workspace App Registrations"?

templates/workspaces/base/terraform/variables.tf
```diff
@@ -32,7 +32,7 @@ variable "deploy_app_service_plan" {
 
 variable "app_service_plan_sku" {
   type        = string
-  default     = "P1v3"
+  default     = "P1v2"
```
tanya-borisova, line 35: Same as above, let's extract out

tanya-borisova (commented):
Looks good! Left some nits, mostly around comments

api_app/services/aad_authentication.py
ross-p-smith: Yes this flow is when we are retrieving the aad info rather than creating it

scripts/db_migrations.py
```diff
@@ -92,19 +92,33 @@ def moveAuthInformationToProperties(self, resources_container_name):
...
+                    item["properties"]["scope_id"] = item["properties"]["client_id"]
+
+                if "authInformation" in item:
+                    print(f'Upgrading authInformation in workspace {item["id"]}')
```
martinpeck, line 109: where is this printed?
is this script still run interactively?

templates/core/terraform/variables.tf
martinpeck, line 38: question: sure it's fine, but why are we changing this?

templates/workspaces/base/.env.sample
martinpeck, line 1: is this a general tidy up of this file?

if so, can we put some space between "stuff you need to change/define" and "stuff that has a default value", and maybe a comment above each?

templates/workspaces/base/terraform/keyvault.tf
```diff
@@ -3,7 +3,7 @@ resource "azurerm_key_vault" "kv" {
   location                 = azurerm_resource_group.ws.location
   resource_group_name      = azurerm_resource_group.ws.name
   sku_name                 = "standard"
-  purge_protection_enabled = var.keyvault_purge_protection_enabled
+  purge_protection_enabled = true
```
martinpeck, line 6: are we removing the ability to turn purge protection off?

templates/workspaces/base/terraform/aad/outputs.tf
ross-p-smith: I've added a comment in the main outputs.tf, not this one (as this whole file is conditional)

templates/workspaces/base/terraform/keyvault.tf
martinpeck: I assume we are...just wondered if we should continue to pass this as a variable, but default to `true`

ross-p-smith: With the latest provider - I think this should always be true and we soft delete and recover secrets at the provider level. There are too many cases where the dev cycles is different to the CI or Prod flow.

templates/core/terraform/variables.tf
ross-p-smith: Because of quota issues in our subscriptions

templates/workspaces/base/terraform/keyvault.tf
martinpeck: fair enough, but what is `KEYVAULT_PURGE_PROTECTION_ENABLED` in the sample env file? Should we remove that too?

scripts/db_migrations.py
ross-p-smith: This is run from the make file. There is a separate api version of this that logs

e2e_tests/helpers.py
```diff
@@ -201,3 +196,60 @@ async def check_deployment(client, operation_endpoint, headers):
...
+    workspace = await get_workspace(client, workspace_id, auth_headers)
+
+    if ("properties" not in workspace):
+        raise Exception("Properties not found in workspace.")
+
+    if ("scope_id" not in workspace["properties"]):
+        raise Exception("Scope Id not found in workspace properties.")
+
+    # Cope with the fact that scope id can have api:// at the front.
+    if workspace["properties"]["scope_id"].startswith("api://"):
```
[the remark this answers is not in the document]
ross-p-smith: This has been cleaned up

docs/tre-admins/auth.md
martinpeck: Also, there's a typo here (`Workpsace` should be `Workspace`)

scripts/db_migrations.py
```diff
@@ -92,19 +92,33 @@ def moveAuthInformationToProperties(self, resources_container_name):
...
+                    updated = True
 
-                # merge authInformation into properties
-                item["properties"] = {**item["authInformation"], **item["properties"]}
-                del item["authInformation"]
-                resources_container.upsert_item(item)
-                print(f'Upgraded authentication info for workspace id {item["id"]}')
+                if "scope_id" not in item["properties"]:
+                    updated = True
+                    item["properties"]["scope_id"] = item["properties"]["client_id"]
```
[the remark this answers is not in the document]
ross-p-smith: client_id is a mandatory workspace field. The check 4 lines above has migrated it from the only other name (app_id).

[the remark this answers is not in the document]
ross-p-smith: the updated=True has been moved

api_app/api/routes/api.py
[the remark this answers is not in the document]
martinpeck: Resolved this as it was an observation on existing code.

scripts/create_aad_assets.sh
martinpeck: The linter *may* complain, but those quotes generate errors in some situations (I hit this recently when deploying)

ross-p-smith: Looks like it doesn't fire on shellcheck - but have added a comment above

templates/workspaces/base/terraform/keyvault.tf
ross-p-smith: Removed all capabilities to disable purge protection

templates/core/terraform/variables.tf
ross-p-smith: I've put it back

api_app/services/aad_authentication.py
tanya-borisova: Leave a comment clarifying this?

scripts/db_migrations.py
ross-p-smith: I  have added #1824 to tidy this up

tanya-borisova (approved):

PR #1798, opened 2022-05-10
Gitea/Guacamole should be able to access AAD_TENANT_ID

### LINKED ISSUE #1795: Guacamole and Gitea fail because Cannot access AAD_TENANT_ID variable
Since we changed AAD_TENANT_ID app setting in the api webapp to be referenced from keyvault, we now cannot access local. aad_tenant_id in both guacamole and gitea, resulting in both services crash.

ross-p-smith:
# Fixes #1795 

## What is being addressed

Gitea and Guacamole should be able to access the AAD_TENANT_ID when building workspace services

## How is this addressed

- Add the AAD_TENANT_ID into the ws KeyVault
- Gitea and Guacamole should read this value from Keyvault and not the core api app settings


marrobi (approved):
As long as it works, LGTM.

marrobi (changes requested):
Are the deploy.sh files not still required?

ross-p-smith:
> Are the deploy.sh files not still required?

They are used for local debugging and I have replaced them with a file called debug.sh which is more explicit what it is used for and added it to the dockerignore file

marrobi (approved):
LGTM. Thanks.

anatbal (approved):
LGTM

PR #1826, opened 2022-05-12
azurerm_app_service_plan to azurerm_service_plan

ross-p-smith:
# PR for issue 1815

## What is being addressed

There is a quota issue at the moment using v3 App Services. This change migrates the azurerm_app_service_plan to azurerm_service_plan so that there is a single variable that you can change in order to bump it to v2
Describe the current behavior you are modifying.  Please also remember to update any impacted documentation.

## How is this addressed

- Change azurerm_app_service_plan to azurerm_service_plan
- Add an environment variable into developer .ENV.sample
- Update porter versions so that it takes effect


tanya-borisova (approved):

marrobi (changes requested):
Sorry to block couple of questions:

1. This seems to be more than an app service plan sku change?
2. Do the changes cause any resources to be deleted and recreated as we need to avoid that?

PR #1854, opened 2022-05-17
Poc/orchestrator as logicapp

eladiw:
**Orchestrator POC**
    
This 'POC' / initial commit is a LogicApp stub version of the Orchestrator.
Since Logic Apps and API Connections are not supported using TF, This POC shows how to combine ARM with TF -  Which is also a reason to decide not to go with this version.

It creates:
1. A Servicebus connection (API connection)
2. An event grid topic API connection
3. A naive orchestrator which is wired to (a) and (b) and all it does is 1) recieve a msg from the sb 2) reply with a msg to the event grid
(simulates the acceptance of a changed_status event and the reply with update_status event)

Logic app:
image


tamirkamara:
A logic app? Is that the intended final implementation as well?

eladiw:
Note to reviewers: this PR also contains many files which are part of the tf resources PR.

eladiw:
> A logic app? Is that the intended final implementation as well?

We agreed that no matter which option we go with, it should be the option future customers will go with.
for example, if we say, you can replace our native orchestrator with one of your own, they can pick it up, change it and continue.

Rather than, disabling ours and go develop something completely new.

marrobi (approved):
LGTM

tamirkamara (changes requested):
Putting a hold on this while we discuss the approach

eladiw:
This PR is abandoned as we are going in different direction

PR #1860, opened 2022-05-18
E2E tests: Fix shared service and performance tests 

tanya-borisova:
# PR for issue

## What is being addressed

In https://github.com/microsoft/AzureTRE/pull/1797, changes were made to E2E tests helper methods, but they weren't changed for performance tests and shared service tests. This PR fixes that.


templates/workspaces/base/terraform/variables.tf
```diff
@@ -32,7 +32,7 @@ variable "deploy_app_service_plan" {
 
 variable "app_service_plan_sku" {
   type        = string
-  default     = "P1v3"
```
ross-p-smith, line 35: This change isn't needed to get the Shared Services tests working

ross-p-smith (changes requested):
We will drop the app service with this change

ross-p-smith (approved):

Issue #517, opened 2021-07-29
[Feature] Azure Virtual Desktop Workspace Service


marrobi:
The ability to enable Azure Virtual Desktop within a workspace is a requirement for some customers.

Now that Azure AD join is in preview this is possible without Azure AD Domain Services: https://docs.microsoft.com/en-us/azure/virtual-desktop/deploy-azure-ad-joined-vm

The two scenarios that may be used are:
- Personal desktops with local user profiles.
- Pooled desktops used as a jump box.

wereinse:
@marrobi Please add some acceptance criteria to this ie
1. Automate the provisioning of a personal desktop to a workspace for a local user and configure access
2.  Automate the provisioning of a pooled set of 3 desktops for use as jump boxes.

PR #1865, opened 2022-05-18
Add Azure Virtual Desktop Workspace Service

SvenAelterman:
#517 

## What is being addressed

This creates an Azure Virtual Desktop deployment in the workspace. Based on work from Paul Yu and @marrobi.

## How is this addressed

- Add new workspace service folder with Porter bundle using Azure Bicep templates.

Pending work:

Additional work noted:

- [x] Remove localAdminPassword param and generate a new unique value to be stored in the workspace's KV.
- [x] Update to use latest AVD artifacts: https://wvdportalstorageblob.blob.core.windows.net/galleryartifacts/Configuration_02-23-2022.zip
- [x] Use the AVD ARM templates to deploy session hosts instead of recreating that in Bicep.
- [x] Create an availability set for the session hosts.
- [ ] Output the needed hub firewall rules as `rules_collection`.
- [x] Create unique ARM deployments by appending the service ID and the deployment timestamp.
- [x] Update the `connection_uri` output from Bicep to a shortlink, or update to the April 2022 AVD endpoint URL.
- [x] Fix the `customRdpProperty` on the hostpool for TRE purposes (disable drive redirection, disable clipboard)
- [ ] Attempt to configure the session hosts to disable drive redirection and clipboard with policy.
- [x] Create a parameter for the friendly name of the workspace and the desktop application because those are user-facing. Provide reasonable defaults.
- [ ] Parameter for images. Using custom images might need to be a separate issue.
- [x] Resolve MD linter errors.
- [ ] Use private endpoints for connection between session hosts and AVD.


marrobi:
@SvenAelterman I've moved the pending work to here and closing my draft.

ukct:
@marrobi - A partial solution to the data exfiltration issue is to make use of windows firewall within the session host itself - this obviously relies on users being unable to modify those firewall rules due to local admin rights or privilege escalation exploits. 
Add a new rule within the firewall to block the IP addresses used by the WindowsVirtualDesktop service tag, applied to local principal 'Everyone', except for 'system', 'local service' and 'network service'. This will prevent users from accessing an external AVD environment but still allow AVD itself to operate.

SvenAelterman:
To anyone following this PR, the best solution to avoid the exfiltration issue is to use private endpoints for AVD, now GA.

I don't know when, but I would be open to restarting the work on providing an AVD workspace service template that leverages this. It would likely require a change to the TRE core so it deploys the required private DNS zone for this new private link service.

marrobi:
This could also be a nice addition https://learn.microsoft.com/en-us/azure/virtual-desktop/uri-scheme

templates/workspace_services/avd-aad/bicep/main.bicep
```diff
@@ -0,0 +1,117 @@
...
+param tags object = {}
+param localAdminName string = 'adminuser'
+param vmSize string = 'Standard_D2as_v4'
+
+param vmCount int = 1
+param deploymentTime string = utcNow()
+@secure()
+param passwordSeed string = newGuid()
+
+var shortWorkspaceId = substring(workspaceId, length(workspaceId) - 4, 4)
```
SvenAelterman, line 15: Use `skip()` instead of `substring()`

martinpeck:
Associated with issue #517

Issue #1852, opened 2022-05-17
ResourceOperations Component


joalmeid:
Create a component that accepts a Resource as a prop, and displays a list of the items in the operations[] array

PR #1867, opened 2022-05-18
Listing workspace operations. for #1852 

joalmeid:
# PR for #1852 

## What is being addressed

Implements the **WorkspaceOperationsPanel**, listing the operations details. 
Accessed from Workspace , and the Operations tab

> Needs validation for the pattern of getting the operations from API.

## How is this addressed

- Lists the operations for the workspace in context
- shows details for each operation.

## Improvements
- No UI research for operations listing
- Identify best Fluent UI component to show the operations list.
- To be used by auditors. Probably not that used.
- Probably should allow to list workspace operations and operations for all services/user resources in a workspace
- Exporting capabilities for auditors


ui/yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
martinpeck, line 1: I suspect you ran `yarn start` from the wrong folder and this got created (I've been there, and done that 😄 ). Can you remove this file please?

ui/app/src/components/workspaces/WorkspaceOperationsPanel.tsx
```diff
@@ -0,0 +1,100 @@
...
+                                <WorkspaceOperationPanelItem header={'Message'} val={op.message} />
+                                <WorkspaceOperationPanelItem header={'User'} val={op.user.name} />
+                            </Stack>    
+                        </Stack>
+                    )
+                })
+            }
+        </>
+    );
+};
```
martinpeck, line 100: nit: add new line to end of file

ui/app/src/components/workspaces/WorkspaceOperationsPanel.tsx
```diff
@@ -0,0 +1,100 @@
...
+    workspace: Workspace
+}
+
+interface WorkspaceOperationPanelItemProps {
+    header: String,
+    val: String
+}
+
+const WorkspaceOperationPanelItem: React.FunctionComponent<WorkspaceOperationPanelItemProps> = (props: WorkspaceOperationPanelItemProps) => {
+
```
martinpeck, line 22: question: Is three-blank-line spacing normal here? I'd have expected just one blank line between function sig and body, but I honestly don't know the convention for .ts. Either way, it's not consistent further down (2 new lines).

(if we have a linter set up, and this is linted, please ignore me 😄 )

ui/app/src/components/workspaces/WorkspaceOperationsPanel.tsx
```diff
@@ -0,0 +1,100 @@
...
+export const WorkspaceOperationsPanel: React.FunctionComponent<WorkspaceOperationPanelProps> = (props: WorkspaceOperationPanelProps) => {
+    const apiCall = useAuthApiCall();
+    const { workspaceId } = useParams();
+    const [workspaceOperations, setWorkspaceOperations] = useState([] as Array<Operation>)
+    
+
+    const getOperations = async () => {
+        let ws = props.workspace && props.workspace.id ?
+        props.workspace :
+        (await apiCall(`${ApiEndpoint.Workspaces}/${workspaceId}`, HttpMethod.Get)).workspace;
```
damoodamoo, line 55: we don't need to get the workspace here, as that's done higher up - we can just use the prop

ui/app/src/components/workspaces/WorkspaceOperationsPanel.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+        await apiCall(`${ApiEndpoint.Workspaces}/${workspaceId}`, HttpMethod.Get, ws.properties.app_id, undefined, ResultType.None, (roles: Array<string>) => {
```
damoodamoo, line 57: don't need this, we have set the roles already in the parent

ui/app/src/models/operation.ts
```diff
@@ -1,28 +1,14 @@
-import { ResourceType } from "./resourceType";
 import { User } from "./user";
 
 export interface Operation {
-    id: string,
     resourceId: string,
-
-    isActive: boolean,
-    isEnabled: boolean,
     resourcePath: string,
```
damoodamoo, line 5: perhaps you didn't have latest from the feature branch but this operation model already existed? suggest you grab the existing one?

ui/app/src/components/workspaces/WorkspaceOperationsPanel.tsx
```diff
@@ -0,0 +1,100 @@
...
+    );
+}
+
+export const WorkspaceOperationsPanel: React.FunctionComponent<WorkspaceOperationPanelProps> = (props: WorkspaceOperationPanelProps) => {
+    const apiCall = useAuthApiCall();
+    const { workspaceId } = useParams();
+    const [workspaceOperations, setWorkspaceOperations] = useState([] as Array<Operation>)
+    
+
+    const getOperations = async () => {
```
damoodamoo, line 52: I think we should wrap this in the `useEffect()` hook here - see the other components for an example. It's a 'lifecycle' hook that means it will fire as the component loads and then only re-fire if it needs to.

ui/app/src/components/workspaces/WorkspaceOperationsPanel.tsx
```diff
@@ -0,0 +1,100 @@
...
+                    : {props.val}
+                </Stack.Item>
+            </Stack>
+        </>
+    );
+}
+
+export const WorkspaceOperationsPanel: React.FunctionComponent<WorkspaceOperationPanelProps> = (props: WorkspaceOperationPanelProps) => {
+    const apiCall = useAuthApiCall();
+    const { workspaceId } = useParams();
```
damoodamoo, line 48: you won't need this if we're just using the resource passed in

ui/app/src/components/workspaces/WorkspaceItem.tsx
```diff
@@ -48,9 +49,11 @@ export const WorkspaceItem: React.FunctionComponent<WorkspaceItemProps> = (props
         </PivotItem>
         <PivotItem headerText="History">
           <h3>--History goes here--</h3>
+
+
         </PivotItem>
         <PivotItem headerText="Operations">
-          <h3>--Operations Log here</h3>
+            <WorkspaceOperationsPanel workspace={props.workspace} />
```
damoodamoo, line 56: I think since we're listing Operations, we could change this to `ResourceOperationsList` rather than it being specific to a workspace - then we could re-use it for every resource type.

ui/app/src/components/workspaces/WorkspaceOperationsPanel.tsx
```diff
@@ -0,0 +1,100 @@
...
+interface WorkspaceOperationPanelProps {
+    workspace: Workspace
+}
+
+interface WorkspaceOperationPanelItemProps {
+    header: String,
+    val: String
+}
+
+const WorkspaceOperationPanelItem: React.FunctionComponent<WorkspaceOperationPanelItemProps> = (props: WorkspaceOperationPanelItemProps) => {
```
damoodamoo, line 21: Bit of a judgement call, but personally I think we could move this to a separate file - `ResourceOperationListItem` (let's not use 'panel' as Fluent's use of 'panel' will confuse us in future). This would help keep each component as small as possible... but again you could fall down either side on this :)

damoodamoo (changes requested):
nice work! Main changes are in using the resource handed to the component rather than getting it again (so it can be nice and dumb :) ) - and refactoring a little to make it a generic component, usable everywhere.

ui/yarn.lock
joalmeid: Yep, did see that. good catch.

ui/app/src/components/workspaces/WorkspaceOperationsPanel.tsx
joalmeid: Yep, linter will definitely structure it. It's fixed for this PR though!

ui/app/src/components/workspaces/WorkspaceOperationsPanel.tsx
joalmeid: Yep those were actually the things I wanted validated. Not sure I would be authed in workspace api.
Thanks!

ui/app/src/components/workspaces/WorkspaceOperationsPanel.tsx
joalmeid: 👍

ui/app/src/components/workspaces/WorkspaceOperationsPanel.tsx
damoodamoo: Id you wanna build just for workspaces that's cool, I have a change coming making it easier to get the client id we need

ui/app/src/components/workspaces/WorkspaceOperationsPanel.tsx
```diff
@@ -0,0 +1,100 @@
...
+        props.workspace :
+        (await apiCall(`${ApiEndpoint.Workspaces}/${workspaceId}`, HttpMethod.Get)).workspace;
+
+        await apiCall(`${ApiEndpoint.Workspaces}/${workspaceId}`, HttpMethod.Get, ws.properties.app_id, undefined, ResultType.None, (roles: Array<string>) => {
+            config.debug && console.log(`Workspace roles for ${ws.properties?.display_name}`, roles);
+        }, true);
+
+        // get workspace operations 
+        config.debug && console.log(`Getting Workspace Operations for workspace:${props.workspace.id}`);
+        const workspaceOperations = await apiCall(`${ApiEndpoint.Workspaces}/${ws.id}/${ApiEndpoint.Operations}`, HttpMethod.Get, ws.properties.app_id);
```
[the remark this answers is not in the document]
joalmeid: Yep, we mentioned that yesterday! Listing operations will probably be for auditing purpose and listing everything for a Workspace (including child services / user resources operations). For now I agree this should be made dumb and ready for any resource.

We'll add a mockup suggestion to give Frankie, and add it to the HX research.

ui/app/src/components/workspaces/WorkspaceOperationsPanel.tsx
joalmeid: Exactly the feedback I wanted! Thanks. adding useEffect for the lifecycle. Also, should we also have a loadingState ?

ui/app/src/components/workspaces/WorkspaceOperationsPanel.tsx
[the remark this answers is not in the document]
damoodamoo: Can we just go with an operations list per resource, for each page, first - the same as the properties and history panel work? Then later we can come back to querying for all operations under a workspace if we want..? @marrobi - that sound reasonable?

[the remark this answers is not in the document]
damoodamoo: Yea agree - different piece of work, using a newly built API endpoint

ui/app/src/components/workspaces/WorkspaceItem.tsx
joalmeid: Agree. Thanks!

ui/app/src/components/workspaces/ResourceOperationsList.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+    useEffect(() => {
+        const getOperations = async () => {
+            try {
+                let ws = props.resource && props.resource.id ?
+                props.resource :
+                (await apiCall(`/${props.resource.resourcePath}`, HttpMethod.Get)).workspace;
+                
+                // get resource operations 
+                const resourceOperations = await apiCall(`${ApiEndpoint.Workspaces}/${ws.id}/${ApiEndpoint.Operations}`, HttpMethod.Get, ws.properties.app_id);
```
ross-p-smith, line 57: I feel that app_id will be moved to client_id if you merge the latest

ui/app/src/components/workspaces/WorkspaceOperationsPanel.tsx
joalmeid: Agree! Seperated `ResourceOperationsListItem` from the `ResourceOperationsList`. Panel is not more.
Thanks

ui/app/src/components/workspaces/ResourceOperationsList.tsx
```diff
@@ -0,0 +1,92 @@
...
+    const { resourceId } = useParams();
+    const [resourceOperations, setResourceOperations] = useState([] as Array<Operation>)
+    const [loadingState, setLoadingState] = useState('loading');
+
+    useEffect(() => {
+        const getOperations = async () => {
+            try {
+                let ws = props.resource && props.resource.id ?
+                props.resource :
+                (await apiCall(`/${props.resource.resourcePath}`, HttpMethod.Get)).workspace;
```
damoodamoo, line 27: We don't need to fetch this from the API as the resource will be passed to us

ui/app/src/components/workspaces/ResourceOperationsList.tsx
```diff
@@ -0,0 +1,92 @@
...
+                
+                // get resource operations 
+                const resourceOperations = await apiCall(`${ApiEndpoint.Workspaces}/${ws.id}/${ApiEndpoint.Operations}`, HttpMethod.Get, ws.properties.app_id);
```
damoodamoo, line 30: instead of `ws.properties.app_id` you should be able to the workspace context - like this:

With the other declarations (after `const apiCall = ...` ) add:
`const workspaceCtx = useContext(WorkspaceContext);`

Then when calling the operations, use `workspaceCtx.workspaceClientId` instead of `ws.properties`.

ui/app/src/components/workspaces/ResourceOperationsList.tsx
```diff
@@ -0,0 +1,92 @@
...
+    const apiCall = useAuthApiCall();
+    // const { workspaceId } = useParams();
+    const { resourceId } = useParams();
+    const [resourceOperations, setResourceOperations] = useState([] as Array<Operation>)
+    const [loadingState, setLoadingState] = useState('loading');
+
+    useEffect(() => {
+        const getOperations = async () => {
+            try {
+                let ws = props.resource && props.resource.id ?
```
damoodamoo, line 25: let's remove this var and use `props.resource`

damoodamoo (changes requested):
nice work - just a couple of changes then we're good to go :)

damoodamoo:
Can we also revert the changes to `operation.ts` too?

ui/app/src/models/operation.ts
joalmeid: updated 👌

ui/app/src/components/workspaces/ResourceOperationsList.tsx
joalmeid: Thanks

damoodamoo (approved):
LGTM!

Issue #1745, opened 2022-04-27
Azure Firewall Logs not appearing in Log Analytics


martinpeck:

martinpeck:
@marrobi are you able to give me an example of a log you would expect to see, and a query that would return that log?

marrobi:
Under log analytics, check for AzureFirewallApplicationRule or AzureFirewallNetworkRule as a category. Neither exist. They used to.

There are also default Azure Firewall queries presented when view logs under the firewall resource in the portal. They fail.

martinpeck:
About 3 months ago we switched logging for the firewall to be "Dedicated".

AFAIK, this means that logs no longer drop into the `AzureDiagnostics` table but, instead, drop into resource specific tables.

https://docs.microsoft.com/en-us/azure/azure-monitor/essentials/resource-logs#resource-specific

I wonder if Azure Firewall properly supports resource-specific log tables. We might want remove this switch for now.

https://github.com/microsoft/AzureTRE/blob/main/templates/shared_services/firewall/terraform/firewall.tf#L37

martinpeck:
I can't find anything that confirms Azure Firewall does, or doesn't, support resource-specific tables. However, in the docs it suggests that when setting up logging for a resource in the portal it will either hide this choice from the user, or offer both. In the case of Azure Firewall, it doesn't offer any choice.

Is it possible that we've had no logs for 3 months (i.e. since we made this change)?

martinpeck:
FYI a total of 4 resources in our deployment use this Dedicated setting...

- keyvault
- service bus
- app gateway
- firewall

martinpeck:
This is the reason we set it to Dedicated:

https://github.com/microsoft/AzureTRE/pull/1313/files#r807969959

marrobi:
Could well be this. It has been a while as we only rely on the firewall logs when debugging new services.

martinpeck:
PR merged. Closing

PR #1870, opened 2022-05-19
Fix Firewall Logging

martinpeck:
# PR for issue #1745 

This switches logging for the firewall back to the generic log tables, rather than dedicated tables, as the dedicated table logging doesn't appear to work!

To test this I needed to...

- delete the firewall service from CosmosDB
- run `make firewall-install`

Once done, I was then able to...

- visit the jumpbox
- use Edge to try and access random website (e.g. `cheese.com`)
- wait a bit, and then see the logs in the Log Analytics workspace

e.g. the following query returned logs...

```
AzureDiagnostics | where
msg_s contains "cheese.com"
```


martinpeck:
Tests passed, but linting picked up an issue in the `firewall.tf`.
Have fixed, and pushed update

marrobi (approved):
LGTM

Issue #1868, opened 2022-05-19
Implement actions and <code class="notranslate">Connect` buttons on the User Resource card, as appropriate. This will need some further planning and sub tasks.


christoferlof:

christoferlof:
Pushing `- [ ] Additional actions for User Resources` to it's own issue. Will require some more thought.

PR #1871, opened 2022-05-19
Connect button on resource card

christoferlof:
# PR for issue #1868

* Adds a Connect button to the `ResourceCard`
* Connect button shown if there's a `connection_uri` available in the resource props.
* `connection_uri` launched in new window.

image

image

User Resource Actions to be addressed in separate PR.

### DIFF
--- a/ui/app/src/components/shared/ResourceCard.tsx
+++ b/ui/app/src/components/shared/ResourceCard.tsx
@@ -81,8 +81,7 @@ export const ResourceCard: React.FunctionComponent<ResourceCardProps> = (props: 
     },
   });
 
-  let canConnect = props.resource.properties && props.resource.properties['connection_uri'];
-  let connectUri = canConnect ? props.resource.properties['connection_uri'] : '';
+  let connectUri = props.resource.properties && props.resource.properties['connection_uri'];
 
   return (
     <>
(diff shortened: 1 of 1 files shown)


ui/app/src/components/shared/ResourceCard.tsx
```diff
@@ -72,24 +81,31 @@ export const ResourceCard: React.FunctionComponent<ResourceCardProps> = (props:
...
+              <Stack.Item><IconButton iconProps={{iconName: 'Info'}} id={`item-${props.itemId}`} onClick={() => setShowInfo(!showInfo)} /></Stack.Item>
+              <Stack.Item>{props.contextMenuElement && props.contextMenuElement}</Stack.Item>
+            </Stack>
+          </Stack.Item>
         </Stack>
-        <Stack.Item grow={3} style={bodyStyles}>
+        <Stack.Item grow={3} style={ (canConnect) ? bodyStylesWithConnect : bodyStyles}>
           <Text>{props.resource.properties.description}</Text>
         </Stack.Item>
+        { canConnect &&
```
martinpeck, line 104: is it normal to put the controls to render on the other side of an `&&`?

martinpeck: wouldn't this usually have a code block that returned the content?

ui/app/src/components/shared/ResourceCard.tsx
```diff
@@ -72,24 +81,31 @@ export const ResourceCard: React.FunctionComponent<ResourceCardProps> = (props:
     },
   });
 
+  let canConnect = props.resource.properties && props.resource.properties['connection_uri'];
+  let connectUri = canConnect ? props.resource.properties['connection_uri'] : '';
```
martinpeck, line 85: do we actually need both? 
it's easier to read, for sure, but could we use `connectUri` for all of the tests below?

i.e. do this here...

``` typescript
let connectUri = props.resource.properties && props.resource.properties['connection_uri'];
```

martinpeck (commented):
some questions/comments, but looks good to me

ui/app/src/components/shared/ResourceCard.tsx
christoferlof: I agree it looks weird and took me some time to accept :) However, seems to be the/an accepted way to do it. We have it across the codebase already. Here's another example https://github.com/microsoft/AzureTRE/blob/e1a302c145159f80faee32879f9e920dd0a9f981/ui/app/src/components/shared/ResourceCard.tsx#L99

damoodamoo: it's a shortcut for if is truthy, render the following element. looks weird but saves a bunch of extra lines

Issue #1819, opened 2022-05-12
Implementing an initial CLI codebase


joalmeid:
Consider an isolated feature branch for the TRE CLI for the earlier implementations. Once matured can be part of main. All dev work to be branched from this feature branch and PR'd back into it.

- [ ]  Create and push branch
- [ ]  Lock branch to prevent direct commits
- [ ] Scaffold initial code base.

joalmeid:
Should take into consideration the spike investigation #1818

joalmeid:
Keeping @Benc-uk  implementation of a standalone CLI

Issue #1820, opened 2022-05-12
[Spike] Investigate having an automated way to generate client SDKs for TRE APIs


joalmeid:
Investigate an automated way to generate a client SDK, that can execute calls to the TRE API, including workspace specific API endpoints.

Given that the TRE API, will gradually have new features, this client SDK, should be autogenerated using tools like [AutoRest](https://github.com/Azure/autorest) or [OpenAPI Generator](https://openapi-generator.tech/)

This process should be documented. Also consider/investigate having SDK automatically genereted with GH Actions

joalmeid:
Addressed in #1879

PR #1879, opened 2022-05-19
CLI PoC using Click and OpenAPI Tools

benc-uk:
# PR for issue #1819 

## What is being addressed

This is a minimal POC of a CLI client for TRE. It is written in Python and uses autogenerated API client which has been created from the published OpenAPI spec (aka Swagger). 

The generator used was:
[OpenAPI tools generator](https://github.com/OpenAPITools/openapi-generator)

## How is this addressed

- It was found we could not use the Azure CLI extension as a way to provide the functionality required, so a standalone Python app was created
- Click provides the framework for the command structure, arguments


joalmeid:
includes client sdk generation for multiple languages. focuses on Python. Addresses #1820

PR #1907, opened 2022-05-24
Resource Actions

### LINKED ISSUE #1904: Resource Actions
Each resource type support custom actions, defined in the template. The UI should query the template + version when displaying a resource, and offer "Actions" in the context menu.

Since actions don't require additional input, a confirm dialog should be shown with Ok/Cancel buttons. Then the operation will be handed to notifications to follow the same path as everything else.

damoodamoo:
# Closes #1904 

If any resource has custom actions, they are represented in the context menu.

image

PR also contains some more stability improvements to notifications + workspace context


marrobi (approved):
LGTM

Issue #1815, opened 2022-05-11
Upgrade azurerm_app_service_plan


ross-p-smith:
## Description

As a TRE Administrator
I want to use the latest Terraform App Service Plan
So that I don't use a deprecated Terraform Module

## Acceptance criteria

- [ ] No longer reference azurerm_app_service_plan

You can use the workspace app service plan as an example for the settings (azurerm_service_plan)

PR #1912, opened 2022-05-24
Upgrade mgmt Azure_rm provider

ross-p-smith:
# PR for issue #1815 

## What is being addressed

We need to upgrade the AzureRM provider so that we can use better keyvault support and paramaterise the AppService Plan. 

## How is this addressed

- This is Part 1 of upgrading the bootstrap/management layer of TRE.


ross-p-smith:
/destroy_branch_env

(If we delete the environment and then create it again - this should prove out the keyvault)

ross-p-smith:
/test-destroy-env

(If we delete the environment and then create it again - this should prove out the keyvault)

ross-p-smith:
/test

(now test recovering the vault)

Issue #1915, opened 2022-05-25
Jumpbox VM OS disk is not tagged as core service


guybartal:
**Describe the bug**
Jumpbox VM OS disk is not tagged as core service, os_disk doesn't support tags in terraform file:

![image](https://user-images.githubusercontent.com/3179107/170229595-8851cb12-6160-44f1-b500-5ab86c4ce90d.png)
[terraform registry - windows vm docs](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/windows_virtual_machine)

**Related issue**
#1888 

**Steps to reproduce**

1. Deploy TRE Instance
1. Go to Azure portal, search for vm-dsk-[your_tre_id] resource
1. The resource won't have the needed tags (tre_id, and tre_core_service_id)

tamirkamara:
@guybartal Is this resolved already?

guybartal:
> @guybartal Is this resolved already?

Terraform doesn't support tagging for os_disk:
[https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/windows_virtual_machine#argument-reference](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/windows_virtual_machine#argument-reference)

Issue #1888, opened 2022-05-20
Tag TRE Core azure services in Terraform


joalmeid:
As a TRE developer,
I want all TRE Core azure services to be tagged
So that I can include TRE Core costs in TRE Cost Reporting

**Acceptance Criteria**
- [x] #1911

guybartal:
PR merged:
https://github.com/microsoft/AzureTRE/pull/1916

Issue #65, opened 2021-05-10
Add bundle registry to TRE templates


christoferlof:
## User story

As a TRE service integrator
I need to place my workspace bundles in the bundle registry
So that the services can be instantiated by researchers and read by the deployment client / resource processor.

## Acceptance critiera

- [x] Bundle registry defined in IaC and integrated into TRE IaC
- [x] Bundle registry deployed as part of TRE deployment
- [x] Sample bundle added to bundle registry in shared environment for demo

christoferlof:
@deniscep  please see whether we can or should repurpose the ACR already in the templates.

tompaana:
The DevOps bits to publish the workspace use custom scripts now run by the `Makefile`, but if at some point we want to move it to typical workflow steps we can refer to https://github.com/getporter/pipeline-demo

PR #1916, opened 2022-05-25
Tag tre core services

guybartal:
# PR for issue #1888 

## What is being addressed

Tag core services with `tre_id` and `tre_core_service_id` need by cost report feature and budget

## How is this addressed

- Added `tre_core_tags` variable in `locals.tf` per moudle
- Added tags property per azure resource of the core services
- Pending work  - #1915

### DIFF
--- a/templates/core/terraform/appgateway/locals.tf
+++ b/templates/core/terraform/appgateway/locals.tf
@@ -24,4 +24,8 @@ locals {
   redirect_configuration_name        = "rdrcfg-tosecure"
 
   certificate_name = "cert-primary"
+  tre_core_tags = {
+    tre_id              = var.tre_id
+    tre_core_service_id = var.tre_id
+  }
 }
--- a/templates/core/terraform/azure-monitor/azure-monitor.tf
+++ b/templates/core/terraform/azure-monitor/azure-monitor.tf
@@ -6,7 +6,7 @@ resource "azurerm_log_analytics_workspace" "core" {
   sku                 = "pergb2018"
   tags                = local.tre_core_tags
 
-  lifecycle { ignore_changes = [tags] }
+  #lifecycle { ignore_changes = [tags] }
 }
 
 # Storage account for Application Insights
@@ -21,7 +21,7 @@ resource "azurerm_storage_account" "app_insights" {
   allow_blob_public_access = false
   tags                     = local.tre_core_tags
 
-  lifecycle { ignore_changes = [tags] }
+  #lifecycle { ignore_changes = [tags] }
 }
 
 data "local_file" "app_insights_arm_template" {
--- a/templates/core/terraform/azure-monitor/azure-monitor.tf
+++ b/templates/core/terraform/azure-monitor/azure-monitor.tf
@@ -56,10 +56,6 @@ resource "azurerm_resource_group_template_deployment" "app_insights_core" {
       value = local.tre_core_tags
     }
   })
-
-  # "tre_id" = {
-  #     value = var.tre_id
-  #   }
 }
 
 data "local_file" "ampls_arm_template" {
--- a/templates/core/terraform/azure-monitor/azure-monitor.tf
+++ b/templates/core/terraform/azure-monitor/azure-monitor.tf
@@ -6,7 +6,7 @@ resource "azurerm_log_analytics_workspace" "core" {
   sku                 = "pergb2018"
   tags                = local.tre_core_tags
 
-  #lifecycle { ignore_changes = [tags] }
+  lifecycle { ignore_changes = [tags] }
 }
 
 # Storage account for Application Insights
@@ -21,7 +21,7 @@ resource "azurerm_storage_account" "app_insights" {
   allow_blob_public_access = false
   tags                     = local.tre_core_tags
 
-  #lifecycle { ignore_changes = [tags] }
+  lifecycle { ignore_changes = [tags] }
 }
 
 data "local_file" "app_insights_arm_template" {
@@ -103,7 +103,7 @@ resource "azurerm_private_endpoint" "azure_monitor_private_endpoint" {
   subnet_id           = var.shared_subnet_id
   tags                = local.tre_core_tags
 
-  #lifecycle { ignore_changes = [tags] }
+  lifecycle { ignore_changes = [tags] }
 
   private_service_connection {
     private_connection_resource_id = jsondecode(azurerm_resource_group_template_deployment.ampls_core.output_content).resourceId.value
--- a/templates/core/terraform/azure-monitor/azure-monitor.tf
+++ b/templates/core/terraform/azure-monitor/azure-monitor.tf
@@ -4,6 +4,7 @@ resource "azurerm_log_analytics_workspace" "core" {
   location            = var.location
   retention_in_days   = 30
   sku                 = "pergb2018"
+  tags                = local.tre_core_tags
 
   lifecycle { ignore_changes = [tags] }
 }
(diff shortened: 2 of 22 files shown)


templates/core/terraform/appgateway/locals.tf
```diff
@@ -24,4 +24,8 @@ locals {
   redirect_configuration_name        = "rdrcfg-tosecure"
 
   certificate_name = "cert-primary"
+  tre_core_tags = {
+    tre_id              = var.tre_id
+    tre_core_service_id = var.tre_id
+  }
```
eladiw, line 30: previously there was an 'environment' tag. which was removed. are we ok with that or we want to add it here?

templates/core/terraform/azure-monitor/azure-monitor.tf
```diff
@@ -50,7 +52,14 @@ resource "azurerm_resource_group_template_deployment" "app_insights_core" {
...
       value = azurerm_storage_account.app_insights.name
     }
+    "tre_core_tags" = {
+      value = local.tre_core_tags
+    }
   })
+
+  # "tre_id" = {
+  #     value = var.tre_id
+  #   }
```
eladiw, line 62: remove

templates/core/terraform/locals.tf
```diff
@@ -21,3 +21,10 @@ data "http" "myip" {
 locals {
   myip = var.public_deployment_ip_address != "" ? var.public_deployment_ip_address : chomp(data.http.myip[0].body)
 }
+
+locals {
+  tre_core_tags = {
+    tre_id              = var.tre_id
+    tre_core_service_id = var.tre_id
```
eladiw, line 28: is this value correct?

templates/core/terraform/network/locals.tf
```diff
@@ -7,4 +7,8 @@ locals {
   shared_services_subnet_address_prefix    = local.core_services_vnet_subnets[4] # .0 - .254
   aci_subnet_address_prefix                = local.core_services_vnet_subnets[5] # .0 - .254
   resource_processor_subnet_address_prefix = local.core_services_vnet_subnets[6] # .0 - .254
+  tre_core_tags = {
+    tre_id              = var.tre_id
+    tre_core_service_id = var.tre_id
```
eladiw, line 12: is this value correct?

templates/core/terraform/azure-monitor/azure-monitor.tf
guybartal: done

eladiw (changes requested):
some comments

templates/core/terraform/locals.tf
guybartal: yes, by design.
we need it to descriminate core service from other services in a TRE instance

templates/core/terraform/appgateway/locals.tf
```diff
@@ -24,4 +24,8 @@ locals {
   redirect_configuration_name        = "rdrcfg-tosecure"
 
   certificate_name = "cert-primary"
+  tre_core_tags = {
+    tre_id              = var.tre_id
+    tre_core_service_id = var.tre_id
```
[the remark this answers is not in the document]
guybartal: yes, same reason.

templates/core/terraform/appgateway/locals.tf
guybartal: not sure, I think if was left there by a mistake.
@marrobi, WDYT?

templates/core/terraform/network/locals.tf
guybartal: yes, same reason.

templates/core/terraform/locals.tf
guybartal: discussted over teams:

when you call Azure Cost Management REST API and group by tag and filter by tre_id tag you get in response aggregated costs by tag.

due to the fact that we don't have a generated id per core service, I chose to give it the same id as the tre id

eladiw (approved):
lgtm

templates/core/terraform/appgateway/locals.tf
marrobi: I can't see the removed tag here - shouldn't be anything with a fixed value, but we need to enable customers to add their own tags. Maybe that should be picked up as another user story?

marrobi:
Have we any way to ensure future PRs keep the appropriate tags/add the appropriate tags to new resources? Again, one for another issue, but think it's important.

marrobi (commented):
Have we any way to ensure future PRs keep the appropriate tags/add the appropriate tags to new resources? Again, one for another issue, but think it's important.

templates/core/terraform/appgateway/locals.tf
guybartal: They can add their own tags via Azure portal / Azure API / TF, I left the ignore_changes in all resources.
Do you think we need a TRE API for that?

guybartal: templates/core/terraform/admin-jumpbox.tf line #65

marrobi: More, how do they add via TF - would be better by setting variables, rather than having their own fork with custom tags.

guybartal: lets create another separate story for that.

guybartal:
> Have we any way to ensure future PRs keep the appropriate tags/add the appropriate tags to new resources? Again, one for another issue, but think it's important.

The only idea I have right now, which is not perfect but may help, is to get all azure resources by resouce group and check if there are any resources without tags.

templates/core/terraform/resource_processor/vmss_porter/locals.tf
```diff
@@ -1,3 +1,9 @@
 locals {
   version = replace(replace(replace(data.local_file.version.content, "__version__ = \"", ""), "\"", ""), "\n", "")
-}
\ No newline at end of file
+}
+locals {
```
ross-p-smith, line 4: Why 2 `locals` sections - can they be merge together

templates/core/terraform/locals.tf
```diff
@@ -21,3 +21,10 @@ data "http" "myip" {
 locals {
   myip = var.public_deployment_ip_address != "" ? var.public_deployment_ip_address : chomp(data.http.myip[0].body)
 }
+
```
ross-p-smith, line 24: Can you merge these 2 locals?
```suggestion
locals {
  myip = var.public_deployment_ip_address != "" ? var.public_deployment_ip_address : chomp(data.http.myip[0].body)
  tre_core_tags = {
    tre_id              = var.tre_id
    tre_core_service_id = var.tre_id
}
}
```

ross-p-smith:
I just have a quick question, did we consider setting the tags on the Resource Group and then everywhere we need to set the tags just something like
`object.tags = data.resource_group.name.tags`

- We would end up with less locals
- We have one place to set tags on everything.

templates/core/terraform/locals.tf
guybartal: > Can you merge these 2 locals?

sure, done

templates/core/terraform/resource_processor/vmss_porter/locals.tf
guybartal: merged, thanks!

guybartal:
> I just have a quick question, did we consider setting the tags on the Resource Group and then everywhere we need to set the tags just something like `object.tags = data.resource_group.name.tags`
> 
> * We would end up with less locals
> * We have one place to set tags on everything.

it may help for tre_id + tre_workspace_id but not for tre_core_service_id and tre_workspace_service_id and tre_user_resource_id becasue they are all share the same resource group.

PR #1924, opened 2022-05-27
Mandatory client-secret when creating a workspace

ross-p-smith:
# PR for issue

## What is being addressed

Currently we only pass in the workspace client id when we create a workspace. When we allow TRE to create the workspace and register it with AAD we will store the new client-secret in the workspace keyvault. We now need to mirror that behaviour for a self registered Workspace.

Also, there has been a change when you create the Guacamole server as it now needs the client id/secret (from the vault). This change means that the workspace-service relies on the creds being in the vault as a pre-requisite.

## How is this addressed

- Creating a workspace now requires client_secret
- Client Id and Client Secret are now added to the vault at workspace creation time
- Guacamole service no longer accepts a client id/secret in the api
- Guacamole service no longer adds the creds into the keyvault
- Added the ability to debug the Guacamole Terraform locally

- KNOWN ISSUE - Client Secret is now a workspace property. This will need to not be visible


templates/workspaces/base/.env.sample
```diff
@@ -1,11 +1,17 @@
...
-AUTH_TENANT_ID="__CHANGE_ME__"
-AUTH_CLIENT_ID="__CHANGE_ME__"
-AUTH_CLIENT_SECRET="__CHANGE_ME__"
 ARM_CLIENT_ID="__CHANGE_ME__"
 ARM_CLIENT_SECRET="__CHANGE_ME__"
 ARM_TENANT_ID="__CHANGE_ME__"
 ARM_SUBSCRIPTION_ID="__CHANGE_ME__"
+AUTH_TENANT_ID="__CHANGE_ME__"
+
+# These are passed in if you let Terraform register the AAD Application
```
marrobi, line 8:
```suggestion
# These are passed in if Terraform will create the Workspace AAD Application
```

templates/workspaces/base/terraform/variables.tf
```diff
@@ -32,7 +32,7 @@ variable "deploy_app_service_plan" {
 
 variable "app_service_plan_sku" {
   type        = string
-  default     = "P1v3"
+  default     = "P1v2"
```
marrobi, line 35: Can we leave this as is? We seem to keep flipping?

marrobi (commented):
Will try do a test deploy of the workspace followed by Guacamole today. Looks good though.

marrobi (approved):
Tested end to end - other than the SKU issue, all worked well.

templates/workspaces/base/terraform/variables.tf
ross-p-smith, line 35:
```suggestion
  default     = "P1v3"
```

ross-p-smith: CHanged it back to P1v3

PR #1926, opened 2022-05-27
Upgrade azurerm provider version to 3.5.0

tanya-borisova:
# PR for issue https://github.com/microsoft/AzureTRE/issues/1925

## What is being addressed

Upgrading azurerm Terraform provider to 3.5.0

## How is this addressed

- Upgrade provider version to "3.5.0" 
- allow_nested_items_to_be_public [is replacing](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/guides/3.0-upgrade-guide) allow_blob_public_access 
- azurerm_service_plan [is replacing](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/guides/3.0-upgrade-guide) azurerm_app_service_plan


templates/core/terraform/keyvault.tf
```diff
@@ -4,7 +4,7 @@ resource "azurerm_key_vault" "kv" {
   location                 = azurerm_resource_group.core.location
   resource_group_name      = azurerm_resource_group.core.name
   sku_name                 = "standard"
-  purge_protection_enabled = var.keyvault_purge_protection_enabled
+  purge_protection_enabled = true
```
stuartleeks: This looks like a behavioural change - is this intentional?

templates/workspace_services/guacamole/terraform/main.tf
```diff
@@ -13,7 +13,10 @@ terraform {
 provider "azurerm" {
   features {
     key_vault {
-      purge_soft_delete_on_destroy = false
+      # Don't purge secrets on destroy (this would fail due to purge protection being enabled on keyvault)
+      purge_soft_deleted_secrets_on_destroy = false
+      # When recreating an environment, recover any previously soft deleted secrets
+      recover_soft_deleted_secrets = true
```
stuartleeks, line 19: Does the azurerm provider now handle secrets existing after recovery? My recollection was that when we had vaults recovered in the past, we had issues as the secrets needed to be imported

tanya-borisova: Ugh this also shouldn't be there -- thanks for spotting, will revert this too

templates/shared_services/firewall/terraform/firewall.tf
```diff
@@ -13,6 +13,8 @@ resource "azurerm_firewall" "fw" {
   name                = "fw-${var.tre_id}"
   resource_group_name = local.core_resource_group_name
   location            = data.azurerm_resource_group.rg.location
+  sku_tier            = "Standard"
+  sku_name            = "AZFW_VNet"
```
stuartleeks, line 17: Could this cause the firewall to be recreated? If so, is that something that we should call out anywhere?

templates/core/terraform/api-webapp.tf
```diff
@@ -6,27 +6,21 @@ locals {
   version = replace(replace(replace(data.local_file.api_app_version.content, "__version__ = \"", ""), "\"", ""), "\n", "")
 }
 
-resource "azurerm_app_service_plan" "core" {
```
stuartleeks, line 9: I guess this causes the app service to be recreated, but assume that's not an issue as users won't have customised it?

stuartleeks (approved):

templates/shared_services/firewall/terraform/firewall.tf
stuartleeks: Cool - might be good to add a custom rule before deploying latest to check it is preserved?

tamirkamara:
Have you tested all the services you've upgraded? Some are not that simple (azureml, innereye) and maybe isn't worth the effort to upgrade...

templates/shared_services/firewall/terraform/firewall.tf
tanya-borisova: Tested that and firewall didn't get recreated 👍

tanya-borisova:
I have tested deploying the version from `main` and then deploying this branch. This has worked and avoided recreating any resources unnecesarily. The only change that forced deletion was this one:

```
      ~ private_service_connection {
          ~ name                           = "psc-web--stwebtanyaupgr4" -> "psc-web-stwebtanyaupgr4" # forces replacement
          ~ private_ip_address             = "10.1.1.23" -> (known after apply)
            # (3 unchanged attributes hidden)
        }
```

I couldn't figure out why there were two dashes in the name, but this doesn't seem to have impacted anything. 

Same with firewall, deploying first from main and then from this branch has made some changes in rule collections, but only order of DNS names in these collections, which I don't think is significant.

tanya-borisova:
Due to a bug I had in a previous version I'm now stuck with a soft-deleted purge-protected keyvault. New PR on the way...

PR #1928, opened 2022-05-29
Setting workspace_owner_object_id when creating workspaces

### LINKED ISSUE #220: AD app registration created when new workspace is created

ross-p-smith:
## What is being addressed

1. We need to know the workspace owner when we are creating a workspace, so that we can grant it the WorkspaceOwner permission so that the tests work correctly.

NB - This is only for workspaces that have a value of "auto_create" for the client_id of the workspace. Fixes #220 

2. Guacamole uses the `api://<identifier_Uri>` to authenticate with the Workspace AAD Application. This can no longer be derived by client_id, but can be configurable; especially if you auto_create the Workspace AD App. Therefore this value is passed in to the swagger when we create the Guacamole workspace service, so it can be used as the scope/audience.

## Outstanding
- At the moment, you cannot pass in workspace_owner_object_id in the Swagger. This will be another PR

## How is this addressed

- At the moment, the current logged in user of the API is passed into the workspace creation process and that user will be the WorkspaceOwner
- Updated the Terraform to assign `WorkspaceOwner` role permission to the passed in object id.
- When we create the workspace we need to wait until the key vault DNS is ready before we add secrets to it. Wait on the null resource provider to complete
- Guacamole service now requires the workspace_identifier_uri to be passed in on creation. This can be found in the workspace properties under scope_id.


templates/workspaces/base/terraform/workspace.tf
```diff
@@ -31,6 +31,6 @@ module "aad" {
   depends_on = [
     azurerm_key_vault_access_policy.deployer,
     azurerm_key_vault_access_policy.resource_processor,
-    azurerm_private_endpoint.kvpe
+    null_resource.wait_for_dns_vault
```
tamirkamara, line 34: What is this for?

e2e_tests/test_workspace_services.py
```diff
@@ -56,7 +56,7 @@ async def test_create_guacamole_service_into_base_workspace(admin_token, verify)
 
 @pytest.mark.extended_aad
 @pytest.mark.timeout(3000)
-async def test_create_guacamole_service_into_aad_workspace(admin_token, workspace_owner_token, verify) -> None:
+async def test_create_guacamole_service_into_aad_workspace(admin_token, verify) -> None:
```
tamirkamara, line 59: Why have you changed the token to the admin one? A workspace service needs to be deployed by the owner... 
Or, am I missing something?

templates/workspaces/base/terraform/workspace.tf
ross-p-smith: When you write items to the keyvault, you have to do it over the private endpoint. However, there is a delay between the private endpoint being created and then that ip address being added to the DNS. This is just a nasty sleep for 2 minutes for the DNS to resolve.

e2e_tests/test_workspace_services.py
ross-p-smith: The E2E tests create the workspace dynamically, and to do that they need the admin token as that route is an admin only route. We used to have a fixture that would get the workspace token by using a convention `api://<app_reg_id`; that convention is not true anymore as we need to read a property of the workspace (scope_id) to tell us what that value is. Once we have that value, then we dynamically get the workspace_token scoped to that application and then use the workspace_owner token from then on in.

Basically, what used to be a test fixture, is now a method called **get_workspace_owner_token** that you call from your test.

api_app/tests_ma/test_db/test_repositories/test_workpaces_repository.py
```diff
@@ -201,3 +203,17 @@ def test_automatically_create_application_registration_returns_false(workspace_r
...
+    expected_object_id = "Expected"
+
+    assert workspace_repo.get_workspace_owner(dictToTest, expected_object_id) is expected_object_id
+
+
+def test_workspace_owner_is_not_overwritten_if_not_present_in_workspace_properties(workspace_repo):
+    dictToTest = {"workspace_owner_object_id": "Expected"}
+    not_expected_object_id = "Not Expected"
+
+    assert workspace_repo.get_workspace_owner(dictToTest, not_expected_object_id) == "Expected"
```
LizaShak, line 219: I think it will be also good to have a positive test where you do have a workspace owner specified in the properties to make sure it is set correctly

e2e_tests/helpers.py
```diff
@@ -246,4 +246,4 @@ async def get_workspace_owner_token(admin_token, workspace_id, verify) -> Option
 
         token = responseJson["access_token"]
 
-        return token if (response.status_code == status.HTTP_200_OK) else None
+        return token, scope_uri
```
LizaShak, line 249: What should happen now if we get a response which is not 200?

templates/workspace_services/guacamole/version.txt
```diff
@@ -1 +1 @@
-__version__ = "0.3.0"
+__version__ = "0.3.1"
```
LizaShak, line 1: Shouldn't it be the same as in : templates/workspace_services/guacamole/porter.yaml? Mmmm I wonder how did it work until now, or maybe I  am missing something....

api_app/tests_ma/test_db/test_repositories/test_workpaces_repository.py
ross-p-smith: Maybe I misunderstood what you mean, but there are 2 new tests
1) Does not have `workspace_owner_object_id` in a dictionary (workspace properties), so therefore returns the default.
2) Does have `workspace_owner_object_id` in a dictionary (workspace properties), so ignores the default

e2e_tests/helpers.py
ross-p-smith: An exception is thrown a few lines above if there is no access token in the response. It would have never met a false condition `response.status_code == status.HTTP_200_OK`.

LizaShak (approved):

Issue #1722, opened 2022-04-21
Add adminuser into the docker group in the Resource Processor


ross-p-smith:
**Describe the bug**
The adminuser on the Linux box in the VMSS is not in the docker group - so you need sudo to work with docker
**Steps to reproduce**

1. Use bastion to get onto the Resource Processor instance within the VMSS
1. Type docker - frustration increases
1. Type `sudo docker` - it works

PR #1929, opened 2022-05-29
Reimage Resource Processor Automatically

### LINKED ISSUE #512: Improve  resource processor VMSS update behavior
**Description**
If the custom data for the resource processor VMSS is updated - for example new docker image. The instances need reimaging before the custom data and docker image is updated ( https://docs.microsoft.com/en-us/azure/virtual-machines/custom-data#can-i-update-custom-data-after-the-vm-has-been-created )

This behavior should be documented and the pipeline reimage existing VMSS instances using CLI.

https://docs.microsoft.com/en-us/cli/azure/vmss?view=azure-cli-latest#az_vmss_reimage

tamirkamara:
Closes #512, #1722 

## What is being addressed

When making a change to the resource processor image and/or config we had to remember to manually reimage the vmss instances.

## How is this addressed

- Include an az cli command to reimage the vmss within our terraform scripts that is triggered when custom data is changed.
- Small: add the vm user to the docker group so it has permissions to interact with the docker daemon. This saves us developers from using sudo when using docker cli.

### DIFF
--- a/templates/core/terraform/.terraform.lock.hcl
+++ b/templates/core/terraform/.terraform.lock.hcl
@@ -57,6 +57,25 @@ provider "registry.terraform.io/hashicorp/local" {
   ]
 }
 
+provider "registry.terraform.io/hashicorp/null" {
+  version = "3.1.1"
+  hashes = [
+    "h1:71sNUDvmiJcijsvfXpiLCz0lXIBSsEJjMxljt7hxMhw=",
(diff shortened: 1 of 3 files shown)


marrobi:
What happens to any jobs in progress?

tamirkamara:
> What happens to any jobs in progress?

I believe it will be stopped/cancelled and later try to be re-runed after the new VM comes online.

ross-p-smith (approved):

eladiw (approved):
lgtm

PR #1933, opened 2022-05-29
Optimize Guacamole docker image

tamirkamara:
## What is being addressed

Update Guacamole docker image to make it smaller

## How is this addressed

- Each archive is unpacked and deleted in a single layer
- OAtuth proxy unpacking moved to docker file to be run a single time in the build phase rather the runtime phase
- Update JRE version to 11.0.15_10
- Move builder (Maven) image source to regular rather openj9 we had before since we're not using this special JVM at runtime anyway
- Use certain values as ENV vars inside the docker file
- Pin tomcat version to 9.0.62 (current) so that it will not change in future builds without us knowing about it
- Add dockerfile linter


Makefile
```diff
@@ -241,6 +241,7 @@ lint:
 		-e VALIDATE_BASH=true \
 		-e VALIDATE_BASH_EXEC=true \
 		-e VALIDATE_GITHUB_ACTIONS=true \
+		-e VALIDATE_DOCKERFILE_HADOLINT=true \
```
eladiw, line 244: 👍

eladiw (approved):
lgtm. just verify it works

PR #1941, opened 2022-05-31
Cost Report - Tag Gitea shared service

LizaShak:
# PR for issue

## What is being addressed

Tag Gitea shared service with required tags for cost reporting.

## How is this addressed

- Add tre_shared_service_id and tre_id tags on Gitea TF scritpts


templates/shared_services/gitea/terraform/mysql.tf
```diff
@@ -14,14 +14,15 @@ resource "azurerm_mysql_server" "gitea" {
   administrator_login_password      = random_password.password.result
   sku_name                          = "GP_Gen5_2"
   storage_mb                        = 5120
-  version                           = "8.0"
+  version                           = "8.0.21"
```
martinpeck, line 17: I'm sure this is fine, but what's the reason for upgrading MySQL, and why this specific version?
This is newer than 8.0, but there are newer versions supported by Azure than this one.

It's also not clear whether this is a value value from a TF point of view. The TF docs suggest you can set this to 5.7 or 8.0 only. The docs also say that changing this value causes a new resource to be created, so are there migration steps needed for any of this?

martinpeck (commented):
Tagging stuff looks good to me.
I had some questions around the MySQL version bump.

templates/shared_services/gitea/terraform/gitea-webapp.tf
```diff
@@ -10,6 +10,7 @@ resource "random_password" "gitea_passwd" {
 resource "azurerm_user_assigned_identity" "gitea_id" {
   resource_group_name = local.core_resource_group_name
   location            = data.azurerm_resource_group.rg.location
+  tags                = local.tre_core_tags
```
guybartal, line 13: ```gitea``` is a shared service and not core service, I would call this local ```tre_shared_service_tags```

guybartal (changes requested):
besides the local name, LGTM.

templates/shared_services/gitea/terraform/mysql.tf
martinpeck: I agree...this is strange.

The portal offers 5.7 or 8.0:

image

... the swagger for the rest api specifies 8.0.21...

image
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/mysql/resource-manager/Microsoft.DBforMySQL/stable/2021-05-01/mysql.json

... while the DOCS suggest that different versions are supported:

https://docs.microsoft.com/en-us/azure/mysql/single-server/concepts-supported-versions

image

martinpeck: My concern is really that the current deployments specific 8.0. If we change this to anything else, will this cause an existing DB to be destroyed/recreated? If it does, migration needs to be addressed.

Can we tell tf lint to ignore this?

LizaShak: Yes. Exactly what I did. I have also opened a [new issue](https://github.com/microsoft/AzureTRE/issues/1944) to remove it in the future once this bug is fixed.

martinpeck: Perfect. Thanks @LizaShak :)

martinpeck (approved):
LGTM! Thanks @LizaShak

Issue #831, opened 2021-09-13
Create E2E tests for user resource


LizaShak:
Origin story: #623 

**Description**
Add an E2E for deploying a user resource, validating that it is deployed and updated successfully. Disabled and deleted at the end.

msebragge:
Waiting for user resource registration.

christoferlof:
@msebragge , please link the issue blocking this one for reference, thanks.

LizaShak:
@marrobi   I think we can remove the blocked label here right?

marrobi:
Agreed, and removed.

PR #1952, opened 2022-05-31
Create user_resource in e2e tests

### LINKED ISSUE #1951: Deleting Guacamole VM results in KeyVault purge protection error
**Describe the bug**
Unable to delete a Guacamole VM because of the key vault purge protection settings

**Steps to reproduce**

1. Create a tre-service-guacamole-windowsvm
1. Delete it from the Swagger Endpoint
.

Error context message = \x1b[31m╷\x1b[0m\x1b[0m \x1b[31m│\x1b[0m \x1b[0m\x1b[1m\x1b[31mError: \x1b[0m\x1b[0m\x1b[1mpurging of Secret "windowsvm7cd6-admin-credentials" (Key Vault "https://xxxxxx-ws-765b.vault.azure.net/") : keyvault.BaseClient#PurgeDeletedSecret: Failure responding to request: StatusCode=403 -- Original Error: autorest/azure: Service returned an error. Status=403 Code="Forbidden" Message="Operation \\"purge\\" is not allowed because purge protection is enabled for this vault. Key Vault service will automatically purge it after the retention period has passed

ross-p-smith:
# PR for issue #831

## What is being addressed

Added the code to create a user_resource in the extended e2e tests

## How is this addressed

- Added the code and tested
- Fixes #1951


damoodamoo (approved):

templates/core/.env.sample
```diff
@@ -5,6 +5,8 @@ TRE_ADDRESS_SPACE="10.0.0.0/12"
 DEPLOY_GITEA=true
 DEPLOY_NEXUS=true
 RESOURCE_PROCESSOR_TYPE="vmss_porter"
+API_APP_SERVICE_PLAN_SKU_SIZE="P1v2"
```
damoodamoo, line 8: are these actually used? can't see where?

ross-p-smith: The default is set in the Terraform as P1v3. Setting this in the env file basically means the developers of TRE will use a SKU that does not encounter Quota issues. The subscription that the build works in has quota setup

PR #1955, opened 2022-06-01
Remove escaped quotes from etag

### LINKED ISSUE #1931: etag header marked as not required in API documentation
etag is required for patch operations, but not marked so in the api documentation
Also, it comes escaped in the resources GET response 

**Steps to reproduce**

1. Try to PATCH a workspace (or shared service)
1.  Do not specify etag (as documentation states)
1. Receive a 'etag is required' response

image

SharonHart:
# PR for issue https://github.com/microsoft/AzureTRE/issues/1931

etag is saved in CosmosDB SQL API with escaped quotes (for example a plain document on a new database, container):
image
There is no apparent way to change it, and even if there is, it will require a db migration as this is also the default indexing policy. 

## How is this addressed

- Using pydantic validator, the etag value is stripped from the escaped quotes during deserialization. 

Fixes #1931


damoodamoo:
just curious - do we actually want to do this? Do we want to care about how Cosmos stores the value? Would we need to add escaped quotes back in when we send the etag to Cosmos?

SharonHart:
@damoodamoo the other end of sending the etag remains the same. This is not fixing the way cosmos saves thd etag, just yhe way it deserializes to a json. Unless we do it, the client of the API would need to know that its escaped and work to unescape it

damoodamoo:
Ok - guess it makes sense then, just didn't occur to me as the value is only ever blindly copied from the GET into the PATCH.
________________________________

tamirkamara (approved):
LGTM

tamirkamara:
/test
extended passed here: https://github.com/microsoft/AzureTRE/actions/runs/2685571353

Issue #1910, opened 2022-05-24
MVP Snagging List


damoodamoo:
- [ ] "Each row needs a unique key" bug in History details list
- [ ] Spinner needed on Form sending 
- [ ] "Some icons were reregistered" warning
- [ ]

PR #1956, opened 2022-06-01
UI Enhancements

damoodamoo:
# PR for issue #1910 

Included:
- Moved create form to central location, use context to control it
- Initial logic for Updates via the create form - likely some stability improvements needed
- Command bar + header for each resource type, so a user can update it from the page they're on
- Some bug fixes around missing keys etc
- Integrated Operations
- Style updated to use some light drop shadows to differentiate areas of the page
- Some restructuring + refactoring

image


ui/app/src/components/root/LeftNav.tsx
```diff
@@ -16,7 +16,7 @@ const navLinkGroups: INavLinkGroup[] = [
         url: '/shared-services',
         key: 'shared-services',
         icon: 'Puzzle'
-      },
```
ross-p-smith, line 19: nit: Should we just delete this?

ross-p-smith (approved):

ui/app/src/components/root/LeftNav.tsx
damoodamoo: Gonna leave them in for now as the next stage will bring them back, and I just want the icon names still :)

PR #1958, opened 2022-06-01
azurerm_app_service_plan is deprecated and we should use azurerm_service_plan

### LINKED ISSUE #1815: Upgrade azurerm_app_service_plan

ross-p-smith:
Fixes #1815 

## What is being addressed

azurerm_app_service_plan is deprecated and we should use azurerm_service_plan

## How is this addressed

- This updates the core resource
- We will upgrade each workspace service in separate PR, so as we can test them individually and add e2e test
- Each resource needs to be AzureRM 3.5


eladiw (approved):
lgtm. what happens if there are no resources to migrate? will the migrate script skip?

ross-p-smith:
> lgtm. what happens if there are no resources to migrate? will the migrate script skip?

Ideally, if there are no resources to migrate, it will just skip. However that's where the bug is at the moment (if there is NO state file, it breaks) :-(

PR #1969, opened 2022-06-02
Cleanup bundle dockerfiles

### LINKED ISSUE #1962: Standardize Porter's docker file templates
## Description

As a TRE Developer I want to have all bundles be as consistent and small as possible

## Acceptance criteria

- [ ] Bundles use the same base image where possible
- [ ] Bundles install az cli in the same way
- [ ] Bundles have apt-get cleanup procedure
- [ ] Bundles don't install things they don't need

tamirkamara:
Fixes #1962 

## What is being addressed

Our bundle docker files were inconsistent and included things we didn't need to deploy a application.
The changes below reduce the installer size from 10MB to a few 100s MB. 

## How is this addressed

- Base image standardized to debian:stretch-slim where possible
- Remove components like jq or az-cli we didn't use in the image
- Include an apt-get cleanup step
- Delete the template dockerfile where there's nothing special needed for an image
- Reformat some porter.yaml (line breaks)
- Include a type for the arm_use_msi parameter to remove a warning from porter
- Include --no-install-recommends on apt-get install commands
- Pin versions installed by apt-get
- Remove Porter's az mixin if not in use


templates/shared_services/certs/Dockerfile.tmpl
```diff
@@ -1,22 +1,25 @@
...
 
 ARG BUNDLE_DIR
 
-RUN apt-get update \
-    && apt-get install -y ca-certificates \
-    && apt-get clean -y && rm -rf /var/lib/apt/lists/*
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
 # Install Azure CLI
+# It's useless to specify azcli version since the mixin installs the latest anyway
```
ross-p-smith, line 8: You don't specify the Az CLI version in this dockerfile, but you do in nexus. What is the difference between the porter mixin

ross-p-smith (approved):

ross-p-smith:
I'm assuming you are not going to test the InnerEye stuff?

templates/shared_services/certs/Dockerfile.tmpl
tamirkamara: If the az mixin is installed then it uses the latest version anyway which overrides anything I will specify here. Nexus doesn't install the mixin which makes it possible to put in a version there, but not here...

There's a PR in progress on the mixin to support versions which will enable us to fix this as well.

tamirkamara:
> I'm assuming you are not going to test the InnerEye stuff?

@ross-p-smith Right. The change there is minimal and only removes a warning Porter outputs...

Issue #1830, opened 2022-05-12
Enable purge protection


tanya-borisova:
TRE customers are usually expected to enable keyvault purge protection. 

However, in the repo we have a flag `keyvault_purge_protection_enabled`, which has been useful in development scenarios, as we were often recreating resources in Azure. 

Now, in the azurerm Terraform provider configuration,one can now set the following:

```

provider "azurerm" {
  features {
    key_vault {
      # Don't purge secrets on destroy (this would fail due to purge protection being enabled on keyvault)
      purge_soft_deleted_secrets_on_destroy = false
      # When recreating a shared service, recover any previously soft deleted secrets
      recover_soft_deleted_secrets = true
    }
  }
}
```

(See also PR https://github.com/microsoft/AzureTRE/pull/1777)

With this, there should not be a problem having purge protection on, because Terraform should not purge secrets on deletion, and should restore previously soft deleted secrets.  So we should have purge protection on. 

Purge protection is already enabled in https://github.com/microsoft/AzureTRE/pull/1797 for workspace keyvault, in this ticket enable it for the keyvault in core as well. 

- [x] #1925

tanya-borisova:
The blocker on this work is that TLS certificate cannot be recovered after being deleted. The crux of the problem is described in this bug report for azurerm: https://github.com/hashicorp/terraform-provider-azurerm/issues/17168  

There might be a workaround possible if we don't import a temporary certificate and instead let Azure create one for us from parameters. I will do a small timeboxed investigation into this.

PR #1973, opened 2022-06-06
Enable purge protection

tanya-borisova:
# PR for issue https://github.com/microsoft/AzureTRE/issues/1830

## What is being addressed

Enable purge protection for all keyvaults used (see the issue linked). 

## How is this addressed

- Enable purge protection for all keyvaults
- Set `azurerm` to not purge soft deleted objects on destroy (all of them) 
- Set `azurerm` to recover soft deleted secrets (which is set to true by default)
- Remove all references to `TF_VAR_keyvault_purge_protection_enabled`


ross-p-smith:
I think you should add the azurerm purge settings into the features block in the /workspaces/AzureTRE/templates/workspace_services/guacamole/user_resources folders. Basically any module that creates a azurerm_key_vault_secret will need to have your settings

tanya-borisova:
I might have missed it, thanks @ross-p-smith 
I'm currently debugging the core anyways...

tanya-borisova:
I've been testing creation, deletion then creation of the TRE (as this is where most errors around purge protection usually manifest themselves). 
Keyvault has keys, secrets and certificates. Keys and secrets get deleted and then re-created as expected, however the letsencrypt certificate doesn't seem to work. When creating, deleting and then creating core of TRE, I'm getting the following error:

image

Even though I state to recover soft deleted certificates [here](https://github.com/microsoft/AzureTRE/pull/1973/files#diff-6b9abdd4e97e9b0e65bc5fbb28b4fdecb6fa5ea8fdfdd296beb5645adc235424), same as for keys and secrets. 

This looks like a bug although I haven't been able to make a smaller reproducible example. The way forward is unclear. I've spent time trying to import the letsencrypt cert into the state instead of creating it, but long story short that didn't work either, as for importing one needs a version of the cert (I can further elaborate on that if anybody is curious). The cert seems to be necessary in terraform, as without it it doesn't seem possible to create AppGateway that is using HTTPS... 
Seems like a dead end :(

tanya-borisova:
I'm going to close this as there _seems_ to be a bug in azurerm implementation of the recover_soft_deleted_secrets flag implementation. I filed the bug here: https://github.com/hashicorp/terraform-provider-azurerm/issues/17168

tanya-borisova:
I've tested locally destroying then recreating core -- everything seems to work

devops/scripts/destroy_env_no_terraform.sh
```diff
@@ -9,7 +9,7 @@
 
 set -o errexit
 set -o pipefail
-# set -o xtrace
+set -o xtrace
```
ross-p-smith, line 12: comment this back out?

templates/core/terraform/appgateway/certificate.tf
```diff
@@ -16,25 +16,43 @@ resource "azurerm_key_vault_certificate" "tlscert" {
...
+        "keyAgreement",
+        "keyCertSign",
+        "keyEncipherment",
+      ]
+
+      subject_alternative_names {
+        dns_names = ["internal.contoso.com", "domain.hello.world"]
+      }
+
+      subject            = "CN=hello-world"
```
marrobi, line 53: can the subject match the TRE FQDN?

templates/core/terraform/appgateway/certificate.tf
```diff
@@ -16,25 +16,43 @@ resource "azurerm_key_vault_certificate" "tlscert" {
...
+        "cRLSign",
+        "dataEncipherment",
+        "digitalSignature",
+        "keyAgreement",
+        "keyCertSign",
+        "keyEncipherment",
+      ]
+
+      subject_alternative_names {
+        dns_names = ["internal.contoso.com", "domain.hello.world"]
```
marrobi, line 50: do we need subject alternate names?

devops/scripts/destroy_env_no_terraform.sh
```diff
@@ -111,17 +111,43 @@ if [[ -n ${SHOW_KEYVAULT_DEBUG_ON_DESTROY:-} ]]; then
...
+
+  certificates=$(az keyvault certificate list --vault-name "${keyvault_name}" | jq -r '.[].id')
+  for certificate_id in ${certificates}; do
+    az keyvault certificate delete --id "${certificate_id}"
+  done
 
+  echo "Removing access policies so if the vault is recovered there are not there"
+  access_policies=$(echo "$keyvault" | jq -r '.properties.accessPolicies[].objectId' )
+  for access_policy_id in ${access_policies}; do
+    az keyvault delete-policy --name "${keyvault_name}" --resource-group "${core_tre_rg}" --object-id "${access_policy_id}" || echo "Not deleting access policy for ${access_policy_id}."
```
ross-p-smith, line 136: if it cant delete the policy, won't this write 1. Not deleting access policy for blah and then 2. Access policy deleted

marrobi (approved):
Couple of comments on the cert - might be useful to use the actaul FQDN, then if people want to trust the cert on their local machine they can.

Providing happy this works, along with destroy, and recreate etc. LGTM.

templates/core/terraform/appgateway/certificate.tf
tanya-borisova: Very good point -- will change that!

templates/core/terraform/appgateway/certificate.tf
tanya-borisova: Apparently we do, if we want to trust this cert on our local machines. Otherwise I get this: 

image

tanya-borisova:
/test-force-approve

I can't test the deletion and recreation of the environment in CI, because it uses destroy_env_no_terraform.sh from main branch and not from mine. I have tested this quite a lot on development TREs though so this should be okay.

PR #1975, opened 2022-06-06
Create Application Administrator

ross-p-smith:
# PR for issue 1716

## What is being addressed

Amend the `make auth` process to create a Service Principal with Application.ReadWrite.All

## How is this addressed

- Create new bash script to create the service principal.
- Chosen not to make the existing script anymore complicated than it needs to be
- Removed any Application.ReadWrite.All logic from the API App


ross-p-smith:
/test-force-approve

There are no tests for this area. You would need to run `make auth` with AUTO_WORKSPACE_APP_REGISTRATION set to true in your env file

scripts/aad/create_application_administrator.sh
```diff
@@ -0,0 +1,191 @@
...
+fi
+
+# Get the directory that this script is in
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+declare grantAdminConsent=0
+declare currentUserId=""
+declare spId=""
+declare msGraphUri="https://graph.microsoft.com/v1.0"
+declare appName="sp_application_admin"
```
marrobi, line 38: Feel this should be TRE specific so know which TRE carried out the operation.

marrobi (approved):
LGTM - one comment re naming.

scripts/aad/create_application_administrator.sh
ross-p-smith: Done!

PR #1978, opened 2022-06-07
Airlock processor handles request Submission

### LINKED ISSUE #1836: Airlock processor handles Request submission
As an Airlock Orchestrator in TRE
I want to be able to consume an Airlock status change signaling submission
So that the process is started and the process status change

**Acceptance Criteria**
- [ ] An event for Airlock process submission is consumed
- [ ] The uploaded data file (zip) is copied from the `External storage` to the `In-progress storage`

~~- [ ] Malware Scanning started with blob storage events~~
~~- [ ] Malware Scanning reports back with Malware Scanning Result~~

eladiw:
# Fixes #1836 

## What is being addressed

Adding support for the StatusChangedEvent in the Airlock processor (for submission - import and export)
Consolidated the blob created queues into a single SB topic
Wired all TF scripts
Added support for the blobCreatedEvent in the AL processor

## How is this addressed

- Added code to extract the msg from the status changed queue
- If data copy is required, extract the source and dest params, and issue a copy command
- added unit tests

How to test:
1. Create container in the import external storage. Give it a name: e.g. 123
2. Put a single file
3. Place this message in the status_changed queue

{ "data": {
    "request_id" : "123",
    "type":"import", 
    "status":"submitted" , 
    "workspace_id":"124" 
    }
}

Check the in progress storage account , it should contain a new container with the same file.
Check the step result queue, you should find a step result event which states the request is now in-progress (transitioned from submitted)


airlock_processor/BlobCreatedTrigger/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+    logging.info("Python ServiceBus topic trigger processed message - A new blob was created!.")
+    body = msg.get_body().decode('utf-8')
+    logging.info('Python ServiceBus queue trigger processed message: %s', body)
+
+    json_body = json.loads(body)
+    # message is due to blob creation in an 'in-progress' blob
+    if "stalipim" in json_body["topic"]:
+        completed_step = "submitted"
+        new_status = "in-progress"
+        request_id = re.search(r'/blobServices/default/containers/(.*?)/blobs', json_body["subject"]).group(1)
```
damoodamoo, line 22: is the request_id not in the service bus message, rather than having to find it here?

airlock_processor/BlobCreatedTrigger/function.json
```diff
@@ -0,0 +1,21 @@
+{
+  "scriptFile": "__init__.py",
+  "entryPoint": "main",
+  "bindings": [
+    {
+      "name": "msg",
+      "type": "serviceBusTrigger",
```
damoodamoo, line 7: if this is triggered by a service bus message why are we calling the function `BlobCreatedTrigger` ?

airlock_processor/StatusChangedQueueTrigger/__init__.py
```diff
@@ -1,18 +1,113 @@
...
+        new_status = json_body["new_status"]
+        request_type = json_body["type"]
+        tre_id = json_body["tre_id"]
+        workspace_id = json_body["workspace_id"]
+    except KeyError:
+        raise
+    except json.decoder.JSONDecodeError:
+        raise
+
+    return req_id, new_status, request_type, tre_id, workspace_id
```
damoodamoo, line 46: is it worth returning a defined object rather than 5 individual properties?

airlock_processor/shared_code/blob_operations.py
```diff
@@ -0,0 +1,44 @@
...
+    # Copy files
+    source_blob_service_client = BlobServiceClient.from_connection_string(sa_source_connection_string)
+    dest_blob_service_client = BlobServiceClient.from_connection_string(sa_dest_connection_string)
+
+    source_container_client = source_blob_service_client.get_container_client(container_name)
+
+    try:
+        blob_name = ""
+        for blob in source_container_client.list_blobs():
+            if blob_name != "":
```
damoodamoo, line 28: think we should use a count / length property to determine if there's more than 1 blob in the list?

airlock_processor/shared_code/blob_operations.py
```diff
@@ -0,0 +1,44 @@
...
+                msg = "Request with id {} contains more than 1 file. flow aborted.".format(request_id)
+                logging.error(msg)
+                raise AirlockInvalidContainerException(msg)
+            blob_name = blob.name
+
+        if blob_name == "":
```
damoodamoo, line 34: Same here - can we use a `if count == 0` type logic?

airlock_processor/shared_code/blob_operations.py
```diff
@@ -0,0 +1,44 @@
...
+            logging.info('Request with id %s did not contain any files. flow aborted.', request_id)
+
+    except Exception:
+        logging.error('Request with id %s failed.', request_id)
+        raise()
+
+    source_blob = source_container_client.get_blob_client(blob_name)
+    source_url = source_blob.url + '?' + sas_token
```
damoodamoo, line 42: bit more readable as `f'{source_blob.url}?{sas_token}'`

damoodamoo (commented):
I'm missing lots of airlock context, but done my best to review. 

My main concern would be using any blob triggers to determine stage transitions - as this will pretty much pin us to one file and require a lot of refactoring if we wanted to support multiple in future.

airlock_processor/shared_code/blob_operations.py
eladiw: it is a lazy generator. so each time you iterate it fetches the info. so there is no simple  property to get it....

airlock_processor/shared_code/blob_operations.py
eladiw: can't because it is a lazy iterator

airlock_processor/BlobCreatedTrigger/__init__.py
eladiw: it is not. the reason is that since blob created events are originated from Azure, we can't control the schema.
But we do have the container it was triggered from. This is why I extract this info from there

airlock_processor/local.settings.json-sample
```diff
@@ -0,0 +1,12 @@
+{
+  "IsEncrypted": false,
+  "Values": {
+    "FUNCTIONS_WORKER_RUNTIME": "python",
+    "AIRLOCK_STATUS_CHANGED_QUEUE_NAME": "status_changed",
+    "SB_CONNECTION_STRING": "Endpoint=sb://XXXX.servicebus.windows.net/;SharedAccessKeyName=.....",
+    "AZURE_SUBSCRIPTION_ID": "",
+    "ENABLE_LOCAL_DEBUG": "true",
```
tamirkamara, line 8: Can this be using an env var we already have (like DEBUG or ENABLE_LOCAL_DEBUGGING)?

airlock_processor/StatusChangedQueueTrigger/__init__.py
```diff
@@ -1,18 +1,136 @@
...
+
+def get_source_dest_env_vars(new_status: str, request_type: str, workspace_id: str):
+
+    # sanity
+    if is_require_data_copy(new_status) is False:
+        raise Exception("Given new status is not supported")
+
+    try:
+        tre_id = os.environ["TRE_ID"]
+        subscription_id = os.environ["AZURE_SUBSCRIPTION_ID"]
```
tamirkamara, line 81: Knowing which subscription this runs in looks odd. Why do we need it?

airlock_processor/StatusChangedQueueTrigger/__init__.py
```diff
@@ -1,18 +1,136 @@
...
+    except KeyError as e:
+        logging.error(f'Missing environment variable: {e}')
+        raise
+
+    request_type = request_type.lower()
+    if request_type != "import" and request_type != "export":
+        raise Exception("Request type must be either import or export")
+
+    if new_status == 'submitted' and request_type == 'import':
+        source_account_name = "stalexim{}".format(tre_id)
```
tamirkamara, line 91: Can you think about a way to store all patterns together and maybe work with ENUMs to ask for them?

airlock_processor/StatusChangedQueueTrigger/__init__.py
```diff
@@ -1,18 +1,136 @@
...
+        # https://github.com/microsoft/AzureTRE/issues/1842
+        pass
+
+
+    if os.environ.get('ENABLE_LOCAL_DEBUG', NONE) is not NONE:
+        # Using the logged-in user
+        credential = DefaultAzureCredential()
+    else:
+        logging.info("using the Airlock processor's managed identity to get build storage management client")
+        credential = DefaultAzureCredential(managed_identity_client_id=os.environ["MANAGED_IDENTITY_CLIENT_ID"], exclude_shared_token_cache_credential=True)
```
tamirkamara, line 120: Why is there a difference in the auth mode?

airlock_processor/shared_code/blob_operations.py
```diff
@@ -0,0 +1,44 @@
...
+            logging.info('Request with id %s did not contain any files. flow aborted.', request_id)
+
+    except Exception:
+        logging.error('Request with id %s failed.', request_id)
+        raise()
+
+    source_blob = source_container_client.get_blob_client(blob_name)
+    source_url = f'{source_blob.url}?{sas_token}'
+
+    dest_blob_service_client.get_blob_client(container_name, source_blob.blob_name).start_copy_from_url(source_url)
```
tamirkamara, line 44: Does this return any operation id? If yes, we should always log it...

airlock_processor/shared_code/blob_operations.py
```diff
@@ -0,0 +1,44 @@
...
+from azure.storage.blob import ContainerSasPermissions, generate_container_sas, BlobServiceClient
+
+from exceptions.AirlockInvalidContainerException import AirlockInvalidContainerException
+
+
+def copy_data(source_account_name: str, source_account_key: str, sa_source_connection_string: str, sa_dest_connection_string: str, request_id: str):
+    container_name = request_id
+
+    # token geneation with expiry of 1 hour. since its not shared, we can leave it to expire (no need to track/delete)
+    sas_token = generate_container_sas(account_name=source_account_name,
```
tamirkamara, line 13: A SAS is usually used for operations without direct access to the storage, which we have in this case, right?
Isn't there a way to copy the blob without using it?

templates/core/terraform/main.tf
```diff
@@ -83,6 +83,7 @@ module "airlock_resources" {
   docker_registry_server   = var.docker_registry_server
   mgmt_resource_group_name = var.mgmt_resource_group_name
   mgmt_acr_name            = var.acr_name
+  arm_subscription_id      = var.arm_subscription_id
```
tamirkamara, line 86: I think this exists already

templates/core/terraform/airlock/variables.tf
```diff
@@ -25,3 +25,9 @@ variable "mgmt_acr_name" {
   type        = string
   description = "Management ACR name"
 }
+
+variable "arm_subscription_id" {
+  description = "The subscription id to create the resource processor permission/role. If not supplied will use the TF context."
```
tamirkamara, line 30: Doesn't look like you use it for permission/role. Is the description up to date?

airlock_processor/StatusChangedQueueTrigger/__init__.py
eladiw: This is what the ctor asks for
https://docs.microsoft.com/en-us/python/api/azure-mgmt-storage/azure.mgmt.storage.storagemanagementclient?view=azure-python#constructor

airlock_processor/BlobCreatedTrigger/function.json
damoodamoo: ok, makes sense with the context, ta.

airlock_processor/shared_code/blob_operations.py
damoodamoo: ok, understood - but might be clearer to read if we added a `blob_count` variable that started at 0 and incremented, then we could use that for checks?

templates/core/terraform/main.tf
eladiw: we talked offline

templates/core/terraform/airlock/variables.tf
eladiw: fixed

airlock_processor/shared_code/blob_operations.py
eladiw: fixed

airlock_processor/StatusChangedQueueTrigger/__init__.py
eladiw: fixed

airlock_processor/StatusChangedQueueTrigger/__init__.py
eladiw: fixed

eladiw:
@damoodamoo @tamirkamara  fixed all comments

airlock_processor/shared_code/blob_operations.py
eladiw: Added a new issue to fix this (if possible). I tried to remove it (simple approach) and it failed

airlock_processor/StatusChangedQueueTrigger/__init__.py
```diff
@@ -1,18 +1,141 @@
...
+    return result
+
+
+def is_require_data_copy(new_status: str):
+    if new_status.lower() in [constants.STAGE_SUBMITTED, constants.STAGE_APPROVED, constants.STAGE_REJECTED, constants.STAGE_BLOCKED]:
+        return True
+    return False
+
+
+def get_source_dest_env_vars(new_status: str, request_type: str, short_workspace_id: str) -> ContainersCopyMetadata:
```
tamirkamara, line 81: IMO the method name isn't telling too much. Given it returns the copy metadata, does it make sense to call it get_copy_operation_metadata or something similar?

tamirkamara (approved):

tamirkamara (approved):
Added a couple of small comments. Up to you...

airlock_processor/StatusChangedQueueTrigger/__init__.py
eladiw: will fix in my next pr

airlock_processor/shared_code/constants.py
```diff
@@ -0,0 +1,17 @@
...
+
+# Export
+STORAGE_ACCOUNT_NAME_EXPORT_INTERNAL = "stalexintws{}"
+STORAGE_ACCOUNT_NAME_EXPORT_INPROGRESS = "stalexipws{}"
+
+# Stages
+STAGE_SUBMITTED = "submitted"
+STAGE_APPROVED = "approved"
+STAGE_REJECTED = "rejected"
+STAGE_BLOCKED = "blocked"
```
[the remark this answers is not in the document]
eladiw: yes. will fix in my next pr

PR #1979, opened 2022-06-07
[cost] Tag firewall and nexus shared services.

LizaShak:
# PR for issue

## What is being addressed

Add relevant tags for Firewall and nexus shared services.

## How is this addressed

- Add tags on each resource TF deployment


guybartal (approved):
LGTM

Issue #1980, opened 2022-06-07
UI Static web app should sit behind application gateway


marrobi:
## Description

As a TRE Administrator
I want to access the UI via the Application Gateway 
So that it can be secured using the same DNS name and SSL and any ip restrictions as the API

## Acceptance criteria
- [ ] UI available on application gateway DNS name

PR #1983, opened 2022-06-07
Move UI to blob storage behind app gw.

marrobi:
# PR for issue #1980

## What is being addressed
 - Move UI to blob storage behind app gw and remove static web app
 - Change makefile to upload to blob storage rather than using swa cli
 - remove swa cli from dev container
 - update auth scripts to use a tre-url so that both root UI and swagger redirect urls can be added
 - update docs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -349,7 +349,7 @@ static-web-upload:
 	&& ${MAKEFILE_DIR}/templates/core/terraform/scripts/upload_static_web.sh
 
 build-and-deploy-ui:
-	$(call target_title, "Use Static Web App CLI to deploy UI") \
+	$(call target_title, "Build and deploy UI") \
 	&& . ${MAKEFILE_DIR}/devops/scripts/check_dependencies.sh nodocker \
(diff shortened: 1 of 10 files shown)


damoodamoo (approved):

Makefile
```diff
@@ -349,15 +349,16 @@ static-web-upload:
 	&& ${MAKEFILE_DIR}/templates/core/terraform/scripts/upload_static_web.sh
 
 build-and-deploy-ui:
-	$(call target_title, "Use Static Web App CLI to deploy UI") \
+	$(call target_title, "Build and deploy UI") \
```
damoodamoo, line 352: does the static-web-upload task need to be conditional on `DEPLOY_UI` ? Need to be able to opt out of it

marrobi: Good point, as the first script will exit? But the following make command will still run?

marrobi: Called the second from the first.

damoodamoo (approved):

PR #1987, opened 2022-06-07
Register VM Bundle for E2E tests

ross-p-smith:
# PR for issue #831

## What is being addressed

We cant create a user_resource in the E2E as none are registered

## How is this addressed

- Register a user_resource bundle type


marrobi (approved):
LGTM

PR #1989, opened 2022-06-07
Registering a user_resource needs the Workspace Service Name

ross-p-smith:
# PR for issue #831

## What is being addressed

The build calls `make bundle-register` which needs the Workspace Service Name

## How is this addressed

- Add this to the matrix


.github/actions/devcontainer_run_command/action.yml
```diff
@@ -104,6 +104,9 @@ inputs:
   BUNDLE_TYPE:
     description: "The Bundle type (workspace / Workspace-service / User Resource)."
     required: false
+  WORKSPACE_SERVICE_NAME:
```
stuartleeks, line 107: Does this also need to be passed through in the `docker run` below?

stuartleeks (approved):
LGTM

PR #1995, opened 2022-06-08
Checking Bundle's parameter.json file

### LINKED ISSUE #1992: Verify bundles can be installed locally
## Description

As a TRE Developer
I want to install a bundle locally
So that I can test my work locally before integrating it with the whole system and save time
(we rely too much on the CI to check things we are able to do locally)

## Acceptance criteria

- [ ] Bundle parameters.json files are up-to-date
- [ ] A validation step is added to make sure these are kept up-to-date with every change

tamirkamara:
Fixes #1992 

## What is being addressed

Parameters.json file is crucial for a developer to be able to deploy a bundle from their own machine, but is often overlooked and not kept up-to-date.

## How is this addressed

- Sync all parameter.json files per their respective porter.yaml
- Add a makefile target that will validate that the file is up-to-date when building & installing the bundle
- Use the new target in the CI to fail a run if there's a mismatch


tamirkamara:
/test-force-approve
The following test shows the right changes in the "Publish Bundle" step (although the rest of the job failed)

PR #1996, opened 2022-06-08
UI Snagging + Migration

damoodamoo:
# PR for issue #1910 

- DB migration to add `deploymentStatus` field from latest operation for each resource
- Securing buttons to roles
- Fixing create form state regression


api_app/db/migrations/resources.py
```diff
@@ -0,0 +1,24 @@
...
+        for op in operations_repository.query("SELECT * from c ORDER BY c._ts ASC"):
+            try:
+                resource = self.get_resource_by_id(uuid.UUID(op['resourceId']))
+                resource.deploymentStatus = op['status']
+                self.update_item(resource)
+            except EntityDoesNotExist:
+                logging.info(f'Resource Id {op["resourceId"]} not found')
+                # ignore errors and try the next one
+
+        return True
```
ross-p-smith, line 24: nit: Is it worth returning a boolean if it's always going to be true? You never test for it. In the calling code you don't test and always set a status of Executed? Is it worth returning a number of how many rows have been updated so you could add that into the status. That might be useful

api_app/api/routes/migrations.py
```diff
@@ -43,6 +47,10 @@ async def migrate_database(resources_repo=Depends(get_repository(ResourceReposit
         migration_status = "Executed" if workspace_migration.moveAuthInformationToProperties() else "Skipped"
         migrations.append(Migration(issueNumber="PR 1726", status=migration_status))
 
+        logging.info("#1406 - Extra field to support UI")
+        resource_migration.add_deployment_status_field(operations_repo)
+        migrations.append(Migration(issueNumber="1406", status="Executed"))
```
ross-p-smith, line 52: Maybe put something more useful than always "Executed" (e.g. number of rows updated)

ross-p-smith (approved):

PR #1998, opened 2022-06-08
[cost] Tag innereye Workspace Service in Terraform

### LINKED ISSUE #1894: Tag innereye  Workspace Service in Terraform
As a TRE developer,
I want innereye workspace service to be tagged
So that I can include costs in TRE Cost Reporting

**Acceptance Criteria**
- [ ] Include `tre_id`, `tre_workspace_id`, `tre_workspace_service_id` in innereye service

pedro-pelegrin-nttdata:
Fixes #1894 

## What is being addressed

Tagging innereye Workspace Service with tre_id, tre_workspace_id and tre_workspace_service_id

## How is this addressed

- Added tre_workspace_tags to locals.tf
- Added variable tre_workspace_tags to variables.tf
- Added tags to tf files


tamirkamara (changes requested):
Will take it offline with @pedro-pelegrin-nttdata

guybartal:
Did you try to deploy it?
looks like you use both local and variable in the same module but you never set the var,
if there's no hierarchy of modules you should simply use locals.

tamirkamara (changes requested):
@pedro-pelegrin-nttdata thank you for going over all these services. Since all items here are workspace services I think you should use tre_workspace_service_tags as the name for the local/variable.

guybartal (approved):
LGTM

PR #2003, opened 2022-06-08
[cost] Create Cost Reporting API stubs

### LINKED ISSUE #1767: Create Cost Reporting API stubs

guybartal:
# Fixes #1767 

## What is being addressed

* Cost report domain model
* Cost report APIs stubs for ```/api/costs``` and ```/api/workspaces/<workspace_id>/costs``` endpoints (will be removed after API implementation)
* Cost report APIs swagger with examples


api_app/api/routes/costs.py
```diff
@@ -0,0 +1,37 @@
...
+from api.dependencies.workspaces import get_workspace_by_id_from_path
+from services.authentication import get_current_admin_user, get_current_workspace_owner_user
+from db.repositories.workspaces import WorkspaceRepository
+from db.repositories.workspace_services import WorkspaceServiceRepository
+from db.repositories.shared_services import SharedServiceRepository
+from models.domain.costs import CostReport, GranularityEnum, WorkspaceCostReport, generate_cost_report_stub, generate_workspace_cost_report_stub
+
+
+costs_core_router = APIRouter(dependencies=[Depends(get_current_admin_user)])
+costs_workspace_router = APIRouter(dependencies=[Depends(get_current_workspace_owner_user)])
```
LizaShak, line 15: add admin to be able to pull workspace cost

api_app/models/domain/costs.py
```diff
@@ -0,0 +1,187 @@
...
+def generate_cost_row_example(granularity: GranularityEnum):
+    return dict({
+        "cost": random.uniform(0, 1000), "currency": "USD", "date":
+            (datetime.today() - timedelta(
+                days=-1 * random.randint(0, 1000))).date() if granularity == GranularityEnum.daily else None
+    })
+
+
+def generate_cost_item_example(name: str, granularity: GranularityEnum):
+    cost_item = dict(
```
LizaShak, line 23: unclear name because it doesn't return cost item - in all example methods

api_app/api/routes/costs.py
guybartal: Hi @marrobi, @LizaShak raises here something that we need to decide upon:
cost report for WS level should be available both for Admin and WS owners,
as we see it there are two options:

1. Duplicate ```/api/workspaces/{workspace_id}/cost endpoint``` and have it also inside the ```cost_core_router``` (main /api)
2. Add admin user to each and every WS app with WS Owner role

What do we prefer?

guybartal: We have decided to implement option 1 as not all customers adds Admin user to WS Owner app registration.

api_app/models/domain/costs.py
guybartal: all ```dict``` typed functions and locals renamed for clarity

api_app/api/routes/costs.py
tamirkamara: Do we have where else an endpoint that both the Workspace Owner and TRE Admin able to execute?

guybartal: Yes, currently three endpoints are set on ```workspaces_shared_router```:
[https://github.com/microsoft/AzureTRE/blob/main/api_app/api/routes/workspaces.py](https://github.com/microsoft/AzureTRE/blob/main/api_app/api/routes/workspaces.py)

LizaShak (approved):

PR #2007, opened 2022-06-09
Fix a pytest error when running only smoke tests in CI

tamirkamara:
## What is being addressed

When running /test via the pr-bot, only smoke tests should run, however the extended job has also started and ended with an error because no tests were "selected" by pytest

## How is this addressed

- Pass in an empty marker selector to the reusable workflow, so that it will not try to run the extended job


tamirkamara:
/test-force-approve
pr-bot isn't tested anyway

guybartal (approved):

PR #2008, opened 2022-06-09
Mandatory Identity with Application.ReadWrite.OwnedBy

ross-p-smith:
# PR for issue 1716

## What is being addressed

There is a need for having an identity that can administer AAD Applications. Depending on the customer need, this identity will need either `Application.ReadWrite.OwnedBy` or `Application.ReadWrite.All`.

## How is this addressed

-  The variable AUTO_WORKSPACE_APP_REGISTRATION determines this, and if true, the identity will be granted Application.ReadWrite.All and can write to any application as this is required by Terraform.
- All users will need to run `make auth` and put this new identity in ./templates/core/.env
-


ross-p-smith:
/test-force-approve

This change does not get tested as it's a pre-requisite for TRE

tamirkamara (approved):

PR #2020, opened 2022-06-09
Fix missing MAKEFILE_DIR

### LINKED ISSUE #2018: deploy-shared-services doesn't support running from a parent makefile
**Describe the bug**
With a setup of a parent project using AzureTRE and referencing the OSS makefile deploy-shared-services fails.

Deployed shared service tre-shared-service-firewall
make[2]: Leaving directory '/workspaces/cprd-tre'
make[1]: Leaving directory '/workspaces/cprd-tre'
/bin/bash: line 1: ./devops/scripts/load_env.sh: No such file or directory
make: *** [/home/vscode/AzureTRE/Makefile:23: deploy-shared-services] Error 1

tamirkamara:
Fixes #2018 

## What is being addressed

Some places were missing MAKEFILE_DIR and that made a parent makefile usage end with errors

## How is this addressed

- Add missing ones
- Tiny fixes and ordering


Makefile
```diff
@@ -423,7 +422,7 @@ setup-local-debugging:
 	&& . ${MAKEFILE_DIR}/devops/scripts/load_env.sh ./templates/core/.env \
```
ross-p-smith, line 422: Why not these other ones too?

ross-p-smith (approved):

Makefile
tamirkamara: We've considered the 2 main env files something a user would interact with and hence be outside the TRE OSS directoy. Everything else (the private one) is kind off "internal" and automated for the user which is why it got that prefix.

PR #2026, opened 2022-06-11
Configure docker hub proxy

### LINKED ISSUE #1089: Problem using Nexus to proxy docker hub

jjgriff93:
# PR for issue

Closes #1089 

## What is being addressed

Add Docker hub configuration to Nexus and automate set-up to consume proxy via docker clients in user resources

## How is this addressed

- Map new port to Nexus container dedicated to docker hub as per requirements
- Configure authentication realm for Docker tokens in Nexus to allow anonymous auth
- Add configuration of registry mirror to user resources so docker clients automatically use nexus on `docker pull`


templates/workspace_services/guacamole/user_resources/guacamole-azure-linuxvm/terraform/vm_config.sh
```diff
@@ -13,10 +13,12 @@ sudo adduser xrdp ssl-cert
...
 sudo apt-get install ca-certificates curl gnupg lsb-release
 # Get Docker Public key from Nexus
-curl -fsSL ${nexus_proxy_url}/repository/docker-public-key/gpg | sudo gpg --dearmor -o /etc/apt/trusted.gpg.d/docker-archive-keyring.gpg
+# shellcheck disable=SC2154
+curl -fsSL "${nexus_proxy_url}"/repository/docker-public-key/gpg | sudo gpg --dearmor -o /etc/apt/trusted.gpg.d/docker-archive-keyring.gpg
 
 # Install desktop environment if image doesn't have one already
-if [ ${install_ui} -eq 1 ]; then
+# shellcheck disable=SC2154
+if [ "${install_ui}" -eq 1 ]; then
```
tamirkamara, line 21: This is very easy to fix without ignoring the warning...
Since install_ui should always get a value you can do something like:
```bash
${install_ui:?}
```
Or if you're unsure just put something that won't change the outcome:
```bash
${install_ui:-0}
```

templates/workspace_services/guacamole/user_resources/guacamole-azure-linuxvm/terraform/vm_config.sh
```diff
@@ -27,13 +29,12 @@ sudo sed -i 's|!/bin/sh|!/bin/bash|g' /etc/xrdp/startwm.sh
 # Make sure xrdp service starts up with the system
 sudo systemctl enable xrdp
 
-
-if [ ${shared_storage_access} -eq 1 ]; then
+# shellcheck disable=SC2154
+if [ "${shared_storage_access}" -eq 1 ]; then
```
tamirkamara, line 33: same comment as above

templates/workspace_services/guacamole/user_resources/guacamole-azure-linuxvm/terraform/vm_config.sh
```diff
@@ -13,10 +13,12 @@ sudo adduser xrdp ssl-cert
 # Required packages for Docker installation
 sudo apt-get install ca-certificates curl gnupg lsb-release
 # Get Docker Public key from Nexus
-curl -fsSL ${nexus_proxy_url}/repository/docker-public-key/gpg | sudo gpg --dearmor -o /etc/apt/trusted.gpg.d/docker-archive-keyring.gpg
+# shellcheck disable=SC2154
+curl -fsSL "${nexus_proxy_url}"/repository/docker-public-key/gpg | sudo gpg --dearmor -o /etc/apt/trusted.gpg.d/docker-archive-keyring.gpg
```
tamirkamara, line 17: Is _nexus_proxy_url_ optional? If not I prefer you check you got a value instead of ignoring the warning.
```bash
${nexus_proxy_url:?}
```

templates/workspace_services/guacamole/user_resources/guacamole-azure-linuxvm/terraform/vm_config.sh
```diff
@@ -42,44 +43,51 @@ if [ ${shared_storage_access} -eq 1 ]; then
...
 
   # Autofs mounts when accessed for 60 seconds.  Folder created for constant visible mount
-  sudo ln -s $mntPath "/$fileShareName"
+  sudo ln -s "$mntPath" "/$fileShareName"
 fi
 
 ### Anaconda Config
-if [ ${conda_config} -eq 1 ]; then
+# shellcheck disable=SC2154
+if [ "${conda_config}" -eq 1 ]; then
```
tamirkamara, line 80: same comment about the variable.

templates/workspace_services/guacamole/user_resources/guacamole-azure-linuxvm/terraform/vm_config.sh
```diff
@@ -42,44 +43,51 @@ if [ ${shared_storage_access} -eq 1 ]; then
   credentialRoot="/etc/smbcredentials"
 
   mntPath="$mntRoot/$fileShareName"
-  smbPath=$(echo $httpEndpoint | cut -c7-$(expr length $httpEndpoint))$fileShareName
+  # shellcheck disable=SC2308
+  smbPath=$(echo "$httpEndpoint" | cut -c7-"$(expr length "$httpEndpoint")")$fileShareName
```
tamirkamara, line 47: Fixing the warning here is a bit more of a change so I understand if you chose to ignore it. 2 options:
1. continue to use cut but replace expr
```bash
smbPath=$(echo "$httpEndpoint" | cut -c7-"${#httpEndpoint}")$fileShareName
```
2. do everything with just bash which is cleaner (I believe this remove the protocol prefix from the url)
```bash
smbPath="${httpEndpoint:7}${fileShareName}"
```

Please note that I haven't tested it.

templates/workspace_services/guacamole/user_resources/guacamole-azure-linuxvm/terraform/vm_config.sh
jjgriff93: A fair point, because it's in fact token replacement by Terraform (`template_file`) I haven't treated them as standard variables - however the null check you propose above is still probably a good idea in case the script finds its way elsewhere in future. Will do the same for the other examples also

jjgriff93: Fixed

templates/workspace_services/guacamole/user_resources/guacamole-azure-linuxvm/terraform/vm_config.sh
jjgriff93: Done

templates/workspace_services/guacamole/user_resources/guacamole-azure-linuxvm/terraform/vm_config.sh
jjgriff93: Fixed

templates/workspace_services/guacamole/user_resources/guacamole-azure-linuxvm/terraform/vm_config.sh
jjgriff93: Done

templates/workspace_services/guacamole/user_resources/guacamole-azure-linuxvm/terraform/vm_config.sh
jjgriff93: Yeah, not my code (most of this script I think was put in before linting was in the codebase) so was wary of changing too much - but given we've now fixed the other stuff I think we may as well address this also. I like the pure bash approach so will add this and re-test the linux VM

jjgriff93:
@tamirkamara unfortunately adding `:?` to the variables results in `failed to render : <template_file>:16,30-31: Extra characters after interpolation expression; Expected a closing brace to end the interpolation expression, but found extra characters.` from the Terraform template renderer, so we'll have to revert. Instead I'll declare the variables at the top and assign them to new local ones, that way we won't get any warnings and we can keep terraform happy

templates/workspace_services/guacamole/user_resources/guacamole-azure-linuxvm/terraform/vm_config.sh
jjgriff93: Because of the template renderer it wont allow `:` in `${}` expressions so will go with option 1

jjgriff93: Neither of these options work unfortunately as the template renderer gives a `failed to render : <template_file>:49,17-18: Invalid character` when you have any expressions within handlebars `${}`. Since this isn't my code I'm not going to spend any more time looking into this and will disable the rule as I did originally

tamirkamara (approved):

PR #2032, opened 2022-06-12
Event Grid uses managed identity instead of access key

### LINKED ISSUE #2019: Change the eventgrid client to use managed identity instead of access key
Change the eventgrid client in the api to use Managed Identity instead of access key

anatbal:
# PR fixes #2019 

## What is being addressed

- Changed evengrid to use managed identity similar to how we do with service bus instead of access key
- In addition, fixed local debugging to work with the new way


templates/core/terraform/identity.tf
```diff
@@ -44,3 +44,11 @@ resource "azurerm_role_assignment" "cosmos_contributor" {
   principal_id         = azurerm_user_assigned_identity.id.principal_id
 }
 
+resource "azurerm_role_assignment" "eventgrid_data_sender" {
+  scope                = data.azurerm_eventgrid_topic.status_changed.id
+  role_definition_name = "EventGrid Data Sender"
+  principal_id         = azurerm_user_assigned_identity.id.principal_id
+  depends_on = [
+    module.airlock_resources
+  ]
```
tamirkamara, line 53: Can you switch this around and do the assignment inside the airlock module?
You will probably need to pass in the api id...

scripts/setup_local_debugging.sh
```diff
@@ -5,8 +5,10 @@ set -e
 : ${RESOURCE_GROUP_NAME?"Check RESOURCE_GROUP_NAME is defined in ./templates/core/private.env"}
 : ${SERVICE_BUS_RESOURCE_ID?"Check SERVICE_BUS_RESOURCE_ID is defined in ./templates/core/private.env"}
 : ${STATE_STORE_RESOURCE_ID?"Check STATE_STORE_RESOURCE_ID is defined in ./templates/core/private.env"}
+: ${EVENT_GRID_TOPIC_RESOURCE_ID?"Check EVENT_GRID_TOPIC_RESOURCE_ID is defined in ./templates/core/private.env"}
```
tamirkamara: Can you give this and all other occurrences (TF files too) a more appropriate name to note which topic is this?
e.g. add the status_changed to its name...

eladiw (approved):
lgtm. Approving so wont block. please address the comment from @tamirkamara

templates/core/terraform/identity.tf
anatbal: Done

templates/core/terraform/identity.tf
```diff
@@ -43,4 +43,3 @@ resource "azurerm_role_assignment" "cosmos_contributor" {
   role_definition_name = "Contributor"
   principal_id         = azurerm_user_assigned_identity.id.principal_id
 }
-
```
ross-p-smith, line 46: nitpick: do we need this change?

ross-p-smith (approved):

templates/core/terraform/identity.tf
anatbal: We definitely not, changing

PR #2035, opened 2022-06-13
Tests for User Resources

### LINKED ISSUE #1163: Deleting user resources through API fails due to ARM dependencies
**Describe the bug**
When trying to use the DELETE API for user resources (i.e. to delete a research VM), it will accept the request, however the deletion will fail after approx 20 mins (querying the GET user resources will show deleting failed status) with:

```
Code=\"NicInUse\" Message=\"Network Interface /subscriptions/*******/resourceGroups/rg-jjgriff93-ws-233d/providers/Microsoft.Network/networkInterfaces/internal-nic-jjgriff93-ws-233d-svc-293a is used by existing resource /subscriptions/*******/resourceGroups/rg-jjgriff93-ws-233d/providers/Microsoft.Compute/virtualMachines/win10vm293a. In order to delete the network interface, it must be dissociated from the resource. 
```

And

```
Error purging of Secret \"win10vm293a-admin-credentials\" (Key Vault \"https://kv-jjgriff93-ws-233d.vault.azure.net/\") : keyvault.BaseClient#PurgeDeletedSecret: Failure responding to request: StatusCode=403 -- Original Error: autorest/azure: Service returned an error. Status=403 Code=\"Forbidden\" Message=\"Operation \\\"purge\\\" is not allowed because purge protection is enabled for this vault. Key Vault service will automatically purge it after the retention period has passed.
```

**Steps to reproduce**

1. Follow the [TRE setup instructions](https://microsoft.github.io/AzureTRE/tre-admins/setup-instructions)
1. Then once the Guacamole Vm user resource is deployed, first use the user resource PATCH API passing its id and set `{ "enabled": false }` so it can be deleted
1. Then use the DELETE user resource API passing the same id
1. Wait 20+ mins and use GET user resources to query deletion status. Should change from 'deleting' to 'deleting_failed' and the error message above

ross-p-smith:
# Fixes #1163

## What is being addressed

Tested all of the User_Resources could be deployed and destroyed from the E2E Tests
- guacamole_azure_windowsvm
- guacamole_azure_linuxvm

## How is this addressed

- Registered the linuxvm bundle for use in the build
- Updated AAD test to use LinuxVM
- Split some of the helper functions out of helper.py so we can diagnose the httpx ReadTimeOut error better
- Split the Register Bundles job out in the build, so that Deploying the Shared Services can be done as soon as possible.

## Tests
image


tamirkamara:
> * Split the Register Bundles job out in the build, so that Deploying the Shared Services can be done as soon as possible.

Isn't build/publish runs in parallel to the core deployment anyway? IIRC only registering the bundles waits and that's anyway a 90 seconds step... so not sure how this saves us time.

tanya-borisova (approved):

PR #2046, opened 2022-06-16
Add curl retry

tamirkamara:
## What is being addressed

We currently experience flaky API behavior like timeouts but also other transient http errors which is causing many CI runs to fail.

## How is this addressed

- Add curl retry to register bundle script as deploy shared service script. This should increase the odds that the action will succeed, although it's a temp fix. 
Settings are: 7 retries with exponential backoff (default), each try is limited to 90 seconds and the entire process is limited to 5 minutes.


LizaShak (approved):

LizaShak:
Good one!

tamirkamara:
@tanya-borisova this is the PR I mentioned.

tamirkamara:
/test-force-approve
This has been tested and passed here: https://github.com/microsoft/AzureTRE/actions/runs/2552141150

PR #2049, opened 2022-06-16
Disable API health check logic

### LINKED ISSUE #2048: [flaky api] Disable components check in the API health check
**Describe the bug**
We suspect that some of the api health check logic is causing timeouts. To test this, we would like to disable the logic testing the different services that will immediately return 200 OK without actually testing anything.

tamirkamara:
Fixes #2048 

## What is being addressed

We want to disable the logic where the API pings downstream services to test an assumption that those cause timeouts in the system.
Preliminary tests showed that they sometimes those take a long time and some are probably blocking so making the server unresponsive to other requests.

## How is this addressed

- Disable downstream services health check


eladiw (approved):

tamirkamara:
/test-force-approve
All tests passed (pr run and branch run) but for some reason the check wasn't marked so had to force it.

PR #2051, opened 2022-06-16
Split user resource registration in CI workflow

### LINKED ISSUE #2050: Users resources should be registered after their parent workspace service
**Describe the bug**

Sometimes we see 404 errors when the CI is trying to register user resources. That's is because the parent service hasn't been registered yet and the API rejects the request.

**Steps to reproduce**

https://github.com/microsoft/AzureTRE/runs/6919307061?check_suite_focus=true#step:3:332

tamirkamara:
Fixes #2050 

## What is being addressed

User resources can't be registered along side other resources in the same job matrix as that causes race condition issues and HTTP 404 errors - they must wait for their parent resource to exist before them.

## How is this addressed

- Split the registration job in the workflow so that all users resources will be registered only after all other resources have been registered.


eladiw (approved):

ross-p-smith:
Nice catch! Sorry!

tamirkamara:
/test-force-approve
Successful branch run: https://github.com/microsoft/AzureTRE/actions/runs/2509955144

PR #2052, opened 2022-06-16
Pipeline Property Substitution

### LINKED ISSUE #1679: Substitutions Mechanism: MVP
Implement the basic substitutions mechanism to translate the properties in the `pipeline: {}` block in the schema into properties to be sent in a `PATCH`/upgrade.

As an MVP this will cover:
- [ ] `type`: `string` | `array`
- [ ] `substitution_action`: `overwrite` | `append` | `remove`
- [ ] Appropriate unit tests
- [ ] Doc updates with examples

Translate the `$resource` string into the valid resource object so a pipeline designer can utilise the properties of the resource document.

damoodamoo:
# Closes #1679 

## What is being addressed
This PR adds support for a template designer to update a secondary resource from a primary, using properties from the primary in the upgrade. 

To do this, a designer can use the syntax `{{ resource.properties.prop_name }}` , which will be translated at runtime to the real value. Using this method, Nexus + Gitea (in this PR) now update the firewall as a secondary resource rather than directly via their own terraform. Since we have queueing in place it should mean that Gitea + Nexus can be deployed in parallel more safely. 

To support this the way outputs are returned from bundles into the cosmos object has been updated. Rather than converting every output to a string in the resource, we now allow a property bag to contain outputs of complex types. 

In the UI, the notifications panel will show the pipeline steps such:
image


templates/shared_services/sonatype-nexus/template_schema.json
```diff
@@ -5,5 +5,114 @@
...
+        "resourceType": "shared-service",
+        "resourceAction": "upgrade",
+        "properties": [
+          {
+            "name": "rule_collections",
+            "type": "array",
+            "arraySubstitutionAction": "replace",
+            "arrayMatchField": "name",
+            "value": {
+              "name": "arc-web_app_subnet_nexus",
```
ross-p-smith, line 69: nit: is there a particular reason why mixed -_? is arc = azure rule collection (just interested!) 
```suggestion
              "name": "arc-web_app_subnet_nexus",
```

docs/tre-templates/pipeline-templates/pipeline-schema.md
```diff
@@ -26,6 +26,51 @@ This document will help you write a valid `pipeline: {}` block in your template.
...
+        }]
+      },
+```
+
+## Working with Properties Containing Arrays
+It's possible that a resource property would actually be an array. As an example, the firewall shared service has the `rule_collections` property. This single property contains
+an array of objects. Since the values inside this array may have been sourced from different resources, it's important to leave other values in tact when modifying the property.
+To do so, the `arraySubstitutionAction` field supports the following values:
+- `append` - just append this object into the array
+- `replace` - find this object in the the array (usign the `arrayMatchField` value), and replace it with this value
```
ross-p-smith, line 71:
```suggestion
- `replace` - find this object in the array (using the `arrayMatchField` value), and replace it with this value
```

api_app/tests_ma/conftest.py
```diff
@@ -316,3 +317,89 @@ def multi_step_operation(test_user, basic_shared_service_template, basic_shared_
...
+def pipeline_step() -> PipelineStep:
+    return PipelineStep(
+        properties=[
+            PipelineStepProperty(
+                name="rule_collections",
+                type="array",
+                arraySubstitutionAction="overwrite",
+                arrayMatchField="name",
+                value={
+                    "name": "arc-web_app_subnet_nexus_api",
```
ross-p-smith, line 365:
```suggestion
                    "name": "arc_web_app_subnet_nexus_api",
```

api_app/tests_ma/test_service_bus/test_resource_request_sender.py
```diff
@@ -49,7 +66,8 @@ async def test_resource_request_message_generated_correctly(resource_template_re
         user=create_test_user(),
         resource_template=template,
         resource_template_repo=resource_template_repo,
-        action=request_action)
+        action=request_action,
```
ross-p-smith, line 69:
```suggestion
        action=request_action
```

api_app/tests_ma/test_service_bus/test_resource_request_sender.py
```diff
@@ -103,19 +137,31 @@ async def test_multi_step_document_sends_first_step(resource_template_repo, reso
         resource_template=basic_shared_service_template,
         etag=basic_shared_service.etag,
         resource_template_repo=resource_template_repo,
-        user=test_user
+        user=test_user,
```
ross-p-smith, line 140:
```suggestion
        user=test_user
```

resource_processor/vmss_porter/runner.py
```diff
@@ -202,6 +202,12 @@ async def get_porter_outputs(msg_body: dict, message_logger_adapter: logging.Log
         outputs_json = {}
         try:
             outputs_json = json.loads(stdout)
+
+            # loop props individually to try to derialise to dict/list, as all TF outputs are strings, but we want the pure value
```
ross-p-smith, line 206:
```suggestion
            # loop props individually to try to deserialise to dict/list, as all TF outputs are strings, but we want the pure value
```

templates/shared_services/sonatype-nexus/template_schema.json
damoodamoo: no reason, that's just what it was called in the TF, so used the same name to avoid disruption

resource_processor/vmss_porter/runner.py
damoodamoo: wow dunno what happened with my typing here

tamirkamara (commented):
I didn't review the logic and would only comment:
1. If gitea/nexues don't need azcli anymore, can you please delete the dockerfile.tmpl as well (from the porter.yaml too)
2. Would merging still make it possible to use the "old" firewall script (as we do so elsewhere), or does it need to be upgrades in the other places immediately?

damoodamoo:
@tamirkamara 
> 1. If gitea/nexues don't need azcli anymore, can you please delete the dockerfile.tmpl as well (from the porter.yaml too)

I have only removed the firewall terraform resources in these bundles, nothing else has changed so no change to tooling needed.

> 2. Would merging still make it possible to use the "old" firewall script (as we do so elsewhere), or does it need to be upgrades in the other places immediately?

By 'old' firewall script do you mean the existing one in `main`? It's still possible for other resources to reference the firewall as before, and existing things should continue to work. It's a good point though, and what i'll do is rename the rules that gitea + nexus now create to avoid clashes with any existing implementations.

ross-p-smith (approved):
Fantastico!! What a PR ❤️

PR #2057, opened 2022-06-19
TRE costs API endpoint (/api/costs)

### LINKED ISSUE #1899: TRE costs API endpoint (/api/costs)
As a TRE developer,
I want to have a costs endpoint at the TRE level
So that I can get the Cost Reporting results from a TRE instance

**Acceptance Criteria**
- [ ] The endpoint should be described in the TRE swagger
- [ ] The endpoint should implement authorization (TRE Admin role only)
- [ ] implementation must query TRE for the existing workspaces
- [ ] implementation must query TRE for the existing shared services
- [ ] implementation must collect costs by calling the Azure Cost Management ARM REST API
- [ ] Implementation should merge the outputs, build a response and return it

guybartal:
# Fixes #1899 

## What is being addressed

Implement Azure TRE cost report

## How is this addressed

- cost service added functionality for azure tre overall cost report
- unit tests


tamirkamara (commented):
@guybartal can you please rephrase the title of the PR to something nicer?

api_app/api/routes/costs.py
```diff
@@ -29,27 +32,38 @@ def __init__(
         self.granularity = granularity
 
 
+def check_time_period(from_date: date, to_date: date):
+    if (to_date - from_date).days > 365:
```
ross-p-smith, line 36: Should we use year?!

api_app/resources/strings.py
```diff
@@ -54,8 +54,10 @@
 API_GET_USER_RESOURCE_TEMPLATES = "Get user resource templates applicable to the workspace service template"
 API_GET_USER_RESOURCE_TEMPLATE_BY_NAME = "Get user resource template by name and workspace service"
 
+# cost report
 API_GET_COSTS = "Get overall costs"
 API_GET_WORKSPACE_COSTS = "Get workspace costs"
+API_GET_COSTS_INVALID_TIME_PERIOD = "The time period for pulling the data cannot exceed 1 year(s)"
```
ross-p-smith, line 60: If you cant pull more than 1 year - do you need the plural on the end `(s)`

api_app/api/routes/costs.py
guybartal: the underlying service (Azure REST API) limits for one year...

api_app/resources/strings.py
guybartal: I don't :) actually copied the string from azure cost management REST API, didn't notice that....
I'll remove the plural, thanks.

api_app/api/routes/costs.py
LizaShak:
```
from dateutil.relativedelta import relativedelta
difference_in_years = relativedelta(to_date, from_date).years
```

I think it would be better to use it this way since you don't know if it was 364 in a year or 365...
doesn't have to be this way exactly but something that calculates the years differences

guybartal: good idea @LizaShak, I'll change it. thanks!

PR #2075, opened 2022-06-21
`make db-migrate` now uses API

### LINKED ISSUE #563: Research and implement database migrations

ross-p-smith:
# Fixes #563

## What is being addressed
- Removed the python database migration script as this relies on network connectivity to Cosmos.
- Moved ./scripts to ./devops/scripts
- After we deploy the shared services, we migrate the state store. (which calls the Health Check anyway)


devops/scripts/upgrade-lock-files.sh
```diff
@@ -7,10 +7,8 @@ list="$(find . -type f -name "*.hcl")"
 
 for i in $list
 do
-    rm $i
-    echo "Updating lock file" + $i
-    dir=$(dirname $i)
+    rm "$i"
+    echo "Updating lock file" + "$i"
```
martinpeck, line 11: I think this line is wrong anyway.
Should update it to....

`echo "Updating lock file $i"`

martinpeck: (it currently echos the `+` character, rather than performing a string concat)

martinpeck (approved):
Approved, but with comments.

PR #2083, opened 2022-06-21
Update parent workspace in pipeline (and set Guac redirect URI after install)

### LINKED ISSUE #2069: Target the parent workspace to update props
Now it is possible to update shared service properties in a pipeline, the current workspace should also be a possible target.

- [x] Target 'this' workspace in a pipeline 
- [x] Use same substitutions mechanism as used for shared services
- [x] Implement the AAD redirect URI property for Guac to demonstrate usage

Should address #1352

damoodamoo:
# Closes #2069 and #1352 

Using the pipeline template, you can now specify `resourceType: workspace`, and the parent workspace of the workspace service / user resource will be found and updated. 

As a use case and test, the Guac connection_uri is used to update the workspace properties, which are fed to some updated workspace TF which sets the URI in AAD.


templates/workspaces/base/terraform/variables.tf
```diff
@@ -48,6 +48,11 @@ variable "register_aad_application" {
   description = "Create an AAD application automatically for the Workspace."
 }
 
+variable "aad_redirect_uris_b64" {
```
ross-p-smith, line 51: why do we accept a dictionary when the terraform just expects a comma seperated list of URIs? Is this a porter thing?

api_app/db/repositories/operations.py
```diff
@@ -86,7 +88,29 @@ def build_step_list(self, steps: List[OperationStep], resource_template_dict: di
...
+
+                        # if it's a workspace service, we must be a user-resource - find the parent
+                        if step["resourceType"] == ResourceType.WorkspaceService:
+                            primary_resource = resource_repo.get_resource_by_id(uuid.UUID(resource_id))
+                            if primary_resource.resourceType != ResourceType.UserResource:
+                                raise Exception("Only user resources can update their parent workspace services")
+                            resource_for_step = resource_repo.get_resource_by_id(uuid.UUID(primary_resource.parentWorkspaceServiceId))
+
+                        if resource_for_step is None:
+                            raise Exception("Error finding resource to update")
```
ross-p-smith, line 112: Can we pop some IDs in the error message to help us debug it in the future?

templates/workspaces/base/terraform/variables.tf
damoodamoo: we need to have a list of objects in the property bag as we need a way of identifying a URI to remove / replace - so it's a key/value pair with the key being something that doesn't change.

damoodamoo:
/test-force-approve
ran extended tests prior to minor change, and shared services tests locally

PR #2095, opened 2022-06-22
UI: Add Power state display to resources

### LINKED ISSUE #2009: Display Azure resource status
- [ ] User Resources such as VMs need to show state - such as Running, Stopped, Deallocated
- [ ] Connect version should only be visible if resource is Running.

jjgriff93:
# PR for issue
Closes #2009 

## What is being addressed

Added power state display to resource cards and header menus, and dynamically disable connect buttons if power status is not 'running'

## How is this addressed

- Fixed bundles for user resources so the Api successfully adds powerStatus information to the resources in a GET response
- Modified ResourceCard to add power badge and disable connect if not powered on
- Added power state badge to ResourceHeader
- Disable connect within resource pages if resource has a power state and is not powered on

image

image


ui/app/src/components/shared/PowerStateBadge.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import { powerStates } from '../../models/resource';
+
+interface PowerStateBadgeProps {
+  state: typeof powerStates[number]
```
damoodamoo, line 5: is it clearer to change `powerStates` to an enum?

damoodamoo:
Lovely piece of work - couple of snags in my testing:

(think?) we should disable the connect button when the resource is in the Lock state? 
image

Also found that my VMs with `"azureStatus":{}`, or the prop missing, were still showing the connect button as active - on the card and the resource page. @marrobi - what shall we do in these situations, where a resource has a `connection_uri` but no Azure status? Given James has fixed the templates the VMs in the repo will always have a status going forward, so maybe we do just disable the button if the status is there and it's 'off' ? 
image
image

marrobi:
Some services, such as Azure ML will not have an `azureStatus` so think the button needs to appear unless it explicitly says off?

jjgriff93:
> Lovely piece of work - couple of snags in my testing:
> 
> (think?) we should disable the connect button when the resource is in the Lock state? image

Good catch! Have amended and bought it in line with how its handled in the resource header context menu. As for handling empty props yes was thinking we err on the side of showing a connect button unless explicitly told not to (power state off) in case resources don't provide it or there's a bug like we had before - at least they'll still be able to attempt to connect to resources instead of having the functionality removed entirely

ui/app/src/components/shared/PowerStateBadge.tsx
jjgriff93: Much clearer, didn't even cross my mind! Adding now

damoodamoo (approved):

martinpeck:
So, are the porter changes required to make the UI power state stuff work?
I wonder if we should have kept those changes separate from the UI changes (and whether we should separate them in future)

jjgriff93:
> So, are the porter changes required to make the UI power state stuff work?
> I wonder if we should have kept those changes separate from the UI changes (and whether we should separate them in future)

Its a bug fix as they are supposed to already be outputting that status, and its required for the UI to work yes. I think for a small change like this I don't see the point doing as two PRs but bigger functionality change in prep for UI would probs be a good idea to separate

PR #2100, opened 2022-06-23
Update workspace redirect uris when auto-aad is disabled

### LINKED ISSUE #1352: App registration's redirect url should be set automatically when creating a workspace

tamirkamara:
Fixes #1352 

## What is being addressed

We need to update a workspace API redirect-uris even when we don't create the app automatically (aka auto-aad).

## How is this addressed

- Use a script with azcli commands to do this in the base workspace bundle
- Add the application admin identity as an owner of the workspace app that is created via a setup script (called in make auth)
- Fix some redundant code in the makefile


ross-p-smith (approved):

devops/scripts/aad/aad-app-reg.sh
```diff
@@ -361,16 +361,15 @@ echo "Setting API permissions (email / profile / ipaddr)"
...
-# Update app registration with redirect urls (SPA)
-# az rest --method PATCH --uri "${msGraphUri}/applications/${apiAppObjectId}" \
-#     --headers 'Content-Type=application/json' \
-#     --body '{"spa":{"redirectUris":["https://localhost:8080"]}}'
-
 # Make the current user an owner of the application.
 az ad app owner add --id "${apiAppId}" --owner-object-id "$currentUserId"
 
+# Add the application admin service principal as an owner of the workspace app
+if [[ $workspace -eq 1 ]]; then
```
ross-p-smith, line 368: Will this get called? Isn't this in a conditional block only being run for API as opposed to Workspace?

tamirkamara: I don't think it runs only for the API and anyway I run this multiple times and it always created the app with the right owner added.

PR #2001, opened 2022-06-08
UI MVP

damoodamoo:
# PR for issue #1406

Issue #1783, opened 2022-05-06
Failed deployment of resources isn't reflected in API / Cosmos


tanya-borisova:
**Describe the bug**

When a resource deployment fails, the status in Cosmos still shows isActive = true, and the resource appears in a list of resources in the API. 

**Steps to reproduce**

I tried this on gitea workspace service and on shared services, but I think it would work on any resource 

1. Trivially break gitea workspace service, e.g. use a non-existing variable in Terraform
2. Build and publish gitea workspace service bundle
3. Run E2E tests and make sure they fail
4. The resource object in Cosmos will still have isActive set to true

martinpeck:
Success/Failure is reflected in the Operations. The overall status isn't being accurately reflected in `isActive`,
This needs more thought.

tanya-borisova:
@martinpeck  I still think this is a bug though, as I don't think we should be returning in API services that have failed to be deployed. `isActive` might not be the field for it, but in this case we can query operation results?

martinpeck:
Agreed it's a bug (or, at very least, not an ideal situation). What's the next step for this? Does this need to be wrapped up in a conversation about how we convert operations into a headline status?

cc @damoodamoo @marrobi

marrobi:
Agree, if initial deployment of a resource fails, shouldn't become active. If we get to updates, conversation might be different.

I do wonder if we need a resource state properly. So Queued, Deploying, Active, Updating, Disabled etc.

tanya-borisova:
Discussed with David, plan is to use the new deploymentStatus field for the shared services logic (i.e.: don't deploy a new shared service, if a previous one has a successful deployment status). The field will be introduced here: https://github.com/microsoft/AzureTRE/pull/2001
We also think that this removed the need for having the isActive field, so we can remove it for simplicity

PR #2101, opened 2022-06-23
Failed deployment of shared service can be attempted again

tanya-borisova:
# PR for issue  https://github.com/microsoft/AzureTRE/issues/1783

## What is being addressed

With the recent work completed in https://github.com/microsoft/AzureTRE/pull/2001/files , isActive field is becoming obsolete. 
It can now be replaced by `deploymentStatus != "deleted"`. 

This also enables us to fix a bug when a shared service that previously failed to be deployed couldn't be deployed again. 
Now we can check the `deploymentStatus` field and attempt the deployment so long as there is no shared service that is already in deployed state. 

## How is this addressed

- Replaced isActive field and use deploymentStatus where necessary 
- Updated devops/scripts/deploy_shared_service.sh 
- Test that a failed deployment of a shared service can now be retried


api_app/models/domain/operation.py
```diff
@@ -13,12 +13,12 @@ class Status(str, Enum):
     """
     Operation status
     """
-    Failed = strings.RESOURCE_STATUS_FAILED
```
tanya-borisova, line 16: I just reordered these lines so that it's a bit easier to read

api_app/tests_ma/test_db/test_repositories/test_resource_repository.py
```diff
@@ -244,7 +243,7 @@ def test_get_resource_dict_by_id_queries_db(resource_repo):
 
     resource_repo.get_resource_dict_by_id(item_id)
 
-    resource_repo.query.assert_called_once_with(query='SELECT * FROM c WHERE c.isActive != false AND c.id = "123"')
+    resource_repo.query.assert_called_once_with(query='SELECT * FROM c WHERE c.deploymentStatus != "deleted" AND c.id = "123"')
```
martinpeck, line 246: any reason we don't use the `IS_NOT_DELETED_CLAUSE` here?

api_app/db/repositories/shared_services.py
```diff
@@ -26,11 +26,11 @@ def shared_service_query(shared_service_id: str):
...
     @staticmethod
     def active_shared_services_query():
-        return f'SELECT * FROM c WHERE {IS_ACTIVE_CLAUSE} AND c.resourceType = "{ResourceType.SharedService}"'
+        return f'SELECT * FROM c WHERE {IS_NOT_DELETED_CLAUSE} AND c.resourceType = "{ResourceType.SharedService}"'
 
     @staticmethod
-    def active_shared_service_with_template_name_query(template_name: str):
-        return f'SELECT * FROM c WHERE {IS_ACTIVE_CLAUSE} AND c.resourceType = "{ResourceType.SharedService}" AND c.templateName = "{template_name}"'
+    def deployed_shared_service_with_template_name_query(template_name: str):
+        return f'SELECT * FROM c WHERE c.deploymentStatus = "deployed" AND c.resourceType = "{ResourceType.SharedService}" AND c.templateName = "{template_name}"'
```
martinpeck, line 33: nit: should we create an `IS_DEPLOYED_CLAUSE` constant?

api_app/db/repositories/shared_services.py
```diff
@@ -26,11 +26,11 @@ def shared_service_query(shared_service_id: str):
 
     @staticmethod
     def active_shared_services_query():
-        return f'SELECT * FROM c WHERE {IS_ACTIVE_CLAUSE} AND c.resourceType = "{ResourceType.SharedService}"'
+        return f'SELECT * FROM c WHERE {IS_NOT_DELETED_CLAUSE} AND c.resourceType = "{ResourceType.SharedService}"'
 
     @staticmethod
-    def active_shared_service_with_template_name_query(template_name: str):
-        return f'SELECT * FROM c WHERE {IS_ACTIVE_CLAUSE} AND c.resourceType = "{ResourceType.SharedService}" AND c.templateName = "{template_name}"'
+    def deployed_shared_service_with_template_name_query(template_name: str):
```
martinpeck, line 32: I guess we're not testing this, as I can't see an associated change in a test.
Not something you need to address...just an observation.

tanya-borisova: Well there is a test `test_create_shared_service_item_with_the_same_name_twice_fails` here: https://github.com/microsoft/AzureTRE/blob/main/api_app/tests_ma/test_db/test_repositories/test_shared_service_repository.py#L107 

It checks that if the query returns existing items, then the shared service isn't created again. 

I wanted to change that test so that I can set the actual items in the container, and then run a query on it, but I haven't found an easy way to do that

tanya-borisova: In azure/cosmos/container.py, there's a method that queries items, and a query is passed in there. So basically there's no way to mock the items it's going to use but supply query from the method being tested.

We could consider using a local Cosmos emulator to be able to write these kinds of tests: https://docs.microsoft.com/en-us/azure/cosmos-db/local-emulator?tabs=ssl-netstd21

martinpeck (approved):
Thanks @tanya-borisova !

api_app/db/repositories/shared_services.py
martinpeck: Leave things as they are for now.
It was just something I spotted while reviewing this, but sounds like there are good reasons (it's non-trivial)

PR #2103, opened 2022-06-23
Set Nexus default to V2

### LINKED ISSUE #2084: Change default Nexus from V1 to V2
## Description

The TRE default installation of the Nexus shared service (through `make all`/`make deploy-shared-services`) should target the V2 instance of Nexus by default as part of preparations for removal as per #1981 

## Acceptance criteria

- [x] `make all` deploys the VM-based Nexus V2 service
- [x] User resources deployed from this point onwards target the V2 service by default in the bundle parameters (with the option remaining to target V1)

jjgriff93:
# Closes #2084

## What is being addressed

Changes the default Nexus shared service deployed by `make deploy-shared-services`/`make all` to the VM-based V2 service

## How is this addressed

- Updated the makefile with cert install/generate and nexus-vm install commands
- Added parameters variable to `deploy_shared_services.sh` to pass additional bundle parameters for cert install
- Updated user resources to use V2 by default
- Updated documentation


devops/scripts/deploy_shared_service.sh
```diff
@@ -3,17 +3,18 @@
...
 set -o pipefail
 # Uncomment this line to see each command for debugging (careful: this will show secrets!)
-# set -o xtrace
+set -o xtrace
 
 function usage() {
     cat <<USAGE
 
-    Usage: $0 [-u --tre_url]  [-c --current] [-i --insecure]
+    Usage: $0 [-u --tre_url]  [-c --current] [-i --insecure] [-p --properties]
```
martinpeck, line 11: The properties thing feels a little "fragile"

What gets passed in are name value pairs that *must* use confirm to a particular format (i.e. be a snippet of JSON with the right quotes etc).

I can see that it works, but it's not obvious (and easy to get wrong). For example, without inspecting the place you pass properties today, you probably wouldn't have guessed that they need to be...

- double quoted name and value
- name/value need to be separated by colon
- name/value pairs need separating by comma

```
"domain_prefix": "nexus","cert_name": "nexus-ssl"
```

Makefile
```diff
@@ -352,10 +352,15 @@ firewall-install:
...
 
 nexus-install:
-	$(MAKE) bundle-build DIR=${MAKEFILE_DIR}/templates/shared_services/sonatype-nexus/ \
-	&& $(MAKE) bundle-publish DIR=${MAKEFILE_DIR}/templates/shared_services/sonatype-nexus/ \
-	&& $(MAKE) bundle-register DIR="${MAKEFILE_DIR}/templates/shared_services/sonatype-nexus" BUNDLE_TYPE=shared_service \
-	&& $(MAKE) deploy-shared-service DIR=${MAKEFILE_DIR}/templates/shared_services/sonatype-nexus/ BUNDLE_TYPE=shared_service
+	$(MAKE) bundle-build DIR=${MAKEFILE_DIR}/templates/shared_services/certs/ \
+	&& $(MAKE) bundle-publish DIR=${MAKEFILE_DIR}/templates/shared_services/certs/ \
+	&& $(MAKE) bundle-register DIR="${MAKEFILE_DIR}/templates/shared_services/certs" BUNDLE_TYPE=shared_service \
+	&& $(MAKE) deploy-shared-service DIR=${MAKEFILE_DIR}/templates/shared_services/certs/ BUNDLE_TYPE=shared_service PROPS='"domain_prefix": "nexus","cert_name": "nexus-ssl"' \
```
martinpeck, line 358: Is this the right place to store these `PROPS`?
I assume it's reasonable that someone might want to change these. Ideally, they shouldn't have to edit the Makefile to do so.

devops/scripts/deploy_shared_service.sh
martinpeck: Having said that, I don't have an obviously better solution to this.
See what other say, but at present I'm not keen on...

1. the values being stored in the `Makefile`
2. the values being passed as a fragment of JSON that must have very specific formatting
3. the help/usage of this script giving no clues on that format

devops/scripts/deploy_shared_service.sh
```diff
@@ -67,6 +72,11 @@ if [ -z "${access_token:-}" ]; then
   access_token=${ACCESS_TOKEN}
 fi
 
+if [ -n "${properties:-}" ]; then
+  # If properties are passed, prepend with a comma so we can add to the payload proeprties
+  properties=", $properties"
```
martinpeck, line 77: prepending with a comma also feels a little brittle

martinpeck (changes requested):

devops/scripts/deploy_shared_service.sh
martinpeck: How about this as an approach. I prefer this because...

- it's in keeping with how you would pass multiple things to a bash script
- the only format you need to document is an easier `name=value`

https://gist.github.com/martinpeck/41bc86dec7e470503d3f1cd27f968846

martinpeck: I've created two options and illustrated them in this Gist:

https://gist.github.com/martinpeck/41bc86dec7e470503d3f1cd27f968846

I prefer option two, but both feel cleaner and more obvious (and allow for extension if other services need properties)

devops/scripts/deploy_shared_service.sh
jjgriff93: I'm open to suggestions...?

devops/scripts/deploy_shared_service.sh
```diff
@@ -139,7 +149,9 @@ if [[ -n "${deployed_shared_service}" ]]; then
   fi
 fi
 
-payload="{ \"templateName\": \"""${template_name}""\", \"properties\": { \"display_name\": \"Shared service ""${template_name}""\", \"description\": \"Automatically deployed ""${template_name}""\" } }"
+payload="{ \"templateName\": \"""${template_name}""\", \"properties\": { \"display_name\": \"Shared service ""${template_name}""\", \"description\": \"Automatically deployed ""${template_name}""\" ${properties}} }"
+echo "$payload"
```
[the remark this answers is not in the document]
jjgriff93: Yep, thanks!

Makefile
jjgriff93: If they were to change these they would also need to modify the deploy-shared-service line for Nexus in the makefile anyway to make sure that the new values are passed via PROPS to override the defaults (the PROPS for the certs service here specify the Nexus defaults). If they were doing that level of customisation they shouldn't be using the makefile command and should be posting to the API manually as the makefile is intended to cover the automated path only.

jjgriff93: If we were to store the props elsewhere the only place I could think of is as default values in the certs bundle, like we do with the other shared services - but I think as we intend for the certs service to be re-used for other ssl requirements in future that feels like the wrong place

devops/scripts/deploy_shared_service.sh
jjgriff93: Yeah I do agree, I guess I didn't think much of it as it's only being used for one of the shared services currently. The second approach seems quite clean, have implemented that. Thanks for the gists, they were helpful!

martinpeck (approved):
Thanks @jjgriff93 !

PR #2082, opened 2022-06-21
airlock doc

joalmeid:
# PR for issue #2081 

## What is being addressed

Goal is to provide documentation for current implementation only. This documentation should focus on providing TRE users guidelines for Airlock usage, but also explain what requirements it tackles.
Finally

As a TRE Administrator/TRE Workspace Owner/Researcher/TRE Service Integrator/Data Engineer...
I want to read about Airlock from a user perspective
So that I can better understand Airlock requirements and capabilities, related concepts and how Airlock works

As an Azure TRE Developer
I want to read about Airlock
So that I can better understand the infrastructure involved, extensibility points, existing implementation, feature roadmap

## How is this addressed

- Create a Documentation Features section
- Include Airlock documentation

Issue #261, opened 2021-06-14
HITRUST/HIPAA compliance


Lybecker:
The Azure TRE solution accelerator should have the relevant compliance assessments for the healthcare industry. Azure TRE will not be certified but use tools to discover and address issues.

## Requirements

- The NHS Digital Cloud good Practice guidelines which is a subset of HITRUST/HIPAA assessment executed
- Issues solved or clearly documented

PR #2108, opened 2022-06-23
Initial Proposal for Restructuring Docs

martinpeck:
This is a proposal for how we might restructure the documentation.

## Goals

- create a top level navigation structure that makes sense
- ensure we have homes for future documentation
- initially guide people based upon the tasks they're performing, rather than their roles

### Top Level Navigation

The idea is that we take the user through an introduction (Overview) and then how to install the AzureTRE (Setup).

We then explain how Researchers and Workspace Owners would use the TRE (Using the TRE for Research).

Templates and Services lists the capabilities that can be deployed into the TRE. This is essentially a catalog, as well as technical docs.

Finally, Technical Guide contains docs relevant to the Azure Administrator and to Developers. Here, technical docs cover general admin, setting up AAD, and then developing (contributing) to the OSS project, or authoring templates.

image

### Overview

Where users/readers start. This section should do enough to explain what this project is, and the overall architecture/approaches. It can then sign post to other sections for further details.

At the end of reading this section a user should have a good understanding of what a TRE is, what the AzureTRE might offer, the architecture, and then whether they want to deploy it.

image

### Setup

This section covers the initial (non-CICD) instructions on deploying the TRE. The steps take the reader from zero to having their first Azure TRE installed.

image

### Using the TRE for Research

This section is aimed at those who will provision workspaces (workspace owners) and those who will use workspaces (researchers). We don't have many documents in this section at the moment, but over time this would be where we'd add general "how to work in the AzureTRE" docs.

image

### Templates and Services

This is essentially our catalog or showcase of the templates, workspace services, user services etc.
image

### Technical Guide

Within the technical guide we have two major sections. 

- The first is aimed at those who will deploy and manage the azure resources.
- The second is aimed at those who will write code that is deployed (be that the AzureTRE itself, or templates)

This section is quite large, but the idea is that everything related to administration and development is in one place, and then shares a single Troubleshooting section. The roles of developer and admin often overlap, hence the grouping, but we also want to keep this section out of the way from those who simply want to understand the project, or *use* the product.

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -237,7 +237,7 @@ lint:
 		github/super-linter:slim-v4.9.4
 
 lint-docs:
-	LINTER_REGEX_INCLUDE=.*docs/.* $(MAKE) lint
+	LINTER_REGEX_INCLUDE='./docs/.*\|./mkdocs.yml' $(MAKE) lint
 
 # check-params is called at the end since it needs the bundle image,
 # so we build it first and then run the check.
--- a/docs/coming-soon.md
+++ b/docs/coming-soon.md
@@ -1,0 +1,3 @@
+# Coming Soon
+
+Sorry, but this content isn't yet available.
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -37,62 +37,92 @@ markdown_extensions:
   - pymdownx.tasklist
 
 nav:
-  - Azure TRE Overview:
-      - Concepts: 'index.md'
-      - User Roles: 'azure-tre-overview/user-roles.md'
-      - Architecture: 'azure-tre-overview/architecture.md'
-      - Network Architecture: 'azure-tre-overview/networking.md'
-      - Azure Resource Breakdown: 'azure-tre-overview/tre-resources-breakdown.md'
-  - TRE Admins:
-      - Setup Instructions:
-          - 1. Getting Started: 'tre-admins/setup-instructions/getting-started.md'
-          - 2. Pre-deployment Steps: 'tre-admins/setup-instructions/pre-deployment-steps.md'
-          - 3. Deploying Azure TRE: 'tre-admins/setup-instructions/deploying-azure-tre.md'
-          - 3b. Deploying Azure TRE with GitHub workflows: 'tre-admins/setup-instructions/workflows.md'
-          - 4. Configuring Shared Services: 'tre-admins/setup-instructions/configuring-shared-services.md'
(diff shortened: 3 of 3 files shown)


marrobi:
In principle I think it's good way forward, I wonder if Setup needs to come under technical guide?

joalmeid:
I agree with the doc approach. 
Looking ahead, TRE will have more complex features for TRE users to leverage. Ex: Airlock 
This would require feature explanation (data movement, notification mechanism and how to extend it)

I'm started Airlock doc in #2082, and happy to align with this. It can be a good exercise to practice the new doc structure.

My view would be:
In `Using TRE for Research` we would describe Airlock import and export, data movement, approval & Notifications.
In `Technical Guide` > `Development` > `Extending Azure TRE` we could add Extending Notification mechanism

damoodamoo:
I like it - makes a clearer journey from zero, and I can see where I'd want to put new docs on features i'm working on

martinpeck:
> In principle I think it's good way forward, I wonder if Setup needs to come under technical guide?

@marrobi yeah, it could live there. I was trying to keep those initial "ok...so you want to give this a try" instructions near the top of the nav.

Maybe moving Setup into Technical Guide makes sense, and then having a single page called "Getting Started" in the Introduction to sign-point people, makes sense.

martinpeck:
In other words, like this...

image

tanya-borisova:
I like it, makes it quite clear where our future docs go. I think I'd have an easier time to onboard on TRE if docs were structured this way

joalmeid:
@martinpeck , @marrobi , one area we should include in this new structure would be related to compliance. There's an issue already: #261

martinpeck:
> @martinpeck , @marrobi , one area we should include in this new structure would be related to compliance. There's an issue already: #261

Agreed. My plan would be that this would appear in the Overview section, either as part of the architecture or in a separate section. The idea is that casual visitors to the docs (those new to the project, those reviewing it for compliance, those reviewing it for architecture/costs/whatever) will find most of what they need in the first section

martinpeck:
I could create a small number of section-specific "coming soon" pages if that softened the blow a little more

joalmeid:
I'm also review the Airlock doc and make the PR Ready. #2082

martinpeck:
Content for Coming Soon page:

image

marrobi (approved):
LGTM. Thanks.

PR #2117, opened 2022-06-26
Workspace app service SKU as top level parameter

### LINKED ISSUE #2085: Accept App Service Plan sku as a workspace parameter
Due to capacity limits and other factors it is not always possible to use `P1v3` instances for the default app service plan sku size in the workspace template.

It should be possible to select from some options for sku size (including `P1v2`)

tamirkamara:
Fixes #2085 

## What is being addressed

Sometimes we like to change the default AppService Plan SKU that is deployed (like due to Azure capacity issues).

## How is this addressed

- Add the sku param in all places so that the api (and ui) will pick it up
- reference the providers.tf file in porter's TF mixin for better caching

### DIFF
--- a/templates/workspaces/base/porter.yaml
+++ b/templates/workspaces/base/porter.yaml
@@ -1,6 +1,6 @@
 ---
 name: tre-workspace-base
-version: 0.3.10
+version: 0.3.11
 description: "A base Azure TRE workspace"
 dockerfile: Dockerfile.tmpl
 registry: azuretre
@@ -93,6 +93,10 @@ parameters:
     type: string
     description: "List of redirect URIs in {name:value} format"
     default: "W10=" # b64 for []
+  - name: app_service_plan_sku
+    type: string
+    description: "The SKU used when deploying an Azure App Service Plan"
+    default: "P1v3"
 
 outputs:
   - name: app_role_id_workspace_owner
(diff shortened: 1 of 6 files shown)


templates/workspaces/base/porter.yaml
```diff
@@ -144,6 +149,7 @@ install:
         app_role_id_workspace_owner: "{{ bundle.parameters.app_role_id_workspace_owner }}"
         app_role_id_workspace_researcher: "{{ bundle.parameters.app_role_id_workspace_researcher }}"
         aad_redirect_uris_b64: "{{ bundle.parameters.aad_redirect_uris }}"
+        app_service_plan_sku: "{{ bundle.parameters.app_service_plan_sku }}"
```
LizaShak, line 152: I know there is a default value to it, but should we also add it maybe in swagger api example? or in its api template....?

tamirkamara: but there're many params we don't list there... what makes this one different?

tamirkamara:
/test-force-approve
tested in the branch [workflow ](https://github.com/microsoft/AzureTRE/actions/runs/2563108565)to run both extended and extended_aad tests.

LizaShak (approved):

PR #2118, opened 2022-06-26
Fix Guacamole's authentication URI to support auto update in workspace app

### LINKED ISSUE #2109: Guacamole's connection_uri output isn't enough for automated e2e provisioning
**Describe the bug**
The automatic implementation of provisioning Guacamole and updating the workspace app-reg redirect-uris is using the returned output of connection_uri, however that isn't where AAD is actually returning the user to.
connection_uri is _something/guacamole_ where the user needs to land on _something/oauth2/callback_

**Steps to reproduce**

1. deploy a workspace
1. deploy guacamole
1. deploy a vm
1. try to connect and you'll get an aad error saying the redirect url is not defined in the app

tamirkamara:
Fixes #2109 

## What is being addressed

The returned connection uri output from the Guacamole bundle isn't the one AAD redirects the user to after authentication which is causing a failure in the flow.

## How is this addressed

- Add a new output for the uri used in the authentication redirect
- Use the uri in the bundle's pipeline definition so that it could be correctly updated in the workspace app
- Reference the providers.tf file for better caching in porter


templates/workspace_services/guacamole/porter.yaml
```diff
@@ -101,11 +101,16 @@ outputs:
     type: string
     applyTo:
       - install
+  - name: authentication_callback_uri
```
anatbal, line 104: It should probably be under uninstall as well, since if we will try to uninstall it will not find the var

templates/workspace_services/guacamole/template_schema.json
```diff
@@ -66,7 +66,7 @@
             "arrayMatchField": "name",
             "value": {
               "name": "{{ resource.id }}",
-              "value": "{{ resource.properties.connection_uri }}"
```
anatbal, line 69: why removing connection uri?

anatbal (approved):

templates/workspace_services/guacamole/porter.yaml
tamirkamara: This is an output param, so it doesn't need to apply to the uninstall too.

templates/workspace_services/guacamole/template_schema.json
tamirkamara: Since connection_uri isn't what is needed when setting the aad app redirect urls.

PR #2123, opened 2022-06-26
Align airlock tags

### LINKED ISSUE #2121: Inconsistent airlock tags
**Describe the bug**
Some of the airlock resources are not tagged correctly and their costs will not be aggregated appropriately.

**Steps to reproduce**

1. Deploy the core infra and see that the storage accounts are missing their tre_* tags

tamirkamara:
Fixes #2121

## What is being addressed

Some of the airlock resources weren't tagged correct which would impact the cost figures.

## How is this addressed

- Make sure all airlock resources have tags needed for cost figures
- Remove data objects for things that could be passed between modules


templates/core/terraform/main.tf
```diff
@@ -80,21 +80,25 @@ module "appgateway" {
...
+  docker_registry_server                = var.docker_registry_server
+  mgmt_resource_group_name              = var.mgmt_resource_group_name
+  mgmt_acr_name                         = var.acr_name
+  api_principal_id                      = azurerm_user_assigned_identity.id.principal_id
+  arm_subscription_id                   = var.arm_subscription_id
+  airlock_app_service_plan_sku_size     = var.api_app_service_plan_sku_size
+  airlock_processor_subnet_id           = module.network.airlock_processor_subnet_id
+  airlock_servicebus                    = azurerm_servicebus_namespace.sb
+  applicationinsights_connection_string = module.azure_monitor.app_insights_connection_string
+  tre_core_tags                         = local.tre_core_tags
```
LizaShak, line 100: shouldn't be var.tre_core_tags?

tamirkamara: no, this is the root module we we call the airlock module. tre_core_tags should be defined once as a local and then passed into child modules

LizaShak (approved):

eladiw (approved):

PR #2134, opened 2022-06-27
resourceTemplateName is only required for targeting shared services

damoodamoo:
Clearing up the guac pipeline document and added a comment to the docs to make it clear that the `resourceTemplateName` field in a pipeline is optional, and only needed when targeting shared services


martinpeck (approved):

damoodamoo:
/test-force-approve
ran on previous commit

Issue #2113, opened 2022-06-24
API redacts sensitive properties in messages to resource processor


tamirkamara:
**Describe the bug**
The API redacts properties in the messages it sends to the resource processor which isn't right.
An example message from service bus:

{
    "operationId": "2e0e51e0-e3a4-4da0-8ed3-7a5ad2c8ecd0",
    "stepId": "main",
    "action": "install",
    "id": "41e9e2c2-348f-438f-8558-279282ff1caf",
    "name": "tre-workspace-base",
    "version": "0.3.10",
    "parameters": {
        "client_id": "7db7f765...",
        "client_secret": "REDACTED",
    }
}

**Steps to reproduce**

1. Create a new base workspace instance
1. Either catch the message sent to RP in the Service Bus explorer or, after deployment is complete look in the workspace keyvault for the value

PR #2066, opened 2022-06-20
Redact secrets before saving resources in Cosmos

tanya-borisova:
# PR for issue https://github.com/microsoft/AzureTRE/issues/1922 

## What is being addressed

Make sure secret values aren't being saved in Cosmos and consequently aren't displayed in the UI 

## How is this addressed

- Added a "sensitive" field to ResourceTemplate 
- Set secret values to be "sensitive" 
- Redact values before being saved to Cosmos but leave the ones being sent to Service Bus intact

PR #2136, opened 2022-06-27
Fix redacting secrets when sending messages to Resource Processor

tanya-borisova:
# PR for issue https://github.com/microsoft/AzureTRE/issues/2113 

## What is being addressed

In `save_and_deploy_resource`, we now manipulate with two resource objects: one is saved to Cosmos (with secrets redacted), another one is being sent to Resource Processor (with secrets intact). (see https://github.com/microsoft/AzureTRE/pull/2066). 
Thus, it is important that what is being sent to Resource Processor remains intact throughout, otherwise secret gets lost. 

In this PR, send the `resource` object provided to `send_resource_request_message` in order to preserve the un-redacted secrets that aren't saved in Cosmos DB.

## How is this addressed

- Pass through the resource object from the `save_and_deploy_resource` method
- Update tests so that the mock is set up on the downstream method, `send_deployment_message`


damoodamoo (approved):
As we've discussed - I think this is fine as a tactical solution, but has the limitations:
- will still fail if a resource is updated (either directly or as part of a pipeline)
- will only work in a pipeline if the primary resource is the first step

tanya-borisova:
/test-force-approve

Some unrelated tests broke when I merged main last, but given there are successful runs (see above) I will merge this

PR #2151, opened 2022-06-29
Migrations were not running

### LINKED ISSUE #1824: Single place to have migrations
At the moment, there is two places where we upgrade the state store. 
- scripts/db-migrate.py
- api/migrate route in the API

## Acceptance Criteria
- Have a way to be able to call migrate from a devcontainer outside the network (dev)
- Have a way to be able to call migrate from a build (e.g. API)
- Have one, preferably the API as that always has access to Cosmos already

ross-p-smith:
# Fixes #1824

## What is being addressed

There were a few bugs in migrations which meant that they were not running.

## How is this addressed

- Fixed the shared services migrations
- Only run the Deployment Status migration if needed.

### DIFF
--- a/api_app/db/migrations/shared_services.py
+++ b/api_app/db/migrations/shared_services.py
@@ -15,8 +15,8 @@ class SharedServiceMigration(SharedServiceRepository):
         for template_name in template_names:
             for item in self.query(query=f'SELECT * FROM c WHERE c.resourceType = "shared-service" AND c.templateName = "{template_name}" \
                                                                 ORDER BY c.updatedWhen ASC OFFSET 1 LIMIT 10000'):
-                logging.INFO(f"Deleting element {item}")
-                self.delete_item(item, partition_key=item["id"])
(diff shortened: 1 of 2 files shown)


api_app/db/migrations/shared_services.py
```diff
@@ -15,8 +15,8 @@ def deleteDuplicatedSharedServices(self) -> bool:
         for template_name in template_names:
             for item in self.query(query=f'SELECT * FROM c WHERE c.resourceType = "shared-service" AND c.templateName = "{template_name}" \
                                                                 ORDER BY c.updatedWhen ASC OFFSET 1 LIMIT 10000'):
-                logging.INFO(f"Deleting element {item}")
-                self.delete_item(item, partition_key=item["id"])
```
tanya-borisova, line 19: This must've been a leftover from some old version of the library used? I remember it required for this parameter to be passed through...

tanya-borisova (approved):
Thanks!

Issue #2150, opened 2022-06-29
Pipeline seem to uninstall a parent instead of upgrading it


tamirkamara:
**Describe the bug**
As described

**Steps to reproduce**

1. Deploy a workspace
1. Deploy Guacamole
1. Disable Guacamole
1. Delete Guacamole
1. See that the deployment has failed for reasons related to WS resources, also see that the WS RG is almost empty (WS resources have been deleted).

PR #2153, opened 2022-06-29
Fix pipeline uninstalling a parent instead of upgrading it

tanya-borisova:
# PR for issue https://github.com/microsoft/AzureTRE/issues/2150 

## What is being addressed

The bug was introduced in https://github.com/microsoft/AzureTRE/pull/2136 
I passed a wrong ID when updating the status in the database, so unsurprisingly the whole pipeline broke after that. 

## How is this addressed

- Fixed the bug
- Made a test that was failing on the code that was there before, but passes now


api_app/tests_ma/test_db/test_repositories/test_airlock_request_repository.py
```diff
@@ -21,7 +21,7 @@
 REJECTION_IN_PROGRESS = AirlockRequestStatus.RejectionInProgress
 REJECTED = AirlockRequestStatus.Rejected
 CANCELLED = AirlockRequestStatus.Cancelled
-BLOCKING_IN_PROGRESS = AirlockRequestStatus.Blocked
+BLOCKING_IN_PROGRESS = AirlockRequestStatus.BlockingInProgress
```
eladiw, line 24: not sure this is correct.

@anatbal ?

eladiw (commented):
looks good (airlock tests fixes) I will let someone else to approve. Thanks

damoodamoo (approved):

PR #2162, opened 2022-06-30
Support unit tests reports from forks

### LINKED ISSUE #2161: Unit-tests failures don't fail the workflow
**Describe the bug**
Our docker build workflow runs the unit tests but seem to complete successfully despite having tests that failed.

**Steps to reproduce**
https://github.com/microsoft/AzureTRE/runs/7124174283?check_suite_focus=true

tamirkamara:
Fixes #2161 

## What is being addressed

We currently don't support reporting of unit tests results for PRs from forks. As a result (probably) we don't require a check to exist on the main branch which means we are introducing changes that don't pass tests.

## How is this addressed

- Move the publish task to a different workflow that would run under the repo context even for forks. More info [here](https://github.com/marketplace/actions/publish-unit-test-results#support-fork-repositories-and-dependabot-branches).


martinpeck:
cc @stuartleeks

tamirkamara:
> Have you run this on your own branch. Can you paste the completed Build Summary so we can see?

No @ross-p-smith as the new workflow can only run once it's in the main branch.

tamirkamara:
/test-force-approve
These changes don't impact deployment and anyway have to be in main to see the impact.

Issue #2164, opened 2022-06-30
EventGrid topics fail to provision due to internal Azure issue


tamirkamara:
**Describe the bug**

Our system Event Grid topics fail to provision due to an underlining Azure issue which is being worked on. 
This is to track a mitigation step we take temporarily while waiting for a proper fix to propagate and be validated in our scenario. 

The error we get is:

> The operation failed due to an internal server error. The initial state of the impacted resources (if any) are restored. Please try again in few minutes. If error still persists, report e075e1e9-d465-452a-8f63-a52dbd61bf64:6/28/2022 9:39:01 PM (UTC) to our forums for assistance or raise a support ticket .

tamirkamara:
Reopening to track fixes on Azure product's side

tamirkamara:
Fixed by the product team.

PR #2165, opened 2022-06-30
[temp fix] Delay between system topics creation

tamirkamara:
## What is being addressed

As part of a temporary mitigation of #2164 we need to add 1 minute delay between two event grid system topics provisioning actions.


eladiw (approved):

tamirkamara:
/test-force-approve
recent commit was only ui so no point in rerunning tests

PR #2172, opened 2022-06-30
Devcontainer on Az CLI 2.37

ross-p-smith:
# PR for issue 1645

Use Az CLI 2.37 in the DevContainer

## Implications

- This means that the az cli will use the Microsoft Graph rather than the AD Graph - https://docs.microsoft.com/en-us/graph/overview.
- Many of the az ad commands now have breaking changes and this PR resolves them


.devcontainer/Dockerfile
```diff
@@ -79,7 +79,7 @@ COPY ["airlock_processor/requirements.txt", "/tmp/pip-tmp/airlock_processor/"]
 RUN pip3 --disable-pip-version-check --no-cache-dir install -r /tmp/pip-tmp/requirements.txt && rm -rf /tmp/pip-tmp
 
 # Install azure-cli
-ARG AZURE_CLI_VERSION=2.36.0-1~buster
+ARG AZURE_CLI_VERSION=2.37.0-1~buster
```
tamirkamara, line 82: IIRC the version is also pinned in the resource processor image and think those should be in-sync.

ross-p-smith: Many thanks! Done.

damoodamoo:
tested locally:
- Dev container built fine
- `make auth` ran fine
- `make all` hit one issue `inconsistent plan` -> but a retry fixed it, thinking it's likely to be confused state on my side.

Nice work!

damoodamoo (approved):

PR #2176, opened 2022-07-01
Add additional Nexus proxy repos

### LINKED ISSUE #2174: Add additional Nexus proxy repos
## Description

As a TRE Service Integrator
I want to have access to almalinux, fedoraproject, and microsoft apt repos from within a workspace
So that I can install packages as part of workspace service deployment and install packages on VMs

marrobi:
Fixes #2174

## What is being addressed
- Add alma linux, fedora, and microsoft apt repos
- Add MM ubuntu and MS edge repos to VM script
- Retrieve and install MS key from nexus


martinpeck:
Adding the Microsoft proxies makes sense (as you then use them within Guac), but why add the other two?
Are they used somewhere, and does this mean that all customers then implicitly trust these package libraries?

martinpeck (approved):

PR #2182, opened 2022-07-03
Tag resource groups with version info

### LINKED ISSUE #2179: Tag core infra with version information
## Description

I want to know which version of the solution is currently deployed on my subscription
So I could plan upgrades and validate deployments

## Acceptance criteria

- [ ] mgmt/devops resource group is tagged
- [ ] core resource group is tagged
- [ ] each deployment updates the tags

tamirkamara:
Fixes #2179 

## What is being addressed

It's hard to know which version of the core infrastructure is currently deployed in an environment.

## How is this addressed

- Include the git source (what repo deployed the environment) and git summary of the last commit, as well as a hard-coded value from a version file.
- Use azcli to update the info on every deployment since we ignore tag info in terraform

## An example of how this looks like
image


templates/core/terraform/update_tags.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
```
tanya-borisova, line 1: Not so keen on this file being an exact copy of `devops/terraform/update_tags.sh`, is it possible to just reference one script?

devops/terraform/update_tags.sh
```diff
@@ -0,0 +1,31 @@
...
+
+mgmt_rg_rid=$(terraform show -json | jq -r '.values.root_module.resources[] | select(.address=="azurerm_resource_group.mgmt") | .values.id')
+echo "Current tags:"
+az tag list --resource-id "${mgmt_rg_rid}"
+
+version=$(cat "${script_dir}"/../version.txt)
+
+# doesn't work with quotes
+# shellcheck disable=SC2206
+version_array=( ${version//=/ } ) # split by =
```
tanya-borisova, line 18: Learned something new today!

tanya-borisova (commented):
Wondering if we have to copy paste the script in two locations. Otherwise, looks great!

templates/core/terraform/update_tags.sh
tamirkamara: It's not an exact copy as it targets a different resource group...
Most folders have a deploy/destroy script that almost identical to each other.
Never the less, I guess I could move this file to devops/scripts and send the full resource id to update. 
Will that be better @tanya-borisova ?

tanya-borisova: Yeah I'm also not keen on the other almost identical scripts but obviously not asking for it to be fixed in this PR. 
Your idea sounds good, IMO that would be better

tamirkamara: After reconsidering, decided to leave as is

tanya-borisova (approved):

PR #2196, opened 2022-07-04
UI should not use client Id, but rather ApplicationIdURI

ross-p-smith:
## What is being addressed

An AAD Application has a Client Id, this is known as Application Id in AAD, or Client Id in OpenId Connect. This is different to the Scope URI that you request tokens/claims from. They can be the same and often are - but they don't have to be. When you request the token you need to define the scope and that uses the `Application Id URI`. This has to be a URI, but in same older flows it will accept a GUID. Which is why they sometimes get muddled up.

Both of these properties are stored in Cosmos within a workpsace (client_id and scope_id)

## How is this addressed
- Update the UI to use the scope_id when requesting tokens.
- Update the workspace context definition to have a ApplicationIdURI property instead of a client_id
- Update documentation


damoodamoo (approved):
awesome job clearing all this up, thanks!

PR #2204, opened 2022-07-06
Reset vm password

### LINKED ISSUE #2087: Need to handle reset of password for Guacamole user resources
At present the VM password is created and stored in Key Vault This is likely to expire at some point in the future, or the local password may be reset by the user meaning the stored password an actual password differ.

A user needs to be able to reset their password in these scenarios.

jjgriff93:
# PR for issue
Closes #2087 

## What is being addressed

Implement reset password custom action on vm user resources

## How is this addressed

- Added custom action in porter bundles to generate a new password, update key vault secret and update user using the VM extension host

Another method using terraform taint was tested however this caused the whole Vm to be re-deployed so have gone in favour of using azure cli


templates/workspace_services/guacamole/user_resources/guacamole-azure-linuxvm/porter.yaml
```diff
@@ -82,6 +82,23 @@ outputs:
...
+    applyTo:
+      - reset_password
+  - name: keyvault_name
+    type: string
+    applyTo:
+      - reset_password
+  - name: new_password
+    type: string
+    applyTo:
+      - reset_password
```
tamirkamara, line 101: I think this will be picked up in the RP response and persisted to the resource's item on Cosmos. Right?
If yes, is there anyway around this? I mean, not to output those here at the bundle level and use only internally. Just a bit concerned we expose too much information which I prefer not be there...

tamirkamara (commented):
I would've approved except for my point around outputting too much information.

templates/workspace_services/guacamole/user_resources/guacamole-azure-windowsvm/terraform/main.tf
```diff
@@ -78,3 +78,15 @@ output "azure_resource_id" {
...
+
+output "vm_username" {
+  value = random_string.username.result
+}
+
+output "vm_password_secret_name" {
+  value = local.vm_password_secret_name
+}
+
+output "keyvault_name" {
```
ross-p-smith, line 90: why do we need teh keyvault name?

templates/workspace_services/guacamole/user_resources/guacamole-azure-linuxvm/porter.yaml
```diff
@@ -221,3 +238,53 @@ stop:
...
+        resource_group_name: "{{ bundle.parameters.tfstate_resource_group_name }}"
+        storage_account_name: "{{ bundle.parameters.tfstate_storage_account_name }}"
+        container_name: "{{ bundle.parameters.tfstate_container_name }}"
+        key: "{{ bundle.parameters.id }}"
+      outputs:
+        - name: azure_resource_id
+        - name: vm_username
+        - name: vm_password_secret_name
+        - name: keyvault_name
+  - az:
```
marrobi, line 257: @jjgriff93  If these steps were in a single shell script would that avoid the need to pass the secret between stages?

@tamirkamara what do you think?

templates/workspace_services/guacamole/user_resources/guacamole-azure-windowsvm/terraform/main.tf
jjgriff93: So that we can persist the new password to key vault - its not outputted currently and isn't set as a parameter so no other way to get it

templates/workspace_services/guacamole/user_resources/guacamole-azure-linuxvm/porter.yaml
jjgriff93: Yep valid concern as discussed in the stand-up - I'm testing to see if handling all in a single script will negate this problem

templates/workspace_services/guacamole/user_resources/guacamole-azure-windowsvm/generate_password.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
jjgriff93: added

templates/workspace_services/guacamole/user_resources/guacamole-azure-linuxvm/porter.yaml
jjgriff93: Working nicely now, no output so no risk of secrets leaking

jjgriff93:
> I would've approved except for my point around outputting too much information.

@tamirkamara thanks, concerns should now be addressed so please take another look when you get chance and approve if happy

templates/workspace_services/guacamole/user_resources/guacamole-azure-linuxvm/porter.yaml
```diff
@@ -82,6 +82,19 @@ outputs:
...
+    applyTo:
+      - reset_password
+  - name: vm_password_secret_name
+    type: string
+    applyTo:
+      - reset_password
+  - name: keyvault_name
+    type: string
+    applyTo:
+      - reset_password
```
tamirkamara, line 97: Per [this ](https://github.com/getporter/porter/discussions/2230#discussioncomment-3093235) looks like you can remove the bundle level outputs and everything will still work.
As we don't need to return and expose these values... worth a try, right?

tamirkamara (commented):
Another ask please.

templates/workspace_services/guacamole/user_resources/guacamole-azure-linuxvm/porter.yaml
jjgriff93: Ah that's good to know - have removed the unnecessary outputs

tamirkamara (approved):

PR #2205, opened 2022-07-06
Remove innereye workspace & devtestlabs workspace service templates

tamirkamara:
## What is being addressed

We no longer need to have templates for an innereye workspace and devtestlabs workspace service, hence deleting those.


marrobi (approved):
Thanks. Worth noting if anybody is looking for th eInnerEye workspace that it is still available as a workspace service.

ross-p-smith (approved):

tamirkamara:
/test-force-approve
Regular deployment tests can't run for this as the workflows have changed. 
Run with a branch workflow: https://github.com/microsoft/AzureTRE/actions/runs/2628615998

PR #2219, opened 2022-07-06
Upgrading AzureRM Provider

### LINKED ISSUE #2218: Upgrade AzureRm to 3.90
There is a bug with using data.azurerm_client_config.current.object_id if we move to az cli 2.37
https://github.com/hashicorp/terraform-provider-azurerm/issues/16982

Upgrade the AzureRm provider first

ross-p-smith:
Fixes #2218 

## What is being addressed

Upgrade AzureRM provider so that we can eventually upgrade az cli.

## How is this addressed

- Upgrade AzureRM in core and workspace


tamirkamara (approved):
Tested. Looks good.

PR #2228, opened 2022-07-07
"Resource does not exist" log errors fix 

### LINKED ISSUE #2167: API errors about missing cosmos resource to update
**Describe the bug**
On June 27th the API has started to output messages like this:
Service bus message refers to resource id = 09342723-e3dd-4821-830a-f570d3602821 which does not exist

Nothing has changed in the message text or exception logging code so this is most probably something new need investigating.

Logs can be fetched from the main environment with:
```kql
traces
| where cloud_RoleName == "uvicorn"
| where message startswith "Service bus message refers to resource id"
| order by timestamp desc
```

damoodamoo:
# Closes #2167

Has always been a bug here but a recent change from using `isActive` to using the `deploymentStatus` field exposed it. The `get_resource_dict_by_id` should not be filtering for 'active' resources only, it's not a list operation and we want to return the resource regardless of the status. 

When resources were being deleted this error was being seen as the status was `deleted` and the new clause was filtering those items out. It only worked previously as updating the status and updating the `isActive` field happened in 2 different steps.


ross-p-smith (approved):

ross-p-smith:
E2E already passed

/test-force-approve

PR #2231, opened 2022-07-10
Remove Temporary EventGrid Fix

### LINKED ISSUE #2164: EventGrid topics fail to provision due to internal Azure issue

tamirkamara:
## What is being addressed

An Azure EventGrid issue should have been resolved so we now can remove a temporary fix. This should close #2164, which will be validated over time.

## How is this addressed

- Remove delays between event grid topics creation
- Add dependency on role assignment
- Small: check that the core version is updated after something changes in that folder


eladiw (approved):
Lgtm

PR #2248, opened 2022-07-12
API Swagger fix: switch lstrip with replace

tamirkamara:
## What is being addressed

Sometimes a scope_id would get malformed when swagger uses it to get a token from AAD.
For example api://a173fe93... is being used as api://173fe93... notice the missing a in the beginning. 
The cause is us using a python lstrip which strips all char occurrences.

## How is this addressed

- Switch lstrip with replace


marrobi (approved):
LGTM

ross-p-smith:
Great spot - that was introduced during Code Review!

tanya-borisova:
Sorry that was my fault :(

PR #2252, opened 2022-07-12
Added actionDisabled states

### LINKED ISSUE #2201: When a custom action fails, delete connect and action buttons get disabled
**Describe the bug**

As stated in the title when you have a custom action that fails, certain context menu buttons get disabled as the resource thinks its in a failed state - but this shouldn't be the case for individual custom actions that aren't install/update/uninstall. I.e. if Stop Vm fails for any reason, you should still have the button to Connect to it or invoke other actions

image

**Steps to reproduce**

1. Push a user resource bundle with a custom action that will fail
1. Deploy that bundle
1. Invoke the action
1. After failure the context buttons will be disabled

jjgriff93:
# PR for issue
Fixes #2201 

## What is being addressed

Fixes a bug where actions are disabled when non-terminal failure resource states are returned by the API (i.e. action_failed)

## How is this addressed

- Created a separate `actionsDisabledStates` array to determine which states to disable actions on
- Modified context menu and resource card `shouldDisable()` logic to check if this array


damoodamoo:
If the only files are changed are under `/ui` then no need to run `/test` - none of the UI code is currently touched or deployed during tests

PR #2254, opened 2022-07-13
UI Schema 

### LINKED ISSUE #895: Enable UI Schema to be provided with bundles and returned with templates
## Description

As a Service Integrator
I want to provide a uischema.json file with a bundle
So that I can control the layout of the bundle input parameters

## Acceptance critiera

- [ ] Optional uischema.json provided with bundle in same way as customer parameters json part of register bundle payload
- [ ] UI Schema stored with the template int he state store
- [ ] UI Schema returned with the template

damoodamoo:
# Closes #895 

- New `uiSchema` object supported at top level of the template. Implemented in the workspace template as an example
- Method to hide fields via CSS (`tre-hidden`) - used to hide `aad_redirect_uris` in  workspace template
- If no ordering is supplied, the UI will default to title / description / overview / all other fields

> Note: For conditional display (if this field is true, show that field) - that's actually in the json schema itself, not the UI schema, and might require a bit more high level change. To be discussed.


ross-p-smith (approved):

damoodamoo:
/test-force-approve
extended tests were successful but no status update?

Issue #2253, opened 2022-07-12
Migrations fail with 504 Gateway Time Out on main


tanya-borisova:
**Describe the bug**

Builds on main fail with 504 Gateway Time-out, most likely because the migration script is taking too long.
In the logs, we see a lot of messages like this:

<img width="689" alt="image" src="https://user-images.githubusercontent.com/52954123/178536102-91e12335-b231-4084-8332-b7298d10e882.png">

This is pointing to api_app/db/migrations/resources.py . 
It seems that getting all operations and then corresponding resources is inefficient. Instead, this migration could be rewritten to only getting the resources which don't already have deploymentStatus set. 

**Steps to reproduce**

One of the runs that failed on main: https://github.com/microsoft/AzureTRE/runs/7304814409?check_suite_focus=true

PR #2257, opened 2022-07-13
Make resource migration more efficient

tanya-borisova:
# Resolves https://github.com/microsoft/AzureTRE/issues/2253

## What is being addressed

Migrations that add deploymentStatus to resources are taking a long time on main (because resources on main have a lot of operations associated with them).

In this PR, making it more efficient, via only querying resources that do not already have a deploymentStatus field.


damoodamoo (approved):

templates/workspaces/base/porter.yaml
```diff
@@ -96,7 +96,7 @@ parameters:
   - name: app_service_plan_sku
     type: string
     description: "The SKU used when deploying an Azure App Service Plan"
-    default: "P1v3"
+    default: "P1v2"
```
ross-p-smith, line 99: I thought that this would need to be a default of P1v3 - as that is what we want customers to use

tanya-borisova: Uh oh, this slipped in from my testing! Thanks for the catch

PR #2266, opened 2022-07-14
Use the config value for app plan in the tests

### LINKED ISSUE #2255: Pass app_service_plan_sku into porter from the E2E tests
**Describe the bug**
When running the E2E tests we default the App Service plan to P1v3 in the porter.yaml. Change the E2E to set the app_service_plan_sku to be whatever is defined in the .env file. This means the E2E will work better locally.

ross-p-smith:
## Fixes #2255 

Use the APP_SERVICE_PLAN_SKU environment variable in the tests so that we can chose P1V2

## How is this addressed

- Pass the value into the tests rather than using the default


ross-p-smith:
/test-force-approve

(Previously passed the tests)

e2e_tests/test_workspace_services.py
```diff
@@ -20,7 +20,8 @@ async def test_create_guacamole_service_into_base_workspace(admin_token, verify)
             "description": "workspace for E2E",
             "address_space_size": "small",
             "client_id": f"{config.TEST_WORKSPACE_APP_ID}",
-            "client_secret": f"{config.TEST_WORKSPACE_APP_SECRET}"
+            "client_secret": f"{config.TEST_WORKSPACE_APP_SECRET}",
+            "app_service_plan_sku": f"{config.TEST_WORKSPACE_APP_PLAN}"
```
stuartleeks, line 24: I have something similar to this in my monitoring PR as I hit the same issue.

Rather than directly adding the `app_service_plan_sku`, I includeed this on the line after `payload` was initialised:

```python
    if config.APP_SERVICE_PLAN_SKU != "":
        payload["properties"]["app_service_plan_sku"] = config.APP_SERVICE_PLAN_SKU
 ```
 
My thinking was that this would skip it in the case where a value isn't provided by the env var and then the default from the porter config would be used?

ross-p-smith: I've just pushed a change where I default the value in the config definition

ross-p-smith: My thoughts are that we shouldn't really have that default in the porter.yaml - it should always be passed in

damoodamoo (approved):

stuartleeks:
/test-extended
(updated from `main` and re-trying due to event grid error)

damoodamoo (approved):

PR #2280, opened 2022-07-17
Fix conflict in webapps vnet integration 

### LINKED ISSUE #2279: API is inaccessible after TF resource upgrade
**Describe the bug**
PR #2232 broke something that made the API inaccessible.

**Steps to reproduce**

1. Deploy the system

tamirkamara:
Fixes #2279 

## What is being addressed

We usually use a azurerm_app_service_virtual_network_swift_connection to define a vnet integration for webapps. But now with azurerm_linux_web_app that can be set internally with virtual_network_subnet_id, and the two seem to conflict sometimes. 

## How is this addressed

- In the API disregard changes on virtual_network_subnet_id. This is done to be backward compatible with existing deployments.
- In Guacamole, remove the swift connection and use virtual_network_subnet_id to define the integration.


LizaShak (approved):
LGTM

PR #2301, opened 2022-07-18
Uncomment xtrace in delete_vm_extensions.sh

tanya-borisova:
# Resolves issue

## What is being addressed

Trying to investigate what is happening by making the script log what it's doing. 

Contrary to the comment, it should't show any secrets, because none are passed in the script.  It will show the resource group name, though. I hope it won't be a big problem


martinpeck (approved):
Suggest you rename the PR (so that it doesn't mention xtrace)

PR #2303, opened 2022-07-18
Airlock - optionally support import/export per workspace

### LINKED ISSUE #2221: Optionally support import/export per workspace
## Description

As a TRE Administrator
I want to have the option to enable/disable airlock functionality per workspace

## Acceptance criteria

- [ ] Base workspace template has the option to turn import/export off (default is on)
- [ ] API returns as error when an import request is opened but the target workspace doesn't have it enabled
- [ ] API returns as error when an export request is opened but the source workspace doesn't have it enabled

API check logic could target the existence of the storage account

yuvalyaron:
# Resolves issue
Fixes #2221 

## What is being addressed
Added the option to choose whether to deploy airlock resources when deploying a workspace.

## How is this addressed

- Added property to the base workspace template to control whether to deploy airlock resources in the workspace or not.
- Added validation in the API - if the "enable_airlock" property is false, a "create airlock request" call will result in HTTP 405.


api_app/tests_ma/test_db/test_repositories/test_airlock_request_repository.py
```diff
@@ -117,7 +117,7 @@ def test_update_airlock_request_status_with_allowed_new_status_should_update_req
 
 
 @pytest.mark.parametrize("current_status, new_status", get_forbidden_status_changes())
-def test_update_airlock_request_status_with_frobidden_status_should_fail_on_validation(airlock_request_repo, current_status, new_status, verify_dictionary_contains_all_enum_values):
```
yuvalyaron, line 120: fixed a typo

api_app/api/routes/airlock.py
```diff
@@ -30,6 +30,8 @@
 # airlock
 @airlock_workspace_router.post("/workspaces/{workspace_id}/requests", status_code=status.HTTP_201_CREATED, response_model=AirlockRequestInResponse, name=strings.API_CREATE_AIRLOCK_REQUEST, dependencies=[Depends(get_current_workspace_owner_or_researcher_user), Depends(get_workspace_by_id_from_path)])
 async def create_draft_request(airlock_request_input: AirlockRequestInCreate, user=Depends(get_current_workspace_owner_or_researcher_user), airlock_request_repo=Depends(get_repository(AirlockRequestRepository)), workspace=Depends(get_deployed_workspace_by_id_from_path)) -> AirlockRequestInResponse:
+    if workspace.properties.get("enable_airlock") is False:
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=strings.AIRLOCK_NOT_ENABLED_IN_WORKSPACE)
```
LizaShak, line 34: maybe 405? because the resource does exist, it is just that you can't create requests underneath it

yuvalyaron: yes it makes more sense... changed

LizaShak (approved):

PR #2307, opened 2022-07-19
Fixed substitution for list of strings

damoodamoo:
When doing the substitutions in a pipeline step, an issue was found when the value object contained a list of strings - as each string was being treated as a dict.


tamirkamara (approved):
LGTM

Issue #2237, opened 2022-07-11
New `shared-services/overview` endpoint + layout


damoodamoo:
Update the `GET` endpoint to the API for TRE Researchers (and other roles). This will return a list of shared services, initially only containing:
- Title
- Description
- Overview

- [ ] Limit shared service fields shown to researchers
- [x] Add "Shared Services" link to the workspace left nav, showing the list of Shared Services. Clicking each will navigate to a simple resource page, with no menu, just showing the Title / Overview fields.

PR #2308, opened 2022-07-19
Restricted Shared Service: API and UI Screens

damoodamoo:
# Resolves issue #2237 

- List endpoint for shared-services now returns a `RestrictedResource` list, which limits the properties that can be returned to `display_name`, `description` and 'overview`.
- Get by shared service ID endpoint now allows TRE users to call it (it was previously limited to TRE Admins only), and checks if the user is a TRE Admin or not. If an admin, returns the full fat object, else a restricted one. This was needed as a user may land on a restricted shared service page in the UI and need to load just that item.
- The resource views in the UI now have an optional `readonly` property, that can be passed in and prevent rendering of all options and menus, regardless of user role.
- Workspace services in the Workspace use the `readonly` property to make sure Admins in the workspace are not confused into thinking that the shared service relates to the workspace, or can be managed separately here.

Card view: 
image

Resource page showing shared service in the workspace:
image


api_app/models/domain/restricted_resource.py
```diff
@@ -0,0 +1,28 @@
...
+from models.domain.azuretremodel import AzureTREModel
+
+
+class RestrictedProperties(AzureTREModel):
+    display_name: str = ""
+    description: str = ""
+    overview: str = ""
+
+
+class RestrictedResource(AzureTREModel):
```
tanya-borisova, line 13: Out of curiousity, what were the fields in shared service that we want masked?

tanya-borisova (commented):
I've only looked at Python part. Looks good, just have one question

api_app/models/domain/restricted_resource.py
damoodamoo: anything in the properties object that's NOT name/description basically - which would include things like firewall rules / certificate names etc potentially

marrobi (approved):
Screen shot looks fine.

damoodamoo:
/test-force-approve
Ran shared service tests previously

Issue #2313, opened 2022-07-21
Tests failing on main build


tanya-borisova:
**Describe the bug**

A recent failed run: https://github.com/microsoft/AzureTRE/actions/runs/2711014023 

There are two failed tests: 

`test_shared_services.py::test_create_shared_service[tre-shared-service-gitea]` seems to fail after 30 minutes which is what the timeout is set to.
`test_shared_services.py::test_create_shared_service[tre-shared-service-nexus]` seems to fail with the following log: 

```
╷
│ Error: A resource with the ID "/subscriptions/*******/resourceGroups/rg-maintre1/providers/Microsoft.Network/azureFirewalls/fw-maintre1/applicationRuleCollections/arc-web_app_subnet_nexus_v2" already exists - to be managed via Terraform this resource needs to be imported into the State. Please see the resource documentation for "azurerm_firewall_application_rule_collection" for more information.
│
│ with azurerm_firewall_application_rule_collection.api_driven_rules["4"],
│ on firewall.tf line 297, in resource "azurerm_firewall_application_rule_collection" "api_driven_rules":
│ 297: resource "azurerm_firewall_application_rule_collection" "api_driven_rules" {
│
╵
```

PR #2314, opened 2022-07-21
Main build: Increase timeout for shared services test

tanya-borisova:
# Resolves issue https://github.com/microsoft/AzureTRE/issues/2313 

## What is being addressed

The last few tests on main failed partially due to test `test_shared_services.py::test_create_shared_service[tre-shared-service-gitea]`  timing out. 
See e.g. https://github.com/microsoft/AzureTRE/actions/runs/2711014023 

Because it now includes an update step as part of pipeline, and because it first _deletes_ gitea shared service (wonder if we still need that logic?), it takes longer than 30 minutes.

### DIFF
--- a/e2e_tests/test_shared_services.py
+++ b/e2e_tests/test_shared_services.py
@@ -92,7 +92,7 @@ shared_service_templates_to_create = [
 
 
 @pytest.mark.shared_services
-@pytest.mark.timeout(30 * 60)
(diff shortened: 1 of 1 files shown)


damoodamoo (approved):

marrobi (approved):

tanya-borisova:
/test-force-approve

It's just a timeout, let's get this in to see if main succeeds...

PR #2318, opened 2022-07-22
Increase timeout for e2e tests in CI

tanya-borisova:
# Resolves issue

## What is being addressed

A recent e2e test run in main took longer than 2 hours 20 minutes: 
https://github.com/microsoft/AzureTRE/runs/7460985823?check_suite_focus=true

I suggest to increase the timeout to 3 hours for now. 
I"m not a fan of tests taking so long. But I would prefer to first get a test run passing in main before fixing the running time


stuartleeks (approved):

tanya-borisova:
/test-force-approve

This is only a timeout

PR #2324, opened 2022-07-24
Core apps use proper cloud_RoleName values

tamirkamara:
## What is being addressed

Reading logs of the api and/or resource processor isn't intuitive since their cloud_RoleName values aren't well known.

## How is this addressed

- Change logging for both apps and set cloud_RoleName to **api** and **resource_processor** which will make finding the right logs clearer.
- Add the version of each app to the telemetry payload

In addition:
- Remove api's usage of appinsights instrumentation key in favor for a connection string (as the key is going to be deprecated)
- Upgrade Opencensus to the latest version
- Optimize the api image build by instructing pip to not save cache (saves around 50MB)


stuartleeks (approved):
LGTM

Issue #2200, opened 2022-07-05
RP cloudinit failure in /test environment causes Firewall deploy timeout


jjgriff93:
**Describe the bug**
When deploying an environment as part of `/test` in CI, the Firewall failed to deploy and after an 1h 15m timed out. This resulted in this message in the GitHub actions logs (returned by the API):

```
Got a non-retrieable HTTP status code: 500
[1015](https://github.com/microsoft/AzureTRE/runs/7196633405?check_suite_focus=true#step:3:1018)
HTTP/1.1 500 Internal Server Error
[1016](https://github.com/microsoft/AzureTRE/runs/7196633405?check_suite_focus=true#step:3:1019)
make: *** [Makefile:336: deploy-shared-service] Error 1
[1017](https://github.com/microsoft/AzureTRE/runs/7196633405?check_suite_focus=true#step:3:1020)
Date: Tue, 05 Jul 2022 13:55:08 GMT
[1018](https://github.com/microsoft/AzureTRE/runs/7196633405?check_suite_focus=true#step:3:1021)
Content-Type: text/plain; charset=utf-8
[1019](https://github.com/microsoft/AzureTRE/runs/7196633405?check_suite_focus=true#step:3:1022)
Content-Length: 25
[1020](https://github.com/microsoft/AzureTRE/runs/7196633405?check_suite_focus=true#step:3:1023)
Connection: keep-alive
[1021](https://github.com/microsoft/AzureTRE/runs/7196633405?check_suite_focus=true#step:3:1024)
Server: uvicorn
[1022](https://github.com/microsoft/AzureTRE/runs/7196633405?check_suite_focus=true#step:3:1025)
[1023](https://github.com/microsoft/AzureTRE/runs/7196633405?check_suite_focus=true#step:3:1026)
Unable to process request
```

After investigating the test environment it seems the Resource Processor cloudinit bootstrapping failed because it was unable to connect to `packages.microsoft.com` which failed the rest of the package installations:

Cloudinit log:
```
Hit:1 http://azure.archive.ubuntu.com/ubuntu bionic InRelease
Get:2 http://azure.archive.ubuntu.com/ubuntu bionic-updates InRelease [88.7 kB]
Get:3 http://azure.archive.ubuntu.com/ubuntu bionic-backports InRelease [74.6 kB]
Get:4 http://azure.archive.ubuntu.com/ubuntu bionic-security InRelease [88.7 kB]
[...]

PR #2326, opened 2022-07-24
Turn on vmss auto-repair

### LINKED ISSUE #2325: Use VMSS auto-repair capabilities
## Description

As a TRE Developer
I want to the VMSS (resource process) to use its auto-repair capabilities
So that the system can come back from a hard failure without a person needing to take any action

This will help in situations where the VM doesn't provision correctly (like docker install fails)

## Acceptance criteria

- [ ] Auto-repair is turned on
- [ ] A new instance is created by the VMSS when the existing one fails its health check

tamirkamara:
Resolves #2325 

## What is being addressed

We currently don't use the ability of VMSS to repair failed instances which just leaves them there despite not doing their job.

## How is this addressed

- Turn on auto-repair with a grace of 10 minutes.

This will help #2200


anatbal (approved):
LGTM

PR #2329, opened 2022-07-24
Resource processor checks its runner processes

### LINKED ISSUE #2327: Resource processor health should check its sub-processes
## Description

As a TRE Developer
I want to check the sub-processes in the resource processor (as those do the actual work)
So that health could be reported more accurately 

## Acceptance criteria

- [ ] Health won't be reported as ok if all processes are down

tamirkamara:
Fixes #2327 

## What is being addressed

We've observed situations where the resource processor main process is online and health (http returns 200) but its runner sub processes are not operational (might be related to opencensys issues).

## How is this addressed

- Add a periodic check on the sub processes. If all are down - bring the http server down as well. This will later cause VMSS to repair the instance as it will be considered unhealthy.


damoodamoo (approved):
like it!

PR #2331, opened 2022-07-24
Bundles build their own runtime image

### LINKED ISSUE #1472: Remove Docker images relating to services from `make all`
## Description

As a TRE Administrator I do not want to have to build and push docker images that I may never use

To be discussed. Given each service are optional and the number of service specific docker images is growing, I propose that they should not be part of make all. Maybe building service specific docker images could be part of the the `bundle-build` stage.

tamirkamara:
# Resolves #1472 

## What is being addressed

We currently have to prebuild and publish any runtime image a bundle might use as a separate step. This adds makes the dev cycle longer as most such images aren't of interest usually.

## How is this addressed

- Add a custom section in bundle's porter.yaml to note how the runtime image should be built
- Add scripts that reads a yaml a execute docker build/push according to the definition
- make bundle-build/publish call these scripts as a pre-step for the relevant porter command
- Remove cicd job that used to built these images, as well as sections of the makefile.

## Future steps
It would be nice to investigate if embedding the runtime image can work for us.
https://getporter.org/bundle/manifest/#images
https://porter.sh/authors/templates/#images


devops/scripts/bundle_runtime_image_build.sh
```diff
@@ -0,0 +1,37 @@
...
+# set -o xtrace
+
+if [ "$(yq eval ".custom.runtime_image.build" porter.yaml)" == "null" ]; then
+  echo "Runtime image build section isn't specified. Exiting..."
+  exit 0
+fi
+
+image_name=$(yq eval ".custom.runtime_image.name" porter.yaml)
+version_file=$(yq eval ".custom.runtime_image.build.version_file" porter.yaml)
+docker_file=$(yq eval '.custom.runtime_image.build.docker_file' porter.yaml)
```
tanya-borisova, line 15: Small nit: use " quotes

tanya-borisova (approved):
Love this!

tamirkamara:
/test-force-approve
Can't test with the bot since there're workflow & makefile changes

ross-p-smith (approved):

PR #2335, opened 2022-07-24
Jumpbox uses a known username

### LINKED ISSUE #1118: Jumpbox Username and Password in KeyVault Not In Obvious Format and Docs Lack Clarity
In order to complete setup you need to set up Shared Services:

https://microsoft.github.io/AzureTRE/tre-admins/setup-instructions/configuring-shared-services/

During this phase you are asked to connect to the jumpbox using credentials stored in KeyVault.
The creds, in keyvault, are not clearly labelled "username" or "password", so it's not obvious which is which. Or, in fact that there's both a username and password present. You simply see something like this (not real)...

```
sd4f
Ds09435hc=&3w4hj@
```
... with no guidance that the first is a username, and the second is a password. They both look like part of a password that has wrapped across 2 lines.

Suggest: 

- add labels into this secret to make it clearer, or
- make the instructional text in the docs explain the format of this secret, or
- or, make the username look much more like a username (e.g. prefix it with `user_`)

Or all of the above

tamirkamara:
Fixes #1118 

## What is being addressed

The admin jumpbox is currently using a random username which makes login a bit more complicated than it needs to.

## How is this addressed

- Use a known username for the jumpbox `adminuser`
- Upgrade jumpbox vm image
- Update random Terraform resources in the core to use numeric instead of number (which is deprecated)


templates/core/terraform/resource_processor/vmss_porter/main.tf
```diff
@@ -30,7 +30,7 @@ resource "random_password" "password" {
   min_lower        = 1
   upper            = true
   min_upper        = 1
-  number           = true
+  numeric          = true
```
anatbal, line 33: 🚀

anatbal (approved):

tamirkamara:
/test-force-approve
only change since last success was version.

PR #2336, opened 2022-07-25
Check status runs after all tests

### LINKED ISSUE #1991: Deploy TRE workflow approval check should depend on extended tests if running
**Describe the bug**
The deploy TRE workflow will send an approval (green check) if the smoke tests are passing while disregarding the extended tests result even when running.
This gives a false positive if developer is running all the tests (including the extended ones). 

**Steps to reproduce**
See this: https://github.com/microsoft/AzureTRE/actions/runs/2454518265

tamirkamara:
# Resolves #1991

## What is being addressed

The E2E check status is reported after smoke tests have passed even if we also run other tests (like extended) in the same run. This might give false positive on the PR as it will turn on the green button despite the extra tests failing.

## How is this addressed

- Move the check status report step into the summary job


stuartleeks (approved):
It feels a little odd to me to have the "E2E Tests (Smoke)" reported at the end of an extended run, but I get why :-)

tamirkamara:
/test-force-approve
This is a workflow change which isn't testable with the bot

Issue #2079, opened 2022-06-21
UI: Properties with complex types rendered as [Object]


damoodamoo:
Now some properties are real complex types rather than just strings, the properties panel needs to reflect this and iterate the object to show the real value rather than `[object]`

![image](https://user-images.githubusercontent.com/35696285/174811295-2edd26c4-f788-4978-a3b4-7b63ee33d480.png)

damoodamoo:
^ this is in the details tab. We can hide all complex values in the details tab if we want, but if i'm looking for the details I think i'd want to see this?

marrobi:
I can't think any reason a  researcher would want to know the redirect URLs, or many of the other fields. As an admin/dev, to debug, maybe... but they could go to cosmos.

We need a UI Schema for the forms, to sort ordering, is it possible to have a UI schema for forms, cards, and details tab?

damoodamoo:
Yea the UI schema for the form is still to-do, we could look to extend that to include a list of props to show on the properties panel, and then use the template to drive what we see on those panels. 

The card is maybe a bit trickier due to the space available.

marrobi:
For the card, the things that spring to mind are VM related, like CPU cores, RAM etc. (actually a single property?).

PR #2340, opened 2022-07-25
Recursive display component for complex props

damoodamoo:
# Resolves issue #2079 

For complex properties - like firewall rules, there was no good way to previously display them, and they'd be rendered as `[Object]` in the UI. This PR pushes any complex property types to a custom modal popup which recurses the object to allow the user to explore the property. 

In the display panel:
image

The popup:
image


marrobi (approved):
LGTM

PR #2349, opened 2022-07-26
Adding Airlock Manager Role - Part I

### LINKED ISSUE #2269: Airlock - add airlock manager role
## Description

As a TRE Developer
I want to define an airlock manager role
So that this role will be given the review permissions for airlock requests instead of ws owner

## Acceptance criteria

Part I:
- [x] airlock manager can review requests

Part II:
- [ ] notification are now sent to an airlock manager(not to ws owner)

### LINKED ISSUE #2380: Airlock fields aren't ordered the same way in Swagger
**Describe the bug**

In some APIs the workspace_id appears first and in others it's second
We need to be consistent with the all the APIs, so the workspace_id should always come first.

anatbal:
# Part 1 of #2269 
# Fixes #2380 

## What is being addressed

- [x]  Adding new airlock manager role
- [x] Allowing only airlock manager to review airlock request
- [x] Edit unit tests to work with the new role


api_app/api/routes/airlock.py
```diff
@@ -15,15 +15,15 @@
 from db.repositories.airlock_requests import AirlockRequestRepository
 from models.schemas.airlock_request import AirlockRequestInCreate, AirlockRequestInResponse
 from resources import strings
-from services.authentication import get_current_workspace_owner_or_researcher_user, get_current_workspace_owner_user
+from services.authentication import get_current_workspace_owner_or_researcher_user_or_airlock_manager, get_current_workspace_owner_or_researcher_user, get_current_airlock_manager_user
```
LizaShak, line 18: I don't see you use the get_current_workspace_owner_or_researcher_user_or_airlock_manager anywhere, also not sure it is needed

LizaShak:
When workspace is created we save auth info regarding it, we should add that role there too...

templates/workspaces/base/porter.yaml
```diff
@@ -89,6 +89,10 @@ parameters:
     type: string
     default: ""
     description: "The id of the application role WorkspaceResearcher in the identity provider"
+  - name: app_role_id_airlock_manager
```
ross-p-smith, line 92: We'd need to bump the porter version to make sure this builds

api_app/api/routes/airlock.py
anatbal: On line 26, it is necessary

templates/workspaces/base/porter.yaml
anatbal: Yes you are right. I will test it again

anatbal:
> When workspace is created we save auth info including the role object ids, we should add that role there too...

Let's speak offline and see where did I miss saving exactly

LizaShak (approved):

ross-p-smith (approved):

anatbal:
/test-force-approve

Tests previously completed

Issue #2358, opened 2022-07-26
Move shared services tests to a nightly build


tanya-borisova:
Shared service tests take quite a long time, but there is currently not a lot of development or changes happening. 
It should be low risk to move these tests to nightly build only.

PR #2359, opened 2022-07-26
Move shared service tests to a nightly build

tanya-borisova:
# Resolves issue https://github.com/microsoft/AzureTRE/issues/2358

## What is being addressed

In order to shorten the time the main build runs for, move shared service tests into a separate nightly workflow.


.github/workflows/deploy_tre_nightly.yml
```diff
@@ -0,0 +1,53 @@
...
+concurrency: "${{ github.workflow }}-${{ github.ref }}"
+
+jobs:
+  run-deploy-tre-main:
+    name: "Deploy main"
+    if: github.ref == 'refs/heads/main'
+    uses: ./.github/workflows/deploy_tre_reusable.yml
+    with:
+      ciGitRef: ${{ github.ref }}
+      e2eTestsCustomSelector: "extended or shared_services or extended_aad"
```
tamirkamara, line 23: This is the only change, right?
Any special reason to split this into its own workflow rather than use an expression for the different event types?

tanya-borisova: No special reason really, I'll look into using event types, thanks!

.github/workflows/deploy_tre.yml
```diff
@@ -22,7 +22,9 @@ jobs:
     uses: ./.github/workflows/deploy_tre_reusable.yml
     with:
       ciGitRef: ${{ github.ref }}
-      e2eTestsCustomSelector: "extended or shared_services or extended_aad"
+      e2eTestsCustomSelector: >-
+        ${{ (github.event_name == 'schedule' && 'extended or shared_services or extended_aad') ||
+        (github.event_name == 'push' && 'extended or extended_aad') }}
```
tamirkamara, line 27: There's another event for this workflow which is the dispatch. I think you should add an "else" at the end to catch every other even with a default...

tamirkamara (approved):

stuartleeks (approved):
lgtm!

tanya-borisova:
/test-force-approve

Github Workflow change - no point running tests on

PR #2362, opened 2022-07-26
Guacamole app check in E2E

### LINKED ISSUE #1602: Guacamole E2E tests are stuck on HTTP 302
**Describe the bug**
Accessing the Guacamole endpoint in the E2E always returns HTTP 302 - a redirect to log-in probably, which means that the credentials are not setup correctly?

**Steps to reproduce**
https://github.com/microsoft/AzureTRE/runs/5688108537?check_suite_focus=true

tamirkamara:
# Resolves #1602 

## What is being addressed

We removed checking the Guacamole app in the tests due to the move to OAuth proxy. It turns out we can't really reach the Guacamole app within a test (by a reasonable investment) due to OAuth flow and the way the proxy works (there's a cookie involved). 
But we really need to check the app is online ([this bug](https://github.com/microsoft/AzureTRE/pull/2361) could have been avoided)

## How is this addressed

- Bring back the check and confirm it redirects to the AAD authentication endpoint. This is at the very least says that the docker image was loaded successfully and the OAuth proxy is online.


tamirkamara:
/test-force-approve
There a failure with the airlock test but everything else passes:
https://github.com/microsoft/AzureTRE/actions/runs/2740605935/attempts/3

LizaShak (approved):

PR #2376, opened 2022-07-31
Guacamole sends logs to app insights

### LINKED ISSUE #1938: Include AppInsights with Guacamole
## Description

I want to view logs and metrics about the Guacamole java application within Application Insights
So that monitor is more easily

## Acceptance criteria

- [ ] AppInsights is included in the image and wired as a java agent
- [ ] A connection string is injected into the runtime container
- [ ] Data is being received in the core Application Insights instance

tamirkamara:
# Resolves #1938 

## What is being addressed

Guacamole logs weren't being sent to app insights since we didn't have one in the workspace level, which we do now...

## How is this addressed

- Include app insights in the guacamole image so that logs will be forwarded
- Move the version file for its docker image into the image directory instead of the root one
- Upgrade Java and Tomcat versions


templates/workspace_services/guacamole/guacamole-server/docker/version.txt
```diff
@@ -0,0 +1 @@
+__version__ = "0.3.6"
```
anatbal, line 1: shouldn't it also be 0.4.0?

anatbal (approved):
Looks good, minor comment

templates/workspace_services/guacamole/guacamole-server/docker/version.txt
tamirkamara: No, only core components follow the main repo tags. Other templates should have versioning done as it makes sense and I thought this is a small change so just a minor version bump.

PR #2382, opened 2022-08-01
Remove usage of storage account key in airlock processor 

### LINKED ISSUE #2185: Airlock processor uses user delegated key to acquire SAS token
## Description

As a Azure TRE Developer
I want to acquire SAS token using user delegated key
So that we use a more secure way to obtain the SAS token.

## Acceptance criteria

- [ ] generate sas token uses a user delegated key to obtain it

tamirkamara:
# Resolves #2185 

## What is being addressed

The airlock processor uses storage account keys for its operations (like copy) which is less secure and complicate the code...

## How is this addressed

- Copy operations are now performed by using a SAS that is signed by a [user delegated key](https://docs.microsoft.com/en-us/rest/api/storageservices/create-user-delegation-sas) rather than the account key
- Create container is done with AD Auth (rather than connection string that includes the account key)
- Align (and simplify) role assignments in Terraform


LizaShak (approved):
LGTM

PR #2385, opened 2022-08-01
Updates to support https://github.com/microsoft/AzureTRE-Deployment

stuartleeks:
## What is being addressed

## How is this addressed

- Fix `.env` file path resolution
- Fix path references in UI scripts

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -344,7 +344,7 @@ setup-local-debugging:
 auth:
 	$(call target_title,"Setting up Azure Active Directory") \
 	&& . ${MAKEFILE_DIR}/devops/scripts/check_dependencies.sh nodocker,env \
-	&& . ${MAKEFILE_DIR}/devops/scripts/create_aad_assets.sh
+	&& ${MAKEFILE_DIR}/devops/scripts/create_aad_assets.sh
 
 show-core-output:
 	$(call target_title,"Display TRE core output") \
--- a/devops/scripts/aad/get_existing_app.sh
+++ b/devops/scripts/aad/get_existing_app.sh
@@ -1,7 +1,6 @@
 #!/bin/bash
-set -euo pipefail
-# Use this for debug only
-# set -o xtrace
+
+# This script is designed to be `source`d to create reusable helper functions
 
 function show_existing_app_usage()
 {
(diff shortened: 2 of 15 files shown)


Makefile
```diff
@@ -344,7 +344,7 @@ setup-local-debugging:
 auth:
 	$(call target_title,"Setting up Azure Active Directory") \
 	&& . ${MAKEFILE_DIR}/devops/scripts/check_dependencies.sh nodocker,env \
-	&& . ${MAKEFILE_DIR}/devops/scripts/create_aad_assets.sh
+	&& ${MAKEFILE_DIR}/devops/scripts/create_aad_assets.sh
```
LizaShak, line 347: Out of curiosity :)  why did you remove the dot here? I understand you don't need the output to be on the current context, but in the future we might add more commands here... So was just wondering what affect did it have?

devops/scripts/aad/get_existing_app.sh
```diff
@@ -1,7 +1,6 @@
 #!/bin/bash
-set -euo pipefail
```
LizaShak, line 2: why we don't need this one ? We still want to have an error if any of the commands failed, don't we?

Makefile
stuartleeks: We have two types of script - those that we want to invoke and those that we want to `source` (the `.` form is shorthand).

When we use `source`, we run the contents of the script in the original shell context. This is useful when we want to set environment variables from files or when we want to load shared function definitions. 

The rest of the time (such as here), we just want to invoke the script which makes the control flow more natural. For example, an `exit` in a script called normally will exit that script and return to the caller. In a `source`d script, the `exit` would exit from the _calling_ script because the script is run in the caller's context.

devops/scripts/aad/get_existing_app.sh
stuartleeks: As discussed above, this script file contains shared function definitions and is intended to be `source`d into the parent context.

When invoking a script normally, these directives only apply to the context of the script they are in. For a `source`d script, they apply to the caller's context and we should leave it to the caller to determine the behaviour they want (the calling scripts _do_ all have this set 😄 )

LizaShak (approved):

PR #2397, opened 2022-08-03
Fix stop automation doc

tamirkamara:
## What is being addressed

1. There're some issues stopping VMs in Azure Automation so updating the doc with a workaround.
2. Remove instruction to build the mlflow image as it's done internally by the bundle.


marrobi (approved):
LGTM

Issue #2389, opened 2022-08-02
API's VM Contributor permissions are too wide


tamirkamara:
**Describe the bug**

We give _Virtual Machine Contributor_ permission to the API identity and that is [too wide](https://docs.microsoft.com/en-us/azure/role-based-access-control/built-in-roles#virtual-machine-contributor). For instance, it gets full access to all storage accounts in the subscription. 

A couple of options I see:
- Remove the subscription level permission - not sure we event need it that wide and move to a workspace RG level one. Or even on the resource itself
- Create a custom role that we can use together with #1.

PR #2398, opened 2022-08-03
Assign VM Contributor role on workspace rg (not subscription)

tamirkamara:
## What is being addressed

We give the API identity this role on the subscription level which is too wide on the permission level and the scope level. 
This is a partial solution to #2389.

## How is this addressed

- Remove the role assignment from the subscription level and assign only to the workspace's resource group.

## Migration Steps (breaking change)

You will loose the ability to stop/start VMs in existing workspaces versioned 0.3.20 or lower. To address this, assign `Virtual Machine Contributor` to `id-api-<TRE_ID>` identity on pre-existing workspace resource groups.


LizaShak (approved):

marrobi:
@tamirkamara this causes a larger issue for existing workspaces. Once the subscription level role assignment is removed users cannot list VM user resources as they rely on this role assignment to query the power state. This means as soon as try to list any resources get 500 errors.

Can we make sure this is covered in breaking changes, and maybe we need a script that adds the role to workspaces, or a way to update workspaces...

PR #2411, opened 2022-08-07
Remove nexus v1 from e2e

### LINKED ISSUE #2379: Nexus versions mismatch in E2Es, Bundles, CI
**Describe the bug**

1. We have 2 versions of Nexus:
    * V1 in template: tre-shared-service-nexus in path: templates/shared_services/sonatype-nexus
    * V2 in template: tre-shared-service-sonatype-nexus in path: templates/shared_services/sonatype-nexus-vm
2. VM Bundles have a default to V2
3. E2E is deploying tre-shared-service-nexus = V1
4. CI is publishing templates/shared_services/sonatype-nexus = V1

It means we never actually test the thing we deploy in bundles

tamirkamara:
# Resolves #2379 

## What is being addressed

We were testing an old version of nexus despite not using it anymore in VMs

## How is this addressed

- Remove it from E2E

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -301,10 +301,6 @@ build-and-deploy-ui:
 prepare-for-e2e:
 	$(call workspace_bundle,base) \
 	&& $(call workspace_service_bundle,guacamole) \
-	&& $(call workspace_service_bundle,azureml) \
-	&& $(call workspace_service_bundle,gitea) \
-	&& $(call workspace_service_bundle,innereye) \
-	&& $(call shared_service_bundle,sonatype-nexus) \
 	&& $(call shared_service_bundle,gitea) \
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -301,10 +301,6 @@ build-and-deploy-ui:
 prepare-for-e2e:
 	$(call workspace_bundle,base) \
 	&& $(call workspace_service_bundle,guacamole) \
-	&& $(call workspace_service_bundle,azureml) \
```
guybartal, line 304: You've removed also azureml, gitea and innereye intentionally?

tamirkamara: yes, they are not tested in any e2e so we don't need to build, publish and register them

guybartal (approved):
LGTM

PR #2423, opened 2022-08-09
Add status validation for unaccesible statuses

### LINKED ISSUE #2422: SAS token should not be generated for failed, blocked and rejected airlock requests.
**Describe the bug**
If the airlock requests is failed, blocked or rejected we shouldn't generate a SAS token for the user to access it. Instead prompt the user with a proper message.

**Steps to reproduce**

1. Create airlock request
1. Make to fail by not uploading any files
1. Submit the request
1. Get SAS token for the request

LizaShak:
# Fixes #2422

## What is being addressed

SAS token should not be generated for failed/blocked/rejected airlock requests

## How is this addressed

- Add status validation


anatbal (approved):
LGTM

PR #2428, opened 2022-08-09
Added placeholder for no templates

### LINKED ISSUE #2080: Show placeholder if no templates registered for resource type
Currently, if you have no registered templates for a resource type and go into the create resource panel for that type, it will just show a blank area. 

We should have a placeholder explaining that you need to publish and register templates for them to appear in the create menu, and can also direct users to the appropriate documentation pages for this process.

jjgriff93:
# Resolves #2080 

## What is being addressed

Adds a placeholder to the create new panel if there are no templates registered for that resource type

## How is this addressed

- Catch for 0 templates being returned by the API and render a helper message

image


marrobi (approved):
LGTM

PR #2434, opened 2022-08-10
Remove support for app insights profiler

### LINKED ISSUE #2421: make all - Diagnostic Services Trusted Storage Access object Id
fresh install
run make all

it will fail at  
module.azure_monitor.azurerm_role_assignment.appinsights_storage_permission: Still creating... 
╷
│ Error: authorization.RoleAssignmentsClient#Create: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="PrincipalNotFound" 
Message="Principal 6243488d10d84ea0884ec2d5d1b7462d does not exist in the directory 2b0d1330-312c-497a-a034-f2374ee0be2a. 
Check that you have the correct principal ID. If you are creating this principal and then immediately assigning a role, this error might be related to a replication delay.
 In this case, set the role assignment principalType property to a value, such as ServicePrincipal, User, or Group.  See https://aka.ms/docs-principaltype"
│ 
│   with module.azure_monitor.azurerm_role_assignment.appinsights_storage_permission,
│   on azure-monitor/azure-monitor.tf line 109, in resource "azurerm_role_assignment" "appinsights_storage_permission":
│  109: resource "azurerm_role_assignment" "appinsights_storage_permission" {
│ 
╵
Script done, file is 1659949004-tre-core.log
Terraform Error

this is because principal_id used for Diagnostic Services Trusted Storage Access is equal to object Id that belongs to Microsoft Tenant.
it should be changed to your object id tenant the installation sits in.

There should be a script to get the object ID from the tenant from the application id which is unique along tenants.

tamirkamara:
# Resolves #2421 

## What is being addressed

App insights Profiler support is causing us issues and not likely to be used away. So we're removing it.


marrobi (approved):
LGTM

Issue #2353, opened 2022-07-26
Build fails with "Failed to query available provider packages"


tanya-borisova:
**Describe the bug**

Main build sometimes fails with an intermittent issue: 

```
INFO     resources.resource:resource.py:74 True, deleting_failed, bf99439c-a834-438b-93aa-0b7013d58d39: Error context message = ╷ │ Error: Failed to query available provider packages │  │ Could not retrieve the list of available versions for provider │ hashicorp/random: could not query provider registry for │ registry.terraform.io/hashicorp/random: the request failed after 2 │ attempts, please try again later: Get │ "https://registry.terraform.io/v1/providers/hashicorp/random/versions": EOF ╵  Error: error running command /cnab/app ./delete_vm_extensions.sh rg-***-mgmt ***mgmt tfstate bf99439c-a834-438b-93aa-0b7013d58d39: exit status 1 Error: mixin execution failed: exit status 1 Error: 1 error occurred: 	* container exit code: 1, message: <nil>   az login --identity -u 261b930f-39a8-4266-bf69-abdb98fddebe && az acr login --name ***mgmt && porter uninstall "bf99439c-a834-438b-93aa-0b7013d58d39"  --reference ***mgmt.azurecr.io/tre-service-guacamole-windowsvm:v0.4.3  --param arm_use_msi="true" --param id="bf99439c-a834-438b-93aa-0b7013d58d39" --param os_image="Windows 10" --param parent_service_id="ecb53982-2437-4ca2-8616-d2de913119b1" --param tfstate_container_name="tfstate" --param tfstate_resource_group_name="rg-***-mgmt" --param tfstate_storage_account_name="***mgmt" --param tre_id="***" --param workspace_id="d5534ecf-7e65-45a4-9c18-4f1f49594b90" --allow-docker-host-access --force --cred ./vmss_porter/arm_auth_local_debugging.json --cred ./vmss_porter/aad_auth.json && porter show bf99439c-a834-438b-93aa-0b7013d58d39
```

![image](https://user-images.githubusercontent.com/52954123/180979547-b69ec365-fbf8-4851-a4be-52de7ca6d2cd.png)

Example runs:

https://github.com/microsoft/AzureTRE/runs/7470499092?check_suite_focus=true
https://github.com/microsoft/AzureTRE/runs/7508258200?check_suite_focus=true
https://github.com/microsoft/AzureTRE/runs/7513578222?check_suite_focus=true

tanya-borisova:
[...]

PR #2446, opened 2022-08-10
Primary bundles mirror Terraform providers

tamirkamara:
## What is being addressed

When we execute a bundle (like via install action) it needs outside communication to terraform domains to initialize itself which isn't ideal (see the issue for more info). 
This is part of #2353 which will be closed in a future PR.

## How is this addressed

- Call a terraform command to mirror providers locally at build time which will prevent any need for external communication at runtime
- Prepare the firewall for dropping the terraform domains more easily in the future
- Upgrade Terraform client to 1.2.6 and AzureRM to 3.16 in all the updated bundles
- In the base workspace, run vnet operations sequentially to prevent _AnotherOperation_ errors
- Minor (unrelated) changes in E2E


LizaShak (approved):

tamirkamara:
/test-force-approve
Tests are currently broken due to unrelated change, but passed on a branch before this change was introduced.

PR #2453, opened 2022-08-11
Jjgriff93/2431 UI remove workspace detail for researcher role

### LINKED ISSUE #2404: UI not able to retrieve Workspace scope and token for non-admin users
**Describe the bug**

When logged in as a user that has TREUser and WorkspaceResearcher roles, I can't retrieve the Workspace, even though as a WorkspaceResearcher I should be able to. 

image

If I then grab a token which was sent to the API, I only see TREUser as a role in it, and not the WorkspaceResearcher, which suggests a wrong token was sent.

image

jjgriff93:
# Resolves #2404 

## What is being addressed

The UI was showing an error when a TREUser / WorkspaceResearcher was trying to view a workspace they were in.

## How is this addressed

- Modified the Api to allow a TREUser to fetch an individual workspace to bring it in line with the list all logic - they are only allowed to retrieve if they are a member of the workspace. This then allows the UI to retrieve it, get the application URI, and proceed to get a workspace-scoped token for the user to view and interact with the workspace as designed


marrobi (approved):
As discussed offline lets get this in, however not sure calling graph on every request is ideal.

If we have a token for the workspace already why do we need to call the endpoint again? Can we not call the workspace specific one that doesn't need the graph call?

jjgriff93:
> As discussed offline lets get this in, however not sure calling graph on every request is ideal.
> 
> If we have a token for the workspace already why do we need to call the endpoint again? Can we not call the workspace specific one that doesn't need the graph call?

I think the main drawback here about using a long-lived token is if someone has their roles removed or modified, it might take a long time before they have their access revoked/modified if we don't get the token each time. Might be one for us to discuss with David when he gets back as he implemented this

marrobi:
> > As discussed offline lets get this in, however not sure calling graph on every request is ideal.
> > If we have a token for the workspace already why do we need to call the endpoint again? Can we not call the workspace specific one that doesn't need the graph call?
> 
> I think the main drawback here about using a long-lived token is if someone has their roles removed or modified, it might take a long time before they have their access revoked/modified if we don't get the token each time. Might be one for us to discuss with David when he gets back as he implemented this

I don't think we need to cache the token any longer, just say if have a token for that workspace already let's not go get the scope again.

jjgriff93:
> > > As discussed offline lets get this in, however not sure calling graph on every request is ideal.
> > > If we have a token for the workspace already why do we need to call the endpoint again? Can we not call the workspace specific one that doesn't need the graph call?
> > 
> > 
> > I think the main drawback here about using a long-lived token is if someone has their roles removed or modified, it might take a long time before they have their access revoked/modified if we don't get the token each time. Might be one for us to discuss with David when he gets back as he implemented this
> 
> I don't think we need to cache the token any longer, just say if have a token for that workspace already let's not go get the scope again.

We're getting the workspace anyway, not just for the scope but to also show the data on the page, so there aren't any duplicate calls from that perspective

PR #2460, opened 2022-08-14
Use delegated key when generating SAS token in API

### LINKED ISSUE #2390: API should generate Airlock SAS with User Delegated Key
## Description

As a TRE Developer
I want to generate airlock sas tokens for users with User Delegated Key
So that I won't use account keys and operation will be more secure

** See how the Airlock processor does this
** This will require the API identity to have blob read and write permission on the relevant storage accounts since Auth is intersection of the sas embedded permission and those on the user delegated key that created it. On first glance it might not look as a security improvement but it is - not using the static account key is the improvement as having it gave the holder (api) full access to the entire storage account.

## Acceptance criteria

- [ ] API doesn't use account keys
- [ ] API doesn't have Read Data Access permission on core or the workspace
- [ ] SAS is generated and working

LizaShak:
# Fixes #2390 

## What is being addressed

Use user delegated key when generating SAS token for airlock request in the API

## How is this addressed

- Remove storage management client and listing account keys
- use blob client to generate user delegated key and get SAS token with it
- Add "Storage Blob Data Contributor" permission to API's managed identity which allows it to generate delegated keys
- Add private endpoints to Import External And Export Approved storage accounts


api_app/api/routes/airlock.py
```diff
@@ -98,6 +97,8 @@ async def get_airlock_container_link(workspace=Depends(get_deployed_workspace_by
                                      user=Depends(get_current_workspace_owner_or_researcher_user)) -> AirlockRequestTokenInResponse:
     validate_user_allowed_to_access_storage_account(user, airlock_request)
     validate_request_status(airlock_request)
+    logging.info("got it")
+    logging.info("got it debug")
```
tamirkamara, line 101: Still need to be here?

api_app/services/airlock.py
```diff
@@ -15,10 +15,12 @@
...
-    return StorageManagementClient(credential=token_credential, subscription_id=config.SUBSCRIPTION_ID)
+def get_credential() -> DefaultAzureCredential:
+    managed_identity = config.MANAGED_IDENTITY_CLIENT_ID
+    if managed_identity:
+        logging.info("Using managed identity credentials.")
+    return DefaultAzureCredential(managed_identity_client_id=config.MANAGED_IDENTITY_CLIENT_ID,
+                                  exclude_shared_token_cache_credential=True) if managed_identity else DefaultAzureCredential()
 
 
 def get_account_and_rg_by_request(airlock_request: AirlockRequest, workspace: Workspace) -> RequestAccountDetails:
```
tamirkamara, line 26: Do we still need the RG to be returned?

api_app/services/airlock.py
```diff
@@ -88,31 +90,30 @@ def validate_request_status(airlock_request: AirlockRequest):
...
 
 
-def get_airlock_request_container_sas_token(storage_client: StorageManagementClient,
-                                            request_account_details: RequestAccountDetails,
-                                            airlock_request: AirlockRequest):
-    account_key = get_storage_account_key(storage_client, request_account_details)
-    required_permission = get_required_permission(airlock_request)
+def get_airlock_request_container_sas_token(request_account_details: RequestAccountDetails, airlock_request: AirlockRequest):
+
+    source_blob_service_client = BlobServiceClient(account_url=get_account_url(request_account_details.account_name),
```
tamirkamara, line 102: I think you can remove the prefix source as there's no destination here...

templates/core/terraform/airlock/identity.tf
```diff
@@ -54,6 +54,6 @@ resource "azurerm_role_assignment" "airlock_blob_data_contributor" {
 resource "azurerm_role_assignment" "api_reader_data_access" {
```
tamirkamara, line 54: can you also change the resource name?

templates/workspaces/base/terraform/airlock/storage_accounts.tf
```diff
@@ -272,6 +272,6 @@ resource "azurerm_role_assignment" "airlock_blob_data_contributor" {
 resource "azurerm_role_assignment" "api_reader_data_access" {
```
tamirkamara, line 272: Can you also change the resource name?

api_app/api/routes/airlock.py
LizaShak: fixed

api_app/services/airlock.py
LizaShak: removed

tamirkamara (approved):
LGTM

PR #2465, opened 2022-08-15
Have a single env file for local debugging

ross-p-smith:
# Relates to #1470

## What is being addressed

Use a single .env file in the root of e2etests and api when you are debugging locally

## How is this addressed

- Single task that merges the various .env file together and copies it into the correct place


ross-p-smith:
/test-force-approve

This change will impact local F5 debugging in Visual Studio Code and is not used in the CI process

tanya-borisova (approved):

PR #2472, opened 2022-08-17
Replace firewall rule with private endpoint in Airlock

### LINKED ISSUE #2098: Airlock - fix export in progress to export approved networking
**Describe the bug**
When the AL processor tries to copy data from the Export inprogress to the Export approved storage it fails on 403 unauthorized.
To overcome this, the fw of the export in progress storage is allowing communication from the airlock processor subnet. 

If possible, remove this fw allow rule and replace with a PE connection in the export approved storage (so they talk over private link).
This user story is about trying to make the communication flow using private link rather public fw.

yuvalyaron:
Resolves #2098 

## What is being addressed

Replaced the firewall rule that allowed airlock processor to access the `export in progress` storage account with a private endpoint for the `export approved` storage account in the `services` subnet of the workspace.

## How is this addressed

- Removed the firewall rule from the `export in progress` storage account
- Added a private endpoint in services subnet (in the workspace's VNet) for the `export approved` storage account

### DIFF
--- a/templates/workspaces/base/terraform/airlock/data.tf
+++ b/templates/workspaces/base/terraform/airlock/data.tf
@@ -23,3 +23,8 @@ data "azurerm_servicebus_topic" "blob_created" {
   resource_group_name = local.core_resource_group_name
   namespace_name      = data.azurerm_servicebus_namespace.airlock_sb.name
 }
+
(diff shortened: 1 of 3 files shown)


yuvalyaron:
was closed as we don't want to enable access to the `export approved` storage account from the workspace.

PR #2475, opened 2022-08-17
Collection of Pipeline Bugs

### LINKED ISSUE #2450: Deleted VM has status Deleting - messages arrive in incorrect order
When deleting a Guacamole VM, the API started to return 500 errors for user resources.

The UI gives the error unable to retrieve workspace details.

Digging in to the API logs the VM status cannot be retrieved as the VM has been deleted. In cosmos status is deleting.

Looking at the logs it appears the service bus messages arrived in the wrong order:

image

To fix, I updated the status of the VM in cosmos to Deleted. However we need to prevent this from happening.

### LINKED ISSUE #2467: Resource status should reflect pipeline status
Currently if a resource is involved in a pipeline as the main step, the status might succeed but the pipeline fail on a subsequent step. This causes an inconsistent way to check whether that resource is 'deployed' or not. 

We should update the primary resource status when the pipeline completes, whether successful or not.

### LINKED ISSUE #2473: Failure to substitute data in a pipeline step not reflected in operation
If a pipeline step fails due to bad data / a schema invalidation, the pipeline effectively 'hangs' and the step doesn't get reported as failed. 

If the data is invalid, we need to catch that and report it back to the step so we can show a failure.

### LINKED ISSUE #2474: Resources with Array types have regressed to strings
Related to #2473 .

A property such as `address_prefixes` should save in Cosmos as a native array type. Instead it's being saved as a string, which is then causing issues in substitutions in pipelines.

damoodamoo:
Resolves #2450 
Resolves #2474 
Resolves #2467 
Resolves #2473


damoodamoo:
/test-force-approve 
known issue on PE

marrobi (approved):
Tested on my deployment, works well. Trust the Python is good!

PR #2478, opened 2022-08-18
Remove AppInsights Profiler support in base workspace bundle

tamirkamara:
## What is being addressed

We have some support for App Insights Profiler but that will likely not work or even needed.

## How is this addressed

- Deploy App Insights in the base workspace to use native resources (rather than ARM). 
- Remove support of the Profiler 
- Support internet ingestion when debugging locally (core & workspace)
- Add `terraform validate` to `bundle-build` to make errors visible locally quicker


stuartleeks (approved):
LGTM!

PR #2479, opened 2022-08-18
Add pipeline for firewall to Azure ML

### LINKED ISSUE #2208: AzureML configures firewall with deployment pipeline
## Description

As a TRE Developer
I want to configure required firewall rules with the deployment pipeline
So that firewall rules will be managed centrally in the shared service

## Acceptance criteria

- [ ] get_firewall_priorities.sh is removed
- [ ] Required rules are present in firewall on azure and in state of the firewall shared service
- [ ] Both install and uninstall work

marrobi:
# Resolves #2208

## What is being addressed
- move firewall rules to pipeline
- fix substitutions bug
- remove firewall from terraform


templates/workspace_services/azureml/template_schema.json
```diff
@@ -1,11 +1,146 @@
...
+      }
     ],
-    "properties": {
-    }
+    "upgrade": [
+      {
+        "stepId": "main"
+      }
+    ],
+    "uninstall": [
```
damoodamoo, line 140: as we discussed - let's put in the uninstall logic to remove the FW rule on resource uninstall (and possibly upgrade if anything could change)

damoodamoo (changes requested):
As we discussed - let's add the template json for uninstall to remove the FW rule when the resource is uninstalled

templates/workspace_services/azureml/template_schema.json
```diff
@@ -1,11 +1,386 @@
...
+      {
+        "stepId": "260421b3-7308-491f-b531-e007cdc0ff48",
+        "stepTitle": "Add network firewall rules for azureml",
+        "resourceTemplateName": "tre-shared-service-firewall",
+        "resourceType": "shared-service",
+        "resourceAction": "upgrade",
+        "properties": [
```
damoodamoo, line 267: can ditch the properties here as we're removing the item

templates/workspace_services/azureml/template_schema.json
```diff
@@ -1,11 +1,386 @@
...
+          {
+            "name": "network_rule_collections",
+            "type": "array",
+            "arraySubstitutionAction": "remove",
+            "arrayMatchField": "name",
+            "value": {
+              "name": "nrc_svc_{{ resource.id }}_azureml",
+              "action": "Allow",
```
damoodamoo, line 275: can ditch everything other than `name` in here as we're removing the item from the array

damoodamoo (approved):

templates/workspace_services/azureml/template_schema.json
marrobi: Done, that look ok?

PR #2486, opened 2022-08-21
Update Resource Processor logging and disk size

tamirkamara:
## What is being addressed

1. Resource processor log warnings when the bundle's image isn't present locally.
2. Its disk is too small and images are cached for too long

## How is this addressed

- If porter output only contains the message about the image missing locally, then set the log to info (rather than warning)
- Double the disk size and docker prune when usage is above 70%


anatbal (approved):
LGTM

PR #2487, opened 2022-08-21
Centralize credentials code in API

tamirkamara:
## What is being addressed

We had multiple places where we got credentials for the API actions on Azure which made it harder to make changes.

## How is this addressed

- Write the logic once and reuse across the application
- If managed identity is present, disable all other forms of getting a token in _DefaultAzureCredentials_ to remove some noise from the logs.
- Remove azure-mgmt-storage and update version of a couple of other packages
- Fix some pytest warnings


api_app/core/credentials.py
```diff
@@ -0,0 +1,41 @@
...
+from azure.identity.aio import (
+    DefaultAzureCredential as DefaultAzureCredentialASync,
+    ManagedIdentityCredential as ManagedIdentityCredentialASync,
+    ChainedTokenCredential as ChainedTokenCredentialASync,
+)
+
+
+def get_credential() -> TokenCredential:
+    managed_identity = config.MANAGED_IDENTITY_CLIENT_ID
+    if managed_identity:
```
yuvalyaron, line 19: maybe it's worth logging whether we use the managed identity or not

yuvalyaron (changes requested):

api_app/core/credentials.py
tamirkamara: Is it?
On Azure it will always be managed, so this is about a local run.

yuvalyaron (approved):

PR #2492, opened 2022-08-22
Paging in ms graph queries

### LINKED ISSUE #2490: MS Graph queries don't follow pages
**Describe the bug**
MS graph implement paging but our queries don't iterate over them. If an identity has many permissions (above ~50) we don't fetch all of them which can cause false 403/permission denied errors. 

This has recently started to happen in the E2E tests, specifically in the AAD enabled ones. The non-AAD ones work correctly because of how the permission results are ordered.

tamirkamara:
# Resolves #2490 

## What is being addressed

MS Graph queries implement paging and we need to do the same. 

## How is this addressed

- Continue to query as long as the result include a _next_ link


api_app/services/aad_authentication.py
```diff
@@ -282,27 +282,42 @@ def _get_app_auth_info(self, client_id: str) -> dict:
...
+        while True:
+            if not url:
+                break
+            logging.debug(f"Making request to: {url}")
+            if json:
+                response = requests.request(method=http_method, url=url, json=json, headers=auth_headers)
+            else:
+                response = requests.request(method=http_method, url=url, headers=auth_headers)
+            url = ""
+            if response.status_code == 200:
```
stuartleeks, line 298: What happens in the error case here?
Prior to this change, I think the error response was returned to the caller

tamirkamara: only the payload was returned and the callers are looking for the "value" property which will be missing anyway.
Should I just throw an exception here?

stuartleeks: In the non-success case previously, I think the payload contained the error details (and callers logged the payload if `value` wasn't found).
With the changes, it looks like the non-success case now returns an empty payload?

tamirkamara: I added a log. The caller is throwing the exception.

api_app/services/aad_authentication.py
```diff
@@ -282,27 +282,44 @@ def _get_app_auth_info(self, client_id: str) -> dict:
...
+                json_response = response.json()
+                graph_data = merge_dict(graph_data, json_response)
+                if '@odata.nextLink' in json_response:
+                    url = json_response['@odata.nextLink']
+            else:
+                logging.error(f"MS Graph query to: {url} failed with status code {response.status_code}")
```
stuartleeks, line 304: Is it worth including the response body? (I think it did this previously to capture the error message)

tamirkamara: done.

tamirkamara:
/test-force-approve
Many runs that are collective fine but none finished ok. https://github.com/microsoft/AzureTRE/actions/runs/2906358327

stuartleeks (approved):
LGTM

PR #2494, opened 2022-08-22
Fix issue with newline in quoted param

stuartleeks:
When there are multiple locks, the ids need to be passed as separate parameters, but the quoted string treated as a single, newline separated value.

This currently works in the GitHub cleanup workflow as we only have a single lock because stateful_resources_locked is false (thanks to @tamirkamara for figuring this out!)


tamirkamara (approved):

stuartleeks:
/test-force-approve

changes not covered in tests

PR #2500, opened 2022-08-23
Update lock files

tamirkamara:
## What is being addressed

Some components have terraform lock files which are out of date with the provider block. This PR updates all files across the solution.
Plus, it change the way the update script does it.


tamirkamara:
/test-force-approve
Not really tested in CI

anatbal (approved):

PR #2504, opened 2022-08-24
Enumerating Airlock request files after request submission

### LINKED ISSUE #2267: Airlock - Enumerating all files when submitting a new request
## Description

As a TRE Developer
I want to enumerate all (currently single) files and save the list inside the request object
So that it can be used by the UI and other validations

## Acceptance criteria

- [ ] The request object inside Cosmos contains the files list (currently it is empty)

yuvalyaron:
# Resolves #2267

## What is being addressed

Populating the `files` field in Airlock requests (using the Airlock processor since it cannot be done in the API)

## How is this addressed

- Added file enumeration to `StatusChangedQueueTrigger` , the function will return a step result with the enumerated files. The enumeration will only happen when the new status is "submitted".
- Refactored the API to handle a "step result" that does not contain a new status - as in the case of file enumeration, only the request files are being sent.
- Refactored the method for `updating` Airlock requests to check the ETag of the request, and retry the update if it's not up to date.
- Added tests.


api_app/db/repositories/airlock_requests.py
```diff
@@ -126,21 +127,17 @@ def get_airlock_request_by_id(self, airlock_request_id: UUID4) -> AirlockRequest
...
-        if self.validate_status_update(current_status, new_status):
-            updated_request = copy.deepcopy(airlock_request)
-            updated_request.status = new_status
-            if new_status == AirlockRequestStatus.Failed:
-                updated_request.errorMessage = error_message
-            if airlock_review is not None:
-                if updated_request.reviews is None:
-                    updated_request.reviews = [airlock_review]
-                else:
-                    updated_request.reviews.append(airlock_review)
```
yuvalyaron, line 140: extracted this part to `_build_updated_request`

airlock_processor/tests/test_status_change_queue_trigger.py
```diff
@@ -62,6 +66,49 @@ def test_wrong_type_raises_when_getting_storage_account_properties(self):
         self.assertRaises(Exception, get_source_dest_for_copy, "accepted", "somethingelse")
 
 
+class TestFileEnumeration(unittest.TestCase):
```
guybartal, line 69: Consider adding another test for asserting the actual returned files array when mocking blob storage client

api_app/db/repositories/airlock_requests.py
```diff
@@ -126,21 +127,17 @@ def get_airlock_request_by_id(self, airlock_request_id: UUID4) -> AirlockRequest
...
-            return self.update_airlock_request_item(airlock_request, updated_request, user, {"previousStatus": current_status})
-        else:
-            raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=strings.AIRLOCK_REQUEST_ILLEGAL_STATUS_CHANGE)
+    def update_airlock_request(self, original_request: AirlockRequest, user: User, new_status: AirlockRequestStatus = None, request_files: List[AirlockFile] = None, error_message: str = None, airlock_review: AirlockReview = None) -> AirlockRequest:
+        updated_request = self._build_updated_request(original_request=original_request, new_status=new_status, request_files=request_files, error_message=error_message, airlock_review=airlock_review)
+        try:
+            db_response = self.update_airlock_request_item(original_request, updated_request, user, {"previousStatus": original_request.status})
+        except CosmosAccessConditionFailedError:
```
guybartal, line 134: should we use a small sleep here to reduce potential error?

guybartal (changes requested):

airlock_processor/tests/test_status_change_queue_trigger.py
yuvalyaron: the method is only extracting properties from the blobs that 
`container_client.list_blobs()` returns. If I mock this method, then I'm only testing my mock - no?

api_app/db/repositories/airlock_requests.py
yuvalyaron: how can it reduce it? if cosmos returns this error, then it was already updated, and when we ask for the updated request from cosmos we query the same cosmos instance

airlock_processor/tests/test_status_change_queue_trigger.py
guybartal: There's no logic to test, so you are correct.

api_app/db/repositories/airlock_requests.py
guybartal: I understand now, then sleep is not relevant here.

guybartal (approved):
LGTM

anatbal (approved):

PR #2505, opened 2022-08-25
Add Airlock Manager workspace

### LINKED ISSUE #2498: Template for review workspace
## Description

As a TRE Administrator
I want to deploy a Review Workspace 
So that there is a way to create VMs and use them to review Airlock requests

## Acceptance criteria

- [ ] Airlock in-progress storage account can be accessed from within an instance of the workspace template

tanya-borisova:
# Resolves #2498 

## What is being addressed

Add a workspace for Airlock Manager that has access to in-review storage account.
Based on work @guybartal did previously in a private repo

## How is this addressed

Base workspace terraform files are downloaded onto the Docker container as a .zip archive. They are then unpacked, and a patch file is applied with the only difference required for the airlock manager workspace.


marrobi:
@tanya-borisova rather than duplicating so many files, can the base template assets be pulled/download from a release as part of the bundle dockerfile? Then build upon it? 

I intend to do that with the task for creating a workspace with unrestricted Internet access.

tanya-borisova:
@marrobi  I can give it a try, though wouldn't it introduce a dependency in the order in which `base` and `airlock_manager` workspaces will need to be built? 
Assuming the `airlock_manager` would depend on the latest version of the `base` image

marrobi:
@tanya-borisova rather than depend on the image, download the terraform, and include in a seperate initial terraform step in the porter.yaml, or add additional TF files and have a single terraform step in the Porter.yaml .

tanya-borisova:
@marrobi  I'm not sure I understand your suggestion. By "download the terraform", you mean terraform files? Where do you suggest to download them from?

marrobi:
Could download a published release, extract the files, or I was thinking of using sparse checkout so just get the files needed.

martinpeck:
Changes we want to see in this PR include:

- remove the shared storage
- ideally, we want to prevent VM to VM networking

We may need to deal with the problem that VMs attempt to mount shared storage.

tanya-borisova:
@martinpeck Could you share a bit more about the VM to VM networking we want to prevent. I don't think it is really feasible to connect between VMs, as I don't think there's an easy way to get Bastion to work, and e.g. RDP or SSH have their ports closed. Is there a use case that we want to stop from working? 

I have addressed the rest of suggestions.

templates/workspace_services/guacamole/user_resources/guacamole-azure-windowsvm/terraform/windowsvm.tf
```diff
@@ -95,7 +95,7 @@ data "template_file" "vm_config" {
     SharedStorageAccess = tobool(var.shared_storage_access) ? 1 : 0
     StorageAccountName  = data.azurerm_storage_account.stg.name
     StorageAccountKey   = data.azurerm_storage_account.stg.primary_access_key
-    FileShareName       = data.azurerm_storage_share.shared_storage.name
+    FileShareName       = var.shared_storage_access ? data.azurerm_storage_share.shared_storage[0].name : ""
```
tanya-borisova, line 98: With these changes, the user resource can be deployed if the shared_storage_access variable is set to False and there is no shared storage in the workspace.

tamirkamara: Should this option also be visible in the UI?

tanya-borisova: I made the change but not actually sure we will need it long term, since we would have a separate user resource template for the review VMs anyway where we can just set it to false by default.

Makefile
```diff
@@ -290,11 +290,7 @@ build-and-deploy-ui:
 	&& if [ "$${DEPLOY_UI}" != "false" ]; then ${MAKEFILE_DIR}/devops/scripts/build_deploy_ui.sh; else echo "UI Deploy skipped as DEPLOY_UI is false"; fi \
 
 prepare-for-e2e:
-	$(call workspace_bundle,base) \
-	&& $(call workspace_service_bundle,guacamole) \
-	&& $(call shared_service_bundle,gitea) \
-	&& $(call user_resource_bundle,guacamole,guacamole-azure-windowsvm) \
-	&& $(call user_resource_bundle,guacamole,guacamole-azure-linuxvm)
+	$(call workspace_bundle,airlock_manager)
```
tamirkamara, line 293: I'm guessing this change shouldn't have been committed?

templates/workspaces/airlock_manager/Dockerfile.tmpl
```diff
@@ -0,0 +1,41 @@
+# We need azurecli 2.37+ which doesn't exist for strech so the minimum is buster
+FROM debian:buster-slim
+
+ARG BUNDLE_DIR
+
+ARG AZURE_TRE_VERSION="0.4.2"
```
tamirkamara, line 6: Can be moved to the latest version?

templates/workspaces/airlock_manager/Dockerfile.tmpl
```diff
@@ -0,0 +1,41 @@
...
+# Use the BUNDLE_DIR build argument to copy files into the bundle
+COPY . $BUNDLE_DIR
+
+RUN patch -p0 < $BUNDLE_DIR/workspace_base.diff
+
+# Mirror plugins to prevent network access at runtime
+# Remove when available from https://github.com/getporter/terraform-mixin/issues/90
+WORKDIR $BUNDLE_DIR/terraform
+RUN terraform init -backend=false \
+  && terraform providers mirror /usr/local/share/terraform/plugins
```
tamirkamara, line 41: Can you sync this with the other templates?
(a line is missing)

templates/workspaces/airlock_manager/update_redirect_urls.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
```
tamirkamara, line 1: What's different about this file compared with the one from the base workspace?
Can't we directly use it here like we do with its terraform folder?

tamirkamara (approved):
Approving but there're a couple of things I think need addressing.

Makefile
tanya-borisova: Ah, I thought I'd reverted that but my push failed... thanks!

marrobi:
@tanya-borisova can we have a docs page for this to go along site the other two workspace templates please.

PR #2526, opened 2022-08-30
Allow longer timeouts for e2e

tamirkamara:
## What is being addressed

Sometimes E2E would fail due to slower run times on azure.

## How is this addressed

- Higher timeouts
- Additional log to make outputs clearer


guybartal (approved):
LGTM

Issue #2323, opened 2022-07-22
Auto install shared service names need to be formatted correctly.


marrobi:
The auto installed shared services have long names that make no sense to users:
![image](https://user-images.githubusercontent.com/17089773/180478353-1bec48be-3eea-47e3-bd87-95b2544dccde.png)

marrobi:
In addition, the certs shared service probably does not want to be listed. Do we need a "show ito users" (as opposed to admins) boolean property?

@jjgriff93 @damoodamoo thoughts?

marrobi:
this would be better:
<img width="185" alt="image" src="https://user-images.githubusercontent.com/17089773/180976039-3388606f-99ad-4cf4-8698-fa2538be87f2.png">

damoodamoo:
yea we can change the display_name in the auto-install, happy to take it

PR #2527, opened 2022-08-30
remove shared service prefix for auto deploy

tamirkamara:
## What is being addressed

The name of shared services installed by the makefile/script can be more friendly. 
Part of #2323.

## How is this addressed

- Shared services installed by scripts don't include "tre-shared-service" prefix so the name would be just `firewall` (for example)


guybartal (approved):
LGTM

Issue #2104, opened 2022-06-23
Workspace status temporarily incorrect


stuartleeks:
Remember to label the issue with the appropriate area, and also assign it to yourself if you plan to work on the issue in the near future.

**Describe the bug**
When a workspace is being deleted, navigating to the workspace view has a delay before showing the "deleting" status. During that delay, the buttons for editing the workspace are enabled.

**Steps to reproduce**

1. Create a workspace
1. Disable and then start deletion of the workspace
1. Navigate to the workpsace view. The status will show as "Action succeeded" for a while and then update to show "deleting"

![tre-workspace-status](https://user-images.githubusercontent.com/1824461/175336519-8988bb89-7250-413e-83fa-f85a91771ab1.gif)

stuartleeks:
To me, this is a bug in the "resource is locked so I won't allow you to make further changes" logic that the UI implements. If we don't want to solve this issue then I'm not clear on the value of the UI-based resource lock behaviour :thinking:

damoodamoo:
Since the Ops now load in one call, and tend to load before the resources, I think this is mitigated. We can reopen if needed.

PR #2530, opened 2022-08-30
My Operations + UI Stability improvements

### LINKED ISSUE #2241: `/my-operations` endpoint
Rather than cycling through every resource to find any active operations, we need a new endpoint to return any operations that were initiated by the current user, that are in-progress. This will be more efficient, and a single call to the API + cosmos. 

- [ ] `my-operations` endpoint
- [ ] Update logic in UI NotificationsPanel to call this endpoint rather than recursing down the tree

damoodamoo:
# Resolves #2241 

## What is being addressed
- Adds a new endpoint to the API to get 'my' operations - those initiated by the current user. This is a single call which executes quickly and I would say mitigates #2104 as the operations are loading before the resource lists. 
- UI Updates:
  - Moved the api-calling logic for individual operations into the Item level rather than the collection level. This improves readability and prevents having to work with collections.
  - Improved error handling when an operation is invalid (bad data / unrecoverable issue in the backend), so it renders an error with the Op Id to help support staff clean up data.
  - Improved stability when locking/refreshing/removing resources. Moved away from a separate 'component actions' collection which was prone to edge cases of re-rendering to the components just subscribing to the operations list directly, and not mutating that collection. 
  - Improved the callout message when an operation completes, to indicate what just happened.
  - The 'bell' icon now shows a mini progress bar underneath when there are running operations, in the same way the Azure Portal does.

image

image

image


ui/app/src/components/shared/notifications/NotificationItem.tsx
```diff
@@ -1,27 +1,78 @@
...
 
   const getRelativeTime = (createdWhen: number) => {
     return (moment.utc(moment.unix(createdWhen))).from(now);
   }
 
-  // update the 'now' time for comparison - only while the item is rendered (panel is open)
+  useEffect(() => {
+    const setupNotification = async (op: Operation) => {
+      // ignore if we've already set this operation up
+      if (notification.resource) return;
```
martinpeck, line 36: stupid question: where is `notification` initialised?

api_app/api/routes/operations.py
```diff
@@ -0,0 +1,17 @@
...
+from api.dependencies.database import get_repository
+from models.schemas.operation import OperationInList
+from resources import strings
+from services.authentication import get_current_tre_user_or_tre_admin
+
+
+operations_router = APIRouter(dependencies=[Depends(get_current_tre_user_or_tre_admin)])
+
+
+@operations_router.get("/operations/my", response_model=OperationInList, name=strings.API_GET_MY_OPERATIONS)
```
martinpeck, line 14: Is there a reason we need the "my" path?
Could the route `/operations/` not do the job?

ui/app/src/components/shared/notifications/NotificationPanel.tsx
```diff
@@ -1,159 +1,61 @@
...
-        if (w.properties.client_id === "auto_create") continue; // skip getting stuff from the workspace if it's auto_create as it doesn't have auth setup yet
-        let workspaceServicesList = (await apiCall(`${w.resourcePath}/${ApiEndpoint.WorkspaceServices}`, HttpMethod.Get, appId)).workspaceServices as Array<Resource>;
-        if (workspaceServicesList && workspaceServicesList.length > 0) (opsToAdd = opsToAdd.concat(await getOpsFromResourceList(workspaceServicesList, appId)));
-        for(let n=0;n<workspaceServicesList.length;n++){
-          let ws = workspaceServicesList[n];
-          let userResourcesList = (await apiCall(`${ws.resourcePath}/${ApiEndpoint.UserResources}`, HttpMethod.Get, appId)).userResources as Array<Resource>;
-          if (userResourcesList && userResourcesList.length > 0) (opsToAdd = opsToAdd.concat(await getOpsFromResourceList(userResourcesList, appId)));
-        }
-      }
+      let opsToAdd = (await apiCall(`${ApiEndpoint.Operations}/my`, HttpMethod.Get)).operations as Array<Operation>;
```
martinpeck, line 21: Couldn't the `ApiEndpoint.Operations` incorporate the `my`?
Or, could the API not require the `my` at all?

I can't see many/any other instances where we've hard-coded part of the API path in the UI

api_app/_version.py
```diff
@@ -1 +1 @@
-__version__ = "0.4.23"
+__version__ = "0.4.24"
```
tamirkamara, line 1: Please add a changelog entry.
(commented here since it's the first file)

api_app/api/routes/operations.py
damoodamoo: yea possibly - this was a bit of speculative future proofing really, but maybe just `/operations` would be clearer, I can change it

ui/app/src/models/operation.ts
```diff
@@ -12,7 +12,8 @@ export interface Operation {
     createdWhen: number,
     updatedWhen: number,
     user: User,
-    steps?: Array<OperationStep>
+    steps?: Array<OperationStep>,
+    dismiss?: Boolean // virtual prop
```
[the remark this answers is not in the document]
damoodamoo: I can make the comment clearer, but by 'virtual' I mean "it's not in the API model, it's a throwaway property we use in the UI only"

api_app/db/repositories/operations.py
```diff
@@ -163,6 +163,11 @@ def get_operation_by_id(self, operation_id: str) -> Operation:
             raise EntityDoesNotExist
         return parse_obj_as(Operation, operation[0])
 
+    def get_my_operations(self, user_id: str) -> List[Operation]:
+        query = self.operations_query() + f' c.user.id = "{user_id}" AND c.status IN ("{Status.AwaitingAction}", "{Status.InvokingAction}", "{Status.AwaitingDeployment}", "{Status.Deploying}", "{Status.AwaitingDeletion}", "{Status.Deleting}", "{Status.AwaitingUpdate}", "{Status.Updating}", "{Status.PipelineRunning}") ORDER BY c.createdWhen ASC'
```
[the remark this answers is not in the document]
damoodamoo: yea i don't like it either - in the UI I use a collection of "in_progress_statuses" - I could bring that here too, then just join the array in the query, rather than listing every in-progress state by hand?

ui/app/src/components/shared/notifications/NotificationItem.tsx
damoodamoo: line 22, in the `useState` call

api_app/api/routes/operations.py
martinpeck: It's fine as is. This comment is linked to the hard-coding of "my" in the UI, and I think that's the more important thing to fix.

ui/app/src/components/shared/notifications/NotificationItem.tsx
damoodamoo: line 22, in the `useState` setup

ui/app/src/models/operation.ts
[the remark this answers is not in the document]
martinpeck: That's what I figured. Is this the only example of a virtual prop? I wondered because it's the only commented one.

api_app/api/routes/operations.py
damoodamoo: i've taken out the `my` - Stuart didn't like it either :)

martinpeck (approved):
Thanks @damoodamoo.

damoodamoo:
/test-force-approve 
tests have run previously

PR #2532, opened 2022-08-30
Create groups for AAD app reg roles

### LINKED ISSUE #2480: When auto creating app registrations, create a user group for each role and assign to the app role
## Description

As a TRE Administrator
I want a group assigning to each application role
So that workspace owners can use AAD group management to manage who is in the groups.

## Acceptance criteria

- [ ] Option to create groups when deploying a workspace
- [ ] Group created for each app role
- [ ] Workspace owner assigned as group owner
- [ ] Group name includes role, workspace name, and short ID in some form (120 char limit, maybe use description for too)

marrobi:
# Resolves #2480

## What is being addressed
- Adds groups to AAD terraform
- Set workspace owner as owner of the groups.
- Adds boolean to create groups to porter.yaml
- Updates aad schema to provide group boolean.
- Add environment variable to enable the feature
- Update app reg scripts to grant required permissions to app registration
- [x] Update changelog


api_app/schemas/azuread.json
```diff
@@ -1,23 +1,29 @@
...
+      "title": "Application (Client) ID",
+      "description": "The AAD Application Registration ID for the workspace. Use 'auto_create' if you wish TRE to create this."
+    },
+    "client_secret": {
+      "type": "string",
+      "title": "Application (Client) Secret",
+      "description": "The AAD Application Registration secret for the workspace. Leave blank if using `auto_create` above. This value will be stored in the Workspace Key Vault.",
+      "sensitive": true
+    },
+    "create_aad_groups": {
```
tamirkamara, line 22: What does it mean to add it here?
Shouldn't this be in the workspace's template schema?

templates/workspaces/base/terraform/aad/aad.tf
```diff
@@ -124,3 +124,25 @@ resource "azuread_app_role_assignment" "workspace_owner" {
   principal_object_id = var.workspace_owner_object_id
   resource_object_id  = azuread_service_principal.workspace.object_id
 }
+
+resource "azuread_group" "workspace_owners" {
+  count        = var.create_aad_groups ? 1 : 0
+  display_name = "${var.workspace_resource_name_suffix} Workspace Owners"
```
tamirkamara, line 130: Since this supposed to be related to the rest of the AAD config, won't it make sense to have it named with the workspace's app id rather than the its TRE resource id?

tamirkamara (changes requested):
What about adding the groups to their respective roles in the workspace app?

api_app/schemas/azuread.json
marrobi: Good question, originally the plan was to have multiple workspace templates, and they could be used with different ID provider confuguration, such as Azure AD, AzureAD B2C, Auth0 etc. The base workspace template has become coupled to Azure AD... as the TF is part of the base workspace. Originally I envisaged the ID piece to be a seperate bundle, that would be a dependancy... but we know the challenges around dependancies.

I don't want to deal with it in this PR, but we need to decide, 1. Is the base workspace become an "AzureAD base workspace"  - I think it is at present, 2. Do we need to seperate out the App Reg creation terraform into a seperate stage. Until we have a firm request (lots talk about it) about using an ID provider other than Azure AD we could move this all into the base workspace template, and remove this json file and associated code.

Maybe create another issue?

templates/workspaces/base/terraform/aad/aad.tf
marrobi: No, as the users need to be able to identify which workspace the group is associated with when working in Azure AD self service tooling. The workspace ID is more appropriate as this wiill be displayed in the UI.

marrobi:
@tamirkamara became a bit more involved than I expected. Appreciate feedback, and if someone has time to test it. I have and got desired result:

image

marrobi:
Might need to output the group IDs too, then can link to a URI such as:

`https://account.activedirectory.windowsazure.com/r#/manageMembership?objectType=Group&objectId=xxxx`

To manage group members.

api_app/schemas/azuread.json
marrobi: I've removed from here an put into the workspace. The properties get merged anyway.

templates/workspaces/base/template_schema.json
```diff
@@ -65,6 +65,23 @@
...
+      "type": "string",
+      "title": "Application (Client) ID",
+      "description": "The AAD Application Registration ID for the workspace. Use 'auto_create' if you wish TRE to create this."
+    },
+    "client_secret": {
+      "type": "string",
+      "title": "Application (Client) Secret",
+      "description": "The AAD Application Registration secret for the workspace. Leave blank if using `auto_create` above. This value will be stored in the Workspace Key Vault.",
+      "sensitive": true
+    },
```
tamirkamara, line 79: Why do we need it here if it's in the azuread schema file?

tamirkamara (approved):
Looks good with a couple of small comments.

PR #2543, opened 2022-09-04
Disable workspace AppInsight's private endpoint

tamirkamara:
## What is being addressed

Due to many failures when creating the workspace's private endpoint for app insights, we need to disable it for now while our support case / PG conversation is still going.

In addition this PR disables the Terraform checkpoint mechanism as that doesn't work in an airgap environment. The error is comes to solve is Terraform unable to query provider/plugin versions.


templates/workspaces/base/terraform/azure-monitor/azure-monitor.tf
```diff
@@ -79,6 +79,7 @@ resource "azurerm_monitor_private_link_scoped_service" "ampls_app_insights" {
 }
 
 resource "azurerm_private_endpoint" "azure_monitor_private_endpoint" {
+  count               = 0 # Remove with https://github.com/microsoft/AzureTRE/issues/2357
```
yuvalyaron, line 82: how is this related to the issue?
should we mention anything about this block there?

yuvalyaron (approved):

templates/workspaces/base/terraform/azure-monitor/azure-monitor.tf
tamirkamara: there 2 things addressed here: the problem with the PE which is happening for some time, and a newer one about the provider (checkpoint) which I also tried to address here since I couldn't get PRs to pass E2Es.

Issue #2534, opened 2022-09-01
Airlock E2E continues to poll failed requests


tamirkamara:
**Describe the bug**

The E2E will wait for a request to change status even if the current status is failed (which is terminal).

**Steps to reproduce**
This run shows this: https://github.com/microsoft/AzureTRE/runs/8128533749?check_suite_focus=true#step:4:1393

PR #2547, opened 2022-09-05
Skip sending step result events for Airlock updates that do not include a status change

### LINKED ISSUE #2540: Airlock request fails in E2E
**Describe the bug**

There's a situation where an airlock request goes into a failed state.
1. We should investigate why - there're errors in the API as well.
2. The E2E shouldn't poll on failed state as that is a terminal one.

**Steps to reproduce**
https://github.com/microsoft/AzureTRE/runs/8143618838?check_suite_focus=true
https://github.com/microsoft/AzureTRE/runs/8174278643?check_suite_focus=true

yuvalyaron:
# Resolves #2540, #2534

## What is being addressed

As part of #2504, Airlock requests might be updated without updating their status (for instance, enumerating the request files adds the files to the request without changing its status), however the method that updates the requests is also responsible for sending `status changed` events, so when there is an update to an Airlock request, a status changed event is triggered, in cases where it was not necessary, such as in file enumeration this resulted in duplicated step results.

## How is this addressed

-  If there is no status change in `update_and_publish_event_airlock_request`, we skip the send of status changed event
- Added unit tests to enforce the previous bullet
- Changed the E2E test to stop polling the request status if the current status is final and will not change.


api_app/api/routes/airlock_resource_helpers.py
```diff
@@ -53,6 +53,10 @@ async def update_and_publish_event_airlock_request(airlock_request: AirlockReque
             raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=strings.AIRLOCK_REQUEST_ILLEGAL_STATUS_CHANGE)
         raise HTTPException(status_code=status.HTTP_503_SERVICE_UNAVAILABLE, detail=strings.STATE_STORE_ENDPOINT_NOT_RESPONDING)
 
+    if not new_status:
+        logging.debug(f"Skipping sending status changed event for airlock request item: {airlock_request.id} - there is no status change")
```
eladiw, line 57:
```suggestion
        logging.debug(f"Skipping sending 'status changed' event for airlock request item: {airlock_request.id} - there is no status change")
```

eladiw (approved):
lgtm. minor suggestion

PR #2563, opened 2022-09-07
api to return commection_uri for shared service

### LINKED ISSUE #2552: In workspace view shared services do not display a connect button.
Users need to connect to shared services to view the shared service configuration, or external URI. The connect button appears on the Core view, but not the workspace view when a `connection_uri` is present.

lenisha:
# Resolves #2552 

## What is being addressed

Return `connection_uri` in shared services REST API to be rendered in UI

## How is this addressed
- add `connection_uri' to the restricted fields


marrobi (approved):
LGTM. Thanks @lenisha .

PR #2564, opened 2022-09-07
Secure Access to Shared Services in workspace ot TREAdmin and Owner

### LINKED ISSUE #2545: Shared Services are shown in workspace view for TREUser
TREUser should not have access to shared services like Firewall etc . They do not show in home screen, but are shown in left tree when user selects workspace 

**Steps to reproduce**

1. Login as TRE User and TREResearcher in one of workspaces
2. Home page shows only allowed workspaces
3. Navigate to Workspace, and Left tree shows shared services

lenisha:
# Resolves #2545 

## What is being addressed

Shared services are not shown in main root navigation if not TREAdmin, but still are shown in Workspace level
Update will secure Shared service in workspace scope similar to Root layout

## How is this addressed

- Added <SecuredByRole> wrapper to navigation links, added permission to WorkspaceOwner as they need to connect to Shared Services - external-url type

### DIFF
--- a/ui/app/src/components/workspaces/WorkspaceProvider.tsx
+++ b/ui/app/src/components/workspaces/WorkspaceProvider.tsx
@@ -16,6 +16,8 @@ import { SharedService } from '../../models/sharedService';
 import { SharedServices } from '../shared/SharedServices';
 import { SharedServiceItem } from '../shared/SharedServiceItem';
 import { Airlock } from '../shared/airlock/Airlock';
+import { SecuredByRole } from '../shared/SecuredByRole';
+import { RoleName, WorkspaceRoleName } from '../../models/roleNames';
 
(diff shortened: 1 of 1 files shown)


marrobi:
@lenisha - thanks, but, hmm, I think we do need them listed here so users can access the shared services from the workspace and read any overview? 

@jjgriff93 thoughts?

damoodamoo:
hiya :) - this functionality is actually by design. We use the root Shared Services screen as the 'admin' screen, and the one within the workspace as a shortcut to show shared service details to researchers. As you know the API restricts the data coming back for shared services based on the user type.

marrobi:
@lenisha as per @damoodamoo will close this and the issue as not planned, if you have a **really** compelling reason for this, then it might be a configuration flag of some kind could be added.

PR #2573, opened 2022-09-08
Remove manual (makefile) install option for Gitea and Nexus

### LINKED ISSUE #2425: Remove installation of Gitea and Nexus from the make file
Shared services are just a type of resource so should not be dealt with in the make file. These should be removed along with the respective environment variables. Also ensure the docs are updated to direct users to using the API/UI.

tamirkamara:
# Resolves #2425

## What is being addressed

We documented a way to install Gitea and Nexus shared services manually by the makefile. Since this option isn't advised anymore we need to remove it.

## How is this addressed

- Remove relevant options from makefile and core .env file 
- Update documentation to install with the UI/API
- Remove other obsolete options from the core .env file


marrobi (approved):
LGTM

Issue #1695, opened 2022-04-14
Include default shared service, workspace service, user resource - names, descriptions and overviews in template schema


marrobi:
As Workspace Owner I do not want to have to enter a description or overview for the workspace service, but I may want to amend it. The Overview should include standard documentation.

The same for shared services. 

The template schema has a name and description, the schema provided by the API should contain these values as defaults.

marrobi:
Although #2576 is merged, this doesn't seem to be working.

marrobi:
Confirmed as working, these need adding for existing and new services.

marrobi:
See https://github.com/microsoft/AzureTRE/blob/1e2177491691fe23101734113d8568361ca02e63/templates/workspace_services/azureml/template_schema.json#L9-L22 

For an example.

PR #2576, opened 2022-09-08
change merge properties sequence for defaults

### LINKED ISSUE #2575: Default values for properties in a bundle overridden by API app
**Describe the bug**
Admin sets default value for Workspace property to be applied for any workspace creation in a bundle
When trying to create it default value is empty

**Steps to reproduce**

1. Update default value in bundle temple.json and build/register bundle
1. Try to create the workspace default is not set
1. Code merging props overrides it https://github.com/microsoft/AzureTRE/blob/main/api_app/services/schema_service.py#L28-L30

lenisha:
# Resolves #2575

## What is being addressed

Defaults set by Admins in bundle template definition are overridden by API default template

## How is this addressed

- Change order of priority for the template properties


marrobi:
@lenisha thanks very much for the PR. It think this an appropriate change as a step towards #1695 . There are a couple of linting errors - trailing white space, and also can you increate the api version in `api_app/_version.py`. Thanks.

tamirkamara:
Purely from the description - is this something we want given the security risk? We sometimes have settings in schemas stored in the api app that we should never override...

marrobi:
We had a discussion around this. The concern is template schema could override values set for system reasons in the API schema, this could be a security issue, or cause functionality issues.

1. I'm not sure this is a major issue, as the role publishing/registering the bundles can go edit the API json schema files if they so wish.
2. The alternative is we only allow overwrite of `default` values. So "if this exists in both, take the default value only from the tempalte_schema.json"

@tamirkamara @stuartleeks is that a reflection of the discussion? Thoughts?

marrobi:
@tamirkamara @damoodamoo can we look at this again, and clarify the issue why this approach isn't ideal?

damoodamoo:
I'm fine with this - @lenisha correct me if I'm wrong as I understand it this just allows a template author to specify a property for in their template and have that override the static json file stored with the API? So I could define my own `display_name` property, for instance, and use that over the top of the default one? 

If so, I think it's good work, and what we need. @tamirkamara not sure where the security risk would come from?

damoodamoo (approved):

tamirkamara:
> I'm fine with this - @lenisha correct me if I'm wrong but as I understand it this just allows a template author to specify a property for in their template and have that override the static json file stored with the API? So I could define my own `display_name` property, for instance, and use that over the top of the default one?
> 
> If so, I think it's good work, and what we need. @tamirkamara not sure where the security risk would come from?

After another looks and chat, this doesn't change the behavior of user provided values so I don't think there's a risk here.

marrobi:
Recreated PR at https://github.com/microsoft/AzureTRE/pull/2953

marrobi:
@tamirkamara how did you get this merged in the end?



@lenisha thank you, it's taken a while, but now merged :)

tamirkamara:
> @tamirkamara how did you get this merged in the end?

The issue was around security of which action is allowed to report UT results. I hardened it a few weeks ago allowing most check just for 1st part actions, but for the UT to support our forks it needs to be "any action" which is how I defined it today.

PR #2583, opened 2022-09-11
Refactor API health endpoint

### LINKED ISSUE #1673: API health endpoint and solution monitoring
## Description

As a TRE Administrator, I want to know the status of my TRE solution so that I can address potential issues with its components.
As an App Gateway, I want to know if the API is healthy so that I can forward it requests.

The current use of the API health endpoint to do both is causing issues and we need to figure out what requirements are important for us and how to address them.

## Acceptance criteria

TBD

anatbal:
# Fixes #1673
## What is being addressed

- Making the api health fully asynchronous
- Still will return 200 OK in every call since we want to see the behavior first.


eladiw (approved):
lgtm

PR #2604, opened 2022-09-15
Update release doc on AzureTRE-Deployment repo

tamirkamara:
## What is being addressed

We need to make sure that when we create a release in this repo we also update the _AzureTRE-Deployment_ one.


tamirkamara:
/test-force-approve
docs only

docs/tre-developers/release.md
```diff
@@ -18,3 +18,8 @@ The process follows these steps:
     1. Include a final line with a link to the full changelog similar to this:
     <!-- markdownlint-disable-next-line MD034 -->
           **Full Changelog**: https://github.com/microsoft/AzureTRE/compare/v0.9.1...v0.9.2
+1. Update [AzureTRE-Deployment](https://github.com/microsoft/AzureTRE-Deployment). The procedure may vary depending on the level of changes introduced in the new version but should include the following steps:
+    1. Update the tag used in [devcontainer.json](https://github.com/microsoft/AzureTRE-Deployment/blob/main/.devcontainer/devcontainer.json).
+    1. Rebuild the container.
+    1. Compare both `.devcontainer` and `.github` folders of the new release with the ones in the repo and make required updates so that only required difference exist.
```
marrobi, line 24: How do we know what are required updates?

The last bit doesn't really make sense - "so that required difference exist"

Does the deployment repo need testing after being updated?

marrobi (approved):
Good to get it in, bit more clarity what needs to be updated would help.

docs/tre-developers/release.md
tamirkamara: On the difference - I mean that some difference will always exist in some of the files, for example in the workflow we delete the e2e job which is fine. 
But the goal should be the smallest difference as possible and the assumption should be that each difference shouldn't be there (unless really necessary) . So if we make a change / improvement in the parent repo it most probably find itself in the deployment repo as well.

I don't have a good way to describe what differences are required. If you know, you know...

tamirkamara: @marrobi I'll add a point of trying to run the devcontainer and a deployment in that repo. 
Although, this still won't test the workflows. Maybe we can configure them to run there too, but it might be too much?

tamirkamara:
/test-force-approve
docs only

PR #2610, opened 2022-09-18
Include untagged workspace costs

### LINKED ISSUE #2468: TRE Cost Management should roll up untagged resources into their parent resource group
If a service outside of the TRE deploys untagged resources into a workspace the costs are not accounted for. Example being Cyclecloud or Cromwell on Azure.

If resources are untagged they should be allocated to the resource group they are inside. This means if the resource group is tagged with a `workspace_id`, the workspace costs, if tagged with a `workspace_service_id` the workspace service costs, if tagged with a `user_resource`, the user resource costs.

This means querying all tagged resources and resources inside any resource groups tagged with the id. This will mean querying the resource groups to get appropriate resource groups prior to the Azure Cost Management Query.

Then using an OR query similar to:

```
{

    "type": "Usage",

    "timeframe": "Custom",

    "timePeriod": {"from":"{{TIME_FROM}}", "to": "{{TIME_TO}}" },

    "dataset": {

        "granularity": "None",

        "aggregation": {

            "totalCost": {

                "name": "Cost",

                "function": "Sum"

            }

        },

        "filter": {

            "or": [

                {

                    "dimensions": {

                        "name": "ResourceGroup",

                        "operator": "In",

                        "values": ["{{RG_NAMES}}"]

                }

                },

                {

                    "tags": {

                        "name": "{{TAG_KEY}}",

                        "operator": "In",

                        "values" : [

                            "{{TAG_VALUE}}",

                            ]

                    }

                }

            ]

        },

        "grouping": [

            {

                "type": "Tag",

            }

        ]

    }

}
```

guybartal:
# Resolves #2468 

## What is being addressed

Include untagged resources cost under workspace

## How is this addressed

- Get resource group by tag
- Get costs filtered by resource group list or tag (before change it was only by tag)
- Pending work - check if required to rollup costs for workspace services and user resources in sperate resource group
- Documentation updated


api_app/services/cost_service.py
```diff
@@ -20,14 +23,16 @@
 class ResultColumnDaily(Enum):
     Cost = 0
     Date = 1
-    Tag = 2
-    Currency = 3
+    ResourceGroup = 2
```
eladiw, line 26: probably ok. but it is ok to add a number in the middle? backward compatibility wise

api_app/services/cost_service.py
```diff
@@ -20,14 +23,16 @@
...
+    Tag = 3
+    Currency = 4
 
 
 class ResultColumn(Enum):
     Cost = 0
-    Tag = 1
-    Currency = 2
+    ResourceGroup = 1
```
eladiw, line 33: same comment

eladiw (commented):
lgtm, I will let someone more familiar with cost approve.

api_app/services/cost_service.py
guybartal: Thanks Elad, those enums are used merely to maps the column positions in the Azure cost management API results, instead of working with hard coded numeric positions. Adding the RresourceGroup group-by column to the result in the query definition resulted in this uncontrollable order.

api_app/services/cost_service.py
guybartal: Same answer

tanya-borisova (approved):
Looks good, thanks for updating the docs and diagrams as you go

PR #2616, opened 2022-09-20
Validate environment variables in bootstrap.sh

### LINKED ISSUE #2615: Validate environment variables in bootstrap.sh.
As `./devops/terraform/bootstrap.sh` does not use terraform the environment variables are not validated up front.

We have had a number of instances where variables have been missing causing the script to fail.

marrobi:
# Resolves #2615

## What is being addressed

Add envionment variable validation to bootstrap.sh.

### DIFF
--- a/devops/terraform/bootstrap.sh
+++ b/devops/terraform/bootstrap.sh
@@ -1,6 +1,27 @@
 #!/bin/bash
 set -e
 
+# validate environment variables are set
+if [ -z "$ARM_SUBSCRIPTION_ID" ]; then
+  echo "ARM_SUBSCRIPTION_ID is not set"
+  exit 1
+fi
+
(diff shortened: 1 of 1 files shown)


devops/terraform/bootstrap.sh
```diff
@@ -1,6 +1,27 @@
 #!/bin/bash
 set -e
 
+# validate environment variables are set
```
tamirkamara, line 4: I think adding the normal header we use elsewhere will accomplish the same thing:

```bash
set -o errexit
set -o pipefail
set -o nounset
```

marrobi: @tamirkamara you are correct, you added this in #2607 so will close the PR and issue.

PR #2623, opened 2022-09-20
Restrict resource templates to specific roles

### LINKED ISSUE #2600: Add ability to restrict resource templates to specific roles
## Description

As a TRE Administrator,
I want to restrict specific resource templates to certain roles,
So that only certain groups of users can deploy certain resources.

For example, only Researcher should be able to deploy research-specific VMs, and only Airlock Manager should be able to deploy a Review VM.

## Acceptance criteria

- [x] Only users with roles listed on the template can perform actions on the template

tanya-borisova:
# Resolves #2600 

## What is being addressed

Add ability to restrict certain user resource templates to be used by certain roles.
This is primarily needed for Airlock Reviews, where the Reviewer should not be able to use user resource templates intended for Researchers. However, it might be used in the future for other resource templates.

## How is this addressed

- Add authorizedRoles field to resource template
- In POST requests for all types of resources, check if the authenticated user has at least one role in the authorizedRoles (if authorizedRoles is not empty, otherwise all roles will be assumed authorized)
- in GET requests to list shared service templates and workspace templates, add a parameter authorized_only (defaulting to False). If true, only return list of templates that the authenticated user is authorized to use. Intended to be used from the UI to reduce a chance a user tries to create a resource they aren't authorized to (and getting an error)
- Added new GET requests to list workspace service templates and user resource templates on Workspace router, which automatically only returns the templates that the user is allowed to see

Note that the template.authorizedRoles are only checked on creation of the resource. This is intentional although may appear somewhat inconsistent


api_app/db/repositories/resource_templates.py
```diff
@@ -97,6 +105,7 @@ def create_template(self, template_input: ResourceTemplateInCreate, resource_typ
             "resourceType": resource_type,
             "current": template_input.current,
             "required": template_input.json_schema["required"],
+            "authorizedRoles": template_input.json_schema["authorizedRoles"] if "authorizedRoles" in template_input.json_schema else [],
```
tanya-borisova, line 108: I think we can't really make authorizedRoles a mandatory field at this point, as all customer templates will need to be migrated (not just the ones already in Cosmos, but any templates in their codebases too). Which is why we need to do these checks

tanya-borisova:
Moved it back to draft temporarily as @damoodamoo  pointed out that there's a problem: the GET methods for templates are on the root router so we won't get workspace-level roles there.
I'm going to mitigate this by having a separate GET method on the workspace router that will be used from the UI with an intention to display to user only the templates that they can deploy

damoodamoo (approved):

PR #2624, opened 2022-09-21
Admin vm shared service

### LINKED ISSUE #2611: Make an admin-vm shared service
## Description

As a TRE Administrator,
I want to install an admin-vm as a shared service
So that it will be created only when required

## Acceptance criteria

- [ ] Extract admin-vm from core infrastructure 
- [ ] Remove tools that aren't required and won't be possible to use (shared services are installed after the fw and most traffic will be blocked)
- [ ] Confirm admin tasks are still possible

tamirkamara:
# Resolves #2611 

## What is being addressed

The admin vm deployed in the core infra isn't used often and sometimes it also fail the provisioning.

## How is this addressed

- Remove it from the core deployment and recreated as an optional shared service that admins could install on demand.
- No tools will be installed on this new service


templates/shared_services/admin-vm/template_schema.json
```diff
@@ -0,0 +1,23 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "https://github.com/microsoft/AzureTRE/templates/shared_services/gitea/template_schema.json",
```
anatbal, line 3: The template shouldn't be this path
```suggestion
  "$id": "https://github.com/microsoft/AzureTRE/templates/shared_services/admin-vm/template_schema.json",
```

templates/shared_services/admin-vm/terraform/deploy.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
anatbal, line 1: Do we need deploy.sh, destory.sh?

anatbal (approved):
Two minor comments

templates/shared_services/admin-vm/terraform/deploy.sh
tamirkamara: yes, for local testing.

PR #2629, opened 2022-09-21
TRE Admin allowed to GET workspace by id

### LINKED ISSUE #2529: GET workspace/<workspace_id> fails on auto_create workspace when no scope_id is set
When creating a new workspace (via the UI, or Swagger) and setting the `client_id` to `auto_create`, any subsequent calls to `workspaces/new_workspace_id` fail whilst the workspace is being created. 

Relevant call stack from API logs: 
```
File "/api/./api/routes/workspaces.py", line 79, in retrieve_workspace_by_workspace_id
2022-08-30T13:46:16.623829724Z if access_service.get_workspace_role(user, workspace, user_role_assignments) != WorkspaceRole.NoRole:
2022-08-30T13:46:16.623833324Z File "/api/./services/aad_authentication.py", line 353, in get_workspace_role
2022-08-30T13:46:16.623836924Z raise AuthConfigValidationError(strings.AUTH_CONFIGURATION_NOT_AVAILABLE_FOR_WORKSPACE)
2022-08-30T13:46:16.623840624Z services.access_service.AuthConfigValidationError: Auth configuration not available for workspace
```

Be good to get any quick thoughts from @stuartleeks / @jjgriff93 as you both made some auth-related changes to the `retrieve_workspace_by_workspace_id` method. I'm happy to pick the bug up but any pointers appreciated.

damoodamoo:
# Resolves #2529 

A regression was introduced such that a TRE admin doesn't automatically get access to GET a workspace by ID. This was causing problems in workspace creation too. 

Change made to check if the user is a TRE admin and return the workspace if they are. If not, do the roles check. This is now in line with the list endpoint.


api_app/api/routes/workspaces.py
```diff
@@ -73,14 +73,19 @@ def _safe_get_workspace_role(user, workspace, user_role_assignments):
...
+async def retrieve_workspace_by_workspace_id(request: Request, user=Depends(get_current_tre_user_or_tre_admin), workspace=Depends(get_workspace_by_id_from_path)) -> WorkspaceInResponse:
+    # check if the user is a TRE admin - if they are, return the workspace. If not, catch it and check their workspace role membership.
+    try:
+        user = await get_current_admin_user(request)
         return WorkspaceInResponse(workspace=workspace)
-    else:
-        logging.debug("User doesn't have roles in workspace.")
-        raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail=strings.ACCESS_USER_IS_NOT_OWNER_OR_RESEARCHER)
+    except Exception:
+        access_service = get_access_service()
```
marrobi, line 82: Can this logic not be done using one of the other methods in `authentication.py` similar to `get_current_admin_user`? Say `get_current_workspace_owner_or_researcher_user_or_airlock_manager` ?

damoodamoo: yes, but - we need to allow any person to call the endpoint, so can't use a `depends` just to get the tre admin user?

marrobi: I meant instead of checking the user roles below, a second call could be used to one of the methods in `authentication.py` rather than a depends.

Or we could create a new entry in `authentication.py` that covers every user to use in a depends? Would make it clearer to read.

Also at the top of this file is already a dependency:

```
workspaces_core_router = APIRouter(dependencies=[Depends(get_current_tre_user_or_tre_admin)])
```
Will this cause an issue?

damoodamoo: The `get_current_xxx` methods check the roles by decoding the token passed to them from the call, so they'd need to be under the workspace router. In this instance we need to be under the main router to get the workspace details in the first place, so the method to get the role membership is by reaching out to AAD.

damoodamoo:
Discussed with @marrobi - we've decided that rather than reach out to AAD to get the role memberships, we can instead infer the scope ID from the TRE id + workspace ID, so we can then authenticate and use that token to GET against the `/workspaces/` router

PR #2631, opened 2022-09-21
POC - Automatically download data onto review VM Using managed identity 

### LINKED ISSUE #2513: Automatically download data onto review VM
In order to simplify the task of the Airlock Manager to review incoming requests, it would be great if the data that they need to review would be pre-populated on the review VM that they are going to use. 

To achieve this, we can accept a SAS URL for downloading request data as a template parameter for exportreview and importreview user resource templates, and have a cloudinit script that on startup downloads the data using the URL passed in.

anatbal:
# Resolves  #2513

## What is being addressed
 This is as POC of the second approach that is described in this issue.
It contains a new user-resource for an import reviewing windows vm, that automatically download the file using managed identity.

This PR is in draft until the discussion on different options of how to download data onto Review VM by @tanya-borisova 
 
## How is this addressed
1. Create a new user resource template, airlock request id (aka container name) should be passed as a parameter.
2. Give the VM's managed identity a **Storage Blob Data Reader** role assignment on the **container**.
3. Add a firewall role under network rule collection (can be added inside the one that is created for guacamole/in a similar way) that allows the review vm subnet access the service tag **AzureResourceManager**on port 443
4. Implement a cloud-init script with these two commands:
```
 az login --identity --allow-no-subscriptions
 az storage blob download-batch --account-name <import_storage_account> --source <container_name> --destination <target_folder> --auth-mode login
```


tanya-borisova:
We went for a different approach which is to provide the SAS URL explicitly to the user resource template: https://github.com/microsoft/AzureTRE/pull/2726

Thank you @anatbal for your POC pull request, it was really helpful in the implementation!

tanya-borisova:
We went for a different approach which is to provide the SAS URL explicitly to the user resource template: https://github.com/microsoft/AzureTRE/pull/2726

Thank you @anatbal for your POC pull request, it was really helpful in the implementation!

PR #2635, opened 2022-09-22
Fix health endpoint output

### LINKED ISSUE #1673: API health endpoint and solution monitoring

anatbal:
# Resolves #1673 

## What is being addressed

After re-enabling the health check again, there are two things which I missed that this PR fixes:
1.  Change the health endpoint output to return the real status (but still always returns 200)
2.  Enable the unit tests again


stuartleeks:
When reading the PR description, I initially thought that this was changing the health endpoint status code. Looking at the code, it looks like this is still always returning a 200 response for the health endpoint regardless of the health of Cosmos etc - is that the case? (I think always returning 200 is a good thing, to be clear 😄 )

anatbal:
> When reading the PR description, I initially thought that this was changing the health endpoint status code. Looking at the code, it looks like this is still always returning a 200 response for the health endpoint regardless of the health of Cosmos etc - is that the case? (I think always returning 200 is a good thing, to be clear 😄 )

You got it right, this intentionally still returns 200, just the json output has changed to reflect the true health status - meaning one can come and debug the system (but the api is still accessible, wont be shut down by Application Gateway)

stuartleeks (approved):
LGTM

PR #2652, opened 2022-09-27
UI: Users only see templates that they are authorized to use

### LINKED ISSUE #2640: UI: Users only see templates that they are authorized to use
In https://github.com/microsoft/AzureTRE/issues/2600, a new GET method was added on Workspace level for workspace service templates and user resource templates, that only returns templates that the user is authorized to use (i.e. they have one of the roles in authorizedRoles list, if that list isn't empty). 

Make change on the UI to make use of the new endpoint.

tanya-borisova:
# Resolves #2640 
## What is being addressed

After work in https://github.com/microsoft/AzureTRE/issues/2600 has been complete, certain user resource templates have become restricted to certain users (such as, review VMs are restricted to Airlock Manager).
On the UI, when creating resources, only return to the user the user resource templates they can use.

## How is this addressed

- Make use on the UI of new API endpoints introduced in https://github.com/microsoft/AzureTRE/pull/2623


ui/app/src/components/shared/create-update-resource/CreateUpdateResource.tsx
```diff
@@ -63,18 +63,26 @@ export const CreateUpdateResource: React.FunctionComponent<CreateUpdateResourceP
...
-      templatesPath = ApiEndpoint.WorkspaceServiceTemplates; break;
+      templateListPath = ApiEndpoint.WorkspaceServiceTemplates; templateGetPath = templateListPath; break;
     case ResourceType.SharedService:
-      templatesPath = ApiEndpoint.SharedServiceTemplates; break;
+      templateListPath = ApiEndpoint.SharedServiceTemplates; templateGetPath = templateListPath; break;
     case ResourceType.UserResource:
       if (props.parentResource) {
-        templatesPath = `${ApiEndpoint.WorkspaceServiceTemplates}/${props.parentResource.templateName}/${ApiEndpoint.UserResourceTemplates}`; break;
+        // If we are creating a user resource, parent resource must have a workspaceId
+        const workspaceId = (props.parentResource as WorkspaceService).workspaceId
```
tanya-borisova, line 81: @damoodamoo  I tried to use workspaceCtx as we discussed but for some reason it was empty when I did my tests, so I did it the other way

damoodamoo: yes of course, apologies that was a bad tip - the Create form is 'global', so is above and outside of the workspace context, good call.

ui/app/src/components/shared/create-update-resource/SelectTemplate.tsx
```diff
@@ -7,6 +7,7 @@ import { ExceptionLayout } from "../ExceptionLayout";
 
 interface SelectTemplateProps {
     templatesPath: string,
+    workspaceApplicationIdURI?: string | undefined,
```
damoodamoo, line 10: `nit` but you don't need undefined here as the prop is optional

damoodamoo (approved):
nice work, LGTM!

PR #2649, opened 2022-09-26
Airlock manager can use user resources

tanya-borisova:
# Resolves #2499 

## What is being addressed

Allowing Airlock Manager to create and operate user resources.
Airlock Manager needs to have this capacity to be able to create VMs to perform Airlock Requests from. 

## How is this addressed

- Add Airlock Manager to the list of allowed roles for all methods on user resources
- Add Airlock Manager to the list of allowed roles for GET for workspace services and workspace service operations (otherwise, on the UI Airlock manager is going to see errors)
- Restrict Research VM templates to only be used by WorkspaceOwner and WorkspaceResearcher
- Made a change on the UI to allow Airlock Manager to see Connect button
- Made a change in Guacamole to allow Airlock Manager to connect to VMs

PR #2664, opened 2022-09-29
Docs: Update permissions map, Airlock Manager can use user resources

tanya-borisova:
## What is being addressed

After #2649 was merged, Airlock Manager can now use user resources. Updating permissions map to reflect this


martinpeck (approved):
Thanks for this!

PR #2670, opened 2022-10-01
UI airlock review requests

jjgriff93:
# Resolves #9325

## What is being addressed

This adds the ability to create an airlock review in the UI as well as view reviews on a request.

## How is this addressed

- Added submit review api method
- Ui display for approval/rejection
- Support for local debugging of UI alongside local API (CORS)
- Fix for UX App Registration not allowing local UI debugging out of the box


api_app/core/config.py
```diff
@@ -7,6 +7,7 @@
 API_PREFIX = "/api"
 PROJECT_NAME: str = config("PROJECT_NAME", default="Azure TRE API")
 DEBUG: bool = config("DEBUG", cast=bool, default=False)
+ENABLE_LOCAL_DEBUGGING: bool = config("ENABLE_LOCAL_DEBUGGING", cast=bool, default=False)
```
tamirkamara, line 10: Historically the API used `DEBUG` as the setting for debug mode. Have you reviewed that? I think we should try not to introduce a new setting and/or rename the existing one.

jjgriff93: Yes I'm aware - I'm re-using this setting because the global DEBUG setting isn't enabled by default even when you're locally developing and you may not want to just to add CORS, whereas if you've prepared your local API for debugging the ENABLE_LOCAL_DEBUGGING setting will be set already so there'll be no extra actions for someone to perform

damoodamoo (approved):
nice work - i've had a side branch with the CORS stuff + local debugging too, but never managed to actually PR it, great to get it in.

api_app/core/config.py
tamirkamara: @jjgriff93 I don't think this is true - I personally have used the DEBUG flag many times and it's globally set up.
As documented in the .env.sample file:
```
# Used by the API and Resource processor application to change log level
# DEBUG=true
```
Granted the text isn't the one that will note this also changes the CORS but it's what's used in the app.

ENABLE_LOCAL_DEBUGGING looks like, and is in fact related to deployment and azure resources:
```
# This setting will enable your local machine to be able to
# communicate with Service Bus and Cosmos. It will also allow deploying
# the base workspace.
# ENABLE_LOCAL_DEBUGGING=true
```

damoodamoo: aren't they different things? `DEBUG` is for boosting the log level, but `ENABLE_LOCAL_DEBUGGING` is all about allowing communication between your dev machine and the cloud resources. I think the `ENABLE_LOCAL_DEBUGGING` makes most sense here.

jjgriff93: Exactly what David says - enable_local_debugging is intended for when you wish to debug locally, thus you’ll need CORS support to do so with the UI - whereas DEBUG is concerned with the log level. I think it’s cleaner to add this additional config setting and not expect users to have to enable two env vars instead of just the one

tamirkamara: IIRC it's true it's mostly about the log (although the var name doesn't note that) but also there're paths of code that use it, I think something about TLS verification and Cosmos DB.

All I'm saying is that as a person who knows different parts of the system `ENABLE_LOCAL_DEBUGGING` isn't something I associate with the API (or the RP) as it's an infra var. It's just not clear and if we build more logic on it, we should make it...
As one that spend hours recently changing docs, env var names, etc. since users where confused I think this is confusing.

PR #2671, opened 2022-10-01
support template version patching

### LINKED ISSUE #1744: Define mechanism for updating template version, and hence infrastructure of deployed workspaces, workspaces services and user resources

guybartal:
# Resolves #1744 

## What is being addressed

Support updating template version when patching resources

## How is this addressed

- check if template version exists, if so update resource document in cosmos db
- pending work: Templates include upgrade-able from version section to check if the request upgrade is valid

### DIFF
--- a/api_app/api/routes/shared_services.py
+++ b/api_app/api/routes/shared_services.py
@@ -90,8 +90,6 @@ async def patch_shared_service(shared_service_patch: ResourcePatch, response: Re
         raise HTTPException(status_code=status.HTTP_409_CONFLICT, detail=strings.ETAG_CONFLICT)
     except ValidationError as v:
         raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=v.message)
-    except ValueError as v:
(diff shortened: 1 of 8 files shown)


api_app/api/routes/shared_services.py
```diff
@@ -93,6 +93,8 @@ async def patch_shared_service(shared_service_patch: ResourcePatch, response: Re
         raise HTTPException(status_code=status.HTTP_409_CONFLICT, detail=strings.ETAG_CONFLICT)
     except ValidationError as v:
         raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=v.message)
+    except ValueError as v:
```
tamirkamara, line 96: Why are all these added?

guybartal: I've added a friendly error which tells the requester when an upgrade fails on target template version existence check.

tamirkamara (commented):
LGTM but will another person take a look too.

marrobi:
Discussed again and @guybartal going to create a design to enable an upgradable from approach as per https://github.com/microsoft/AzureTRE/issues/1744#issuecomment-1250234039

PR #2673, opened 2022-10-02
Stop using `upgrade` in terraform init calls

### LINKED ISSUE #2672: Terraform "Failed to marshal state to json"
**Describe the bug**

When trying to upgrade an older TRE release to a newer one this errors occur.

Examples:

*** Migrating TF Resources ***
Failed to marshal state to json: schema version 3 for random_password.password in state does not match version 2 from the providermake: *** [/home/vscode/AzureTRE/Makefile:103: deploy-core] Error 1

*** Migrating TF Resources ***
Failed to marshal state to json: unsupported attribute "method"make: *** [/home/vscode/AzureTRE/Makefile:103: deploy-core] Error 1

*** Migrating TF Resources ***
Failed to marshal state to json: schema version 0 for azurerm_servicebus_namespace_network_rule_set.servicebus_network_rule_set in state does not match version 1 from the providermake: *** [/home/vscode/AzureTRE/Makefile:103: deploy-core] Error 1

tamirkamara:
# Resolves #2672 

## What is being addressed

Unmarshal terraform errors in `migrate.sh`.
This is a result of providers being upgraded and their schema changes. This is usually handled by `terraform apply` but we call other commands in the migrate script before the apply which might produce an error.

## How is this addressed

- Call `terraform refresh` before we call `show` to sort out any schema changes
- Upgrade all providers in core to their latest version (apart from azurerm). The reason is we didn't have all listed in `required_providers` block until recently and some deployments might see providers version move backward when then upgrade to the latest AzureTRE (mostly the random provider).
- Remove all uses of `-upgrade` when calling `terraform init` as that causes an upgrade and disregard the locks file. We don't need that behavior to happen automatically.


guybartal (approved):
LGTM

Issue #2518, opened 2022-08-28
Review and add private endpoints to storage accounts in core


tamirkamara:
2 accounts currently don't have private endpoints:
- [x] airlock function
- [x] azure monitor

PR #2679, opened 2022-10-02
Airlock function private endpoint for internal storage

tamirkamara:
## What is being addressed

The Airlock Function uses a storage account internally but does it over public network.
This is part of #2518 

## How is this addressed

- Add Private Endpoint(s) for the storage account
- Block public internet access to the storage account


guybartal (approved):
LGTM

yuvalyaron:
@tamirkamara 
why does the airlock function need a private endpoint?
isn't a firewall rule in the storage account enough?

tamirkamara:
> @tamirkamara why does the airlock function need a private endpoint? isn't a firewall rule in the storage account enough?

@yuvalyaron I imagine you'd define a rule with the subnet/vnet internal address, right? A private endpoint is essencially a way to give the storage account an ip address in the vnet and by such we make sure all communication is done over private network. IIRC, you need this anyway for the firewall rule mentioned above.

PR #2698, opened 2022-10-03
Bugfix: Airlock Manager should only see their own resources

### LINKED ISSUE #2697: Airlock Manager can see resources of other users
**Describe the bug**

In the UI, Airlock Manager can see resources of other users, such as VM. This privilege should only be granted to Workspace Owner.

tanya-borisova:
# Resolves #2697 

## What is being addressed

In https://github.com/microsoft/AzureTRE/pull/2649, a bug was made that meant that Airlock Manager is treated as Workspace Owner when it comes to checking permissions on workspace user resources, and could see all resources, not just their own.
Fixing the bug so that they can do operations on their own resources only. 

Additionally, made a change so that Airlock Manager can only see their own resources in list response. 

Tested in a dev environment.


martinpeck (approved):

tanya-borisova:
/test-force-approve

There was already a successful test run earlier

PR #2700, opened 2022-10-04
Re-working Status Update Queue Message Handling

### LINKED ISSUE #2647: Resource hangs with "pipeline running" during update
We have had a couple of instances where resource updates do not complete, and the status remains on updating. This means the resource remains locked in the UI. The operations look like:

image

We need to get the resource processor logs from an instance of this happening.

damoodamoo:
# Resolves #2647

### DIFF
--- a/api_app/service_bus/deployment_status_update.py
+++ b/api_app/service_bus/deployment_status_update.py
@@ -13,7 +13,7 @@ from db.repositories.resource_templates import ResourceTemplateRepository
 from service_bus.helpers import send_deployment_message, update_resource_for_step
 from azure.servicebus import NEXT_AVAILABLE_SESSION
 from azure.servicebus.exceptions import OperationTimeoutError, ServiceBusConnectionError
-from azure.servicebus.aio import ServiceBusClient, AutoLockRenewer, ServiceBusReceiver
+from azure.servicebus.aio import ServiceBusClient, AutoLockRenewer
 from db.repositories.operations import OperationRepository
 from core import config, credentials
 from db.errors import EntityDoesNotExist
--- a/api_app/service_bus/deployment_status_update.py
+++ b/api_app/service_bus/deployment_status_update.py
@@ -1,8 +1,9 @@
+import asyncio
 import json
 import logging
+import threading
 import uuid
 
-from azure.servicebus.aio import ServiceBusClient
 from pydantic import ValidationError, parse_obj_as
 
 from api.dependencies.database import get_db_client
@@ -10,6 +11,9 @@ from api.routes.resource_helpers import get_timestamp
 from models.domain.request_action import RequestAction
 from db.repositories.resource_templates import ResourceTemplateRepository
 from service_bus.helpers import send_deployment_message, update_resource_for_step
+from azure.servicebus import NEXT_AVAILABLE_SESSION
+from azure.servicebus.exceptions import OperationTimeoutError, ServiceBusConnectionError
+from azure.servicebus.aio import ServiceBusClient, AutoLockRenewer
 from db.repositories.operations import OperationRepository
 from core import config, credentials
 from db.errors import EntityDoesNotExist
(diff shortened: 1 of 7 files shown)


tanya-borisova (commented):
Does adding session_id require any kind of migration plan for the queue?

damoodamoo:
> Does adding session_id require any kind of migration plan for the queue?

The queue will just recreate. I've called it out in the changelog, but there's no 'migration' possible really. just a case of making sure you deploy when the TRE is idle.

api_app/services/health_checker.py
```diff
@@ -50,7 +50,7 @@ async def create_service_bus_status(credential) -> Tuple[StatusEnum, str]:
     try:
         service_bus_client = ServiceBusClient(config.SERVICE_BUS_FULLY_QUALIFIED_NAMESPACE, credential, retry_total=0)
         async with service_bus_client:
-            receiver = service_bus_client.get_queue_receiver(queue_name=config.SERVICE_BUS_DEPLOYMENT_STATUS_UPDATE_QUEUE)
```
tanya-borisova, line 53: Why has this changed?

api_app/service_bus/deployment_status_update.py
```diff
@@ -18,217 +22,227 @@
...
-        resource_id = uuid.UUID(step_to_update.resourceId)
+                                complete_message = await self.process_message(msg)
 
-        resource = resource_repo.get_resource_by_id(resource_id)
-        resource.deploymentStatus = step_to_update.status
-        resource_repo.update_item(resource)
+                                if complete_message:
+                                    await receiver.complete_message(msg)
+                                else:
+                                    # could have been any kind of transient issue, we'll abandon back to the queue, and retry
```
tanya-borisova, line 55: Do we have / do we need any kind of dead letter queue here?

api_app/service_bus/deployment_status_update.py
```diff
@@ -18,217 +22,227 @@
...
+            logging.info(f"Update status in DB for {message.operationId} - {message.status}")
+        except (json.JSONDecodeError, ValidationError) as e:
+            complete_message = True
+            logging.error(f"{strings.DEPLOYMENT_STATUS_MESSAGE_FORMAT_INCORRECT}: {msg.correlation_id} - {e}")
+        except Exception as e:
+            logging.info(f"Exception for: {msg.correlation_id} - {e}")
+
+        return complete_message
+
+    async def update_status_in_database(self, message: DeploymentStatusUpdateMessage):
```
tanya-borisova, line 90: From here on the code hasn't changed, right?

tanya-borisova (commented):
Looks good, just had one question and I haven't read part of the diff under the assumption that it hadn't changed

api_app/services/health_checker.py
damoodamoo: because this was previously connecting to the status update queue in order to test connectivity for the health endpoint - but since that now uses sessions the mechanism to connect to it is more complex - so I opted to move this code to connect to the other 'normal' queue, ie. the airlock one.

api_app/service_bus/deployment_status_update.py
damoodamoo: correct, the `update_status_in_database` does the same thing - the only stuff that's changed is that now we're using a class we have some class level members (the repositories) that we set up once at startup, so we use `self.repo` instead of passing `repo` around

api_app/service_bus/deployment_status_update.py
```diff
@@ -18,217 +22,227 @@
...
 
-async def receive_message():
-    """
-    This method is an async generator which receives messages from service bus
-    and yields those messages. If the yielded function return True the message is
-    marked complete.
-    """
-    async with credentials.get_credential_async() as credential:
-        service_bus_client = ServiceBusClient(config.SERVICE_BUS_FULLY_QUALIFIED_NAMESPACE, credential)
+class DeploymentStatusUpdate(threading.Thread):
```
tanya-borisova, line 25: Nit: DeploymentStatusUpdater (ending with r), as it's a class?

damoodamoo: harsh...but fair :)

tanya-borisova (approved):

PR #2608, opened 2022-09-18
fix various terraform issues / warnings

tamirkamara:
## What is being addressed

Since we upgraded some of the Terraform providers we use, we see deprecation warnings. 
This sorts all of those.

PR #2702, opened 2022-10-04
Disable network policies for private endpoints

### LINKED ISSUE #2667: Get 500 error from api when calling /link for export airlock request
**Describe the bug**
Both from the UI and the API directly I get a 500 Unable to process request error when trying to get a SAS link on a draft export airlock request. I'm authenticated as a researcher role and am able to generate links for import requests without any issues.

image

**Steps to reproduce**

1. Run the UI
1. Create an export airlock request as a researcher within a workspace
1. Press Generate on the files link field

tanya-borisova:
# Resolves #2667 

## What is being addressed

It seems that in https://github.com/microsoft/AzureTRE/pull/2608, a degradation was introduced: API can no longer connect to Services subnet in Workspaces, thus creation of Airlock export links is broken.

It seems that we do not yet intend enabling network policies for private endpoints, so in this PR I'm disabling all of them. 

I have tested that disabling network policies on Services subnet in a Workspace fixes the problem of creating export Airlock requests.

### DIFF
--- a/templates/core/terraform/network/network.tf
+++ b/templates/core/terraform/network/network.tf
@@ -27,7 +27,7 @@ resource "azurerm_subnet" "app_gw" {
   virtual_network_name                          = azurerm_virtual_network.core.name
   resource_group_name                           = var.resource_group_name
   address_prefixes                              = [local.app_gw_subnet_address_prefix]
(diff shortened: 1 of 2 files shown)


stuartleeks (approved):
LGTM based on the discussion earlier :-)

PR #2704, opened 2022-10-05
Add Explicit Dependency to App Gateway

martinpeck:
# Resolves 2598

## What is being addressed

This adds an explicit dependency to the app gateway so that it waits until the private link has been set up. The hope is that this prevents the situations where deployments result in an app gateway returning 502 errors.


marrobi (commented):
Rather than an explicit dependancy, if the app gw backend was to reference the private endpoint fqdn, it should wait. I expect at present its using the web app fqdn, hence waiting for the web app, not private endpoint. 

https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/private_endpoint#fqdn

marrobi:
Yup, see https://github.com/microsoft/AzureTRE/blob/main/templates/core/terraform/main.tf#L89.

martinpeck:
As per conversation elsewhere let's keep this as a dependency, see if the issue goes away, and then look to improve the fix later.

stuartleeks (approved):
LGTM

PR #2705, opened 2022-10-05
New /workspaces/<workspace-id>/scopeid endpoint

### LINKED ISSUE #2662: Update Auth Script to name WS App Scope IDs the same as auto_create does
Now the UI infers the scope ID based on convention, the `create_workspace_application.sh` script need to be updated to name the Scope ID in the format:

`<TRE_ID>-ws-<last-4-chars-of-ws-id>`

Also need to highlight this breaking change in the Changelog.

damoodamoo:
# Resolves #2662 

Few notes on this:
- We need to get the workspace by id - for which we need a workspace token. The `/workspaces` endpoint uses Graph calls to check whether the user can see a given workspace or not - but this requires `Directory.ReadAll` permission, which some customers are not comfortable with. In this case they will not use the dashboard page, but provide links direct to the `/workspace/<id>` page in the UI.
- A TRE user has no way to get the workspace by id without getting the token - for which we need to know the `scope_id`
- A previous approach used the fact that AAD apps are generated in a consistent way by terraform, so we could infer the `scope_id` by convention... BUT this failed when considering manually created / scripted AAD apps.

The new approach allows any TRE User to query the `/workspaces/<id>/scopeid` endpoint, which will return _just_ the scope ID for that workspace, and no other properties. There are unit tests here to cover this. The scope ID is then used to get a token and authenticate against the workspace to get it.


marrobi (approved):
LGTM

PR #2677, opened 2022-10-02
Upgrade Terraform Mixing to 1.0.0-rc.1

tamirkamara:
# Resolves #2659

## What is being addressed

The terraform mixin in porter now includes mirroring of providers internally so we can remove the process we did in our dockerfile.tmpl in our templates.

## How is this addressed

- Resource processor image used the 1.0.0.rc.1 of the terraform mixin
- Remove terraform mixin commands from all templates

PR #2590, opened 2022-09-13
Block TRE access to Terraform/Hashicorp domains

tamirkamara:
# Resolves #2445 

## What is being addressed

With all bundles implementing Terraform provider mirroring it's time to start blocking TRE communication to Terraform/Hasihcorp domains. 

## Migration

It's advised to upgrade your firewall to the new `0.5.0` version if all your deployed tre resources are at (or above) the versions below.
If you have created and deployed custom templates that use Terraform, you need to make sure you mirror providers. Look at the _dockerfile.tmpl_ of one our templates for an example.

The method to do this upgrade is manual - upgrade the _templateVersion_ of `tre-shared-service-firewall` resource in Cosmos to `0.5.0`

| template name | version |
| ----- | ----- |
| tre-workspace-base | 0.3.28 |
| tre-workspace-unrestricted | 0.1.9 |
| tre-service-mlflow | 0.3.7 |
| tre-service-innereye | 0.3.5 |
| tre-workspace-service-gitea | 0.3.8 |
| tre-workspace-service-mysql | 0.1.2 |
| tre-service-guacamole-linuxvm | 0.4.13 |
| tre-service-guacamole-windowsvm | 0.4.8 |
| tre-service-guacamole | 0.4.5 |
| tre-user-resource-aml-compute-instance | 0.3.2 |
| tre-service-azureml | 0.4.8 |
| tre-shared-service-cyclecloud | 0.2.6 |
| tre-shared-service-gitea | 0.3.14 |
| tre-shared-service-airlock-notifier | 0.1.2 |
| tre-shared-service-certs | 0.1.3 |
| tre-shared-service-sonatype-nexus | 2.1.6 |
| tre-shared-service-firewall | 0.4.3 |

PR #2716, opened 2022-10-10
Allow access to Terraform/Hashicorp domains

tanya-borisova:
## What is being addressed

For release 0.5.0, we're temporarily reverting change that blocked TRE access to hashicorp domain. 
https://github.com/microsoft/AzureTRE/pull/2590/files 

Having the change in the release means that customers wanting to upgrade to TRE 0.5.0 need to upgrade all bundles they use in TRE at the same time as they upgrade the firewall, which can be difficult for some customers (e.g. there is no clear path to upgrade the Nexus shared service). 

By reverting temporarily the blocking of Hashicorp domains, we give customers time to upgrade the resources they use to the new bundles that use the new mixin (see https://github.com/microsoft/AzureTRE/pull/2677).
After that, we block access to hashicorp domain in one of the future releases. 

## How is this addressed

- Revert PR https://github.com/microsoft/AzureTRE/pull/2590
- Move version of firewall bundle forward to 0.6.1
- Updated CHANGELOG


stuartleeks (approved):
LGTM

templates/shared_services/firewall/porter.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 name: tre-shared-service-firewall
-version: 0.6.0
+version: 0.6.1
```
tamirkamara, line 3: @tanya-borisova why would one upgrade to the newer firewall versions in the first place and run into this issue?
Iirc the changes in the recent versions are exactly that so a way to avoid the issue is not to upgrade the firewall. You can resolve this after considering.

tanya-borisova: @tamirkamara  You are correct that if firewall isn't upgraded, customers don't run into this issue. Currently, in the CHANGELOG, customers are advised to upgrade firewall, at which point, if they have any old bundles still deployed, these will stop working. 

We have discussed two options with the team and others, for the way forward with the release: 

Option 1 is what you suggest. We make the release with code changes as-is, but we advise customers to not upgrade firewall until all their bundles are upgraded (to the ones that are built with the new mixin).

Option 2. We make a code change before we do the release, which would temporarily revert blocking of hashicorp domains in firewall. This means that customers have time to upgrade their bundles to use the new mixin, which is only available in this release. 

We think that doing option 2 provides a better upgrading strategy for customers, as they can take time to upgrade their bundles and be confident everything is upgraded, before they upgrade the firewall in the future. It also means we have a way to release any other change or fix to firewall, should the need arise, before customers finished upgrading their bundles.

PR #2718, opened 2022-10-10
Simplify makefile commands

tanya-borisova:
## What is being addressed

There are helpful functions defined in Makefile to build, publish and register different types of bundles.
Unfortunately they are defined as functions which means they can't be called easily from the command line.

Convert them to targets so that a bundle can be easily built, published, and registered in one line with a command like this:

```
make user_resource_bundle BUNDLE=guacamole-azure-export-reviewvm WORKSPACE_SERVICE=guacamole
```

or

```
make workspace_bundle BUNDLE=base
```


Makefile
```diff
@@ -251,17 +251,21 @@ bundle-register:
 		--current --insecure --tre_url "$${TRE_URL:-https://$${TRE_ID}.$${LOCATION}.cloudapp.azure.com}" --verify \
 		--workspace-service-name "$${WORKSPACE_SERVICE_NAME}"
 
-workspace_bundle = $(MAKE) bundle-build bundle-publish bundle-register \
-	DIR="${MAKEFILE_DIR}/templates/workspaces/$(1)" BUNDLE_TYPE=workspace
+workspace_bundle:
```
tanya-borisova, line 254: These targets shouldn't be used anywhere apart from `prepare-for-e2e` target below, so it should be okay to change them

Makefile
```diff
@@ -290,11 +294,11 @@ build-and-deploy-ui:
 	&& if [ "$${DEPLOY_UI}" != "false" ]; then ${MAKEFILE_DIR}/devops/scripts/build_deploy_ui.sh; else echo "UI Deploy skipped as DEPLOY_UI is false"; fi \
 
 prepare-for-e2e:
```
tanya-borisova, line 289: Not sure this target is needed anymore. I am probably the only person using it... I would probably just replace the reference to it here: https://github.com/microsoft/AzureTRE/blob/main/docs/tre-developers/end-to-end-tests.md?plain=1#L17  with concrete commands to run

martinpeck (approved):
LGTM

tanya-borisova:
/test-force-approve

No point taking it through CI as these are the targets that aren't used there

Issue #2503, opened 2022-08-24
Add Title field to airlock requests


jjgriff93:
## Description

The Airlock request object should include a Title field (a short descriptive name to help reviewers identify what the request pertains to, i.e. "Clinical trial data for ML model training"). This will be used by the UI when displaying a list of requests to help differentiate them (the longer businessJustification field does not cover this purpose).

## Acceptance criteria

- [ ] Airlock request objects processed by API have new "title" field and this gets returned by API
- [x] Older deployments of airlock have migration path to new structure
- [x] Show the new field in the airlock UI (list and in view pane)

PR #2722, opened 2022-10-11
add title to airlock request (#2503)

hizni:
# Resolves:  add a title to airlock request (#2503)

## What is being addressed

Adds a title to the airlock request to have a terse description that allows the request to be more identifiable. 

## How is this addressed

- Adds the ability to add a title to the airlock request
- This title is then visible when viewing the airlock request
- Updated the interaction with the API to store this new data


jjgriff93:
Closing this as it is continued in #2731

PR #2725, opened 2022-10-11
Fix dev container CLI install path and shared services install jq command

### LINKED ISSUE #2724: Shared services pipeline install fails with 409 if status isn't deployed
If try to run the pipeline and a shared service is installed with a status other than deployed the install will proceed and error with a 409. The status should not matter - redeployment should fail.

marrobi:
# Resolves #2724

## What is being addressed
- Makes the CLI install path relative in the dev container
- Deploy shared services script not concerned about deployment status

### DIFF
--- a/devops/scripts/deploy_shared_service.sh
+++ b/devops/scripts/deploy_shared_service.sh
@@ -49,7 +49,7 @@ if [[ "$last_result" != 0 ]]; then
 fi
 
 deployed_shared_service=$(echo "${get_shared_services_result}" \
-  | jq -r ".sharedServices[] | select(.templateName == \"${template_name}\" and .deploymentStatus == \"deployed\")")
+  | jq -r ".sharedServices[] | select(.templateName == \"${template_name}\")")
 
 if [[ -n "${deployed_shared_service}" ]]; then
   # Get template version of the service already deployed
--- a/devops/scripts/deploy_shared_service.sh
+++ b/devops/scripts/deploy_shared_service.sh
@@ -49,7 +49,7 @@ if [[ "$last_result" != 0 ]]; then
 fi
 
 deployed_shared_service=$(echo "${get_shared_services_result}" \
-  | jq -r ".sharedServices[] | select(.templateName == \"${template_name}\" and (.deploymentStatus != \"deleted\" or .deploymentStatus != \"deploymentFailed\"))")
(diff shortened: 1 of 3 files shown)


devops/scripts/deploy_shared_service.sh
```diff
@@ -49,7 +49,7 @@ if [[ "$last_result" != 0 ]]; then
 fi
 
 deployed_shared_service=$(echo "${get_shared_services_result}" \
-  | jq -r ".sharedServices[] | select(.templateName == \"${template_name}\" and .deploymentStatus == \"deployed\")")
```
tanya-borisova, line 52: Yeah this condition is something I missed when I was working on this https://github.com/microsoft/AzureTRE/pull/2235/files 
Sorry!

I think the right condition is to check that the status is not "deleted". As if it is "deleted", we can attempt the deployment again, but not in any other case

tanya-borisova (changes requested):
Requested changes as I think the condition in shared services script needs to be changed and not removed

devops/scripts/deploy_shared_service.sh
tanya-borisova: Actually, if the status is deploymentFailed, we can also attempt the deployment again

marrobi: @tanya-borisova and we won't get a 409 in either of those cases?

marrobi: Made that change.

tanya-borisova (approved):

devops/scripts/deploy_shared_service.sh
```diff
@@ -49,7 +49,7 @@ if [[ "$last_result" != 0 ]]; then
...
+  | jq -r ".sharedServices[] | select(.templateName == \"${template_name}\" and (.deploymentStatus != \"deleted\" or .deploymentStatus != \"deploymentFailed\"))")
```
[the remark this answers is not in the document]
marrobi: Good catch? I was going by @tanya-borisova comment above. Can you let me know.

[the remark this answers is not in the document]
tanya-borisova: it's correct, good spot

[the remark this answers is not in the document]
marrobi: I've updated, as think you are correct.

PR #2740, opened 2022-10-16
Create and delete Airlock Review VMs

### LINKED ISSUE #2735: Create Review VM from the Airlock UI screen
## Description

As a TRE Airlock Manager
I want to have a button next to the Airlock Request to spin up the Review VM
So that I don't have to remember which workspace and workspace service to create it in and I don't have to copy SAS URL

### Configuration

We propose to add an additional nested block to Research Workspace (=base workspace) properties for configuring the Airlock Review VMs. 

This could look like this: 

```
{
    "id": "<id>",
    "templateName": "tre-workspace-base",
    "templateVersion": "0.3.28",
    "properties": {
        "enable_airlock": true,
        "airlock_review_config": {
            "import": {
                "workspace_id": "<guid>",  // ID for the already deployed Import Review Workspace to use for reviews
                "workspace_service_id": "<guid>",  // ID for the already deployed Workspace Service to deploy review VMs to
                "user_resource_template_name": "guacamole-azure-import-reviewvm"  // the user resource template to create
            },
            "export": {
                "workspace_serivce_id": "<guid>",  // ID for the already deployed Workspace Service to deploy review VMs to
                "user_resource_template_name": "guacamole-azure-import-reviewvm"  // the user resource template to create
            },
        },
    // other properties as normal
    },
```

The `export` section does not have `workspace_id` because for export it will always be the Research Workspace the data is exported from. 
The corresponding resources need to already be deployed before this configuration can be set, which is why we should recommend users to issue an update on an existing resource if they wish to configure Airlock Review. 

### UI screens

We then propose to add a button / screen on the UI next to the Airlock Review requests ([updated in Figma](https://www.figma.com/file/Ncl8TJsFf2WKf3frrqTbtq/AzureTRE-OSS-UI?node-id=306%3A21280)): 

image

image

image

The button will be greyed out if the above configuration is not set for the workspace, with a message for the user to ask their TRE admin to configure this for them. 

image

If there already is a review VM created for the request by this user, it could show a link to that created review VM. 
If the review VM was created by another user, it could show a message saying that another user is currently reviewing this request. For this, the `review_user_resources` field in Airlock Request object is used (see below).

image

### API endpoint

Add a corresponding API endpoint that will create a review User Resource

`/api/workspaces/{workspace_id}/requests/{airlock_request_id}/create-review-user-resource`

This will accept the Research Workspace ID and Request ID as parameters, from the properties outlined above will determine in which Workspace and Workspace Service to create them in, will create the user resource and return the Operation to watch, which will be tracked in the usual way on the UI. 

The API endpoint will also update the Request object with the information on this VM: 

```
{
    "id": "45e1d9d8-f669-47bf-b0eb-a11bb34aa456",
    "workspaceId": "26f41c2a-2bed-4f78-a745-c41ca8f326fe",
    "requestType": "export",
    "files": [
        {
            "name": "research_result.txt",
            "size": 7
        }
    ],
    "review_user_resources": [
        {
            "id": "<guid>",
            "workspace_id": "<guid>",
            "workspace_service_id": "<guid>"
        }
    ],
    // other properties as normal
}
```

## Acceptance criteria

- [x] #2736
- [x] #2737
- [x] #2738
- [ ] #2830

### LINKED ISSUE #2737: API endpoint is added for creating Review user resources
See the parent task for details

tanya-borisova:
# Resolves #2737

## What is being addressed

Add endpoints for creating an Airlock Review user resources.
Add cleaning up of these user resources in `/review` method. 
For details on the design, please see the parent issue https://github.com/microsoft/AzureTRE/issues/2735 

The methods can be used so long as the workspace configuration is set up appropriately. 

Additionally, made changes to tre-service-guacamole-import-reviewvm and tre-service-guacamole-export-reviewvm to make the correct configuration the default one.


api_app/api/routes/airlock_resource_helpers.py
```diff
@@ -42,12 +50,28 @@ async def save_and_publish_event_airlock_request(airlock_request: AirlockRequest
         raise HTTPException(status_code=status.HTTP_503_SERVICE_UNAVAILABLE, detail=strings.EVENT_GRID_GENERAL_ERROR_MESSAGE)
 
 
-async def update_and_publish_event_airlock_request(airlock_request: AirlockRequest, airlock_request_repo: AirlockRequestRepository, user: User, new_status: AirlockRequestStatus, workspace: Workspace, request_files: List[AirlockFile] = None, status_message: str = None, airlock_review: AirlockReview = None):
+async def update_and_publish_event_airlock_request(
+        airlock_request: AirlockRequest,
+        airlock_request_repo: AirlockRequestRepository,
+        user: User,
+        workspace: Workspace,
+        new_status: AirlockRequestStatus = None,
```
tanya-borisova, line 58: new_status can now be None to account for the case when the status isn't changed

tanya-borisova:
@marrobi @damoodamoo  This is how creation of the workspace screen looks like now, possibly something here is broken as the description text repeats twice. However it doesn't currently ask for any inputs (maybe that also is a bug?)

image

api_app/api/routes/airlock.py
```diff
@@ -67,18 +75,123 @@ async def retrieve_airlock_request_by_id(airlock_request=Depends(get_airlock_req
...
+                            Please ask your TRE administrator to check the configuration. Details: {str(e)}")
+
+    # Getting the SAS URL
+    airlock_request_sas_url = get_airlock_container_link(airlock_request, user, workspace)
+
+    # Now have all components for user resource, create an object for it
+    user_resource_create = UserResourceInCreate(
+        templateName=user_resource_template_name,
+        properties={
+            "display_name": "Review VM",
```
jjgriff93, line 139: Should we maybe say Airlock Review VM to be explicit? For cases where its an Export request they'll have a number of VMs showing up perhaps alongside regular VMs

api_app/api/routes/airlock.py
```diff
@@ -90,19 +203,35 @@ async def create_airlock_review(airlock_review_input: AirlockReviewInCreate, air
     elif airlock_review.reviewDecision.value == AirlockReviewDecision.Rejected:
         review_status = AirlockRequestStatus.RejectionInProgress
 
-    updated_airlock_request = await update_and_publish_event_airlock_request(airlock_request=airlock_request, airlock_request_repo=airlock_request_repo, user=user, new_status=review_status, workspace=workspace, airlock_review=airlock_review)
+    # If there was a VM created for the request, clean it up as it will no longer be needed
+    # In this request, we aren't returning the operations for clean up of VMs,
+    # however the operations still will be saved in the DB and displayed on the UI as normal.
+    _ = await delete_review_user_resources(
```
jjgriff93, line 209: Should this maybe be moved to after the update airlock request completes, as if it fails for whatever reason, the review VM will have been deleted but the request would still be in a review status

jjgriff93 (approved):
Really great work! Just some minor suggestions etc., up to you which you want to look at/ignore

api_app/api/routes/airlock.py
tanya-borisova: Yes good point, will rename. By the way, researchers will not see the Review VMs, as they only can see their own VMs. Same for Airlock Managers, they won't see the research VMs. But in any case it's best to be explicit!

api_app/api/routes/airlock.py
```diff
@@ -67,18 +75,123 @@ async def retrieve_airlock_request_by_id(airlock_request=Depends(get_airlock_req
...
+    try:
+        workspace_service = workspace_service_repo.get_workspace_service_by_id(workspace_id=workspace_id, service_id=workspace_service_id)
+    except EntityDoesNotExist as e:
+        logging.error(f"Failed to get workspace serivce {workspace_service_id} for workspace {workspace_id}: {str(e)}")
+        raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+                            detail=f"Failed to retrive Airlock Review configuration for workspace {workspace.id}.\
+                            Please ask your TRE administrator to check the configuration. Details: {str(e)}")
+
+    # Getting the SAS URL
+    airlock_request_sas_url = get_airlock_container_link(airlock_request, user, workspace)
```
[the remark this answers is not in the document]
tanya-borisova: The method raises HTTPException in case something goes wrong, I'll leave a comment about that

api_app/api/routes/airlock.py
tanya-borisova: A very good point! Will do that

tanya-borisova:
/test-force-approve

Successful test run in a comment above

PR #2742, opened 2022-10-18
Remove shared service name prefix

tamirkamara:
## What is being addressed

A previous change introduced a prefix for the shared service display_name when deployed via script/cli. This PR fixes that back so that the firewall (for example) shared service will be listed as just "firewall" with nothing extra.


marrobi (approved):

stuartleeks (approved):
LGTM - thanks!

PR #2745, opened 2022-10-18
Fix Airlock notifier deployment

tanya-borisova:
## What is being addressed

- Fix deployment of Airlock notifier (previously was failing with ` StatusCode=400 -- Original Error: Code="BadRequest" Message="The parameter 'MinimumElasticInstanceCount' has an invalid value. `)


damoodamoo (approved):

tanya-borisova:
/test-force-approve

This bundle isn't used in tests

Issue #2707, opened 2022-10-06
Add missing table format queries for tre CLI commands


stuartleeks:
## Description

As a TRE Administrator
I want to have default table renderings for CLI commands
So that I am productive when interactively using the `tre` CLI

## Acceptance criteria

- [ ] All CLI commands that allow tabular output have default table queries specified

stuartleeks:
The remaining queries here are in the costs commands and require [JMESPath lexical scopes](https://jmespath.site/#wiki-lexical-scopes) which are a new proposal and not yet implemented.

Issue #2706, opened 2022-10-06
Ensure full API coverage in the tre CLI


stuartleeks:
## Description

As a TRE Administrator
I want to be able to use the full TRE API via the `tre` CLI
So that I can automate all aspects of the TRE

## Acceptance criteria

- [ ] All API endpoints are reflected in the `tre` CLI

PR #2747, opened 2022-10-18
CLI Updates

stuartleeks:
# Work towards #2706 and #2707

## What is being addressed

## How is this addressed

- Add invoke action commands
- Add shell completion for shared service IDs
- Add basic table query for overall costs
- Add migrations command
- Add raw output option
- Add airlock request title
- airlock -> airlock-request and fix error text
- Fix airlock request id shell completion
- Add table format for airlock request submit response and review response
- Fix up airlock request review URL
- Add example airlock-request creation to README.md
- Update API: Handle missing app role name
- Add airlock-request cancel command
- Add workspace service delete command
- Update output command to take response and exit on error (for more consistent exit code handling)


cli/tre/api_client.py
```diff
@@ -1,10 +1,12 @@
 import sys
+from typing import Union
 import click
 import json
 import msal
 import os
 
 from httpx import Client, Response
+from httpx._types import QueryParamTypes
```
martinpeck, line 9: Looks like you're importing a "private" module. Is QueryParamTypes a stable API to depend upon?
Do you need to type it like this, or can it just be typed as a Dict?

cli/tre/commands/shared_services/shared_service.py
```diff
@@ -10,8 +10,17 @@
 from .operations import shared_service_operations
 
 
+def shared_service_id_completion(ctx, param, incomplete):
```
martinpeck, line 13: nit: This is fine....it's just that you've been type-hinting all the other methods, so I wondered "why not this one?"

cli/tre/api_client.py
stuartleeks: I used that type as that is what the `params` property on the (public) `request` method is typed as.
 
 I've updated to a `dict` now

cli/tre/commands/shared_services/shared_service.py
stuartleeks: I've pushed a commit with type annotations added to the completion methods

martinpeck (approved):
LGTM!

api_app/services/aad_authentication.py
```diff
@@ -265,7 +265,7 @@ def get_workspace_role_assignment_details(self, workspace: Workspace):
 
             if principal_type == "User" and principal_id in user_emails:
                 app_role_id = role_assignment["appRoleId"]
-                app_role_name = inverted_app_role_ids[app_role_id]
+                app_role_name = inverted_app_role_ids.get(app_role_id)
```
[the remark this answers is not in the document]
martinpeck: Yep. Makes sense.

PR #2759, opened 2022-10-19
Airlock UI: create review VM

### LINKED ISSUE #2735: Create Review VM from the Airlock UI screen

### LINKED ISSUE #2738: UI for Airlock Review VM
https://www.figma.com/file/Ncl8TJsFf2WKf3frrqTbtq/AzureTRE-OSS-UI?node-id=1293%3A27999

jjgriff93:
# Resolves
#2738

## What is being addressed

Adds ability to deploy a review Vm in airlock review UI and open to view airlock request data securely in the review workspace

## How is this addressed

- Redesigned review experience as multi-step modal
- Implemented create-review-vm endpoint
- Display relevant resource state (created, not created, creating) 
- Added workspace scope option to componentManager hook so we can check for updates on resource that's in a different workspace to the current context (needed for import review Vms that reside in a different workspace)
- Changed create-review-vm api endpoint to return airlock request as well as create vm operation to reduce calls and latency on front-end and potential state mismatch

image

image

image

Note: this only supports single review VMs currently. A follow-up story adds support for multiple reviewers' Vm resources: #2764


api_app/_version.py
```diff
@@ -1 +1 @@
-__version__ = "0.4.54"
+__version__ = "0.4.57"
```
tamirkamara, line 1: Commenting here as there's no better place for this:
1. Can you update the changelog
2. [nit] I think it's best to use a vector image (and not a raster one). It should be simple enough to find an SVG version of it...

jjgriff93: Yep good shout thanks, have addressed both

damoodamoo (approved):
LGTM!

PR #2762, opened 2022-10-20
Pin porter components versions

### LINKED ISSUE #2761: New Porter Plugin/Mixin versions break the system
**Describe the bug**

The Porter Azure plugin has recently been updated with a breaking change - they removed support in Azure Tables for state storage.
This breaks the Resource Processor if its image is being rebuilt from scratch.

tamirkamara:
Resolves #2761 

## What is being addressed

We use "latest" for various mixins/plugins we install in porter. A recent version of azure plugin removed the ability to store state in Azure Tables which we're using.

## How is this addressed

- Pin version of mixins and plugins to specific ones
- [small] remove devcontainer extensions we don't need anymore


marrobi (approved):
LGTM. If needs further testing, let me know, otherwise when merged will deploy as soon as I can.

PR #2763, opened 2022-10-20
Add missing quote in .env.sample

stuartleeks:
## What is being addressed

Missing quote in value in `.env.sample`

## How is this addressed

- Add missing quote in value in `.env.sample`


damoodamoo (approved):

stuartleeks:
/test-force-approve

The only real change is to the `.env.sample` file

PR #2778, opened 2022-10-24
Add API ping endpoint

### LINKED ISSUE #2777: Resource Processor Not OK
**Describe the bug**

We have seen errors in the `make db-migrate` deployment step with the output from the API healthcheck showing the Resource Processor being unhealthy:

```json
{"services":[{"service":"Cosmos DB","status":"OK","message":""},{"service":"Service Bus","status":"OK","message":""},{"service":"Resource Processor","status":"Not OK","message":"Unspecified error"}]}
```

There is no additional information in the App Insights logs to help diagnose.

**Steps to reproduce**

This issue has surfaced in environments where we are making performing a high number of builds/deployments.
Running `curl <API>/api/health` in a bash loop will trigger the issue, and then it will be observed in a deployment to that environment

stuartleeks:
# Resolves #2777

## What is being addressed

## How is this addressed

- Add richer error logging on healthcheck tests to aid any future diagnosis of healthcheck issues
- Add simple ping endpoint (and make `/` an alias for this instead of the health check endpoint)
- Update App Gateway health probe to /api/ping. Since the App Gateway probe was ignoring the content of the health check response, it is only testing that it can reach the API. The new ping endpoint will allow achieving this without incurring the cost of the Azure Management calls


damoodamoo (approved):
lovely jubbly!

PR #2779, opened 2022-10-24
Restructure Airlock requests & add createdBy field

### LINKED ISSUE #2765: Airlock incorrectly assigns latest update user to initiator of request
When an airlock request is updated, the creator_user_id field is overwritten with the user that's just updated the request. This field should be untouched by subsequent updates after creation.

**Steps to reproduce**

1. Create an airlock request in draft
1. Sign in as a different user with appropriate permissions and submit the draft request
1. It will change the initiator to the latter user

jjgriff93:
Resolves #2765 

## What is being addressed

This PR fixes the querying of airlock requests by the initiator of the request by adding a createdBy field to the requests as well as reworking several fields to be more descriptive

## How is this addressed

- added a createdBy field to keep the original request creator value
- updated Api queries to use this field
- renamed some airlock properties to more descriptive and sensible values, such as user to updatedBy (both in the root and the history objects)
- De-coupled the airlock requests model in the UI from the base resource model to reflect the structure in the API models
- Added DB migration to fill in the createdBy value from the history if present and also rename the other fields to their new values
- Also corrected some things that have been bothering me such as consistent snake casing and camel casing where appropriate


api_app/db/repositories/airlock_requests.py
```diff
@@ -103,12 +103,13 @@ def create_airlock_request_item(self, airlock_request_input: AirlockRequestInCre
...
-        query = self.airlock_requests_query() + f' where c.workspaceId = "{workspace_id}"'
+    def get_airlock_requests(self, workspace_id: str, initiator_user_id: str = None, type: AirlockRequestType = None, status: AirlockRequestStatus = None, order_by: str = None, order_ascending=True) -> List[AirlockRequest]:
+        query = self.airlock_requests_query() + f' WHERE c.workspaceId = "{workspace_id}"'
 
         # optional filters
-        if user_id:
-            query += ' AND c.user.id=@user_id'
+        if initiator_user_id:
+            # If the resource has history, get the first user object (initiator)
+            query += ' AND (ARRAY_LENGTH(c.history) > 0? c.history[0].user.id=@user_id: c.user.id=@user_id)'
```
stuartleeks, line 112: Do we need to do this to get the user who initiated the request? Or is that the value that is stored in the top-level `user` property?

jjgriff93: So I've opted not to change the behaviour of adding the latest user to the root user object as this is what all other resources do in TRE and is how the history is populated, so instead I've adapted the Cosmos query to query for the first user in the history if it exists, or the root level if not

jjgriff93: Yeah I’m with you - currently its coupled with the state of the resource at that time, so anything in the root of a resource is the current state, and history records the state as it was alongside the user that modified it, so would be a bigger undertaking to change this. It may well be we should think about renaming some fields to make this more obvious in the very least. As per our Teams discussion shall resolve this and raise a discussion on our next stand-up

damoodamoo:
I understand the approach but if the longer term plan is to add an initiator field - shall we not just do that now? 

I'm also nervous about coupling any logic to the History array as we've previously discussed moving that out to a separate Cosmos collection, so this would have to be updated in that case.

damoodamoo:
Not sure why we'd need to update all Resource types - as the Airlock Request is it's own model? The shape of it is very similar to a `Resource` but it's defined separately and lives in a separate collection. Maybe we'd want to update resource types at some point but I don't think we'd need to do that in this work. What do you think about just adding the field we need to the airlock request model? 

For existing requests we _could_ add a migration which would move the top history item user to the initiator field - or we could not worry about it as I can't imagine there are many inflight requests that we need to worry about at this stage? @marrobi ?

jjgriff93:
> Not sure why we'd need to update all Resource types - as the Airlock Request is it's own model? The shape of it is very similar to a `Resource` but it's defined separately and lives in a separate collection. Maybe we'd want to update resource types at some point but I don't think we'd need to do that in this work. What do you think about just adding the field we need to the airlock request model?
> 
> For existing requests we _could_ add a migration which would move the top history item user to the initiator field - or we could not worry about it as I can't imagine there are many inflight requests that we need to worry about at this stage? @marrobi ?

It's an extension of the Resource model so inherits user from that - however discussed this on the standup this morning and have decided to go ahead and add the additional `initiator` field to the AirlockRequest model. Oxford haven't raised any concerns over any previous requests missing this field and I think we're early enough along for this not to be a problem - and like you said it then covers us for any changes to the history model at a later date. Also means later on should we decide to we can adopt the initiator field into the core resource model if we're moving the history out and want quick access to the person who created the resource, so this will act as a nice test case for this

jjgriff93:
Have reworked this as per our discussions:
- added a createdBy field to keep the original request creator value
- updated Api queries to use this field
- renamed some airlock properties to more descriptive and sensible values, such as user to updatedBy (both in the root and the history objects)
- De-coupled the airlock requests model in the UI from the base resource model to reflect the structure in the API models
- Added DB migration to fill in the createdBy value from the history if present and also rename the other fields to their new values
- Also corrected some things that have been bothering me such as consistent snake casing and camel casing where appropriate

api_app/models/domain/airlock_request.py
```diff
@@ -81,16 +81,17 @@ class AirlockRequest(AzureTREModel):
...
     resourceVersion: int = 0
-    user: dict = {}
+    createdBy: dict = {}
+    createdWhen: float = Field(None, title="Creation time of the request")
+    updatedBy: dict = {}
     updatedWhen: float = 0
     history: List[AirlockRequestHistoryItem] = []
     workspaceId: str = Field("", title="Workspace ID", description="Service target Workspace id")
-    requestType: AirlockRequestType = Field("", title="Airlock request type")
+    type: AirlockRequestType = Field("", title="Airlock request type")
```
stuartleeks, line 90: Does this affect the payload shape in the API? I.e. do callers now need to pass `type` instead of `requestType`?

jjgriff93: It does indeed - it changes from:
```
{
    "requestType": "import",
    "requestTitle": "a request title",
    "businessJustification": "some business justification"
}
```

to 

```
{
    "type": "import",
    "title": "a request title",
    "businessJustification": "some business justification"
}
```

api_app/db/migrations/airlock.py
```diff
@@ -0,0 +1,33 @@
...
+class AirlockMigration(AirlockRequestRepository):
+    def __init__(self, client: CosmosClient):
+        super().__init__(client)
+
+    def add_created_by_and_rename_in_history(self) -> int:
+        num_updated = 0
+        for request in self.container.read_all_items():
+            # Only migrate if createdBy isn't present
+            if 'createdBy' in request:
+                break
```
damoodamoo, line 14: think you want `continue` here to move and try the next request?

api_app/db/repositories/airlock_requests.py
```diff
@@ -85,42 +85,43 @@ def validate_status_update(self, current_status: AirlockRequestStatus, new_statu
...
         airlock_request = AirlockRequest(
             id=full_airlock_request_id,
             workspaceId=workspace_id,
-            requestTitle=airlock_request_input.requestTitle,
+            title=airlock_request_input.title,
             businessJustification=airlock_request_input.businessJustification,
-            requestType=airlock_request_input.requestType,
-            creationTime=datetime.utcnow().timestamp(),
+            type=airlock_request_input.type,
+            createdBy=user,
```
damoodamoo, line 99: do we need to add `updatedBy` here, as we then copy that into the history?

api_app/db/repositories/airlock_requests.py
```diff
@@ -31,18 +31,18 @@ def get_resource_base_spec_params():
     def get_timestamp(self) -> float:
         return datetime.utcnow().timestamp()
 
-    def update_airlock_request_item(self, original_request: AirlockRequest, new_request: AirlockRequest, user: User, request_properties: dict) -> AirlockRequest:
+    def update_airlock_request_item(self, original_request: AirlockRequest, new_request: AirlockRequest, updated_by: User, request_properties: dict) -> AirlockRequest:
         history_item = AirlockRequestHistoryItem(
             resourceVersion=original_request.resourceVersion,
             updatedWhen=original_request.updatedWhen,
-            user=original_request.user,
+            updatedBy=original_request.updatedBy,
```
damoodamoo, line 38: as per other comment - looks like the first time this creates the History Item, the updatedBy will be empty? unless i've missed something.

damoodamoo (approved):
looks great + a much better shape of the object :)

api_app/models/domain/airlock_request.py
stuartleeks: Ok, I think we should call this out in breaking changes in the CHANGELOG.

Also, can you update the CLI as this change will break it

jjgriff93: Yes good shout, will do :)

jjgriff93: And already updated the CLI ;)

jjgriff93: Changelog updated

api_app/db/migrations/airlock.py
jjgriff93: Spot on, thank you - updated

api_app/db/repositories/airlock_requests.py
jjgriff93: Again great spot you're absolutely right. Fixed!

api_app/db/repositories/airlock_requests.py
jjgriff93: Fixed, thanks!

Issue #2620, opened 2022-09-20
A researcher shouldn't be able to access operations on workspace-service


anatbal:
**Describe the bug**
A researcher cannot do any actions on workspace-service but query it, but still can access operations, while TRE User cannot access operation on core API	 

Endpoints | Researcher | Workspace Owner
-- | -- | --
GET /workspaces/{workspace_id}/workspace-services/{service_id}/operations | V | V
GET /workspaces/{workspace_id}/workspace-services/{service_id}/operations/{operation_id} | V | V

Issue #2619, opened 2022-09-20
 A researcher shouldn't be able to PATCH workspace-service


anatbal:
**Describe the bug**
 A researcher cannot create a workspace-service but can PATCH it, which he shouldn't be able to do.

Endpoints | Researcher | Workspace Owner
-- | -- | -- 
POST /workspaces/{workspace_id}/workspace-services | X | V
PATCH /workspaces/{workspace_id}/workspace-services/{service_id} | V | V

Issue #2618, opened 2022-09-20
Airlock manager doesn't have permission to get a specific airlock request 


anatbal:
**Describe the bug**
Airlock manager can get all requests but cannot get a single airlock request .

Endpoints | Researcher | Workspace Owner | Airlock Manager
|-- | -- | -- | --
|GET /workspaces/{workspace_id}/requests | V | V | V|
|GET /workspaces/{workspace_id}/requests/{airlock_request_id}  | V | V | X|

PR #2781, opened 2022-10-25
Fix API permissions bugs 

### LINKED ISSUE #2621: TRE User should be able to get workspace-templates
**Describe the bug**
TRE User cannot query workspace-templates but can query workspace-service-template, user-resource-template, shared-service-templates.

Endpoints | TRE Admin | TRE User 
-- | -- | -- |
GET /workspace-templates | V | **_X_** |   |  
GET /workspace-templates/{workspace_template_name} | V | **_X_** |   |  
GET /workspace-service-templates | V | V
GET /workspace-service-templates/{workspace_service_template_name} | V | V
GET /workspace-service-templates/{service_template_name}/user-resource-templates | V | V
GET /workspace-service-templates/{service_template_name}/user-resource-templates/{user_resource_template_name} | V | V

anatbal:
# Fixes #2621 #2620 #2619 #2618

## What is being addressed

After the permissions discussion that was conducted, there was an agreement on a few bugs regarding the existing permissions today. This PR handles those minor changes.


api_app/api/routes/airlock.py
```diff
@@ -68,7 +68,7 @@ async def get_all_airlock_requests_by_workspace(
     return AirlockRequestWithAllowedUserActionsInList(airlockRequests=airlock_requests_with_allowed_user_actions)
 
 
-@airlock_workspace_router.get("/workspaces/{workspace_id}/requests/{airlock_request_id}", status_code=status.HTTP_200_OK, response_model=AirlockRequestInResponse, name=strings.API_GET_AIRLOCK_REQUEST, dependencies=[Depends(get_current_workspace_owner_or_researcher_user), Depends(get_workspace_by_id_from_path)])
```
tanya-borisova, line 71: @jjgriff93  this would be the endpoint that the Airlock Manager uses to review Airlock requests? Feels like a pretty major bug, wondering why we didn't run into this, although maybe we just were always Workspace Owners in our testing environments :/

jjgriff93: Yes I think it's because we just added the airlock manager role to our main Microsoft identities, so we would have both been workspace managers

api_app/api/routes/workspaces.py
```diff
@@ -302,12 +303,12 @@ async def invoke_action_on_workspace_service(response: Response, action: str, us
 
 
 # workspace service operations
-@workspace_services_workspace_router.get("/workspaces/{workspace_id}/workspace-services/{service_id}/operations", response_model=OperationInList, name=strings.API_GET_RESOURCE_OPERATIONS, dependencies=[Depends(get_current_workspace_owner_or_researcher_user_or_airlock_manager), Depends(get_workspace_by_id_from_path)])
```
tanya-borisova, line 305: I'm wondering if this would break anything on the UI side

tanya-borisova (approved):

Issue #2708, opened 2022-10-06
Automated validation of tre CLI


stuartleeks:
## Description

As an Azure TRE Developer
I want to have automated validation of the `tre` CLI
So that changes to the API that affect the CLI are caught early

## Acceptance criteria

- [ ] Key areas of the CLI are validated via automated testing

PR #2784, opened 2022-10-26
CLI: Add missing commands and bug fixes

### LINKED ISSUE #2706: Ensure full API coverage in the tre CLI

stuartleeks:
# Resolves #2706 #2707

## What is being addressed

## How is this addressed

- Add shell completion for workspace id in get-token
- Add --decode option to get-token command
- Add template registration commands
- Update register bundle script to use the CLI (additional step towards #2708 )
- Fix default table query
- Add shared-service update/set-enabled
- Update ACR login step in cli-package.yml

### DIFF
--- a/cli/tre/commands/shared_services/shared_service.py
+++ b/cli/tre/commands/shared_services/shared_service.py
@@ -156,7 +156,7 @@ def shared_service_set_enabled(shared_service_context: SharedServiceContext, eta
               flag_value=True,
               default=False)
 @click.option('--ensure-disabled',
-              help="Disable before deleting if not currently enabled",
+              help="Disable before deleting if not currently disabled (required to delete)",
               flag_value=True,
               default=False)
 @click.pass_context
--- a/cli/tre/commands/shared_services/shared_service.py
+++ b/cli/tre/commands/shared_services/shared_service.py
@@ -156,7 +156,7 @@ def shared_service_set_enabled(shared_service_context: SharedServiceContext, eta
               flag_value=True,
               default=False)
 @click.option('--ensure-disabled',
-              help="Disable before deleting if not currently disabled (required to delete)",
+              help="Ensure disabled before deleting (resources are required to be disabled before deleting)",
               flag_value=True,
               default=False)
 @click.pass_context
--- a/cli/tre/commands/shared_services/shared_service.py
+++ b/cli/tre/commands/shared_services/shared_service.py
@@ -1,3 +1,4 @@
+import json
 import click
 import logging
 
@@ -42,8 +43,6 @@ def shared_service_show(shared_service_context: SharedServiceContext, output_for
     output(response, output_format=output_format, query=query, default_table_query=r"sharedService.{id:id,name:templateName, version:templateVersion, is_enabled:isEnabled, status: deploymentStatus}")
 
 
-# TODO - add PATCH (and ?set-enabled)
-
 @click.command(name="invoke-action", help="Invoke an action on a shared_service")
 @click.argument('action-name', required=True)
 @click.option('--no-wait',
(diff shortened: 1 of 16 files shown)


cli/tre/commands/shared_services/shared_service.py
```diff
@@ -74,16 +73,97 @@ def shared_service_invoke_action(shared_service_context: SharedServiceContext, c
...
+        operation_url = response.headers['location']
+        operation_show(log, operation_url, no_wait=False, output_format=output_format, query=query, suppress_output=suppress_output)
+
+
 @click.command(name="delete", help="Delete a shared_service")
 @click.option('--yes', is_flag=True, default=False)
 @click.option('--no-wait',
               flag_value=True,
               default=False)
+@click.option('--ensure-disabled',
```
tanya-borisova, line 158: Nit: I kind of think that if I didn't know what it does and hadn't look at the code, I would have thought that an --ensure-disabled would check that the resource is disabled, and raise if it isn't. I wouldn't think that it would do the disabling
But yeah it's a nit so up to you if you act on it!

devops/scripts/register_bundle_with_api.sh
```diff
@@ -109,64 +100,64 @@ explain_json=$(porter explain --reference "${acr_name}".azurecr.io/"$(yq eval '.
...
+get_result=$(get_template)
+if [[ -n "$(echo "$get_result" | jq -r .id)" ]]; then
+  # 'id' was returned - so we successfully got the template from the API. Now check the version
+  if [[ "$(echo "$get_result" | jq -r .version)" == "$template_version" ]]; then
+    echo "Template with this version already exists"
+    exit
+  fi
+else
+  error_code=$(echo "$get_result" | jq -r .status_code)
+  if [[  "$error_code" = "404" ]]; then
```
tanya-borisova, line 140: We could check that error_code is not 200?  To catch any other potential errors (500?)

tanya-borisova (approved):
Looks great! Only have two very minor comments

devops/scripts/register_bundle_with_api.sh
stuartleeks: Ah - actually.... it should be `!= 404`.

If it's a 404 then we didn't find the template and want to continue and install it. Otherwise, we should error.  Have pushed an update

devops/scripts/register_bundle_with_api.sh
```diff
@@ -109,64 +100,64 @@ explain_json=$(porter explain --reference "${acr_name}".azurecr.io/"$(yq eval '.
...
+if [[ -n "$(echo "$get_result" | jq -r .id)" ]]; then
+  # 'id' was returned - so we successfully got the template from the API. Now check the version
+  if [[ "$(echo "$get_result" | jq -r .version)" == "$template_version" ]]; then
+    echo "Template with this version already exists"
+    exit
+  fi
+else
+  error_code=$(echo "$get_result" | jq -r .status_code)
+  if [[  "$error_code" != "404" ]]; then
+    echo "Error checking for existing template: $get_result"
```
tanya-borisova, line 141: Nit: leave a comment explaining why 404 is fine (I mean it might be almost obvious but I think a comment would still be useful)

stuartleeks:
/test-force-approve

just added comment to bash script

devops/scripts/register_bundle_with_api.sh
stuartleeks: Added 😄

tanya-borisova (approved):

Issue #2783, opened 2022-10-25
Document importing/exporting data with Airlock for Researchers


jjgriff93:
Adding a section on "Importing and exporting data" in https://microsoft.github.io/AzureTRE/using-tre/tre-for-research/ which would be a user-facing doc describing the process of importing and exporting data from a researcher's point of view

This should include guidance on
- [x] Creating a draft request
- [x] Generating a SAS and examples of different methods to add blob (CLI & Storage Explorer)
- [x] Submitting the request
- [x] Viewing status of a request
- [x] Accessing the output file

PR #2786, opened 2022-10-26
Document mechanism for reviewing airlock requests

### LINKED ISSUE #2665: Document mechanism for reviewing Airlock requests
Provide documentation on how to set up and customise the Airlock review process in a TRE

tanya-borisova:
# Resolves #2665 

## What is being addressed

Add user docs for Airlock Review feature.

## How is this addressed

- [x] Document the process for reviewing a request
- [x] Add docs for all related templates
- [x] Document configuring Airlock Review
- Also updated some fields in Base Workspace template to include the word Review in them


mkdocs.yml
```diff
@@ -72,6 +72,8 @@ nav:
       - Using AzureTRE for Research:
           - Introduction: using-tre/tre-for-research/index.md
           - Set up of a Virtual Machine: using-tre/tre-for-research/using-vms.md
+      - Using AzureTRE as an Airlock Manager:
+          - Reviewing Airlock Requests: using-tre/tre-for-airlock-reviewer/review-airlock-request.md
```
tamirkamara, line 76: I think that should be another doc in the **Research** category.

tanya-borisova: My thinking was to separate this section by roles

tamirkamara: We anyone have very few docs and a group with just 1 doc isn't needed IMO. It's not like the researcher group has much info anyone.
I'd opt for a simplified approach in the docs as well.

tanya-borisova: We do have more docs coming in the researcher group as James is working on adding docs for using Airlock for researchers: https://github.com/microsoft/AzureTRE/issues/2783 
Also if we keep the structure then existing links won't break.

The new Reviewer group isn't likely to get more docs though I think, so I could move it to tre-for-research folder. The only reason I don't do it is because I wanted to have roles separation, as I think customer was also saying that it would be useful to have docs for researchers in a single place.

marrobi: I don't think we need the `Using AzureTRE as an Airlock Manager` heading. You call out to do reviews you need to be an airlock manager at the top of the doc.

It might be in time we need to split it down, but not sure needed at the moment.

tanya-borisova: @marrobi  Fair enough, so do you think it should go to "Using TRE for research" section? If anything goes into this section, do we need it at all?

tanya-borisova: I've moved the new file into "Using TRE for research", @marrobi  PTAL

marrobi (approved):
LGTM. Thanks!

tanya-borisova:
/test-force-approve

As this is a docs PR.

PR #2802, opened 2022-10-31
GET specific template version

### LINKED ISSUE #2803: Unable to get specific version of a template
Using the API it's currently not possible to retrieve a specific template version. This is needed for resource updates, else the UI can only pull the latest template version - yet the resource is validated against the previous version. It can introduce scenarios where the resource is 'stuck' and unable to be updated.

damoodamoo:
Closes #2803 

Part of work needed to support #2795 

Makes it possible for clients to call the API to retrieve specific versions of the templates, rather than just the latest version.


tanya-borisova (approved):

api_app/_version.py
```diff
@@ -1 +1 @@
-__version__ = "0.5.9"
```
tanya-borisova, line 1: Don't forget the changelog :)

damoodamoo: gaaaaaaaaaaaaaaa David!

PR #2806, opened 2022-11-01
Make number of Resource Processor Processes Configurable

damoodamoo:
The number of processes in the RP is currently configurable in a TF Var, and defaulted to 2. It was not configurable in the build, and not clear to a dev user how/where to change that default. 

This PR adds the boilerplate to the `.env.sample` and adds the `RESOURCE_PROCESSOR_NUMBER_PROCESSES_PER_INSTANCE` as a var in the build pipelines, to get the value from the build secret.


.github/actions/devcontainer_run_command/action.yml
```diff
@@ -174,5 +177,6 @@ runs:
             && inputs.CORE_APP_SERVICE_PLAN_SKU) || 'P1v2' }}" \
           -e WORKSPACE_APP_SERVICE_PLAN_SKU="${{ (inputs.WORKSPACE_APP_SERVICE_PLAN_SKU != ''
             && inputs.WORKSPACE_APP_SERVICE_PLAN_SKU) || 'P1v2' }}" \
+          -e TF_VAR_resource_processor_number_processes_per_instance="${{ inputs.RESOURCE_PROCESSOR_NUMBER_PROCESSES_PER_INSTANCE }} \
```
tamirkamara, line 180: You probably need to specifiy a default here as well, otherwise you must always create the github secret (which isn't something we want...)

docs/tre-admins/environment-variables.md
```diff
@@ -36,3 +36,4 @@
 | `RESOURCE_PROCESSOR_VMSS_SKU` | The SKU of the VMMS to use for the resource processing VM. |
 | `CORE_APP_SERVICE_PLAN_SKU` | The SKU of AppService plans created for the core infrastructure. |
 | `WORKSPACE_APP_SERVICE_PLAN_SKU` | Optional. The SKU used for AppService plan used in E2E tests unless otherwise specified. Default value is `P1v2`. |
+| `RESOURCE_PROCESSOR_NUMBER_PROCESSES_PER_INSTANCE` | Optional. The number of processes to create in the Resource Processor. Equates to the number of parallel deployment operations possible in your TRE. Defaults to `2`. |
```
tamirkamara, line 39: You should also explain this in the doc explaining how to setup the workflow/pipeline. It's about to be merged by Liza.

templates/core/.env.sample
```diff
@@ -64,3 +64,6 @@ ENABLE_AIRLOCK_MALWARE_SCANNING=false
 
 # TODO: move to RP default with https://github.com/microsoft/AzureTRE/pull/2634
 WORKSPACE_APP_SERVICE_PLAN_SKU="P1v2"
+
+# The number of processes to start in the resource processor VMSS image
+RESOURCE_PROCESSOR_NUMBER_PROCESSES_PER_INSTANCE=5
```
tamirkamara, line 69: If we recommend 5 here, why do we have a default of 2?

templates/core/version.txt
```diff
@@ -1 +1 @@
-__version__ = "0.4.38"
+__version__ = "0.4.39"
```
tamirkamara, line 1: Can you please change the type of the TF variable to int?

tamirkamara (changes requested):

templates/core/.env.sample
damoodamoo: fair - have changed default to 5 as well

PR #2809, opened 2022-11-02
Update base workspace to allow multiple address spaces

### LINKED ISSUE #1846: Workspace Service needs address space for additional Vnet
## Description

As a TRE Service Integrator 
I want to get the next available address space when deploying a workspace service
So that I can create a new virtual network with address space that does nor overlap

## Acceptance criteria

- [ ] Can deploy a workspace service with parameter address_space

marrobi:
# Resolves #1846

## What is being addressed

Add additional address_spaces property that contains an array or address spaces. This means that we can add additional address spaces in future to a workspace vnet.

## How is this addressed

- Add additional address_spaces property in the api when creating a workspace, left address_space for backwards compatibility
- Update base workspace template to take multiple address spaces for a vnet
- Added missing env load into make file for terraform and bundle install
- [ ] Update CHANGELOG.md if needed

Future Work
- [ ] Update unrestricted template to take `address_spaces`


marrobi:
Need to looking for existing address spaces as well as address space when getting a new address space

marrobi:
Blocked until updating of workspaces is reenabled.

api_app/db/repositories/workspaces.py
```diff
@@ -67,14 +67,18 @@ def create_workspace_item(self, workspace_input: WorkspaceInCreate, auth_info: d
 
         template = self.validate_input_against_template(workspace_input.templateName, workspace_input, ResourceType.Workspace, user_roles)
 
+        # allow for workspace template taking a single address_space or multiple address_spaces
         address_space_param = {"address_space": self.get_address_space_based_on_size(workspace_input.properties)}
+        address_spaces_param = {"address_spaces": [self.get_address_space_based_on_size(workspace_input.properties)]}
```
tamirkamara, line 72: I think we need to be clear what address_spaces mean.
We expect a primary one in address_space, right?
If so, maybe we can name the new field additional_address_spaces?

tamirkamara, line 72: Will this always validate that the value provided in the new field is free?
get_address_space_based_on_size is doing that internally, but only if address_space_size=custom. What will happen if the primary field will get assigned automatically (e.g. size=small)?

api_app/db/repositories/workspaces.py
```diff
@@ -124,7 +128,10 @@ def validate_address_space(self, address_space):
         return is_network_available(allocated_networks, address_space)
 
     def get_new_address_space(self, cidr_netmask: int = 24):
-        networks = [x.properties["address_space"] for x in self.get_workspaces()]
+        workspaces = self.get_workspaces()
+        networks = [[x.properties.get("address_space")] for x in workspaces]
+        networks = networks + [x.properties.get("address_spaces", []) for x in workspaces]
+        networks = [i for s in networks for i in s if i is not None]
```
tamirkamara, line 134: What does this line do?

api_app/db/repositories/workspaces.py
marrobi: It's actually all the address spaces, including the primary. Reason being that's the name of the property in terraform is a single one and needs all of them. Otherwise would have to add `address_space` to `additional_address_spaces` in the terraform, and assign that to `address_space`...

The property is actually called `address_space` - https://learn.microsoft.com/en-us/rest/api/virtualnetwork/virtual-networks/create-or-update?tabs=HTTP#addressspace . I didn't want to break existing functionality so left the existing property as is.

Thoughts?

marrobi: This looks wrong, will take another look, it should be:

```
        address_spaces_param = {"address_spaces": [address_space_param]}
```

tamirkamara: so this is either-or?

api_app/db/repositories/workspaces.py
marrobi: flattens the `address_spaces` property, joins with `address_space`. So have a list of used address spaces so can find a free address space.

api_app/db/repositories/workspaces.py
marrobi: Yes, so the "new" bundle I have just takes in `address_spaces`, it could also use `address_space` I guess, but I have indexed the first as `address_spaces[0]`

api_app/db/repositories/workspaces.py
tamirkamara: I might be mistaken but a list + a list in python is still a list. 
So why is this line required?

api_app/db/repositories/workspaces.py
```diff
@@ -67,14 +67,18 @@ def create_workspace_item(self, workspace_input: WorkspaceInCreate, auth_info: d
 
         template = self.validate_input_against_template(workspace_input.templateName, workspace_input, ResourceType.Workspace, user_roles)
 
+        # allow for workspace template taking a single address_space or multiple address_spaces
         address_space_param = {"address_space": self.get_address_space_based_on_size(workspace_input.properties)}
+        address_spaces_param = {"address_spaces": [address_space_param]}
```
tamirkamara, line 72: I'm missing the plan here...
You're adding a new field to the database workspace object but it's value is the existing/single address space we have been using.

marrobi: When a workspace service is deployed that needs an address space, an additional address will be added to the `address_spaces` array. The PR will follow this one, but includes:

```
    # if template has address_space get an address space
    if resource_template.properties.get("address_space"):
        # check workspace has address_spaces property
        if not workspace.properties.get("address_spaces"):
            raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=strings.WORKSPACE_DOES_NOT_HAVE_ADDRESS_SPACES_PROPERTY)
        workspace_service.properties["address_space"] = workspace_repo.get_address_space_based_on_size(workspace_service_input.properties)
        workspace_patch = ResourcePatch()
        workspace_patch.properties = {"address_spaces": workspace.properties["address_spaces"] + [workspace_service.properties["address_space"]]}
        workspace_repo.patch_workspace(workspace, workspace_patch, etag, resource_template_repo, user)

```

api_app/db/repositories/workspaces.py
marrobi: @tamirkamara which line specifically? 133?

tamirkamara: I mean that it looks like this is enough
```python
        workspaces = self.get_workspaces()
        networks = [[x.properties.get("address_space")] for x in workspaces]
        networks = networks + [x.properties.get("address_spaces", []) for x in workspaces]
```

marrobi: It flattens the lists, without that line the list of existing address spaces has multiple dimensions.

As in its a list address_space and address_spaces, where each address_spaces entry is actually another list. It needs to be flattened to be passed to the function that finds an unused address space.

api_app/db/repositories/workspaces.py
tamirkamara: Ok, I understand. 
Just a question to consider for when this extra functionality will be introduced - do you ensure the workspace_service operation wait for the workspace update one? 
(you can resolve this after reading)

marrobi: I'm handling the workspace upgrade in the workspace service pipeline. Let's discuss it in the next PR.

tamirkamara (approved):

PR #2811, opened 2022-11-02
Fix nexus bootstrapping key server error

### LINKED ISSUE #2785: Nexus fails to deploy with cloud-init failure (Failed to import key from key server)
**Describe the bug**
I deployed the most recent release (v0.6.0) of AzureTRE from the AzureTRE-Deployment repo. Any Linux VM - I've tested both the available images - I create in a workspace is unreachable via Guacamole. Attempts to connect are simply rejected with the message "The remote desktop server is currently unreachable". The Linux VMs are definitely up and running. In contrast, Windows VMs function as expected - no issues connecting.

**Steps to reproduce**

1. Create any Linux VM in a workspace
2. Attempt to connect to it through Guacamole

jjgriff93:
# Resolves #2785 

## What is being addressed

Due to the introduction of template pipelines and changing the firewall rule addition/removal step to occur outside the main bundle's terraform, a race condition was introduced where the Nexus bundle was being deployed and then the firewall rules (that it relied on to connect to ubuntu key server, packages.microsoft.com and docker.com) were being applied afterwards in the pipeline. As the Nexus clouding bootstrapping starts up, this firewall step wasn't always completed in time for it to work, causing regular failures.

## How is this addressed

I tried swapping the order so that the firewall pipeline step happens first, however it depends on outputs from the main bundle terraform. After experimenting with a few workarounds the cleanest solution seems to be whitelisting the key server, Microsoft packages and docker fqdns in the main shared-subnet fqdn exceptions as part of the firewall bundle, as these will be whitelisted anyway as per the nexus bundle and are core trusted repositories already whitelisted by the resource processor.


damoodamoo:
I think this looks great and is a really low-impact way of solving the problem - but @marrobi should probably be the one to hit the approve button

templates/shared_services/firewall/terraform/firewall.tf
```diff
@@ -101,7 +101,10 @@ resource "azurerm_firewall_application_rule_collection" "shared_subnet" {
       "graph.microsoft.com",
       "login.microsoftonline.com",
       "aadcdn.msftauth.net",
-      "graph.windows.net"
+      "graph.windows.net",
```
tamirkamara, line 104: This is beginning to look like a catch all list and we loose the ability to know which domain is required for what.
I think it makes more sense to create a rule specifically for Nexus (targeting the shared subnet as a whole). Please discuss and resolve this comment per the conclusion in the team.
cc: @marrobi

jjgriff93: Good idea, have amended

marrobi (approved):
LGTM - can you add an issue referencing this as a case for needing to be able to handle pre steps prior to terraform having run. So using inputs, rather than outputs?

PR #2814, opened 2022-11-03
Configurable number of processors in RP

damoodamoo:
THIRD attempt at new branch/PR to get a fresh CI build.


.github/actions/devcontainer_run_command/action.yml
```diff
@@ -103,6 +103,9 @@ inputs:
     description: "The AppService plan sku used by tests"
     required: false
     default: ""
+  RESOURCE_PROCESSOR_NUMBER_PROCESSES_PER_INSTANCE:
```
tanya-borisova, line 106: Could also name it RESOURCE_PROCESSOR_PARALLELISM to make it slightly less of a mouthful

templates/core/terraform/variables.tf
```diff
@@ -60,8 +60,8 @@ variable "terraform_state_container_name" {
 }
 
 variable "resource_processor_number_processes_per_instance" {
-  type        = string
-  default     = "2"
+  type        = number
+  default     = 5
```
tanya-borisova, line 64: Why 5? :)  Seems odd not to have power of 2

tanya-borisova (commented):
If you've tested this locally then I would be fine with getting this in with force-approve, given how long it takes to get the tests to work :(

.github/actions/devcontainer_run_command/action.yml
damoodamoo: yes, could have... but I think the long name is ok and also explains what it does :)

templates/core/terraform/variables.tf
damoodamoo: no particular reason - and given the 'host' process also exists this would make the number of python processes an even number

tanya-borisova (approved):

PR #2813, opened 2022-11-02
Import/export docs

jjgriff93:
# Resolves #2783

## What is being addressed

Add documentation for researchers with steps on importing and exporting data using airlock.

PR #2817, opened 2022-11-03
Add preface to airlock docs

jjgriff93:
# Resolves #2813 

Adds preface to the researcher import/export data docs that makes it clear it pertains to Airlock and that it is intended for the ad-hoc use case as per @marrobi's request.


marrobi (approved):
Thanks.

Issue #2820, opened 2022-11-04
Singleton resource


tanya-borisova:
## Description

As a TRE Administrator, 

I would like to be able to deploy resources that are singletons. 
For example, Nexus, Firewall and Gitea are currently deployed as singletons (because they are shared services).

To extend the singleton concept to resources beyond shared services, and allow to deploy multiple instances for certain shared services, introduce a concept of a singleton template.

Template could have a flag "singleton" that, when set to True, would indicate that there can't be more than one resource of this template deployed.

PR #2828, opened 2022-11-07
Add extra dns zones and links

### LINKED ISSUE #2827: Include extra private DNS zones in core
## Description

As a  TRE Developer
I want to be able to resolve synapse/purview/hds services
So that they can be accessed from within a TRE workspace

Our current approach of adding these zones in our terraform comes with the issue that when running a `destroy` with a workspace deployed the zones created cannot be removed because of dependent links. One possibility is to have a flag like `var.include_extra_dns_zones`, as to not include them if they're not required

## Acceptance criteria

- [ ] Extra DNS zones are managed in the TRE terraform state

cc: @marrobi (feel free to modify)

t-young31:
# Resolves #2827 

## How is this addressed

- Adds extra DNS zones + links to core network


t-young31:
@microsoft-github-policy-service agree [company="UCL"]

templates/core/terraform/network/dns_zones.tf
```diff
@@ -257,6 +247,150 @@ resource "azurerm_private_dns_zone_virtual_network_link" "eventgridlink" {
   lifecycle { ignore_changes = [tags] }
 }
 
+resource "azurerm_private_dns_zone" "purview_account" {
```
tamirkamara, line 250: Instead of copy/pasting all these new sections... have you considered using the for_each method already present at the bottom of this file?

tamirkamara (commented):
DNS zones are taking more time in the deployment, so I just want to know if other options have been considered... Depending how the azure services are planned to be deployed it might be possible to use shared services for this?

templates/core/terraform/network/dns_zones.tf
t-young31: I hadn't! that is much nicer – thanks

t-young31:
> Depending how the azure services are planned to be deployed it might be possible to use shared services for this?

good question – not sure. will need @marrobi 's input

marrobi:
> > Depending how the azure services are planned to be deployed it might be possible to use shared services for this?
> 
> good question – not sure. will need @marrobi 's input

Each one could be done as a separate shared service, or could do something similar to what I've done with External URI, single shared service, can be deployed multiple time with different DNS zone name. or even an array? PR is open, and @tanya-borisova is looking at how to handle singleton services as part of #2820 .

The other option is we have an environment variable with a list of all the DNS zones passed into the core deployment. 

@tamirkamara thoughts? A single DNS zones shared service with array is maybe a nice solution for these "optional" DNS zones.

marrobi:
Also worth noting @t-young31 has followed the approach used for AML, so maybe merge this and then take the "improved" approach, as a new issue?

marrobi:
One more thing, if it was a single shared service with an array, a pipeline in the workspace could add the DNs zones as per the same method used for firewall rules.

tamirkamara:
> A single DNS zones shared service with array is maybe a nice solution for these "optional" DNS zones.
Yes, something like this. Another approach to the lack of state for the core system, is to add a shared service exactly for that, something like common state (store) and have the pipeline procedure interact with it. It seems a bit "hacky" but will work...

tamirkamara (approved):

PR #2842, opened 2022-11-08
E2E tests: Extract setting up test workspaces

### LINKED ISSUE #2839: Create one set of test workspaces for all E2E tests
Tests for Airlock and tests for Guacamole user resource can reuse the same workspace. Any future tests can also reuse that workspace, unless they need a specific workspace (such as the test that creates AAD app registrations).

tanya-borisova:
# Resolves #2839 

## What is being addressed

Reuse test workspaces between tests via introducing `setup_test_workspace` fixture.

## How is this addressed

- Added fixtures `setup_test_workspace` and `setup_test_aad_workspace` 
- Added TEST_WORKSPACE_ID and TEST_WORKSPACE_AAD_ID env variables to e2e tests to make it easier to iterate on tests locally 
- Simplify the `test_create_guacamole_service_into_aad_workspace` test: only check that the workspace was successfully deployed and can be returned from the API


e2e_tests/test_workspace_services.py
```diff
@@ -63,76 +44,21 @@ async def test_create_guacamole_service_into_base_workspace(verify) -> None:
...
-
-    # patch the guac service. we'll just update the display_name but this will still force a full deployment run
-    # and essentially terraform no-op
-    patch_payload = {
-        "properties": {
-            "display_name": "Updated Guac Name",
-        }
-    }
-
-    await post_resource(patch_payload, f'/api{workspace_service_path}', workspace_owner_token, verify, method="PATCH")
```
tanya-borisova, line 117: I don't think doing this is necessary for this test, as if AAD creation doesn't work, it should fail at the step of workspace creation

tamirkamara: Are you referring to the reason why it should be enough to just test the workspace creation step without any services etc?

tanya-borisova: Yes

tamirkamara: I think it does - auto_create impacts the workspace_app settings and that is used by guacamole. 
If something doesn't work right there, the ping should fail (in theory at least). We can skip on VM but I think deploying/pinging Guacamole is still valuable.

tanya-borisova: Ah okay, good to know. I'll add Guacamole back, and add a ping for Guacamole

tamirkamara (approved):

tanya-borisova:
/test-force-approve

A successful run on an identical branch:

https://github.com/microsoft/AzureTRE/actions/runs/3438476485

PR #2854, opened 2022-11-09
Redacting nested or conditional properties

### LINKED ISSUE #2836: Secret properties of workspace are no longer being redacted

**Describe the bug**

Properties of workspace such as client_secret are no longer being redacted. 

This is due to the fact that they used to be under "properties" field, but are no longer there as per https://github.com/microsoft/AzureTRE/pull/2805 , and are instead on a level above. 

The secret redaction code should be changed to account for this fact, and the [test](https://github.com/microsoft/AzureTRE/blob/main/api_app/tests_ma/test_api/test_routes/test_resource_helpers.py#L278) should be updated too. 

**Steps to reproduce**

1. Deploy a workspace with client_secret set 
1. Secret does not show "REDACTED" on the UI: 

image

damoodamoo:
# Resolves #2836 

## What is being addressed
Following the move on the base workspace to conditional properties via `allOf`, the client secret was no longer being redacted as it was assumed to be in the main template `properties` and the input property was assumed to be a flat list with no nesting. This fixes that and supports conditional properties in the template as well as nested properties.


api_app/api/routes/resource_helpers.py
```diff
@@ -40,26 +55,68 @@ async def save_and_deploy_resource(resource: Resource, resource_repo: ResourceRe
...
+                for p in prop:
+                    if isinstance(p, dict):
+                        flatten_template_props(p)
+
+            if isinstance(prop, dict) and prop_name != "if":
+                flatten_template_props(prop)
+
+    flatten_template_props(template.dict())
+
+    def recurse_input_props(prop_dict: dict):
```
tanya-borisova, line 102: nit: maybe name it `recurse_mask_sensitive_props`? Or something else that hints at sensitive properties being redacted in this method

tanya-borisova (approved):
LGTM % one naming comment :)

api_app/api/routes/resource_helpers.py
damoodamoo: it doesn't just recurse sensitive props though, it recurses _all_ props, and just redacts the sensitive ones

api_app/_version.py
```diff
@@ -1 +1 @@
-__version__ = "0.5.12"
+__version__ = "0.5.13"
```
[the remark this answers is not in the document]
damoodamoo: there are actually only a handful of formatting changes where long lines have been sent to multiple lines - that's also from the VS Code "Format Document" command, so should be consistent with other developers

api_app/api/routes/resource_helpers.py
damoodamoo: (it's also an inner method, inside the `mask_sensitive_props` method

api_app/_version.py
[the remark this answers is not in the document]
damoodamoo: think we've talked about changing the linter to check for line lengths already - let's raise again in a standup

damoodamoo:
killing this to try again

PR #2865, opened 2022-11-11
Update mlflow and fix terraform issues

### LINKED ISSUE #2863: ML Flow bundle fails with undefined `tre_workspace_service_tags`
When deploying ML flow it fails with an error due to undefined tags.

marrobi:
# Resolves #2863

## How is this addressed

- Update ml flow version
- Fix terraform issues
- [x] Update CHANGELOG.md


guybartal:
some resources which were in the files you've updated are missing tags, for example
image

tamirkamara (approved):

PR #2871, opened 2022-11-14
E2E: Simplify smoke and extended tests

tanya-borisova:

## Description TBA


e2e_tests/test_shared_services.py
```diff
@@ -91,6 +92,7 @@ async def test_patch_firewall(verify):
 ]
 
 
+@pytest.mark.extended
```
tamirkamara, line 95: this currently fails 100% of the time on the main branch and is very likely to fail on repeat runs at some point in PR environments.

e2e_tests/test_shared_services.py
```diff
@@ -9,6 +9,7 @@
 LOGGER = logging.getLogger(__name__)
 
 
+@pytest.mark.extended
```
tamirkamara, line 12: Since almost every other service does an internal update on the firewall, do we even need it?

.devcontainer/devcontainer.json
```diff
@@ -89,36 +89,6 @@
             "false"
           ]
         },
-        {
-          "name": "E2E Extended AAD",
```
tamirkamara, line 93: How does one debug a specific test?

.github/workflows/pr_comment_bot.yml
```diff
@@ -135,9 +135,7 @@ jobs:
     needs: [pr_comment]
     if: |
       needs.pr_comment.outputs.command == 'run-tests' ||
-      needs.pr_comment.outputs.command == 'run-tests-extended' ||
-      needs.pr_comment.outputs.command == 'run-tests-extended-aad' ||
-      needs.pr_comment.outputs.command == 'run-tests-shared-services'
+      needs.pr_comment.outputs.command == 'run-tests-extended'
```
tamirkamara, line 138: Shouldn't the command be still available for cases where developers know what they want to test?
For example, if I change a shared service being tested I will call the run-shared-services without any of the workspace level stuff?

tamirkamara (changes requested):
Requesting changes as I don't want this PR merged before we can address my main concern: running more tests on PRs is very likely to make them harder to merge given the instability we see in Azure provisioning resources.

.github/workflows/pr_comment_bot.yml
damoodamoo: sort of agree - however now that workspaces + services can update those shared services, and do, there's also a case to be made that we should be running all the tests as we need to test the pipeline update from a workspace -> shared service?

tanya-borisova: > Shouldn't the command be still available for cases where developers know what they want to test?

I think a more future-proof is would be to run a workflow with a custom selector. This way we can easily extend selectors without having to update PR bot each time. 

>  there's also a case to be made that we should be running all the tests as we need to test the pipeline update from a workspace -> shared service?

Well other bundles only really update firewall, and that will be tested by deploying said bundles. I don't think we would need to test shared services each time

.devcontainer/devcontainer.json
tanya-borisova: You couldn't do that before either, could you? There is still a target for E2E Extended tests, there's just no separate target for AAD.

e2e_tests/test_shared_services.py
tanya-borisova: Good point, I think we can remove this one

tanya-borisova:
@tamirkamara 

> running more tests on PRs is very likely to make them harder to merge given the instability we see in Azure provisioning resources

This doesn't add any more tests that are required for a PR to pass though. Only running smoke tests is required. Given that running `/ test-extended` is optional anyway, and we can run a custom subset of tests using a selector, is this a big concern?

tamirkamara:
> This doesn't add any more tests that are required for a PR to pass though. Only running smoke tests is required. Given that running `/ test-extended` is optional anyway, and we can run a custom subset of tests using a selector, is this a big concern?

Granted - the `test` one hasn't changed much but I meant the `extended` one which is our "baseline" for the system to be functional within the PR validation. 
Now `extended` becomes everything: gitea, firewall patch (and nexus, etc. in the future), two workspace flavors, guacamole (times 2), 2 types of VMs (linux and windows). 
The question here is do we really want to do that? When we discussed this last (some months ago) we agreed to keep a balance between what runs in PR and what runs in a nightly test.
At any case, this is going to cause much more frictions to get PRs to merge and I'm not sure it's a good idea for now.

As for the custom selector, it's not relevant for many PRs as the bot doesn't (and shouldn't) expose that functionality. You must have a branch in this repo to make use of it.
Also, using a selector doesn't make the process "simpler" IMO.

tanya-borisova:
@tamirkamara  I agree that a balance should be kept, my intention was mostly bringing all these additional things to extended set of tests and only mandate to run it nightly. 
Is there maybe a case of bringing some tests that were previously in `extended` test into `smoke`, to really make it, as you say, baseline for the system to be functional. 

Alternatively we could create a new selector, call it `baseline` and only add tests there that are absolutely essential for the PR to pass. (Though wouldn't there be overlap with `smoke`, would we need both?) 

Let's discuss on the morning call tomorrow?

e2e_tests/test_shared_services.py
tanya-borisova: Yeah it still seems to be failing with the same problem "secret being recovered and can't be deleted". I will mark this test as skipped for now.

PR #2883, opened 2022-11-15
Airlock UI: support multiple review vms

### LINKED ISSUE #2764: Support multiple review resources in Airlock Review VM UI
As a follow on from #2738 we should support multiple review user resources in the UI to create a screen like this: https://www.figma.com/file/Ncl8TJsFf2WKf3frrqTbtq/AzureTRE-OSS-UI?node-id=1293%3A27999

This will involve modifying the reviewUserResources property of the airlock request to include the reviewerId in the form of a dictionary:

```json
"reviewUserResources": {
    "<user_object_id>": {
        "workspaceId": "XXX",
        "workspaceServiceId": "XXX",
        "userResourceId": "XXX"
    }
}
```

This will allow the UI to then filter for the signed-in user id and also list the other ids to show who else is reviewing.

The API will also need to be modified to only allow one review VM per user. On /Create-review-resource, the API should:
- [x] Check if the current signed in user has a resource already in the reviewUserResources dict
- [x] If so, do a call to get the status of the resource. If it is running and healthy, return 409 Conflict
- [x] If not, delete it and create a new one, then replace the userResource details in the request
- [x] Update UI

jjgriff93:
# Resolves #2764

## What is being addressed

This PR adds support in the API and UI for multiple reviewers being able to create review VMs for an airlock request. This also adds functionality for re-deploying a review resource if it's in a failed state and automatic clean-up of the previous resource.

image

## How is this addressed

- Changed reviewUserResources from array to dictionary, using the signed-in user's userId as the key to ensure one user resource per user
- Added migration to update from previous format
- Expanded /create-review-resource API to do a check for an existing resource, in which case it checks its health, returns a 409 conflict if healthy, otherwise redeploys and removes the old failed resource
- Also made some general improvements to the airlock experience - allowed actions are now returned with most airlock API calls meaning the user doesn't need to refresh the UI for some options to appear when a request is updated
- Review user resource block shows loading shimmers/spinner instead of defaulting to not created view when looking up an existing resource
- Updated to React 18 and updated dependencies to latest versions
- Update CHANGELOG.md


api_app/api/routes/airlock.py
```diff
@@ -164,13 +201,14 @@ async def create_review_user_resource(
         templateName=user_resource_template_name,
         properties={
             "display_name": "Airlock Review VM",
-            "description": f"Airlock Review VM for request {airlock_request.title} (ID {airlock_request.id})",
```
tanya-borisova, line 167: Maybe keep the "Airlock Review VM" part there? Just so it's immediately obvious to eg TRE Admin, which VMs are Airlock Review VMs

api_app/api/routes/airlock_resource_helpers.py
```diff
@@ -73,8 +74,9 @@ async def update_and_publish_event_airlock_request(
     except Exception as e:
         logging.error(f'Failed updating airlock_request item {airlock_request}: {e}')
         # If the validation failed, the error was not related to the saving itself
-        if e.status_code == 400:
-            raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=strings.AIRLOCK_REQUEST_ILLEGAL_STATUS_CHANGE)
+        if hasattr(e, 'status_code'):
```
tanya-borisova, line 77: Good catch!

CHANGELOG.md
```diff
@@ -24,6 +27,7 @@ COMPONENTS:
 }
 ```
 * Fields in AirlockNotification event have changed without backward compatibility. If Airlock Notifier shared service is deployed, it needs to be re-deployed. Any other consumers of AirlockNotification event need to be updated. For more details, see [#2798](https://github.com/microsoft/AzureTRE/pull/2798)
+* The model for `reviewUserResources` in airlock requests has changed from being a list to a dictionary. A migration has been added to update your existing requests automatically; please make sure you run the migrations as part of updating your API and UI. [#2883](https://github.com/microsoft/AzureTRE/pull/2883)
```
tanya-borisova, line 30: Should we mention that the in-flight requests will display an "UNKNOWN" user? (This is what will happen, right?)

api_app/api/routes/airlock_resource_helpers.py
```diff
@@ -140,6 +142,35 @@ def enrich_requests_with_allowed_actions(requests: List[AirlockRequest], user: U
     return enriched_requests
 
 
+async def delete_review_resource(
```
tanya-borisova, line 145: Nit: maybe name it `delete_one_review_user_resource` for parity with `delete_review_user_resources`?

tanya-borisova (approved):
The API part looks great! 
Left some small nits / comments but up to you whether to apply them

api_app/api/routes/airlock.py
jjgriff93: I removed this because it's already in the display_name, so that should be how they'd quickly identify? Then means they can quickly glance at the description for the request ID

api_app/api/routes/airlock_resource_helpers.py
jjgriff93: I'll rename it to `delete_review_user_resource`, then rename the other method to `delete_all_review_user_resources` to make it clear

jjgriff93:
/test-force-approve (no e2e changes wince test pass: https://github.com/microsoft/AzureTRE/actions/runs/3575859656)

PR #2889, opened 2022-11-17
Remove docker mixin

### LINKED ISSUE #2771: Re-examine if we need the Docker Mixin
The Docker Mixin (porter) is only installed in the Innereye bundle, but doesn't seem to be used.
If we indeed don't need it we should remove it from:
1. porter.yaml of the Innereye bundle
2. devcontainer
3. Resource Processor image

tamirkamara:
# Resolves #2771 

## What is being addressed

We don't need the docker mixin anymore.


marrobi (approved):
LGTM

tamirkamara:
/test-force-approve
changes after a successful test-extended run are only docs.

PR #2887, opened 2022-11-16
handle cost management API throttling

guybartal:
# Resolves #2835 

## What is being addressed

handle 429 and 503 status codes returned from azure cost management when rate limiting is applied and when service is temporarily unavailable. 

## How is this addressed

- propagate HTTP status codes 429, 503 from azure cost management api to client side with retry-after X seconds 
- when TRE cost api returns 429 or 503, UI will retry to fetch results after given number of seconds from the response 
- add Shimmer effect while loading costs in UI

Loading cost (Shimmer):

MicrosoftTeams-image

When loading cost fails -  cost tag is hidden (In the future we might display in notification pane):
MicrosoftTeams-image (1)

When loaded:
MicrosoftTeams-image (2)

PR #2894, opened 2022-11-20
cache cost results

guybartal:
# Related to: #2887 

## What is being addressed

Reduces probability for Azure cost management throttling issue by caching result,
caching will improve performance by returning faster response.

## How is this addressed

- cache Azure cost management result if not exists yet in cache for two hours
- clear expired cache items
- store ```CostService``` in ```FastAPI``` application state ```on_start``` event to create a singleton instance


api_app/api/routes/costs.py
```diff
@@ -83,3 +107,20 @@ async def workspace_costs(workspace_id: UUID4, params: CostsQueryParams = Depend
...
+    except ServiceUnavailable as e:
+        return JSONResponse(content={
+                            "error": {
+                                "code": "503",
+                                "message": strings.API_GET_COSTS_SERVICE_UNAVAILABLE,
+                                "retry-after": str(e.retry_after)
+                            }}, status_code=503, headers={"Retry-After": str(e.retry_after)})
+    except Exception as e:
+        logging.error("Failed to query Azure TRE workspace costs", exc_info=e)
+        raise e
```
LizaShak, line 126: Why raise and not return 500?

guybartal: you are correct, on the router I should have just return a 500 and not rethrow the exception.
I'll fix that. thanks

api_app/api/routes/costs.py
```diff
@@ -15,7 +16,8 @@
 from models.domain.costs import CostReport, GranularityEnum, WorkspaceCostReport
 from resources import strings
 from services.authentication import get_current_admin_user, get_current_workspace_owner_or_tre_admin
-from services.cost_service import CostService, SubscriptionNotSupported, WorkspaceDoesNotExist
+from services.cost_service import CostService, ServiceUnavailable, SubscriptionNotSupported, TooManyRequests, WorkspaceDoesNotExist
+from starlette.responses import JSONResponse
```
martinpeck, line 20: Why are you importing from starlette?
Generally, FastAPI has it's own types (based upon the starlette ones) that you should use, rather than importing the underlying starlette type. In this case, the one in fastapi.responses.

api_app/models/schemas/costs.py
```diff
@@ -30,6 +30,34 @@ def get_cost_report_responses():
...
             }
+        },
+        429: {
+            "description": "Too Many Requests",
+            "content": {
+                "application/json": {
+                    "example": {
+                        "error": {
+                            "code": "429",
+                            "message": "Too many requests to Azure cost management api. Please retry.",
```
martinpeck, line 41: nit: uppercase `api` to `API`

api_app/models/schemas/costs.py
```diff
@@ -30,6 +30,34 @@ def get_cost_report_responses():
...
+            }
+        },
+        503: {
+            "description": "Service Unavailable",
+            "content": {
+                "application/json": {
+                    "example": {
+                        "error": {
+                            "code": "503",
+                            "message": "Azure cost management api is temporarly unavaiable. Please retry.",
```
martinpeck, line 55: nit: uppercase `api` to `API`

api_app/models/schemas/costs.py
```diff
@@ -62,5 +90,33 @@ def get_workspace_cost_report_responses():
...
             }
+        },
+        429: {
+            "description": "Too Many Requests",
+            "content": {
+                "application/json": {
+                    "example": {
+                        "error": {
+                            "code": "429",
+                            "message": "Too many requests to Azure cost management api. Please retry.",
```
martinpeck, line 101: nit: uppercase `api` to `API`

api_app/models/schemas/costs.py
```diff
@@ -62,5 +90,33 @@ def get_workspace_cost_report_responses():
...
+            }
+        },
+        503: {
+            "description": "Service Unavailable",
+            "content": {
+                "application/json": {
+                    "example": {
+                        "error": {
+                            "code": "503",
+                            "message": "Azure cost management api is temporarly unavaiable. Please retry.",
```
martinpeck, line 115: nit: uppercase `api` to `API`

api_app/resources/strings.py
```diff
@@ -72,6 +72,9 @@
 API_GET_COSTS_TO_DATE_NEED_TO_BE_LATER_THEN_FROM_DATE = "to_date needs to be later than from_date"
 API_GET_COSTS_FROM_DATE_NEED_TO_BE_BEFORE_TO_DATE = "from_date needs to be before to_date"
 API_GET_COSTS_SUBSCRIPTION_NOT_SUPPORTED = "Azure subscription doesn't support cost management"
+API_GET_COSTS_TOO_MANY_REQUESTS = "Too many requests to Azure cost management api. Please retry."
```
martinpeck, line 75: nit: uppercase `api` to `API`

martinpeck: (following line as well)

api_app/services/cost_service.py
```diff
@@ -45,29 +46,106 @@ class SubscriptionNotSupported(Exception):
...
+    result: QueryResult
+    ttl: datetime
+
+    def __init__(self, item: QueryResult, ttl: datetime) -> None:
+        self.result = item
+        self.ttl = ttl
+
+
+# make sure CostService is singleton
+@lru_cache(maxsize=None)
```
martinpeck, line 78: I've generally seen lru_cache used with a class factory (i.e. a method that returns an instance) rather than on a class definition. 

In fact, maxsize=None generally means it has an ever-growing cache.

Do you have something I can read that explains how this makes it a singleton?

martinpeck: Ah...ok...so does this work because the __init__ has no parameters?

guybartal: exactly, it works because there are no parameters.
at the beginning I've instantiated it on the on_start event of the app and stored it inside the app.state,
but then I saw this implementation of lru_cache and I've changed the code,
I'll search for class factory example, probably that are nicer ways to implement that in python with less potential bugs.

api_app/api/routes/costs.py
guybartal: you are correct I didn't notice that, I saw we use those types elsewhere in the code inside: ```/api_app/api/errors/generic_error.py``` ```starlette.responses.PlainTextResponse```

I'll replace this type to FastAPI's types.

api_app/services/cost_service.py
martinpeck: A simple example showing both approaches might be...

```python
from functools import lru_cache
import datetime

@lru_cache(maxsize=None)
class Test:

    the_time = None
    
    def __init__(self) -> None:
        self.the_time = datetime.datetime.now()
    
    def GetCounter(self) -> int:
        return self.the_time

@lru_cache(maxsize=None)
def test_factory() -> Test:
    return Test() 
```

With the factory approach it's a little more obvious how the caching is working.

api_app/api/routes/costs.py
martinpeck: Yeah....these types slip in to our code from time to time, but it's generally not needed. FastAPI exposes types that wrap these things, and it's safer (for us) to try and stick to using the FastAPI types.

api_app/services/cost_service.py
guybartal: I've added a factory method, and it works, can you please take a look?

martinpeck: Looks good. I suspect you can then remove the LRU from the class definition, but it's fine to leave it there as well I guess.

martinpeck (approved):

api_app/api/routes/costs.py
```diff
@@ -117,5 +117,5 @@ async def workspace_costs(workspace_id: UUID4, params: CostsQueryParams = Depend
                                 "retry-after": str(e.retry_after)
                             }}, status_code=503, headers={"Retry-After": str(e.retry_after)})
     except Exception as e:
-        logging.error("Failed to query Azure TRE workspace costs", exc_info=e)
-        raise e
+        logging.error("Failed to query Azure TRE costs", exc_info=e)
+        raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail=strings.API_GET_COSTS_INTERNAL_SERVER_ERROR)
```
LizaShak, line 121: Sorry for changing, but we saw in all the other places 503 code is used...

PR #2902, opened 2022-11-23
Add ability for workspace service to request an address space

### LINKED ISSUE #2810: modify workspace service item creation to get an address space if specified

marrobi:
# Resolves #2810

## What is being addressed
- If workspace service template requires an address_space then add one to the workspace
- Fix error in prior PR (test was failing) as was adding a property to a property, rather than just its value
- [x] Update documentation
- [x] Update CHANGELOG.md

## Note:
I have done the logic bit in the route - not sure if this is the best place but was easiest for me to implement.

The update of the workspace itself is down to the workspace service bundle such as:

```
  "pipeline": {
    "install": [
      {
        "stepId": "12ba0dad-ea6c-4d0d-9255-d316212f5ffa",
        "stepTitle": "Upgrade to ensure aware of address space",
        "resourceType": "workspace",
        "resourceAction": "upgrade",
        "properties": [
        ]
      },
      {
        "stepId": "main"
      },
```


api_app/api/routes/workspaces.py
```diff
@@ -229,6 +229,16 @@ async def create_workspace_service(response: Response, workspace_service_input:
         logging.error(f"User not authorized to use template: {e}")
         raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail=str(e))
 
+    # if template has address_space get an address space
+    if resource_template.properties.get("address_space"):
+        # check workspace has address_spaces property
+        if not workspace.properties.get("address_spaces"):
+            raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=strings.WORKSPACE_DOES_NOT_HAVE_ADDRESS_SPACES_PROPERTY)
```
tamirkamara, line 236: Is this really about the current request?
Since we expect every workspace to have an address space anyway should it be an HTTP 500?

api_app/schemas/workspace.json
```diff
@@ -6,8 +6,7 @@
   "description": "These parameters are required for all workspaces",
   "required": [
     "display_name",
-    "description",
-    "address_space_size"
```
tamirkamara, line 10: Don't we want a size prop in every workspace request?

api_app/api/routes/workspaces.py
```diff
@@ -229,6 +229,16 @@ async def create_workspace_service(response: Response, workspace_service_input:
...
+        workspace_service.properties["address_space"] = workspace_repo.get_address_space_based_on_size(workspace_service_input.properties)
+        workspace_patch = ResourcePatch()
+        workspace_patch.properties = {"address_spaces": workspace.properties["address_spaces"] + [workspace_service.properties["address_space"]]}
+        workspace_repo.patch_workspace(workspace, workspace_patch, etag, resource_template_repo, user)
```
tamirkamara, line 240: I think you should explain why you go directly to the repo and bypass the operation (i.e. why not send to RP).

It's confusing since now the API made the decision that the workspace template only "cares" about the first address space in the list which is not generic. Can't it be that a workspace will need all the spaces?
Not sure it's a good thing to do...

api_app/api/routes/workspaces.py
marrobi: I think it is the current request, what we are doing is saying if this is a workspace without an `address_spaces` (not `address_space`) property it cannot succeed. At present there are TREs deployed with workspaces, and workspace templates without the `address_spaces` property.

api_app/schemas/workspace.json
marrobi: It's set in the required section of the workspace specific `tempalte_schema.json`, which is where the `address_space_size` property also sits. Having it here means if there is no address space size in the workspace specific JSON you get an error. 

I can put this back, as doesn't directly affect this PR, but the required should be in the same file as the property itself?

api_app/api/routes/workspaces.py
marrobi: I've added the workspace changes to this PR.

marrobi:
Have an issue with etags on this, works through swagger, not the UI.

marrobi:
Sorted the etag issue.

api_app/api/routes/workspaces.py
marrobi: @tamirkamara added a comment, is that sufficient?

api_app/schemas/workspace.json
marrobi: Also it defaults to small within the api if not included in the request.

api_app/_version.py
```diff
@@ -1 +1 @@
-__version__ = "0.6.4"
+__version__ = "0.6.5"
```
tamirkamara, line 1:
```suggestion
__version__ = "0.7.0"
```

templates/workspaces/base/terraform/network/providers.tf
```diff
@@ -5,5 +5,9 @@ terraform {
       source  = "hashicorp/azurerm"
       version = ">= 3.8.0"
     }
+    null = {
+      source  = "hashicorp/null"
+      version = "=3.2.0"
```
tamirkamara, line 10: You're missing a lock file update in this PR. Also:
```suggestion
      version = "=3.2.1"
```

templates/workspaces/base/terraform/network/security.tf
```diff
@@ -53,25 +53,25 @@ resource "azurerm_network_security_rule" "deny_all_inbound_override" {
   source_port_range           = "*"
 }
 
-resource "azurerm_network_security_rule" "allow_inbound_within_services_subnet" {
+resource "azurerm_network_security_rule" "allow_inbound_within_workspace_vnet" {
```
tamirkamara, line 56: This opens up the whole "workspace" address space compared with just the "services" subnet of the workspace network. Has this been evaluated for the implications on the Airlock Review where I remember a discussion on which VM can access what?

templates/workspaces/base/terraform/network/network.tf
```diff
@@ -75,6 +75,57 @@ resource "azurerm_subnet_route_table_association" "rt_services_subnet_associatio
...
+  provisioner "local-exec" {
+    command = "az login --identity -u '${data.azurerm_client_config.current.client_id}'"
+  }
+
+  triggers = {
+    timestamp = timestamp()
+  }
+}
+
+resource "null_resource" "ws_core_peer_sync" {
```
tamirkamara, line 105: Why is it important to be done in TF and not via a porter step?

marrobi: See: https://github.com/hashicorp/terraform-provider-azurerm/issues/18444

Did it here as only wants to execute when the address spaces change.

templates/workspaces/base/terraform/network/security.tf
marrobi: The VMs themselves for export have a MV NSG blocking access to other than the rules defined in the VM NSG. so this shouldn't affect this.

templates/workspaces/base/terraform/network/network.tf
tamirkamara: I see. Why did you choose the scripted way vs the _replace_triggered_by_ option?

marrobi: As this way avoids any downtime for existing services - the peer isn't disconnected.

tamirkamara (approved):

tamirkamara (changes requested):
Something isn't right and I'm not sure where it came from: https://github.com/microsoft/AzureTRE/actions/runs/3734344062

So blocking this PR for now.

marrobi:
@tamirkamara not sure a need to block if the tests fail?

I can see:
> linkedstorageaccounts.LinkedStorageAccountsClient#CreateOrUpdate: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="Bad Request" Message="Failed to upsert linked storage account due to invalid property storageAccountIds due to storage accounts [/subscriptions/*******/resourceGroups/rg-***-ws-cd9b/providers/Microsoft.Storage/storageAccounts/stai***wscd9b] which might be considered faulted due to mismatch between workspace region *** and some storage accounts regions, or due to the fact some of the listed storage accounts are not assigned as bypass for AzureService

tamirkamara:
@marrobi, the issue is being address here: https://github.com/microsoft/AzureTRE/pull/2978

tamirkamara:
Still doesn't work: https://github.com/microsoft/AzureTRE/actions/runs/3738252473/jobs/6344598367

tamirkamara (approved):

anatbal:
This extended e2e passed on this run https://github.com/microsoft/AzureTRE/actions/runs/3775797437
was not automatically merged since there was a conflict with api version

Issue #2913, opened 2022-11-27
API sends secret as REDACTED to the RP on install action


tamirkamara:
**Describe the bug**
The API will send a secret as REDACTED on the initial deployment and that shouldn't happen as the RP should get the actual value on the first run to persist in KV.

**Steps to reproduce**
1. Submit a new workspace request with a manual auth setup - provide the id & secret
1. Wait for the deployment to finish and check the workspace keyvault where you will see a single version for the secret and its value will be REDACTED

tamirkamara:
I think this is the second time we seen an issue around this procedure.
We need to make sure we have a UT that validates: 
1. Secret is persisted as REDACTED to db
2. Secret is sent in the clear to the queue

PR #2938, opened 2022-12-04
Disable base workspace upgrade

tamirkamara:
## What is being addressed

This [issue](https://github.com/microsoft/AzureTRE/issues/2913) prevents a normal upgrade flow for the base workspace so we need to disable it for the time being.

### DIFF
--- a/templates/workspaces/base/porter.yaml
+++ b/templates/workspaces/base/porter.yaml
@@ -1,6 +1,6 @@
 ---
 name: tre-workspace-base
-version: 0.7.3
+version: 0.7.4
 description: "A base Azure TRE workspace"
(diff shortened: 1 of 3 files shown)


LizaShak (approved):

marrobi:
@tamirkamara @LizaShak this will break Guacamole as it relies on the upgrade step to add redirect URIs.

How hard is it to fix the Redaction issue?

tamirkamara:
It's hard and @damoodamoo is fixing it.
In the meantime, I'll bring back the step calling the script directly.

tamirkamara:
I now see you deleted the update redirect urls script in that workspace upgrade PR which I don't get (and I approved it). Anyway, I'll work on reverting it.

PR #2940, opened 2022-12-05
Destroy script support for azcli 2.42

### LINKED ISSUE #2939: Environment cleanup in CI fails to remove certain resources

**Describe the bug**
Running `devops/scripts/destroy_env_no_terraform.sh` in Github Actions uses Azure CLI 2.42.0 while our devcontainer used 2.37.0. In one of the version between those there's a change in how the output of `az monitor diagnostic-settings list` is returned and now it's a plain array in the root of the object, no `value` property is present.
Hence, this line will not do anything which is causing duplicate resources errors when we try to recreate a CI environment:
`az monitor diagnostic-settings list --resource "$1" --query "value[].name" -o tsv 2> /dev/null |`

tamirkamara:
# Resolves #2939 

## What is being addressed

Diagnostic settings were not being found and deleted when running the cleanup procedure in Github Actions due to Azure CLI output format change.

## How is this addressed

- Include another query to support the new format too. I tried to call `az` once and use `jq` for the 2 queries but could find a good expression that could handle errors (missing keys) it procedues.
- Add a few prints as we're seeing some issues with KeyVault objects too (still unclear why)

### DIFF
--- a/devops/scripts/destroy_env_no_terraform.sh
+++ b/devops/scripts/destroy_env_no_terraform.sh
@@ -84,7 +84,9 @@ fi
 
 delete_resource_diagnostic() {
   # the command will return an error if the resource doesn't support this setting, so need to suppress it.
-  az monitor diagnostic-settings list --resource "$1" --query "value[].name" -o tsv 2> /dev/null |
+  # first line works on azcli 2.37, second line works on azcli 2.42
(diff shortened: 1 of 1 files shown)


tamirkamara:
/test-force-approve
cleanup isn't tested in e2e

anatbal (approved):

PR #2942, opened 2022-12-06
Auto upgrade firewall

tamirkamara:
## What is being addressed

Since we now support upgrading the template version of deployed resources we should also have the makefile/cicd support upgrading the firewall currently installed in the system.

## How is this addressed

- `devops/scripts/deploy_shared_service.sh` to call `update`

### DIFF
--- a/devops/scripts/deploy_shared_service.sh
+++ b/devops/scripts/deploy_shared_service.sh
@@ -51,6 +51,7 @@ fi
 deployed_shared_service=$(echo "${get_shared_services_result}" \
   | jq -r ".sharedServices[] | select(.templateName == \"${template_name}\" and (.deploymentStatus != \"deleted\" or .deploymentStatus != \"deployment_failed\"))")
 
+is_update=0
 if [[ -n "${deployed_shared_service}" ]]; then
   # Get template version of the service already deployed
   deployed_version=$(echo "${deployed_shared_service}" | jq -r ".templateVersion")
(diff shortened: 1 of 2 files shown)


devops/scripts/deploy_shared_service.sh
```diff
@@ -84,9 +86,19 @@ if ! deploy_result=$(cat << EOF | tre shared-services new --definition-file -
...
-  exit 1
+  ); then
+    echo "Failed to deploy shared service:"
+    echo "${deploy_result}"
+    exit 1
+  fi
+
+else
+
+  echo "An older version is already deloyed. Trying to update..."
```
LizaShak, line 97: typo in deployed

anatbal (approved):

devops/scripts/deploy_shared_service.sh
```diff
@@ -59,22 +60,23 @@ if [[ -n "${deployed_shared_service}" ]]; then
...
+    value=${property_values[$index]}
+    additional_props="$additional_props, \"$name\": \"$value\""
+  done
 
-echo "Not currently deployed - deploying..."
-display_name="${template_name#tre-shared-service-}"
-if ! deploy_result=$(cat << EOF | tre shared-services new --definition-file -
+  echo "Not currently deployed - deploying..."
+  display_name="${template_name#tre-shared-service-}"
+  if ! deploy_result=$(cat << EOF | tre shared-services new --definition-file -
```
LizaShak, line 79: I think the indent here is incorrect

guybartal (approved):
LGTM

LizaShak (commented):
Minor comments

devops/scripts/deploy_shared_service.sh
tamirkamara: I understand why you think that but due to the EOF is has to be like that.

PR #2944, opened 2022-12-06
Environment files consolidation

### LINKED ISSUE #1470: ENV Files Need Consolidating and Refactoring

LizaShak:
## What is being addressed

Fixes #1470 
Setup all configuration in one config.yaml file 

## How is this addressed

- New config.yaml is used for loading the configuration
- This config file will replaces the .env files we had at /templates/core and /devops
- The auth setup adds its configuration under the authentication setup in the new config.yaml file
- private.env file remaisn the same as it is right now. api and e2e also keep their .env files and will get their config as part of the devcontainer setup (as it is now) from private.env and the config.yaml
-  Added a json schema which validates the config prior running the make commands
- load env will be changed and load the env vars from the config.yaml


marrobi:
@chboudry would welcome your feedback on this.

devops/scripts/env_to_yaml_config.sh
```diff
@@ -0,0 +1,34 @@
...
+cp config.sample.yaml config.yaml
+# Loop over the relevant lines in the file specified in $1 (passed in after the loop)
+# The loop source filters the lines in the source file to those that should be treated
+# as variable definitions
+for f in "devops/auth.env" "devops/.env" "templates/core/.env"
+do
+  while read -r line
+  do
+    # split the line into name/value
+    name=$(echo "$line" | cut -d= -f1| tr '[:upper:]' '[:lower:]')
```
martinpeck, line 20: how do you skip/ignore commented lines?

devops/scripts/load_and_validate_env.sh
```diff
@@ -0,0 +1,42 @@
...
+    if [ -n "${has_dupes:-}" ]; then
+      echo -e "\e[31m»»» 💥 There are duplicate keys in your config, please fix and try again!\e[0m"
+      exit
+    fi
+
+    # Validate config schema
+    pajv validate -s config_schema.json -d config.yaml
+
+    # declare the variable and export to the caller's context
+    STRIP_COMMENTS="... comments=\"\""
```
martinpeck, line 28: this is very similar to the code in another script
is there a way to package this up so that we don't duplicated this code?

docs/tre-admins/auth.md
```diff
@@ -22,9 +22,9 @@ You can build all of the Identity assets by running the following at the command
 ```bash
 make auth
 ```
-This will create five identities, and if successful will write a new file; `/devops/auth.env`. If you are building locally, these values will be used when building your TRE. If you are setting this up for CI/CD, then these values will be needed by your Build Orchestrator.
+This will create five identities, and if successful will write the outputs to athentication section in `config.yaml` file. If you are building locally, these values will be used when building your TRE. If you are setting this up for CI/CD, then these values will be needed by your Build Orchestrator.
 
-The contents of your `/devops/auth.env` file should contain :
+The contents of your athentication section in `config.yaml` file should contain :
```
martinpeck, line 27: typo `athentication` should be `authentication`

martinpeck (commented):
Generally looks good!
I've made a few comments / questions.

config.sample.yaml
```diff
@@ -0,0 +1,87 @@
...
+  # public_deployment_ip_address: __CHANGE_ME__
+  defaults:
+    core_address_space: "10.1.0.0/22"
+    tre_address_space: "10.0.0.0/12"
+
+    core_app_service_plan_sku: "P1v2"
+    resource_processor_vmss_sku: "Standard_B2s"
+    enable_airlock_malware_scanning: false
+
+  # TODO: move to RP default with https://github.com/microsoft/AzureTRE/pull/2634
```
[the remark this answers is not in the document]
LizaShak: Not sure what it means, checking with @tamirkamara

devops/scripts/aad/create_api_application.sh
```diff
@@ -352,11 +352,10 @@ JSON
   fi
 fi
 
-{
-  echo "API_CLIENT_ID=\"${appId}\""
-  echo "API_CLIENT_SECRET=\"${spPassword}\""
-  echo "SWAGGER_UI_CLIENT_ID=\"${uxAppId}\""
-} >> "devops/auth.env"
+# Set outputs in configuration file
+yq -i ".authentication.api_client_id |= \"${appId}\"" config.yaml
```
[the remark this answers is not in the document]
LizaShak: Sure will add

devops/scripts/check_dependencies.sh
```diff
@@ -20,36 +20,17 @@ if ! command -v az &> /dev/null; then
...
-fi
 
 if [[ "${1:-?}" == *"env"* ]]; then
   if [ -z "${USE_ENV_VARS_NOT_FILES:-}" ]; then
     # We only do this for local builds
     echo -e "\n\e[96mLoading local environment variables\e[0m..."
-    if [ ! -f "devops/.env" ]; then
-      echo -e "\e[31m»»» ⚠️ Your devops environment has not been setup! 😥 Please create a ./devops/.env file."
+    if [ ! -f "config.yaml" ]; then
+      echo -e "\e[31m»»» ⚠️ Your config.yaml file has not been setup! 😥 Please create a config.yaml file."
```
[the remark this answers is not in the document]
LizaShak: Not sure at what stage the user is here, if it is his first time or not. Maybe deleted it accidently, or didn't do the migration, or it is his first time ever. We didn't provide any guidance before, but I agree maybe we should... In my opinion provide a link here to relevant doc. WDYT?

devops/scripts/env_to_yaml_config.sh
LizaShak: at line 30 we source the file without the comments

docs/tre-admins/auth.md
LizaShak: Fixed, I think we should install a spellchecker in our devcontainer. Maybe create an issue for that and add it... WDYT?

martinpeck: yeah...maybe....I'm not a big fan of spell checkers as you end up fighting with them a lot when spellchecking code

.devcontainer/Dockerfile
```diff
@@ -88,6 +88,9 @@ ARG YQ_VERSION="v4.27.2"
 RUN curl -L --fail -o /usr/local/bin/yq "https://github.com/mikefarah/yq/releases/download/${YQ_VERSION}/yq_linux_amd64" \
     && chmod +x /usr/local/bin/yq
 
+ARG PAJV_VERSION="1.2.0"
+RUN npm install -g pajv@${PAJV_VERSION}
```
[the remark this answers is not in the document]
LizaShak: It is part of the suggested command line tools:  https://json-schema.org/implementations.html
others didn't support yaml, or were for windows

docs/tre-admins/auth.md
LizaShak: You convinced me right away :)

config.sample.yaml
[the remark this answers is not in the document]
LizaShak: Created new issue: https://github.com/microsoft/AzureTRE/issues/2948 updated the comment

config.sample.yaml
```diff
@@ -0,0 +1,92 @@
+---
+# After you have replaced all the __CHANGE_ME__ values in this file, you need to
+# run `make auth` to setup the Auth.
+management:
+  location: __CHANGE_ME__
```
tamirkamara, line 5: should be common - top level

config.sample.yaml
```diff
@@ -0,0 +1,92 @@
...
+  # Azure Resource Manager credentials used for CI/CD pipelines
+  arm_subscription_id: __CHANGE_ME__
+
+  # If you want to override the currently signed in credentials
+  # You would do this if running commands like `make terraform-install DIR=./templates/workspaces/base`
+  # arm_tenant_id: __CHANGE_ME__
+  # arm_client_id: __CHANGE_ME__
+  # arm_client_secret: __CHANGE_ME__
+
+tre:
```
tamirkamara, line 19: Offline I said I think we don't need this subsection and can move all children to be under the root.

config.sample.yaml
```diff
@@ -0,0 +1,92 @@
...
+  # stateful_resources_locked: false
+
+# If you want to use TRE_URL to point to your local TRE API instance,
+# uncomment and set this variable
+#   tre_url: __CHANGE_ME__
+
+# This setting will enable your local machine to be able to
+# communicate with Service Bus and Cosmos. It will also allow deploying
+# the base workspace.
+  enable_local_debugging: false
```
tamirkamara, line 84: false is the default behavior, shouldn't this be true and defined as a comment

config.sample.yaml
```diff
@@ -0,0 +1,92 @@
...
+
+  # Used by the API and Resource processor application to change log level
+  debug: false
```
tamirkamara, line 87: false is the default behavior, shouldn't this be true and defined as a comment

config.sample.yaml
```diff
@@ -0,0 +1,92 @@
...
+
+# To avoid creating + deleting a workspace + service in tests set the vars for existing ones
+performance_in_e2e_test:
```
tamirkamara, line 90: the e2e folder (like many others) have its own .env file. What makes these settings be defined here where all the rest are defined in the specific .env file?

.devcontainer/scripts/consolidate_env.sh
```diff
@@ -0,0 +1,30 @@
...
+# Uncomment this line to see each command for debugging (careful: this will show secrets!)
+# set -o xtrace
+
+# usage: consolidate_env.sh [workdir] [file]
+
+WORKDIR=${1:-"automatic"}
+FILE=${2:-"automatic"}
+
+# done with processing args and can set this
+set -o nounset
```
tamirkamara, line 13: I think this can move to the top since you handle params with defaults.
Please also add the 3rd header we normally use about pipe...

.devcontainer/Dockerfile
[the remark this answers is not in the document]
martinpeck: lgtm

martinpeck (approved):
you still have some comments from @tamirkamara , but otherwise this lgtm

config.sample.yaml
martinpeck: Agreed. That's how we generally set up config that is "off by default" but needs "illustrating" in the config.

config.sample.yaml
martinpeck: are you saying that all of these values should be moved to the top level, or just the location?

If it's just the location, that probably makes sense.
If it's the whole section, I'm not sure I agree.

LizaShak: I've updated the file structure, so the tre_id and the location are at top level and and everything else under "tre"

config.sample.yaml
```diff
@@ -0,0 +1,92 @@
...
+tre:
+  tre_id: "__CHANGE_ME__"
+  # If your local machine/build agent cannot get the public IP
+  # address from https://ipecho.net/plain, then you can circumvent
+  # this by setting this Environment variable. This blockage can
+  # be caused by trying to deploy TRE in an Office environment where
+  # this website is blocked. This value is the public facing IP
+  # address of the deploying machine.
+  # public_deployment_ip_address: __CHANGE_ME__
+  defaults:
```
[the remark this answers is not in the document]
LizaShak: fixed as mentioned in previous comment

config.sample.yaml
LizaShak: As we talked offline, opened a new issue to handle it: https://github.com/microsoft/AzureTRE/issues/2957

tamirkamara (approved):

config.sample.yaml
```diff
@@ -0,0 +1,87 @@
...
+  swagger_ui_client_id: ""
+  workspace_api_client_id: ""
+  workspace_api_client_secret: ""
+
+resource_processor:
+    # The number of processes to start in the resource processor VMSS image
+    # resource_processor_number_processes_per_instance: 5
+
+    # This setting provides a way to pass environment values to the resource processor to use as a source of bundle parameter values
+    # For example, to specify your image_gallery_id for use in VM user resources with custom VM images:
```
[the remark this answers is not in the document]
LizaShak: Created a new issue: https://github.com/microsoft/AzureTRE/issues/2959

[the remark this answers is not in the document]
LizaShak: > is it worth mentioning that this value must be JSON formatted?

I don't think so, it is quite clear that it is the format, and we already have a lot of text, but if you think it is unclear then I can add maybe something like "Fill in the following json your subscription_id, resource group, and the gallery name"

config.sample.yaml
```diff
@@ -0,0 +1,87 @@
...
+  test_account_client_secret: ""
+  api_client_id: ""
+  api_client_secret: ""
+  swagger_ui_client_id: ""
+  workspace_api_client_id: ""
+  workspace_api_client_secret: ""
+
+resource_processor:
+    # The number of processes to start in the resource processor VMSS image
+    # resource_processor_number_processes_per_instance: 5
```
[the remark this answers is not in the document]
martinpeck: I still think this one should be addressed (either here, or in a bug).
Generally, we don't want people to set this number unreasonably high. I assume the number should be based upon the number of CPUs present on each instance.

Plus, I'm not sure of the values we've tested for this.

We probably need something that says "recommend setting this to X times number of CPUs".

Also, I'm still not sure what the default is. Is it 5, or some other number?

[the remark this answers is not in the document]
martinpeck: Actually, this is already a bit of a mess.

- The terraform defaults to 5
- The code defaults to 1
- Neither has guidance on what to set

Can you please raise a bug that guidance for setting this config is not good enough, and then you can resolve this comment and merge this PR?

PR #2952, opened 2022-12-08
Enable workspace upgrade

### LINKED ISSUE #2847: Enable terraform step in workspace bundle upgrade
## Description

As a TRE Administrator 
I want to update workspace properties after it's initial provisioning AND those will be sent to terraform
So that the workspace infrastructure could be updated.

## Acceptance criteria

- [ ] A property sent to terraform and updates are reflected in Azure

tamirkamara:
# Resolves #2847 

## What is being addressed

Workspace upgrade has been disabled due to a bug with the client_secret value. Since that has been resolved we can reenable the upgrade.

## How is this addressed

- Uncomment the terraform step in the bundle upgrade action
- Redirect URLs script gets the parameter of `register_aad_application` which turns it off in the case where terraform handles the AAD application

### DIFF
--- a/templates/workspaces/base/porter.yaml
+++ b/templates/workspaces/base/porter.yaml
@@ -1,6 +1,6 @@
 ---
 name: tre-workspace-base
-version: 0.7.5
+version: 0.7.6
 description: "A base Azure TRE workspace"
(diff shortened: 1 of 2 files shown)


marrobi (approved):
LGTM

PR #2953, opened 2022-12-08
change merge properties sequence for defaults

### LINKED ISSUE #2575: Default values for properties in a bundle overridden by API app

marrobi:
# Resolves #2575

## What is being addressed

Defaults set by Admins in bundle template definition are overridden by API default template

## How is this addressed

- Change order of priority for the template properties


marrobi:
@tamirkamara looks like this one has the same issue with unit test results?

tamirkamara:
Original PR was merged

PR #2960, opened 2022-12-12
Enabling http2 + requesting client cert

eladiw:
## What is being addressed

Enabling HTTP2 + requesting client certificate for the airlock function and api app service

## How is this addressed

- Enabled http2 and requesting client cert as described here:
- https://learn.microsoft.com/en-us/azure/app-service/app-service-web-configure-tls-mutual-auth
- https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/linux_function_app#client_certificate_enabled

### DIFF
--- a/templates/core/terraform/airlock/airlock_processor.tf
+++ b/templates/core/terraform/airlock/airlock_processor.tf
@@ -31,13 +31,16 @@ resource "azurerm_storage_account" "sa_airlock_processor_func_app" {
 }
 
 resource "azurerm_linux_function_app" "airlock_function_app" {
-  name                      = local.airlock_function_app_name
-  resource_group_name       = var.resource_group_name
-  location                  = var.location
-  https_only                = true
-  virtual_network_subnet_id = var.airlock_processor_subnet_id
-  service_plan_id           = azurerm_service_plan.airlock_plan.id
-  storage_account_name      = azurerm_storage_account.sa_airlock_processor_func_app.name
+  name                       = local.airlock_function_app_name
+  resource_group_name        = var.resource_group_name
+  location                   = var.location
+  https_only                 = true
+  http2_enabled              = true
+  client_certificate_enabled = true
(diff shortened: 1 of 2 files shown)


marrobi:
@eladiw 

> App Service injects an X-ARR-ClientCert request header with the client certificate. App Service does not do anything with this client certificate other than forwarding it to your app. Your app code is responsible for validating the client certificate.

Two things:

1. The app gateway will need to be configured to send a client certificate to the API?
2. The app needs to validate the client certificate?

Otherwise this can't see how it is useful to enable requiring client certificates? As the web app can only be accessed from the private network and App GW (worth checking the NSGs), not sure it really adds much additonal security.

marrobi (commented):
@eladiw 

> App Service injects an X-ARR-ClientCert request header with the client certificate. App Service does not do anything with this client certificate other than forwarding it to your app. Your app code is responsible for validating the client certificate.

Two things:

1. The app gateway will need to be configured to send a client certificate to the API?
2. The app needs to validate the client certificate?

Otherwise this can't see how it is useful to enable requiring client certificates? As the web app can only be accessed from the private network and App GW (worth checking the NSGs), not sure it really adds much additional security.

eladiw:
> @eladiw
> 
> > App Service injects an X-ARR-ClientCert request header with the client certificate. App Service does not do anything with this client certificate other than forwarding it to your app. Your app code is responsible for validating the client certificate.
> 
> Two things:
> 
> 1. The app gateway will need to be configured to send a client certificate to the API?
> 2. The app needs to validate the client certificate?
> 
> Otherwise this can't see how it is useful to enable requiring client certificates? As the web app can only be accessed from the private network and App GW (worth checking the NSGs), not sure it really adds much additional security.

For now, removed the client cert and only enabled http2

marrobi (approved):
Lgtm

anatbal (approved):
LGTM

PR #2963, opened 2022-12-13
Nexus VM disk size

JaimieWi:
# Resolves Nexus disk space issue

## What is being addressed

We have two researchers in the TRE, after using the TRE for around a week they had installed around 24GB of packages into the Nexus VM. This then caused errors that did not initially seem related. Such as:
image (1)
image

Running `df -h` on the Nexus VM showed that only 3.2GB was available. Therefore, as this problem occurred so quickly from researcher's first use. The decision was made to increase the OS disk size to 64GB. 

This PR is a suggestion, as it seems it would be beneficial to all for the standard size of the Nexus VM to be at least 64GB.

## How is this addressed
- Add `disk_size_gb: 64` to the vm.tf file for Nexus.


JaimieWi:
> 

@marrobi I didn't have to reinstall Nexus. But these are the steps I did have to take:
1. Edit the version of the Nexus Resource in Cosmos 
2. Stop the Nexus VM
3. Choose `Update` in the UI

Based on that, and if you were going to introduce the drop-down, it may also be best to provide the `Stop` and `Start` actions for the Nexus VM. The OS disk size could then be changed via the UI with the steps being:
1. Stop Nexus shared service
2. Update OS disk size via drop-down
3. Restart service

marrobi:
@JaimieWi thanks, we will get this merged, and we will create an issue created to parameterise this.

Look like is also a linting issue, are you good to fix this? Otherwise will pull your branch and do this.

JaimieWi:
> @JaimieWi thanks, we will get this merged, and we will create an issue created to parameterise this.
> 
> Look like is also a linting issue, are you good to fix this? Otherwise will pull your branch and do this.

Hi @marrobi, Thanks, that sounds great. I have fixed the linting issue so should be ok to merge now

marrobi (approved):
LGTM

marrobi:
#2968

marrobi:
/test-force-approve 036c485

Nexus not tested in e2e.

PR #2967, opened 2022-12-15
add guacamole restrictions

eladiw:
## What is being addressed

Configure guacamole to have stricter rules

## How is this addressed

- Update Guacamole config file 
- https://guacamole.apache.org/doc/gug/configuring-guacamole.html

### DIFF
--- a/templates/workspace_services/guacamole/guacamole-server/docker/guacamole/guacamole.properties
+++ b/templates/workspace_services/guacamole/guacamole-server/docker/guacamole/guacamole.properties
@@ -5,3 +5,8 @@
 
 enable-environment-properties:true
 enable-websocket:true
+enable-sftp:false
+enable-drive:false
+sftp-disable-upload:true
+disable-download:true
+disable-upload:true


templates/workspace_services/guacamole/guacamole-server/docker/guacamole/guacamole.properties
```diff
@@ -5,3 +5,8 @@
 
 enable-environment-properties:true
 enable-websocket:true
+enable-sftp:false
+enable-drive:false
+sftp-disable-upload:true
+disable-download:true
+disable-upload:true
```
marrobi, line 12: Download is overridden here - https://github.com/microsoft/AzureTRE/blob/a64d1d7e3d69abf205fb32f42d3e8968b433cf6b/templates/workspace_services/guacamole/guacamole-server/guacamole-auth-azure/src/main/java/org/apache/guacamole/auth/azuretre/connection/ConnectionService.java#L112

Doesn't hurt to have defaults in this file, but can we add a checkbox for upload too.  Both should default to being disabled.

marrobi (changes requested):

templates/workspace_services/guacamole/guacamole-server/docker/guacamole/guacamole.properties
eladiw: maybe we should remove the option to override it . wdyt?

marrobi: Some customers use that to download files prior to airlock. We should also have the upload blocked/optional - that's been overlooked, but was required until we had the airlock.

marrobi (approved):
Providing it works, LGTM!

templates/workspace_services/guacamole/guacamole-server/docker/version.txt
```diff
@@ -1 +1 @@
-__version__ = "0.5.0"
+__version__ = "0.5.1"
```
tamirkamara, line 1:
```suggestion
__version__ = "0.6.0"
```

Please update the bundle version as well.

anatbal (approved):
LGTM

PR #2977, opened 2022-12-19
Porter v1 core changes

tamirkamara:
Main changes:
1. Add porter v1 to devcontainer and RP runtime image
2. Add cosmos-mongo resource as that is now the only option for state in porter
3. Use credential/parameter sets (locally and in RP runtime)
4. A script used by local porter commands to complete optional=missing environment variables (work pending in porter to remove this)
5. porter v0 -> v1 migration is added into the RP startup script
6. Changes in API and RP as outputs model has changed a bit
7. RP is able to overcome missing installation error8. 

Pending - convert TODOs in the code to stories and add the links


resource_processor/vmss_porter/runner.py
```diff
@@ -168,13 +169,25 @@ async def invoke_porter_action(msg_body: dict, sb_client: ServiceBusClient, mess
 
     # Handle command output
     if returncode != 0:
-        error_message = "Error context message = " + " ".join(err.split('\n')) + " ; Command executed: ".join(porter_command)
-        resource_request_message = service_bus_message_generator(msg_body, statuses.failed_status_string_for[action], error_message)
+        error_message = "Error message: " + " ".join(err.split('\n')) + "; Command executed: " + " ".join(porter_command)
+
+        pass_despite_of_error = False
+        if "uninstall" == action and "could not find installation" in err:
```
marrobi, line 175: What happens in the scenario where the install has failed part way through - are Azure resources left behind?

resource_processor/vmss_porter/runner.py
```diff
@@ -168,13 +169,25 @@ async def invoke_porter_action(msg_body: dict, sb_client: ServiceBusClient, mess
...
+            message_logger_adapter.warning("The installtion doesn't exist. Treating as a succussful action to allow the flow to proceed.")
+            pass_despite_of_error = True
+            error_message = f"A success despite of underlining error. {error_message}"
```
marrobi, line 178:
```suggestion
            error_message = f"A success despite of underlying error. {error_message}"
```

resource_processor/vmss_porter/runner.py
tamirkamara: That is handled today correctly - TF state includes the provisioned resources and when porter will try to uninstall it will be handled else where and won't reach this block of code as there is an installation

Makefile
```diff
@@ -234,7 +252,7 @@ bundle-publish:
 	&& cd ${DIR} \
 	&& FULL_IMAGE_NAME_PREFIX=${FULL_IMAGE_NAME_PREFIX} \
 		${MAKEFILE_DIR}/devops/scripts/bundle_runtime_image_push.sh \
-	&& porter publish --registry "$${ACR_NAME}.azurecr.io" --debug
+	&& porter publish --registry "$${ACR_NAME}.azurecr.io" --force
```
jjgriff93, line 255: Is the use of `--force` here to overwrite a bundle version if it already exists in the registry?

resource_processor/shared/config.py
```diff
@@ -30,19 +31,51 @@ def get_config(logger_adapter) -> dict:
...
+        config["application_admin_client_id"] = ""
+        config["application_admin_client_secret"] = ""
 
     # Create env dict for porter
     config["porter_env"] = {
         "HOME": os.environ["HOME"],
         "PATH": os.environ["PATH"],
+        "KEY_VAULT_NAME": config["key_vault_name"],
+
+        # TODO: why do we need these
```
jjgriff93, line 57: From memory these were defined for the Porter env dict but looks like now you're doing it directly on line 65, so can probably be removed

resource_processor/vmss_porter/runner.py
```diff
@@ -168,13 +169,25 @@ async def invoke_porter_action(msg_body: dict, sb_client: ServiceBusClient, mess
 
     # Handle command output
     if returncode != 0:
-        error_message = "Error context message = " + " ".join(err.split('\n')) + " ; Command executed: ".join(porter_command)
-        resource_request_message = service_bus_message_generator(msg_body, statuses.failed_status_string_for[action], error_message)
+        error_message = "Error message: " + " ".join(err.split('\n')) + "; Command executed: " + " ".join(porter_command)
+
+        pass_despite_of_error = False
+        if "uninstall" == action and "could not find installation" in err:
+            message_logger_adapter.warning("The installtion doesn't exist. Treating as a succussful action to allow the flow to proceed.")
```
jjgriff93, line 176:
```suggestion
            message_logger_adapter.warning("The installation doesn't exist. Treating as a successful action to allow the flow to proceed.")
```

resource_processor/vmss_porter/runner.py
```diff
@@ -168,13 +169,25 @@ async def invoke_porter_action(msg_body: dict, sb_client: ServiceBusClient, mess
 
     # Handle command output
     if returncode != 0:
-        error_message = "Error context message = " + " ".join(err.split('\n')) + " ; Command executed: ".join(porter_command)
-        resource_request_message = service_bus_message_generator(msg_body, statuses.failed_status_string_for[action], error_message)
+        error_message = "Error message: " + " ".join(err.split('\n')) + "; Command executed: " + " ".join(porter_command)
+
+        pass_despite_of_error = False
```
jjgriff93, line 174: `grammar nit` should probably change to `pass_despite_error`

templates/core/terraform/resource_processor/vmss_porter/cloud-config.yaml
```diff
@@ -59,18 +59,23 @@ write_files:
...
       echo "Used disk space percent: $${used_percent}"
-      if (( used_percent > 70 )); then
+      if (( used_percent > 75 )); then
         echo "Free space too low, pruning..."
         docker system prune -f
       fi
     permissions: '0755'
 
 runcmd:
+  # Those are useful live debug commands. Check the docs for details
```
jjgriff93, line 69: Perhaps add link to doc?

jjgriff93 (approved):
Looks great! Just some minor points

Makefile
tamirkamara: Yes. Porter now checks if the version is already in the registry and fail the publish action.

templates/core/terraform/resource_processor/vmss_porter/cloud-config.yaml
tamirkamara: Added a TODO as I'll handle all docs in a separate PR.

resource_processor/shared/config.py
tamirkamara: I think to comment is because it doesn't work on Azure (with MSI) without them. I'll check again.

tamirkamara: We actually need most of those settings on Azure too because they are referenced as credentials in all bundles and also in the credential set. I've updated the comments around those section to make it clearer for the next person.

Issue #1796, opened 2022-05-10
Add more documentation for Guacamole oauth proxy extension


anatbal:
## Description

Some documentation is still missing regarding oauth proxy which needs to be added:
1. Add the guacamole callback url to the workspace app registration redirect.
2. A user must have an email field in order to be able to authenticate with the oauth proxy

tamirkamara:
This should be done alongside upgrading the proxy version

PR #2980, opened 2022-12-20
Upgrade OAuth Proxy to 7.4.0

### LINKED ISSUE #2732: Upgrade OAuth Proxy to 7.4.X
We current use oauth proxy 7.2.1, we need to upgrade to 7.3.0see

https://github.com/marrobi/AzureTRE/blob/4fa3e3dd5a08204fff1c171eabf3550713bbe80e/templates/workspace_services/guacamole/guacamole-server/docker/Dockerfile#L67

anatbal:
# Fixes #2732 #1796 

## What is being addressed

- Upgrading oauth2 proxy to 7.4.0. It required a few more run time arguments:
  -  set oidc_groups_claim="roles" since microsoft does not send groups when user is member of more than 5 groups. 
  - set insecure_oidc_allow_unverified_email="true" since microsoft does not issue email_verified claim but rather claims that all emails issued are valid by default.
- Allowing guacamole workspace service bundle to be upgraded
- Adding documentation for oauth2 proxy
**TBD**
- Test if upgrade completes successfully


templates/workspace_services/guacamole/guacamole-server/docker/services/oauth/run
```diff
@@ -21,4 +21,6 @@ cookiesecret=$(dd if=/dev/urandom bs=32 count=1 2>/dev/null | base64 | tr -d --
 --set-xauthrequest true \
 --pass-basic-auth true \
 --cookie-refresh 50m \
---scope "openid offline_access ${AUDIENCE}/user_impersonation"
+--insecure-oidc-allow-unverified-email true \
+--oidc-groups-claim "roles" \
```
LizaShak, line 25: I think roles is not the one holding the groups info. Maybe we can do something like this: https://learn.microsoft.com/en-us/answers/questions/300079/how-to-get-group-information-as-part-of-claims-in.html. I'd be happy to look into this together if you'd like.

templates/workspace_services/guacamole/guacamole-server/docker/services/oauth/run
```diff
@@ -21,4 +21,6 @@ cookiesecret=$(dd if=/dev/urandom bs=32 count=1 2>/dev/null | base64 | tr -d --
...
+--scope "openid offline_access ${AUDIENCE}/user_impersonation profile"
```
LizaShak, line 26: Is profile required now?

templates/workspace_services/guacamole/guacamole-server/docker/services/oauth/run
anatbal: Sure we can take it offline

templates/workspace_services/guacamole/guacamole-server/docker/services/oauth/run
anatbal: Yes, it's required since 7.4.0

templates/workspace_services/guacamole/guacamole-server/docker/version.txt
```diff
@@ -1 +1 @@
-__version__ = "0.3.8"
+__version__ = "0.3.9"
```
tamirkamara, line 1:
```suggestion
__version__ = "0.4.0"
```
This is a meaningful change that should be noted by at least a minor version bump. Please read the instructions in the PR template.

tamirkamara (approved):

PR #2981, opened 2022-12-20
Add option to disable Swagger

### LINKED ISSUE #2925: Add option to disable Swagger UI
Customer would like to make swagger API docs less accessible to ordinary users.

yuvalyaron:
# Resolves #2925

## What is being addressed

Added the option to disable Swagger UI

## How is this addressed
- Added 'ENABLE_SWAGGER' option in the config.yaml
- When swagger is disabled, the `/docs` endpoint will return an error messages saying that Swagger is disabled.
- The Swagger UI is disabled by default, but the config.yaml has a default value of true, so developers will still have Swagger.


api_app/api/routes/api.py
```diff
@@ -157,5 +158,20 @@ async def get_workspace_swagger(workspace_id, request: Request, workspace_repo=D
...
     return swagger_ui_html
 
-workspace_router.include_router(workspace_swagger_router)
+
+swagger_disabled_router = APIRouter()
+
+
+@swagger_disabled_router.get("/docs", include_in_schema=False, name="swagger_disabled")
+async def get_disabled_swagger():
+    return PlainTextResponse("Swagger is disabled. Set 'ENABLE_SWAGGER' to true in order to access Swagger.", status_code=HTTP_403_FORBIDDEN)
```
LizaShak, line 167: Can you use HTTPException instead of PlainTextResponse? Also I have a few questions on the routes, can we have a short chat about it?

templates/core/terraform/api-webapp.tf
```diff
@@ -44,6 +44,7 @@ resource "azurerm_linux_web_app" "api" {
     "MANAGED_IDENTITY_CLIENT_ID"                     = azurerm_user_assigned_identity.id.client_id
     "TRE_ID"                                         = var.tre_id
     "RESOURCE_LOCATION"                              = azurerm_resource_group.core.location
+    "ENABLE_SWAGGER"                                 = var.enable_swagger
```
tamirkamara, line 47: Github actions are using secrets and it's not ideal for us to put values of true/false there as it obfuscate that value across all log outputs. 
Can you do a change that would effectively allow any value other than false to turn this option on?

api_app/api/routes/api.py
```diff
@@ -1,6 +1,9 @@
 from collections import defaultdict
 from typing import Any, DefaultDict, Dict, Optional
 
+from starlette.responses import PlainTextResponse
+from starlette.status import HTTP_403_FORBIDDEN
```
tamirkamara, line 5: are we using these starlette imports in other places too?

api_app/api/routes/api.py
```diff
@@ -157,5 +158,20 @@ async def get_workspace_swagger(workspace_id, request: Request, workspace_repo=D
 
     return swagger_ui_html
 
-workspace_router.include_router(workspace_swagger_router)
+
+swagger_disabled_router = APIRouter()
+
+
+@swagger_disabled_router.get("/docs", include_in_schema=False, name="swagger_disabled")
```
tamirkamara, line 165: Swagger is also active in the workspace scope, i.e. /api/workspaces/some_guid/docs. 
Does this handle it as well?

templates/core/terraform/api-webapp.tf
yuvalyaron: let's talk have a quick chat about this, what do you mean by obfuscating the logs?
why boolean values in particular cause more obfuscation?

yuvalyaron: update: we've decided to keep it as boolean for now

LizaShak (approved):

PR #2985, opened 2022-12-25
Update templates for Porter v1

tamirkamara:
## What is being addressed

Updates all templates to Porter v1 spec:

1. Replace `"{{ }}"` with `${ }` as that is the new porter syntax for value substitution 
2. Update bundles' dockerfile 
3. Update bundles' terraform cli to 1.3.6
4. Pin bundles' azure cli to 2.37.0 (same as the dev container)
5. When running `make bundle-check-param` verify that the _ParameterSet_ has the same name as the bundle
6. Fix workflow to return the correct exit code when doing "bash retry"
7. Stop installing python with certbot as images have it preinstalled
8. Remove locks from the firewall bundle

TODO: drop temp commits


.github/workflows/deploy_tre_branch.yml
```diff
@@ -60,7 +60,7 @@ jobs:
       environmentName: ${{ github.event.inputs.environment }}
     secrets:
       AAD_TENANT_ID: ${{ secrets.AAD_TENANT_ID }}
-      ACR_NAME: ${{ format('tre{0}', needs.prepare-not-main.outputs.refid) }}
+      ACR_NAME: tre233a1e59  # ${{ format('tre{0}', needs.prepare-not-main.outputs.refid) }}
```
anatbal, line 63: I guess this is temp?

templates/shared_services/firewall/porter.yaml
```diff
@@ -44,59 +45,49 @@ parameters:
...
 
 mixins:
   - terraform:
-      clientVersion: 1.2.6
+      clientVersion: 1.3.6
 
 install:
   - terraform:
       description: "Deploy shared service"
-      input: false
```
anatbal, line 59: Why is this line being deleted?

templates/shared_services/firewall/terraform/firewall.tf
```diff
@@ -26,13 +26,13 @@ resource "azurerm_firewall" "fw" {
...
 }
 
-resource "azurerm_management_lock" "fw" {
-  count      = var.stateful_resources_locked ? 1 : 0
-  name       = azurerm_firewall.fw.name
-  scope      = azurerm_firewall.fw.id
-  lock_level = "CanNotDelete"
-  notes      = "Locked to prevent accidental deletion"
-}
+# resource "azurerm_management_lock" "fw" {
```
anatbal, line 29: I guess temp?

.github/workflows/deploy_tre_reusable.yml
```diff
@@ -203,8 +203,12 @@ jobs:
         # failure in the first attempt indicates a new ACR, so we need to try again after it's been created
         if: steps.ci_cache_cr_login.outcome != 'success'
         run: |
-          # shellcheck disable=SC2034,SC2015
-          for i in {1..3}; do az acr login --name "${{ secrets.CI_CACHE_ACR_NAME }}" && break || sleep 10; done
+          # shellcheck disable=SC2034,SC2015,SC2125
```
anatbal, line 206: Should it still be disabled?

templates/shared_services/firewall/Dockerfile.tmpl
```diff
@@ -1,19 +1,10 @@
...
-# dockerfile: Dockerfile.tmpl
-
-# You can control where the mixin's Dockerfile lines are inserted into this file by moving "# PORTER_MIXINS" line
-# another location in this file. If you remove that line, the mixins generated content is appended to this file.
 # PORTER_MIXINS
 
 # Use the BUNDLE_DIR build argument to copy files into the bundle
-COPY . $BUNDLE_DIR
+COPY --link . ${BUNDLE_DIR}/
+
```
anatbal, line 10: redundant line

.github/workflows/deploy_tre_branch.yml
tamirkamara: yes, I will delete it when I rebase.

templates/shared_services/firewall/porter.yaml
tamirkamara: It's not a valid property per the new porter.yaml schema. All terraform commands have no input by default.

templates/shared_services/firewall/terraform/firewall.tf
tamirkamara: No, I removed the firewall lock as it isn't that useful anymore (all rules are anyway in the DB).
I just forgot to delete it here. I will do so...

.github/workflows/deploy_tre_reusable.yml
tamirkamara: yes, it doesn't work any other way.

templates/shared_services/firewall/Dockerfile.tmpl
tamirkamara: removed

.github/workflows/deploy_tre_branch.yml
tamirkamara: done

PR #2988, opened 2022-12-25
dependabot config

tamirkamara:
## What is being addressed

We only get alerts on npm dependencies. This PR adds specific configuration for dependabot to see if this will make a difference and we'll start getting more.


tamirkamara:
/test-force-approve
This change isn't tested by the validation suite.

anatbal (approved):

PR #3016, opened 2022-12-26
Move core folder to root

### LINKED ISSUE #2870: Core infrastructure under the root folder
## Description

As a TRE Developer
I want to have the core infrastructure under the root folder
So that it would be easily found and separated from _real template_ assets. 

This could be done **around EOY 2022 or beginning of 2023**

## Acceptance criteria

- [ ] core folder is removed from templates and placed under the root
- [ ] everything work...
- [ ] deployment repo is updated as well
- [ ] docs are updated

tamirkamara:
Resolves #2870 

## What is being addressed

The _core_ folder isn't a template like all the other definitions under the _templates_ folder.

## How is this addressed

- Move it to the root folder


yuvalyaron (changes requested):
what about line 18 in `env_to_yaml_config.sh` was this not changed on purpose?

`
for p in "devops/auth.env" "devops/.env" "templates/core/.env"
do
  if [ -r "$p" ]
  then
    env_files+=("$p")
  else
    echo -e "\e[31m»»» ⚠️ Your $p file has not been setup! 😥 Make sure to fill in the missing configration in config.yaml."
  fi
done
`

tamirkamara:
> what about line 18 in `env_to_yaml_config.sh` was this not changed on purpose?
> 
> `for p in "devops/auth.env" "devops/.env" "templates/core/.env" do if [ -r "$p" ] then env_files+=("$p") else echo -e "\e[31m»»» ⚠️ Your $p file has not been setup! 😥 Make sure to fill in the missing configration in config.yaml." fi done`
@yuvalyaron 
This is a migration script from .env files to the new config.yaml. As such it's find for it to look for files in the old location, we don't expect to find them in the new.

yuvalyaron (approved):

PR #3023, opened 2022-12-27
Update API dependencies (dependabot)

yuvalyaron:
## What is being addressed

Updated the dependencies of the API.

### Related dependabot PRs:
https://github.com/microsoft/AzureTRE/pull/3002
https://github.com/microsoft/AzureTRE/pull/3006
https://github.com/microsoft/AzureTRE/pull/3007
https://github.com/microsoft/AzureTRE/pull/3008
https://github.com/microsoft/AzureTRE/pull/3005 (partly)


cli/setup.py
```diff
@@ -44,7 +44,7 @@
         "httpx~=0.23.0",
         "msal >= 1.17.0",
         "jmespath==1.0.1",
-        "tabulate==0.8.10",
+        "tabulate==0.9.0",
```
yuvalyaron, line 47: @tamirkamara tabulate was updated in the requirements.txt file in https://github.com/microsoft/AzureTRE/pull/3005 - but this file was not updated. Should it be updated here as well?

yuvalyaron: and if so, I also noticed that requirements.txt has
`PyJWT==2.6.0` which is missing from here

yuvalyaron: talked offline, the answer is yes

requirements.txt
```diff
@@ -1,6 +1,6 @@
 # Dev requirements
-flake8==5.0.4 # same as super linter
-pre-commit==2.20.0
+flake8==6.0.0 # same as super linter
```
tamirkamara, line 2: Is this still the same version super linter is using?

tamirkamara (approved):

requirements.txt
yuvalyaron: yes
https://github.com/github/super-linter/blob/f0321c38d468cfa3589ed0427a619959fe19d145/dependencies/python/flake8.txt#L1

yuvalyaron: you were right, they released the latest version on October 6, but updated flake version only last week, so we shouldn't update it yet - reverting

Issue #2491, opened 2022-08-22
Resource history blows up document size


tamirkamara:
**Describe the bug**

Resource history is saved indefinitely along side the current resource version. This causes the document to get larger and larger overtime but CosmosDB is limited to 2MB per item.

This is happening in our CICD environment when many runs change the firewall share service until it gets too large and the API starts to output errors (from cosmos) about the document size. As a temp measure I deleted some items from the history object.

Probably best to think of a way to address this internally (either save last X versions, archive older history items elsewhere, etc.).

There was an agreement about going on and crearing a new container for archiveing older history items.
The tasks that are planned are:

- [ ] Create a new container for history
- [ ] Create history item in a new container
- [ ] Change history item schema to include also parent id 
- [ ] Change the query (but still return it as part of the resource as a list) 
- [ ] Create an API to return only history for UI usage and undo the list in the resource schema 
- [ ] Fix unit tests 
- [ ] Write migration script

anatbal:
We decided not to change the airlock archiving mechanism

PR #3032, opened 2023-01-01
Archive resource history - part I

anatbal:
# Resolves part of #2491 

This is the first part of the implementation for archiving a resource history

Creation of a new container name resources_history
Creation of ResourceHistoryRepository class
Save each history item in the new container (maintain the existing behavior, this will be removed in the next PR together with migrations)
Note - These changes are not touching airlock just yet


api_app/db/repositories/resources_history.py
```diff
@@ -0,0 +1,46 @@
...
+class ResourceHistoryRepository(BaseRepository):
+    @classmethod
+    async def create(cls, client: CosmosClient):
+        cls = ResourceHistoryRepository()
+        await super().create(client, config.STATE_STORE_RESOURCES_HISTORY_CONTAINER)
+        return cls
+
+    @staticmethod
+    def resource_history_query(resourceId: str):
+        return f'SELECT * FROM c WHERE c.resourceId = "{resourceId}"'
```
yuvalyaron, line 19: do we need this method?
since we know the ID, it's better to not use SQL, and just read it directly -
it's cheaper in terms of RU.

yuvalyaron (changes requested):

api_app/models/domain/resource.py
```diff
@@ -16,14 +16,28 @@ class ResourceType(str, Enum):
     SharedService = strings.RESOURCE_TYPE_SHARED_SERVICE
 
 
+class NewResourceHistoryItem(AzureTREModel):
```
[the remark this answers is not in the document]
anatbal: No. This is strictly for making it easier for review

api_app/api/routes/airlock_resource_helpers.py
```diff
@@ -147,6 +148,7 @@ async def delete_review_user_resource(
         workspace_service_repo: WorkspaceServiceRepository,
         resource_template_repo: ResourceTemplateRepository,
         operations_repo: OperationRepository,
+        resource_history_repo: ResourceHistoryRepository,
```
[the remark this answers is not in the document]
anatbal: I strongly disagree.  Since the UI is using the history this PR will become huge and no way anyone can review it. Had this discussion with @tamirkamara  as well that this is going to be divided.

api_app/db/repositories/resources.py
```diff
@@ -112,7 +113,8 @@ async def patch_resource(self, resource: Resource, resource_patch: ResourcePatch
             templateVersion=resource_copy.templateVersion
         )
         resource.history.append(history_item)
-
+        # create the history item in the resource history repository
```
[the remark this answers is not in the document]
anatbal: Again, only for reviewer to be able to see the difference

api_app/models/domain/resource.py
```diff
@@ -16,14 +16,28 @@ class ResourceType(str, Enum):
...
+    """
+    Resource History Item - to preserve history of resource properties
+    """
+    id: str = Field(title="Id", description="GUID identifying the resource request")
+    resourceId: str = Field(title="Id", description="GUID identifying the resource request")
```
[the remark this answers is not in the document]
anatbal: Typo will fix

api_app/db/repositories/resources_history.py
anatbal: I understand what you are saying, but here it's different. we are doing the query by **resourceId** which means it will potentially bring a list of items, since these are all the history items that are related to this resource, each will have a different **id**. The thing you commented on is not the function but strictly a  const for the sql query which is necessary.

api_app/models/domain/resource.py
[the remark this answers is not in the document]
eladiw: so, what is the name? ResourceHistoryItem?
I suggest you already rename

api_app/models/domain/resource.py
```diff
@@ -16,14 +16,28 @@ class ResourceType(str, Enum):
...
 class ResourceHistoryItem(AzureTREModel):
     """
     Resource History Item - to preserve history of resource properties
     """
-    properties: dict = {}
+    properties: dict = Field({}, title="Resource template parameters", description="Parameters for the deployment")
     isEnabled: bool
     resourceVersion: int
-    updatedWhen: float
+    updatedWhen: float = 0
```
eladiw, line 40: why is this needed

api_app/core/config.py
```diff
@@ -26,6 +26,7 @@
 STATE_STORE_DATABASE = "AzureTRE"
 STATE_STORE_RESOURCES_CONTAINER = "Resources"
 STATE_STORE_RESOURCE_TEMPLATES_CONTAINER = "ResourceTemplates"
+STATE_STORE_RESOURCES_HISTORY_CONTAINER = "ResourcesHistory"
```
eladiw, line 29: up to debate.
ResourceTemplates --> implies --> ResourceHistories?

api_app/db/repositories/resources_history.py
eladiw: please make sure. at least somewhere along the process, this resource id is sanitized.

api_app/db/repositories/resources_history.py
```diff
@@ -0,0 +1,46 @@
...
+
+    @staticmethod
+    def resource_history_with_resource_version_query(resourceId: str, resourceVersion: str):
+        return f'SELECT * FROM c WHERE c.resourceId = "{resourceId}" AND c.resourceVersion = "{resourceVersion}"'
```
eladiw, line 23: same comment

api_app/db/repositories/resources_history.py
```diff
@@ -0,0 +1,46 @@
...
+
+    async def get_resource_history_by_resource_id(self, resourceId: str) -> List[NewResourceHistoryItem]:
+        query = self.resource_history_query(resourceId)
+        resource_history_items = await self.query(query=query)
```
eladiw, line 27: can we add some logs? e.g. number of items in the list. or more useful debug info?

api_app/db/repositories/resources_history.py
```diff
@@ -0,0 +1,46 @@
...
+        return parse_obj_as(List[NewResourceHistoryItem], resource_history_items)
```
eladiw, line 28: what happens if this parse fails?

eladiw (changes requested):
Looks good. some comments.
in general can you add more logs?

api_app/api/routes/airlock_resource_helpers.py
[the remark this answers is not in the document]
anatbal: Since the UI is using the history this PR will become huge and no way anyone can review it. Had this discussion with @tamirkamara as well that this is going to be divided.
Regarding the usage, it has all the details in the PR description

api_app/db/repositories/resources_history.py
yuvalyaron: oh, I see, but if you query based on the resource ID, and the partition key is the ID -  shouldn't we choose the resource ID to be the partition key? this way the history items for a resource will be stored on the same partition, which will be better in terms of cost and performance

api_app/db/repositories/resources.py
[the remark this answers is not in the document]
yuvalyaron: your call, but I don't find it necessary

api_app/db/repositories/resources_history.py
anatbal: I agree. changing the partition key to resourceID.
Just documenting here that even though the operation repo acts in a similar manner it has a partition key of id, most likely because no one noticed.

api_app/models/domain/resource.py
anatbal: To match it with default values like the rest of the models

api_app/core/config.py
anatbal: after a discussion ResourceHistory was picked

api_app/db/repositories/resources_history.py
anatbal: Will add validation for special chars and guid

api_app/db/repositories/resources_history.py
anatbal: An error will be thrown similar to all our routes

anatbal:
I am closing this since I opened a new PR. All the comments are fixed there

PR #3045, opened 2023-01-02
Align performance test config with regular e2e

### LINKED ISSUE #2957: Performance tests should use same params as other tests
## Description

As a Azure TRE Developer
I want to performance tests to use existing test_worspace_service_id and test_workspace_id used by other tests (instead of perf_test_workspace_id, perf_test_worspace_service_id)
So that it can be managed and defined once

tamirkamara:
Resolves #2957

## What is being addressed

Performance E2E tests had environment settings used exclusivly by them.

## How is this addressed

- Remove performance tests environment variables and align with the TEST* ones used by other E2Es.


LizaShak (approved):
LGTM

tamirkamara:
/test-force-approve
performance tests aren't checked

PR #3044, opened 2023-01-02
Add ExceptionTracebackFilter + Replace logging.error with logging.exception in API routes

yuvalyaron:
# Resolves #2921 

## What is being addressed

Improved the logging in the API routes.

## How is this addressed

-  Replaced `logging.error` with `logging.exception` calls in API routes wherever it was applicable - this way we can have the traceback of the exception, and not only the exception message.
-  Added ExceptionTracebackFilter - a logger filter that logs again logs that that have exception info.  This is necessary because if a log has exception info (such as logs that were done using logging.exception) Application insights will only show them in 'exceptions' section without showing in the 'traces' section.

(https://microsoft.github.io/AzureTRE/tre-workspace-authors/authoring-workspace-templates/#versioning)

PR #3046, opened 2023-01-02
Replace logging.error with logging.exception in the API

### LINKED ISSUE #2921: All API routes should log errors with stack trace
Many routes log errors with just the exception message which makes it hard to understand some errors and where they come from.
This is especially true in situations where the route returns an HTTP exception to the client.

As example from the log:
```
Failed to create workspace model instance: Only decimal digits permitted in "{'address_space': '10" in "{'address_space': '10.1.177.0"
```
Custom dimension data will show:
```json
{"process":"MainProcess","module":"workspaces","fileName":"/api/./api/routes/workspaces.py","lineNumber":"100","level":"ERROR"}
```
But that doesn't help too much since workspace.py:100 is the line where the log was generated in the route

yuvalyaron:
# Resolves #2921

## What is being addressed

Improved the logging in the API.
Continuation of #3044.

## How is this addressed

- Replaced logging.error with logging.exception calls in API routes wherever it was applicable - this way we can have the traceback of the exception, and not only the exception message.


api_app/service_bus/deployment_status_updater.py
```diff
@@ -175,9 +175,9 @@ async def update_status_in_database(self, message: DeploymentStatusUpdateMessage
             # Marking as true as this message will never succeed anyways and should be removed from the queue.
             result = True
             error_string = strings.DEPLOYMENT_STATUS_ID_NOT_FOUND.format(message.id)
-            logging.error(error_string)
-        except Exception as e:
-            logging.error(strings.STATE_STORE_ENDPOINT_NOT_RESPONDING + " " + str(e))
+            logging.exception(error_string)
```
tamirkamara, line 178: I think the error string variable is redundant since it's not used elsewhere. I think there're other places where we have it and can consider using the expression directly.

tamirkamara (approved):

api_app/service_bus/deployment_status_updater.py
yuvalyaron: removed

PR #3061, opened 2023-01-05
Lint TRE cost tags

tamirkamara:
## What is being addressed

We have been checking that each TRE resource has the tre_id tag as that is the common one between all entities (like workspace, workspace service, etc.).
Ideally we should lint each TRE level for its respective tag requirements.

## How is this addressed

- Upgrade SuperLint to 4.10 to include a new TFLint version which is required for complex tags like we do (with a `local` definition)
- Add TFLint definition files for each level that describe the required tags
- Add CI steps to run SuperLinter with the TFLint definitions when needed
- A couple of fix-ups on missing tags

### DIFF
--- a/.github/linters/.tflint_shared_services.hcl
+++ b/.github/linters/.tflint_shared_services.hcl
@@ -1,0 +1,19 @@
+# This is used for TRE tags validation only.
+
+config {
+  module = true
+  force = false
+}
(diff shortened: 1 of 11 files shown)


guybartal (approved):
Looks great, much better now!

PR #3063, opened 2023-01-05
Fix running actions on a user resource

### LINKED ISSUE #3062: Execute any actions on VMs fails
can't execute any actions on VMs, this is due to a change in the async cosmos client.

anatbal:
# Resolves #3062 
## What is being addressed

Fix the route for running actions on a user resource


marrobi (approved):
Confirmed fixed. thanks!

PR #3069, opened 2023-01-10
Add additional log deploy management

dusan-ilic-mhra:
Add additional log for deploy management job


dusan-ilic-mhra:
created by accident

PR #3071, opened 2023-01-11
Fix odd API behavior to illegal patching

### LINKED ISSUE #2310: Odd API behavior to "illegal" patching
**Describe the bug**
API allows requests to patch certain properties to go through but:
1. Doesn't commit the change to DB
2. Doesn't log any error or at least a warning of the "dropped" update
3. Sends an update/upgrade to RP

**Steps to reproduce**
1. Patch the firewall shared service with:
```json
{
  "templateVersion": "0.3.999"
}
```
or: 
```json
{
  "abc": "1.2.3.4"
}
```
2. Get back a good response code, resource history entry and an upgrade request.

yuvalyaron:
# Resolves #2310 

## What is being addressed

- Patching with a field that is not defined in 'ResourcePatch', such as `"abc": "1.2.3.4"`, will now result in a HTTP 422 error, instead of the field being ignored and the patch applied without it.
- Patching with a template version in invalid format, such as `"templateVersion": "0.3.1.1"` will now result in a HTTP 400 Bad Request error instead of a HTTP 500 Internal Server Error.

## How is this addressed

- Added `Extra.forbid` to the `Config` of  `ResourcePatch` - this will throw error if we send a field that is not defined in `ResourcePatch`, the default behavior is to just ignore the field
- Added a "try except" block to `semantic_version.Version`
- Added a test to enforce the behavior of sending 422 with an invalid field


guybartal (approved):
LGTM

Issue #2984, opened 2022-12-22
Deploy management failed - AZURE_CREDENTIALS set as GitHub secret


dusan-ilic-mhra:
**Describe the bug**

We upgraded the OSS version to 0.7.0 and now we are facing this issue when we deploy using CI:

![image](https://user-images.githubusercontent.com/104429461/209239729-7b5b67ca-f30c-4673-8b79-e1681ccf3ae9.png)

I reviewed this PR:
https://github.com/microsoft/AzureTRE/pull/2654

and created AZURE_CREDENTIALS as a GitHub secret in the mentioned format:

{
  "clientId": "xxx",
  "clientSecret": "xxx",
  "tenantId": "xxx",
  "subscriptionId": "xxx",
"resourceManagerEndpointUrl": "management.azure.com"
}

**Steps to reproduce**

1. Run Deploy Azure TRE workflow

Do I miss something? Could you please help us because this is currently a blocker for us?

tamirkamara:
@dusan-ilic-mhra please verify that the format & values corrosponds to the example given [here](https://github.com/marketplace/actions/azure-login#configure-deployment-credentials).
I think you should try the change the _resourceManagerEndpointUrl_

dusan-ilic-mhra:
@tamirkamara Where I can find resourceManagerEndpointUrl?

dusan-ilic-mhra:
Following your comment to take a look at this link https://github.com/marketplace/actions/azure-login#configure-deployment-credentials

I run a command for the existing App Reg: 

az ad sp create-for-rbac --name "**App_Reg_Name**" --role contributor --scopes /subscriptions/**Subscription_ID**/resourceGroups/**management_rg_name** --sdk-auth

and I’ve got this output:

{
  "clientId": "xxxxxxxxxxxxxxxxxxxx",
  "clientSecret": "xxxxxxxxxxxxxxxxxxxx",
  "subscriptionId": "xxxxxxxxxxxxxxxxxxxx",
  "tenantId": "xxxxxxxxxxxxxxxxxxxx",
  "activeDirectoryEndpointUrl": "xxxxxxxxxxxxxxxxxxxx",
  "resourceManagerEndpointUrl": "https://management.azure.com/",
  "activeDirectoryGraphResourceId": "xxxxxxxxxxxxxxxxxxxx",
  "sqlManagementEndpointUrl": "xxxxxxxxxxxxxxxxxxxx",
[...]

PR #3073, opened 2023-01-12
Validate required secrets have value

tamirkamara:
## What is being addressed

Some secrets must exist and have values, but something people forget to define them. In this case the errors might not be easy to understand.
An example issue: https://github.com/microsoft/AzureTRE/issues/2984

## How is this addressed

- Add a step in the reusable deployment workflow that checks that required secrets exist with non-empty values.


guybartal (changes requested):
required secret is missing in test: ```MS_TEAMS_WEBHOOK_URI```

tamirkamara:
> required secret is missing in test: `MS_TEAMS_WEBHOOK_URI`

tamirkamara:
> required secret is missing in test: `MS_TEAMS_WEBHOOK_URI`

It's not really missing since we sometimes remove the step that uses it (especially in the deployment repo)

guybartal (approved):

PR #3075, opened 2023-01-12
Add Pylance extension and fix [str | None] issue

tamirkamara:
## What is being addressed

Mismatch types in Python can cause runtime issues. We currently don't use any tool to alert us from those kind of problems.

## How is this addressed

- Add Pylance VSCode extension
- Fix the most frequent issue of assigning default _None_ to optional _str_ params


anatbal (approved):
Thanks for adding that!

PR #3081, opened 2023-01-15
Add firewall rule to allow AAD for health services

LizaShak:
# Resolves 3085

## What is being addressed

Add firewall rule to allow Az Portal Login . This will allow the user to acquire a token in order to access the deployed health services.


templates/workspace_services/health-services/terraform/firewall.tf
```diff
@@ -0,0 +1,35 @@
+resource "azurerm_firewall_application_rule_collection" "health_services_authentication" {
```
marrobi, line 1: Is there a reason that you are you not using the pipeline functionality in the the template schema to add the rules? This avoids clashing operations.

templates/workspace_services/health-services/terraform/firewall.tf
```diff
@@ -0,0 +1,35 @@
...
+    protocol {
+      port = "443"
+      type = "Https"
+    }
+    protocol {
+      port = "80"
+      type = "Http"
+    }
+
+    # Taken from the whitelist for Azure portal authentication:
```
marrobi, line 19: Think these are covered by the AzureActiveDirectory service tag.

tamirkamara (approved):

PR #3083, opened 2023-01-15
Add workspace services to e2e

tamirkamara:
## What is being addressed

Most workspace services aren't tested in E2Es.

## How is this addressed

- Add workspace services deployment tests
- Create a new pytest marker for workspace services
- Remove certs shared service from tests as it doesn't play nice with the firewall (it doesn't use the pipeline)
- MLFlow workspace service updates
- Gitea workspace template
- [unrelated temp fix] disable bundle-check-params in CI as porter is working on a fix


tamirkamara:
/test-force-approve
was tested a bunch of times via branch workflow

PR #3084, opened 2023-01-15
Move Github secrets to variables

### LINKED ISSUE #3070: Convert Github Secrets to Variables where possible
## Description

As a TRE Developer
I want to use the new feature of Github Actions Variables and not secrets where values aren't sensitive
So that logs will be clearer and existing values could be reviewed

https://github.blog/changelog/2023-01-10-github-actions-support-for-configuration-variables-in-workflows/

tamirkamara:
Resolves #3070 

## What is being addressed

We store non-secret values in Github Secrets but that can create issues since these values are redacted from logs (try to put a value of 1 and see what happens), and can't be viewed on the website to understand the current configuration.

## How is this addressed

- Move certain values to github variables (new feature)
- Define a default for address spaces as that is usually enough
- Run MS Teams messages only if secret has value (useful for those who don't want to use it)
- Update documentation


tamirkamara:
So the linter doesn't know about the new vars collection 😔

tamirkamara:
The PR is waiting for the next superlinter version to be release. So will close for now.

PR #3086, opened 2023-01-16
Aligning airlock processor unittests to use pytest

### LINKED ISSUE #3026: Convert Airlock UTs to pytest
## Description

As a TRE Developer
I want the Airlock function tests to be written in pytest
So that they will be consistent with all the other tests we have in the solution

## Acceptance criteria

- [ ] Airlock UTs use pytest (and not unittest)

eladiw:
# Resolves #3026 

## What is being addressed

Aligning the Airlock processor tests to use the pytest (same as all other tests)

## How is this addressed

- Updated the tests and used pytest assert where needed


airlock_processor/tests/test_status_change_queue_trigger.py
```diff
@@ -1,6 +1,6 @@
 from json import JSONDecodeError
 import os
-import unittest
+import pytest
 from unittest import mock
 from unittest.mock import MagicMock, patch
```
tamirkamara, line 5: should this just be 
```suggestion
from mock import MagicMock, patch
```
like you/we do elsewhere

airlock_processor/tests/shared_code/test_blob_operations.py
```diff
@@ -1,6 +1,6 @@
 from collections import namedtuple
 import json
-from unittest import TestCase
+import pytest
 from unittest.mock import MagicMock, patch
```
tamirkamara, line 4:
```suggestion
from mock import MagicMock, patch
```
?

tamirkamara (approved):

PR #3088, opened 2023-01-17
Update Nexus template docs for V2

### LINKED ISSUE #2617: Nexus Shared Service docs need updating for v2
The Nexus docs listed here - https://microsoft.github.io/AzureTRE/tre-templates/shared-services/nexus/#deploy are not appropriate for the VM based nexus deployment. 

They do not cover certificate installation and refer to running setup scripts.

### LINKED ISSUE #2701: Documentation: Shared service docs don't include makefile deployment step
**Describe the bug**
Currently the docs at https://github.com/microsoft/AzureTRE/blob/main/docs/tre-admins/setup-instructions/configuring-shared-services.md mention configuring shared services and how to deploy using Swagger, but they don't include instructions using the make command, `make deploy-shared-service` (which was removed from the `make all` path). This has led a few people now to miss this step when configuring CICD scenarios and wonder why things aren't set up for Guacamole and other things that depend on one of the shared services.

We should include the deployment steps as part of the set up guide for each shared service and change the title from configuring shared services to deploying them. We should also consider highlighting the dependencies on Nexus for the Guacamole Workspace Service to ensure there is not confusion during set up.

jjgriff93:
# Resolves #2617 

Also resolves #2701 

## What is being addressed

This PR updates the template docs for Nexus to cover the V2 service instead of V1. It also moves a lot of unnecessary information out of the Quickstart and into the template doc. I've also updated the Gitea docs as part of this so it's consistent.

## How is this addressed

- Update Nexus template doc with V2 URLs, new repositories and deploy steps
- Move upgrade path notes and other unnecessary info from the Quickstart to the template doc
- Do the same for Gitea


martinpeck (approved):
Thanks @jjgriff93

PR #3089, opened 2023-01-18
Support ip group names in dynamic firewall rules

### LINKED ISSUE #2328: NexusV2 & Certs configures firewall with deployment pipeline
## Description

As a TRE Developer
I want to configure required firewall rules with the deployment pipeline
So that firewall rules will be managed centrally in the shared service

## Acceptance criteria

- [ ] Hardcoded rules are removed from Terraform files
- [ ] Required rules are present in firewall on azure and in state of the firewall shared service
- [ ] Both install and uninstall work

tamirkamara:
Resolves #2328 

## What is being addressed

Some services require firewall rules while deploying, i.e. open a new rule for the resource processor. The Pipeline method of creating rules dynamicly didn't allow for this so templates like Certs, Airlock-Notifier created rules directly via terraform which caused issues when running in parallel with another firewall change.

## How is this addressed

- Create an ip group for the resource processor subnet
- Include a _source_ip_groups_in_core_ property for the rules defined by another template to be able to specify ip groups present in the core resource group
- Update the firewall template to be able to use these referenced ip groups
- Update Certs and Airlock Notifier templates to use this new capability
- Re-enable Certs and Nexus to run in the shared services E2Es


eladiw (approved):

tamirkamara:
/test-force-approve
gave up with the certs bundle in the ci environments. will pick it up separately.

PR #3096, opened 2023-01-19
secretName pattern error in Airlock Notifier deployment

### LINKED ISSUE #3077: secretName pattern error in Airlock Notifier deployment
**Describe the bug**

3fc56153-03bb-4cc9-8d38-243b00090e74: Error message: rpc error: code = Unknown desc = failed to set secret for key 01GPRDS7A5P71EC0MH3DGP59RN-smtp_password in azure-keyvault: keyvault.BaseClient#SetSecret: Invalid input: autorest/validation: validation failed: parameter=secretName constraint=Pattern value="01GPRDS7A5P71EC0MH3DGP59RN-smtp_password" details: value doesn't match pattern ^[0-9a-zA-Z-]+$ failed to save sensitive param to secrete store: rpc error: code = Unknown desc = failed to set secret for key 01GPRDS7A5P71EC0MH3DGP59RN-smtp_password in azure-keyvault: keyvault.BaseClient#SetSecret: Invalid input: autorest/validation: validation failed: parameter=secretName constraint=Pattern value="01GPRDS7A5P71EC0MH3DGP59RN-smtp_password" details: value doesn't match pattern ^[0-9a-zA-Z-]+$ ; Command executed: az login --identity -u 1d9ec1e8-f9bb-4be4-aabd-d4a35edad118 && az acr login --name acrtamir02ops && porter install "3fc56153-03bb-4cc9-8d38-243b00090e74" --reference acrtamir02ops.azurecr.io/tre-shared-service-airlock-notifier:v0.3.4 --param arm_use_msi="true" --param id="3fc56153-03bb-4cc9-8d38-243b00090e74" --param smtp_from_email="a@a.com" --param smtp_password="REDACTED" --param smtp_server_address="a.b.com" --param smtp_username="username" --param tfstate_container_name="tfstate" --param tfstate_resource_group_name="rg-tamir02-ops" --param tfstate_storage_account_name="tamir02ops" --param tre_id="tk21a" --allow-docker-host-access --force --credential-set arm_auth --credential-set aad_auth

**Steps to reproduce**

1. Deploy an airlock notifier

### LINKED ISSUE #3307: [GOV] Disable RuntimeAuditLogs and ApplicationMetricsLogs on Gov Cloud
## Description

We saw that in `core/terraform/servicebus.tf` the `"RuntimeAuditLogs", "ApplicationMetricsLogs"` failed on Gov Cloud with the following error: 
`│ Error: creating Monitor Diagnostics Setting "diagnostics-sb-lishaktre" for Resource "/subscriptions/2f921dae-fe64-472a-b21b-100a0ccf57ab/resourceGroups/rg-lishaktre/providers/Microsoft.ServiceBus/namespaces/sb-lishaktre": diagnosticsettings.DiagnosticSettingsClient#CreateOrUpdate: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="BadRequest" Message="Category 'ApplicationMetricsLogs' is not supported."`

As a Azure TRE Developer
I don't want to enable these logs when deploying on Azure Gov Cloud
So that the deployment won't fail . 

Technical Notes:
There is a way to check what the service supports. See as reference the app services we deploy.

## Acceptance criteria

- [ ] RuntimeAuditLogs and ApplicationMetricsLogs enabled on Azure Public Cloud 
- [ ] RuntimeAuditLogs and ApplicationMetricsLogs are not enabled/passed on Azure Gov Cloud

anatbal:
# Resolves #3077 
## What is being addressed

This bug is caused due to the `sensitive: true`. According to porter [documentation](https://getporter.org/blog/persist-sensitive-data-safely/), since we migrated to version 1.0.0, the behavior of the secret plugin has changed. It is now trying to save the smtp password in the kv, but since the name of the var is `SOMEID-smtp_password`, the underscore is not a leagl char in a kv secret name.
Hence, we have two options:
1. Take the approach implemented in this PR: changing smtp_password to smtpPassword. I made the changes everywhere so it will be consistent, but it still different than our convention.
2. Consider deleting sensitive: true attribute. I am not sure how necessary it is, since the porter documentation only states it save it in kv, but nothing more. In addition, it's the only place in the code we use this attribute. 

@tamirkamara , @marrobi , @SharonHart  WDYT?

### DIFF
--- a/templates/shared_services/airlock_notifier/porter.yaml
+++ b/templates/shared_services/airlock_notifier/porter.yaml
@@ -1,7 +1,7 @@
 ---
 schemaVersion: 1.0.0
 name: tre-shared-service-airlock-notifier
-version: 0.3.0
+version: 0.3.1
 description: "A shared service notifying on Airlock Operations"
 registry: azuretre
 dockerfile: Dockerfile.tmpl
@@ -43,7 +43,7 @@ parameters:
   - name: smtp_username
     env: SMTP_USERNAME
     type: string
-  - name: smtp_password
+  - name: smtpPassword
     env: SMTP_PASSWORD
     type: string
     sensitive: true
@@ -71,7 +71,7 @@ install:
         tre_url: ${ bundle.parameters.tre_url }
         smtp_server_address: ${ bundle.parameters.smtp_server_address }
         smtp_username: ${ bundle.parameters.smtp_username }
-        smtp_password: ${ bundle.parameters.smtp_password }
+        smtpPassword: ${ bundle.parameters.smtpPassword }
         smtp_from_email: ${ bundle.parameters.smtp_from_email }
       backendConfig:
         resource_group_name: ${ bundle.parameters.tfstate_resource_group_name }
@@ -121,7 +121,7 @@ uninstall:
         tre_url: ${ bundle.parameters.tre_url }
         smtp_server_address: ${ bundle.parameters.smtp_server_address }
         smtp_username: ${ bundle.parameters.smtp_username }
-        smtp_password: ${ bundle.parameters.smtp_password }
+        smtpPassword: ${ bundle.parameters.smtpPassword }
         smtp_from_email: ${ bundle.parameters.smtp_from_email }
       backendConfig:
         resource_group_name: ${ bundle.parameters.tfstate_resource_group_name }
--- a/templates/shared_services/airlock_notifier/porter.yaml
+++ b/templates/shared_services/airlock_notifier/porter.yaml
@@ -71,7 +71,7 @@ install:
         tre_url: ${ bundle.parameters.tre_url }
         smtp_server_address: ${ bundle.parameters.smtp_server_address }
         smtp_username: ${ bundle.parameters.smtp_username }
-        smtpPassword: ${ bundle.parameters.smtpPassword }
+        smtp_password: ${ bundle.parameters.smtpPassword }
         smtp_from_email: ${ bundle.parameters.smtp_from_email }
       backendConfig:
         resource_group_name: ${ bundle.parameters.tfstate_resource_group_name }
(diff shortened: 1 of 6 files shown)


templates/shared_services/airlock_notifier/porter.yaml
```diff
@@ -43,7 +43,7 @@ parameters:
   - name: smtp_username
     env: SMTP_USERNAME
     type: string
-  - name: smtp_password
+  - name: smtpPassword
     env: SMTP_PASSWORD
     type: string
     sensitive: true
```
anatbal, line 49: This is the attribute causing the change.

tamirkamara: I don't remember why @SharonHart marked it as sensitive at the time, but now:
1. If not marked as sensitive, it will be saved in clear text into the porter database which I think isn't desired.
2. We can (should) use this property when generating the template sent to the API and not just count on the developer to mark the property as sensitive in the template_schema.json file.

templates/shared_services/airlock_notifier/porter.yaml
```diff
@@ -71,7 +71,7 @@ install:
         tre_url: ${ bundle.parameters.tre_url }
         smtp_server_address: ${ bundle.parameters.smtp_server_address }
         smtp_username: ${ bundle.parameters.smtp_username }
-        smtp_password: ${ bundle.parameters.smtp_password }
+        smtpPassword: ${ bundle.parameters.smtpPassword }
```
tamirkamara, line 74: I'm not sure which option is best...
but you can keep the terraform variable name consistent with the rest of them.

tamirkamara (approved):

templates/shared_services/airlock_notifier/porter.yaml
anatbal: done

PR #3106, opened 2023-01-23
Fix certs shared service and add back e2e test

tamirkamara:
## What is being addressed

The certs shared service would fail sometimes to install and to renew the certificate.

## How is this addressed

- Update the terraform output step
- Application gateway stop/start actions as porter steps (remove from terraform and script)
- Add back the tests


jjgriff93 (approved):
Looks great!

PR #3107, opened 2023-01-23
Firewall Policy

### LINKED ISSUE #2723: Investigate if Basic Azure Firewall SKU can be used

### LINKED ISSUE #3100: Use Azure Firewall Policies
## Description

I want to use _Firewall Policies_
So that the solution will be up-to-date with the latest and recommended way to define firewall rules.

tamirkamara:
Resolves #3100 Resolves #2723

## What is being addressed

Update the firewall implementation to use policies (instead of classic rules).

## How is this addressed

- Use a firewall policy resource and define rules in it
- Support for `Basic` firewall SKU. The default is still `Standard` as in most cases a standard one will be cheaper (since a basic one can't be deallocated/stopped).

An example policy:
image


templates/shared_services/firewall/porter.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 schemaVersion: 1.0.0
 name: tre-shared-service-firewall
-version: 0.8.0
+version: 0.8.1 # TODO: update before merge
```
tamirkamara, line 4: keeping this open to remember

```suggestion
version: 1.0.0
```

eladiw (approved):

tamirkamara:
/test-force-approve
test was successful but produced an error in teardown which is unrelated to this PR.

PR #3113, opened 2023-01-24
Fix airlock assigned only with groups

### LINKED ISSUE #3093: Error creating airlock request when groups assigned to roles
**Describe the bug**
Having groups only assigned to Researcher application role, when trying to create an Airlock Request an error returned: "No researchers with an email address"

**Steps to reproduce**

1. Create a group in AAD
2. Assign it to Researcher application role
3. Make sure it is the only one assigned
4. Try to create an airlock request

LizaShak:
# Resolves #3093 

## What is being addressed

Fix airock requests creation when assigned only with groups.

## How is this addressed

- Add the id of group as part of emails response


api_app/_version.py
```diff
@@ -1 +1 @@
-__version__ = "0.9.1"
+__version__ = "0.9.2"
```
tamirkamara, line 1:
```suggestion
__version__ = "0.10.0"
```

tamirkamara (approved):

PR #3119, opened 2023-01-25
Move non-core DNS zones outside network module

tamirkamara:
## What is being addressed

We create many objects in the network module that cause delays on its completion but aren't really required to be provisioned there.

## How is this addressed

- Split out all DNS zones not required for the core applications, and move them out of the network module. This allows TF to better manage dependencies and on some scenarios expedite deployment (as many things are waiting for the network module).


core/terraform/locals.tf
```diff
@@ -12,4 +12,24 @@ locals {
...
+  private_dns_zone_names_non_core = toset([
+    "privatelink.purview.azure.com",
+    "privatelink.purviewstudio.azure.com",
+    "privatelink.sql.azuresynapse.net",
+    "privatelink.dev.azuresynapse.net",
+    "privatelink.azuresynapse.net",
+    "privatelink.dfs.core.windows.net",
+    "privatelink.azurehealthcareapis.com",
+    "privatelink.dicom.azurehealthcareapis.com",
+
```
guybartal, line 27: is there a reason for this empty line?

tamirkamara: there was but I don't remember. I'll remove it.

core/terraform/migrate.sh
```diff
@@ -189,6 +189,43 @@ if [ -n "${rp_subnet}" ]; then
...
+do
+  dns_zone_rid=$(echo "${terraform_show_json}" \
+    | jq -r "select(.values.root_module.child_modules != null) .values.root_module.child_modules[] | select (.address==\"module.network\") | .resources[] | select(.address==\"${zones[$value]}\") | .values.id")
+  if [ -n "${dns_zone_rid}" ]; then
+    terraform state rm "${zones[$value]}"
+    terraform import azurerm_private_dns_zone.non_core[\""$value"\"] "${dns_zone_rid}"
+  fi
+done
+
+# module.network.azurerm_private_dns_zone_virtual_network_link.mysql
```
guybartal, line 211: maybe change comment to `import state if module.network.azurerm_private_dns_zone_virtual_network_link.mysql exists`

guybartal (approved):
LGTM

core/terraform/migrate.sh
tamirkamara: I'll delete it too since every single block works in the same way here.

PR #3130, opened 2023-01-26
Clean up review VMs for cancelled Airlock requests

### LINKED ISSUE #2829: Clean up any review VMs if the request is cancelled
## Description

As a TRE Administrator / Airlock Manager
I want to make sure the review VM is cleaned up if the request is canceled
So that it is not left hanging and using up resources

## Acceptance criteria

- [ ] Review VM is cleaned up if the request is canceled

yuvalyaron:
# Resolves #2829 

## What is being addressed

Review VMs are being cleaned up when an Airlock request is canceled.

## How is this addressed
- Moved the logic of cancelling airlock requests from the route to the service and added a call to delete review VMS
- Added "disable VM" call to the method that deletes the review VM - this is necessary as disabling the resource might carry important logic, and this is what we force users to do in the UI (you can't delete a resource without disabling first).


api_app/_version.py
```diff
@@ -1 +1 @@
-__version__ = "0.9.2"
+__version__ = "0.9.3"
```
tamirkamara, line 1: This is an added functionality:
```suggestion
__version__ = "0.10.0"
```

api_app/services/airlock.py
```diff
@@ -401,6 +407,31 @@ async def delete_review_user_resource(
     return operation
 
 
+async def disable_user_resource(
```
anatbal, line 410: why not reuse the patch_user_resource func?

yuvalyaron: what do you mean by reuse? `patch_user_resource` is used here

anatbal: I mean the function that is under the route, which is basically copied here , in this file /workspaces/AzureTRE/api_app/api/routes/workspaces.py

yuvalyaron: changed, moved the shared code to one method and used in both places

e2e_tests/test_airlock.py
```diff
@@ -73,10 +73,10 @@ async def submit_airlock_import_request(workspace_path: str, workspace_owner_tok
...
 @pytest.mark.timeout(50 * 60)
 @pytest.mark.airlock
-async def test_airlock_review_vm_flow(setup_test_workspace_and_workspace_service, setup_test_airlock_import_review_workspace_and_workspace_service, verify):
-    workspace_path, workspace_id, workspace_service_path, workspace_service_id, workspace_owner_token = setup_test_workspace_and_workspace_service
-    _, import_review_workspace_id, _, import_review_workspace_service_id, _ = setup_test_airlock_import_review_workspace_and_workspace_service
-    LOGGER.info("Workspace and workspace service set up")
+async def test_airlock_review_vm_flow(setup_test_workspace, setup_test_airlock_import_review_workspace_and_guacamole_service, verify):
+    workspace_path, workspace_id, workspace_owner_token = setup_test_workspace
+    _, import_review_workspace_id, _, import_review_workspace_service_id, _ = setup_test_airlock_import_review_workspace_and_guacamole_service
+    LOGGER.info("Base workspace set up")
```
tamirkamara, line 79:
```suggestion
    LOGGER.info("Import review workspace & Guacamole setup is done.")
```

yuvalyaron: at this point both workspaces are ready, shouldn't we mention both of them? 
`Workspaces setup is done.`

Or the elaborated one:

`Base workspace and Review workapace with Guacamole are set up`

tamirkamara: isn't the fixture(s) anyway log that it did its work?

yuvalyaron: yes, they do, so we can remove it,
but on the other hand, it's sometimes convenient to look for a log that was generated by the test itself, to know where the real test starts

anatbal (approved):
LGTM

e2e_tests/test_airlock.py
yuvalyaron: removed

PR #3137, opened 2023-01-29
CLI to Reauthenticate when TRE endpoint changes

### LINKED ISSUE #3095: tre cli doesn't refresh old environment token and api uri when creating a new environment
**Describe the bug**
tre cli doesn't refresh old environment token and api uri when creating a new environment

**Steps to reproduce**

1. deploy a TRE environment.
1. change tre id in config.yml and deploy a new environment.
1. deployment will fail on registering the FW because the tre uri and token still points to the old environment (see error below).

**Workaround until fixed:**
change the condition `if [[ $got_token_from_cli == "0" ]]; then` to `if [[ $got_token_from_cli == "1" ]]; then` in file `ensure_cli_signed_in.sh` line number 10 to force tre cli to login to the new environment.

image

tamirkamara:
Resolves #3095 

## What is being addressed

CLI saves the TRE endpoint in a local config file. If one changes their TRE_ID then the CLI will continue to use the old endpoint which might result in errors.

## How is this addressed

- In `ensure_cli_signed_in.sh`, check that the cached endpoint is equal to the one correctly configured in the environment.
- Delete the CLI config file after each `make deploy-core` as that might be also an TRE_ID (endpoint) change. This to catch cases where people run a tre cli command directly (without the script mentioned in the previous step)


guybartal (approved):
LGTM

PR #3143, opened 2023-01-30
Consolidate http status assertion in E2Es

tamirkamara:
## What is being addressed

On some of the E2E tests an assertion has been made over the http status code without logging the message that came with it. This made it hard to understand the exact error.

## How is this addressed

- Build a single method to handle all http status checks that will also print out the text in case of an error.

### DIFF
--- a/e2e_tests/test_workspace_templates.py
+++ b/e2e_tests/test_workspace_templates.py
@@ -33,4 +33,4 @@ async def test_get_workspace_templates(template_name, verify) -> None:
 async def test_get_workspace_template(template_name, verify) -> None:
     admin_token = await get_admin_token(verify)
     async with get_template(template_name, strings.API_WORKSPACE_TEMPLATES, admin_token, verify) as response:
-        assert_status(response, [status.HTTP_200_OK], f"Failed to create template: {template_name}")
+        assert_status(response, [status.HTTP_200_OK], f"Failed to GET template: {template_name}")
--- a/e2e_tests/test_workspace_templates.py
+++ b/e2e_tests/test_workspace_templates.py
@@ -4,7 +4,7 @@ from httpx import AsyncClient
 from starlette import status
 
 import config
-from helpers import get_auth_header, get_template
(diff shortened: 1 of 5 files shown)


eladiw (approved):

e2e_tests/test_workspace_templates.py
```diff
@@ -33,4 +33,4 @@ async def test_get_workspace_templates(template_name, verify) -> None:
 async def test_get_workspace_template(template_name, verify) -> None:
     admin_token = await get_admin_token(verify)
     async with get_template(template_name, strings.API_WORKSPACE_TEMPLATES, admin_token, verify) as response:
-        assert (response.status_code == status.HTTP_200_OK), f"GET Request for {template_name} creation failed"
+        assert_status(response, [status.HTTP_200_OK], f"Failed to create template: {template_name}")
```
yuvalyaron, line 36: does it necessarily apply that the creation has failed?

yuvalyaron (changes requested):

e2e_tests/test_workspace_templates.py
tamirkamara: @yuvalyaron what do you mean?

yuvalyaron: if we fail to get the workspace template - we see a message that says that the creation of the template has failed,
I'm asking if there are other options. 
i.e. the creation was ok, but there's a problem with the token, or with the endpoint.
Is it ok to assume that it's the creation that failed?

yuvalyaron: because the test is not creating the template, it just calls GET on it

tamirkamara: I made a mistake in the error message. it should be:
```suggestion
        assert_status(response, [status.HTTP_200_OK], f"Failed to GET template: {template_name}")
```

yuvalyaron (approved):

PR #3147, opened 2023-01-30
Set peer to exclude VirtualNetwork tag and allow all traffic within a workspace.

### LINKED ISSUE #3065: Simplify NSGs applied to workspace VNets by removing peer from virtual network tag
At present the workspace to core vnet peer has this setting set to Allow:
image

This makes the core VNet part of the `VirtualNetwork` tag used by network security groups. This means within NSGs we need to keep track of workspace address spaces, and if the  `VirtualNetwork`  is used inadvertently it also allows access to the core network.

As we have explicit allow and deny rules on the NSGs changning this setting to `Block all traffic to the remote virtual network` should have no adverse affect. However this should be verified.

Once disabled NSG configurations can revert to using the `VirtualNetwork` tag which leads to easier to read definitions and simplifies Terraform.

marrobi:
# Resolves #3065

## What is being addressed

## How is this addressed

- Update documentation


marrobi:
This currently breaks connectivity from workspaces.

PR #3152, opened 2023-01-30
Migrate reviewDecision of Airlock Reviews

### LINKED ISSUE #3121: make db-migrate is not working
**Describe the bug**
We upgraded TRE CORE to v0.7.0 and in the release notes we saw this:

[0.7.0](https://github.com/microsoft/AzureTRE/releases/tag/v0.7.0)
BREAKING CHANGES & MIGRATIONS:

The airlock request object has changed. Make sure you have ran the db migration step after deploying the new API image and UI (which runs automatically in make all/make tre-deploy but can be manually invoked with **make db-migrate**) so that existing requests in your DB are migrated to the new model.

We executed **make db-migrate** command and we got this output:
image

But, still we have an issue when we want to review Airlock requests:

image

**Steps to reproduce**

1. Run **make db-migrate** on TEST environment
2. Log in to swagger for workspace and run GET /api/workspaces/{workspace_id}/requests from the airlock section

yuvalyaron:
# Resolves #3121 

## What is being addressed

Airlock reviews used to have one of these values: `approval_in_progress` and `rejection_in_progress`.
But since #2670 , instead of using these values, we use  `approved` and `rejected`,
This causes API type errors when trying to GET requests that have been created with older TRE versions, as these contain the older values.

## How is this addressed

- Added a DB migration to update each old value to it's updated version


api_app/api/routes/migrations.py
```diff
@@ -75,6 +75,10 @@ async def migrate_database(resources_repo=Depends(get_repository(ResourceReposit
         num_updated = await airlock_migration.change_review_resources_to_dict()
         migrations.append(Migration(issueNumber="XXXX", status=f'Updated {num_updated} airlock requests with new reviewUserResources format'))
 
+        logging.info("PR 3152 - Migrate reviewDecision of Airlock Reviews")
```
eladiw, line 78: I thing the PR xxx should not be in the log itself (maybe a comment)

api_app/api/routes/migrations.py
```diff
@@ -75,6 +75,10 @@ async def migrate_database(resources_repo=Depends(get_repository(ResourceReposit
...
+        num_updated = await airlock_migration.update_review_decision_values()
+        migrations.append(Migration(issueNumber="3121", status=f'Updated {num_updated} airlock requests with new reviewDecision value'))
```
eladiw, line 80: 3152 != 3121

api_app/api/routes/migrations.py
```diff
@@ -75,6 +75,10 @@ async def migrate_database(resources_repo=Depends(get_repository(ResourceReposit
         num_updated = await airlock_migration.change_review_resources_to_dict()
         migrations.append(Migration(issueNumber="XXXX", status=f'Updated {num_updated} airlock requests with new reviewUserResources format'))
```
eladiw, line 76: same here xxxx != 2883

eladiw (approved):
lgtm

api_app/api/routes/migrations.py
yuvalyaron: fixed

api_app/api/routes/migrations.py
yuvalyaron: fixed

PR #3176, opened 2023-02-01
UI roles fix

### LINKED ISSUE #3175: UI: Unhandled exception at app root when user not in a TRE role
The UI contains logic to check the user's role and nicely report any permission denied errors from the API. This doesn't seem to be applied / working at the root level. If the user is not in any TRE roles the UI fails and the `GenericErrorBoundary` "oops!" kicks in. 

image

**Steps to reproduce**

1. Take your user out of all TRE app roles
1. Open the UI, logging in from fresh

damoodamoo:
# Resolves #3175 

Added a null check to let the access denied logic continue to flow and cater for users in no roles.

### DIFF
--- a/ui/app/package.json
+++ b/ui/app/package.json
@@ -1,6 +1,6 @@
 {
   "name": "tre-ui",
-  "version": "0.3.0",
+  "version": "0.3.1",
   "private": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


marrobi (approved):
LGTM

PR #3204, opened 2023-02-07
Cascade disable and delete operations

### LINKED ISSUE #2152: Cascade disable and delete operations
When a workspace is disabled all child workspace services and user resources should be disabled.

- [ ] As a Workspace Owner I want to disable a workspace service, and all child user resources get disabled.
- [ ] As a TRE Administrator I want to disable a workspace, and all workspace services get disabled.

The same should be the case for deletion.

anatbal:
# Resolves #2152 

## What is being addressed

- Utilizing the existing way steps and operations today work with minimum changes to enable cascading for both disable and deletion.
- Fix existing unit tests - Need to add more coverage


api_app/_version.py
```diff
@@ -1 +1,2 @@
-__version__ = "0.11.4"
+__version__ = "0.12.0"
+
```
anatbal, line 2:
```suggestion
```

api_app/api/routes/workspaces.py
```diff
@@ -118,18 +117,23 @@ async def create_workspace(workspace_create: WorkspaceInCreate, response: Respon
...
+async def patch_workspace(resource_patch: ResourcePatch, response: Response, user=Depends(get_current_admin_user), workspace=Depends(get_workspace_by_id_from_path), workspace_repo: WorkspaceRepository = Depends(get_repository(WorkspaceRepository)), resource_template_repo=Depends(get_repository(ResourceTemplateRepository)), operations_repo=Depends(get_repository(OperationRepository)), resource_history_repo=Depends(get_repository(ResourceHistoryRepository)), etag: str = Header(...), force_version_update: bool = False) -> OperationInResponse:
     try:
-        patched_workspace, resource_template = await workspace_repo.patch_workspace(workspace, workspace_patch, etag, resource_template_repo, resource_history_repo, user, force_version_update)
+        cascade_enabled = not resource_patch.isEnabled
+        if cascade_enabled:
```
eladiw, line 123: I think cascade enabled is a bit misleading.
When I read it I thought that the user or someone decides whether or not they enable cascade.

The more acccurate terminology at least according to the code is 'should_perform_cascade' or 'should_propogate' 'is_cascade/propogate' something like this. Because it's not 'enablement' it's more about should you propagate the action

api_app/db/repositories/resources.py
```diff
@@ -124,6 +124,22 @@ async def patch_resource(self, resource: Resource, resource_patch: ResourcePatch
         await self.update_item_with_etag(resource, etag)
         return resource, resource_template
 
+    async def get_resource_dependecny_graph(self, resource: Resource) -> List:
+        # Get the parent resource path and id
+        parent_resource_path = resource.resourcePath
+        dependentResourcesList = []
+
+        # Get all related resources
+        related_resources_query = f"SELECT * FROM c WHERE CONTAINS(c.resourcePath, '{parent_resource_path}') AND c.deploymentStatus != '{Status.Deleted}' and c.deploymentStatus != '{Status.DeploymentFailed}'"
```
eladiw, line 133: should you sanitize the input before creating the query string?

api_app/api/routes/workspaces.py
anatbal: I agree. Changed. Let me know what you think

api_app/db/repositories/resources.py
anatbal: The input is the resource which is satanize by the method signature (line 127)

damoodamoo (approved):
LGTM!

api_app/api/routes/resource_helpers.py
```diff
@@ -26,6 +27,33 @@
 from services.authentication import get_access_service
 
 
+async def delete_validation(resource: Resource, resource_repo: ResourceRepository):
+    dependency_graph = await resource_repo.get_resource_dependecny_graph(resource)
+    for resource in dependency_graph:
```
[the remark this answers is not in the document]
anatbal: Yeah, I like dependency_list. changed

PR #3208, opened 2023-02-08
Align 'Airlock Get Container URL' permissions with 'Create Airlock Request's permissions

### LINKED ISSUE #3187: Align the permissions of Airlock GET container URL
## Description

As a workspace owner, the API will let you create an airlock request but will not let you get the container URL. 
We need to make sure that every role that is able to create an airlock request is able to upload files to it.

## Acceptance criteria
- [x] Every role that is able to create an airlock request, should also be able to get the container URL
- [x] Unit tests that verifies the previous requirement for all roles

yuvalyaron:
# Resolves #3187 

## What is being addressed

Verified that every role that is able to create an Airlock request is also able to upload a file to the request.

## How is this addressed

- Added `Workspace Owner` to the roles that can access the container link when the status is not `in_review`
- Added tests to verify that the roles are aligned


api_app/tests_ma/test_services/test_airlock.py
```diff
@@ -138,19 +139,12 @@ def sample_airlock_review(review_decision=AirlockReviewDecision.Approved):
     return airlock_review
 
 
-def get_required_roles(endpoint):
-    dependencies = list(filter(lambda x: hasattr(x.dependency, 'require_one_of_roles'), endpoint.__defaults__))
-    required_roles = dependencies[0].dependency.require_one_of_roles
-    return required_roles
```
yuvalyaron, line 144: moved to `conftest`

api_app/tests_ma/test_api/conftest.py
```diff
@@ -76,6 +76,12 @@ def override_get_user():
     return user
 
 
+def get_required_roles(endpoint):
+    dependencies = list(filter(lambda x: hasattr(x.dependency, 'require_one_of_roles'), endpoint.__defaults__))
+    required_roles = dependencies[0].dependency.require_one_of_roles
+    return required_roles
+
+
```
yuvalyaron, line 84: moved from `test_services/test_airlock.py`

LizaShak (approved):
LGTM :)

PR #3220, opened 2023-02-13
Remove the .diff file from airlock-import-review Template

yuvalyaron:
## What is being addressed

Replaced the ".diff" in the import review workspace with a Terraform file that contains the necessary resources for the review workspace.

## How is this addressed

- Replaced the .diff file with a .terraform file (see PR comments)
- Set the enable_airlock to false as Airlock is not required for this workspace type.


templates/workspaces/airlock-import-review/terraform/import_review_resources.terraform
```diff
@@ -0,0 +1,41 @@
+# This file has a .terraform file extension in order to avoid 'terraform init's validation checks that are executed by the 'make bundle-build' command.
+# The Dockerfile includes a RUN command to change the extension from .terraform to .tf after the files from the base workspace are copied to this directory.
```
yuvalyaron, line 2: as the comment explains, this terraform file is not complete as is, it requires the files from the base workspace template in order to pass the validations of `terraform init`

tamirkamara (approved):

PR #3224, opened 2023-02-14
Validate template pipeline has unique step IDs

### LINKED ISSUE #3212: Registering a template with duplicated stepId value succeed
**Describe the bug**
As for today, the stepId for a step in a pipeline that is described in the template json schema has an hardcoded stepId value. If giving the same stepId to more than one step (which will cause malfunction in the operation logic) it manages to register the template without any errors.

Need to allow Registering only if all stepIds are unique.

yuvalyaron:
# Resolves #3212 

## What is being addressed

Added a validation for pipeline templates to have unique step IDs.
A registration of a template with duplicated stepIds will result in HTTP 422.

## How is this addressed

- Added validation
- Added tests


anatbal (approved):
Looks really good!

PR #3227, opened 2023-02-16
Fix E2E shared services tests

### LINKED ISSUE #3226: E2E fail to uninstall shared services
**Describe the bug**

Looks like a recent change caused all shared services E2E tests to fail on uninstall.

https://github.com/microsoft/AzureTRE/actions/runs/4189772916/jobs/7262707823#step:3:1575

LizaShak:
# Resolves #3226 

## What is being addressed

Fix E2E shared services tests

Successful branch run for shred services e2e: https://github.com/microsoft/AzureTRE/actions/runs/4192515006


martinpeck:
Is this the right fix?
Isn't the better fix to re-order the parameters of the function, rather than reorder the variables passed to the function?

The param `verify` is an optional parameter for the things that use it. It's also not the "primary" thing you'd expect to pass to a method called `disable_and_delete_tre_resource`. I'd expect the resource to be the first thing I pass into `disable_and_delete_tre_resource`.

If you changed the order of the parameter in the definition of `disable_and_delete_tre_resource` then you'd also need to change the order of params for `get_workspace_owner_token` and  `disable_and_delete_ws_resource` but then all of these would be consistent with what I'd expect, and also consistent with other methods that have `verify` as the final parameter.

tamirkamara (approved):

tamirkamara (commented):
Need to address the comment from Martin.

martinpeck (approved):
Thanks @LizaShak!

e2e_tests/test_shared_services.py
```diff
@@ -134,7 +134,7 @@ async def test_create_shared_service(template_name, verify) -> None:
     )
 
     await disable_and_delete_tre_resource(
-        shared_service_path, verify
+        verify, shared_service_path
```
tamirkamara, line 137: Is this the right new order?

e2e_tests/test_shared_services.py
```diff
@@ -184,11 +184,11 @@ async def test_create_certs_nexus_shared_service(verify) -> None:
     )
 
     await disable_and_delete_tre_resource(
-        nexus_shared_service_path, verify
+        verify, nexus_shared_service_path
```
tamirkamara, line 187: Is this the right new order?

e2e_tests/test_shared_services.py
```diff
@@ -184,11 +184,11 @@ async def test_create_certs_nexus_shared_service(verify) -> None:
...
     )
 
     await disable_and_delete_tre_resource(
-        certs_shared_service_path, verify
+        verify, certs_shared_service_path
```
tamirkamara, line 191: Is this the right new order?

tamirkamara (approved):
I think there're places with the wrong order of vars being passed.

PR #3231, opened 2023-02-19
standardize airlock storage private endpoints

tamirkamara:
## What is being addressed

Private endpoint names for the airlock's storage account weren't using the same-consistent convention. 

## How is this addressed

- Update names so all will follow a similar pattern.
- Update private dns zone group name to reduce a chance of accidental conflict (a bug we had recently).


guybartal (approved):
LGTM

Issue #3118, opened 2023-01-25
Allow users to upgrade resources through the UI


guybartal:
## Description

As a TRE Administrator/TRE Workspace Owner/Researcher 
I want to be able to upgrade resources that have been deployed through the UI,
so that I can easily keep them up-to-date and take advantage of new features and bug fixes. 

## Acceptance criteria
- [ ] List the versions that we can upgrade to when we GET a resource
- [ ] Allow users to choose the version they want to upgrade to within the resource card, based on their role.

marrobi:
@guybartal moved onto the feature roadmap. Would be useful, but not sure its an immediate priority.

yuvalyaron:
For now, we will not let users upgrade to major versions through the UI, since it might break the resource so we should not make it too accessible.

PR #3234, opened 2023-02-20
Add 'availableUpgrades' field to Resources

yuvalyaron:
# Part of #3118 

## What is being addressed

Added `availableUpgrades` property to Resources.

The field lists versions of the template that are available for upgrade - these are versions higher than the current template version of the resource.

Each element in the list contains 2 fields:
`version` - the template version of the upgrade
`forceUpdateRequired` - whether the upgrade requires the `forceUpdate` flag to be sent with the update request (`true` should be sent for major upgrades and `false` otherwise).

## How is this addressed

- Added the `availableUpgrades` field
- Refactored all of the `GET` and `GET all` routes and added a call to the enrichment logic to populate the value
-  Fixed and added tests


guybartal (commented):
The logic you've implemented is very well, one thing I would consider changing - 
If we want to hide the logic behind the API when it's required to pass the force update flag from the client then I would either combine the two version lists with another property besides the version called `forceUpdateRequired` or at least change the names of the lists to `availableVersions{forceUpdateRequired[], nonForceUpdateRequired[]}`.

WDYT?

guybartal (approved):
Looks great! 👌

PR #3243, opened 2023-02-22
Support dynamic resource suffix in Airlock

### LINKED ISSUE #3237: Airlock should support dynamic resource suffix
## Description

As a TRE Administrator developer
I want to make sure that when the resources suffix will be dynamic (i.e. won't always be 4 chars) the airlock processor will be able to find target resources
So that it will continue to function

## Acceptance criteria

- [ ] airlock processor should not assume the storage account suffix, but get it from the message itself

yuvalyaron:
# Resolves #3237 

## What is being addressed

- The code in the airlock processor references certain storage accounts by assuming that their names have the workspace short ID as a suffix. Rather than making this assumption, the suffix was changed to be passed in the 'status changed' message.

- Added Migration to add the `unique_identifier_suffix` field to all resources. As the changes in #2893 expect this field to exist in all resources. The value that the migration is adding is the last 4 digits of the resource as this was the value that we used before the changes.

## How is this addressed

- Added migration to add the `unique_identifier_suffix` field.
- Replaced workspace ID with `unique_identifier_suffix` in airlock processor
- Fixed unit tests


airlock_processor/StatusChangedQueueTrigger/__init__.py
```diff
@@ -18,7 +18,7 @@ class RequestProperties(BaseModel):
     new_status: str
     previous_status: Optional[str]
     type: str
-    workspace_id: str
+    unique_identifier_suffix: str
```
tamirkamara, line 21: Across much of this change, isn't the _unique_identifier_suffix_ always the **workspace** one?
If so, shouldn't we call it like that?

yuvalyaron: changed

eladiw (approved):
Lgtm

marrobi:
@yuvalyaron @tamirkamara can you recall where this got to? We are hitting some storage account name conflicts on the airlock accounts...

` import_approved_storage_name = lower(replace("stalimapp${substr(local.workspace_resource_name_suffix, -8, -1)}", "-", ""))`

This needs to include say the `tre_id` to reduce the risk of a conflict.

tamirkamara:
> @yuvalyaron @tamirkamara can you recall where this got to? We are hitting some storage account name conflicts on the airlock accounts...
> 
> ` import_approved_storage_name = lower(replace("stalimapp${substr(local.workspace_resource_name_suffix, -8, -1)}", "-", ""))`
> 
> This needs to include say the `tre_id` to reduce the risk of a conflict.

This PR depends on another large development we paused for now. It looks for a workspace property that notes the random suffix of resource names like the storage accounts. 
So this PR specifically should remain agnostic to what value is chosen for that suffix.

martinpeck:
Adding label `spike` as the code still has value.
Associated with issue #3243 3237

PR #3264, opened 2023-02-28
Fix updating a workspace due to cascade bug

### LINKED ISSUE #3263: Updating a workspace fails when workspace has workspace service/user resources
**Describe the bug**
image
Cannot update workspace with the following error.
**Steps to reproduce**

1. Have a workspace with at least one workspace service
2 Try updating the workspace

anatbal:
# Resolves #3263
## What is being addressed

The cascade disable condition was malformed and when a workspace had a workspace service it tried to do a cascade update with malformed values.


marrobi (approved):
LGTM

Issue #3287, opened 2023-03-01
[GOV]  Authority passed to services should be dynamically sourced


LizaShak:
## Description

As a Azure TRE Developer
I want to eliminate the hardcoded `login.microsoftonline.com`  passed as authority
So that I can deploy our solution in multiple clouds

Technical notes:
- Search for all the places where `login.microsoftonline.com` is passed as authority/issuer
- Think of a design how to source it dynamically or with a config that easy to change when deployed on another cloud

## Acceptance criteria

- [ ] login.microsoftonline.com is sourced  based on the target cloud type

tamirkamara:
Is this the same Authority that corresponds to one in which the workspace app was create in?

PR #3300, opened 2023-03-02
Add support for setting the cloud environment in TRE CLI

### LINKED ISSUE #3293: [GOV]  Use auth libraries to authenticate TRE cli
## Description

As a Azure TRE Developer
I want to replace hardcoded login endpoints calls and use existing libraries which we use in other places to authenticate
So that our code is easier to maintain.

Technical notes: 
- Replace the REST calls to `login.microsoftonline.com`
- Consider the following as a replacement: https://learn.microsoft.com/en-us/python/api/azure-identity/azure.identity.clientsecretcredential?view=azure-python

## Acceptance criteria

- [ ] No hardcoded login endpoints used in cli

yuvalyaron:
# Resolves #3293 

## What is being addressed

Added support for multiple cloud environments in TRE CLI

## How is this addressed

- The cloud environment is set according to the cloud environment that was set for Azure CLI (and will be the public Azure cloud if Azure CLI not installed)
- Replaced REST calls that were used to retrieve the tokens with ClientSecretCredential
- Not related to the task: renamed the CLI project, as the project `tre` already exists in PyPI


cli/tre/api_client.py
```diff
@@ -189,7 +180,7 @@ def get_auth_token(self, log, scope):
 
         app = msal.PublicClientApplication(
             client_id=self._client_id,
-            authority=f"https://login.microsoftonline.com/{self._aad_tenant_id}",
+            authority=AuthorityBuilder(AZURE_PUBLIC, self._aad_tenant_id),
```
tamirkamara, line 183: But this won't work in gov. Shouldn't we fix that now as well?

cli/tre/commands/login.py
```diff
@@ -253,30 +255,19 @@ def _get_auth_token_client_credentials(
...
+
+        credential = ClientSecretCredential(aad_tenant_id, client_id, client_secret, connection_verify=verify)
+        token = event_loop.run_until_complete(credential.get_token(f'{api_scope}/.default'))
+        event_loop.run_until_complete(credential.close())
+
+        event_loop.close()
+        return token.token
+    except Exception as ex:
+        log.error(f"Failed to authenticate: {ex}")
+        raise RuntimeError("Failed to get auth token")
```
tamirkamara, line 270: This looks similar to the code block above. Have you thought about making a method out of this?

cli/tre/commands/login.py
```diff
@@ -95,7 +97,7 @@ def login_device_code(base_url: str, client_id: str, aad_tenant_id: str, api_sco
 
     app = msal.PublicClientApplication(
         client_id=client_id,
-        authority=f"https://login.microsoftonline.com/{aad_tenant_id}",
+        authority=AuthorityBuilder(AZURE_PUBLIC, aad_tenant_id),
         token_cache=cache)
```
tamirkamara, line 101: This looks similar to the code block above. Have you thought about making a method out of this?
When we introduce support for the gov cloud, it will make even more sense...

cli/tre/api_client.py
yuvalyaron: The goal of this task is to remove the hardcoded URLs,
the next task is about supporting the gov cloud.
the usage of ClientSecretCredential (see code bleow) also requires modifications to support the gov cloud

tamirkamara: Not sure I understand. I get that this way looks better, but why would we do it if not to support the gov cloud? Why should we split it to 2 PRs and 2 tasks?

cli/tre/commands/login.py
yuvalyaron: moved to shared method

cli/tre/commands/login.py
yuvalyaron: moved to shared method

cli/tre/api_client.py
yuvalyaron: The rationale was to start with a task without waiting for the decision of how to handle the parameter that controls the cloud type.
It can be done as part of this task, but after the modifications, the work that has to be done is the same as the work that has to be done as part of #3287.

cli/tre/commands/login.py
```diff
@@ -118,6 +119,7 @@ def login_device_code(base_url: str, client_id: str, aad_tenant_id: str, api_sco
         'aad-tenant-id': aad_tenant_id,
         'api-scope': api_scope,
         'verify': verify,
+        'cloud': cloud,
```
tamirkamara, line 122: So the new param is only relevant for login?
Have you tested that a logged-in cli of the previous version is still functional after this new param has been added without doing a new login?

tamirkamara (approved):

cli/tre/commands/login.py
yuvalyaron: it's only relevant for login and it will fail if we have an already logged-in cli session.
However, I have found a better way to get the cloud environment - we can get it from az CLI using this package:
https://pypi.org/project/azure-cli-core/

Issue #3308, opened 2023-03-04
[GOV]  Porter should source keyvault uri


LizaShak:
## Description

As a Azure TRE Developer
I want to key vault url param in porter 
So that I can pass it dynamically.

Technical notes:
Currently in resource processor we get the secrets from key vault by using its name only: `AzureTRE/resource_processor/vmss_porter/config.yaml`. We want to use the new key vault url param instead. 
For that we will need to output it as part of key vault creation and source it in the porter env `resource_processor/shared/config.py` in ` config["porter_env"]`

## Acceptance criteria

- [ ] Porter uses key vault url and sources the url dynamically

PR #3312, opened 2023-03-05
Bump azure plugins version for porter and use the new valut-url param

anatbal:
# Resolves #[3308](https://github.com/microsoft/AzureTRE/issues/3308)

## What is being addressed

- Bump azure plugins version from 1.1.0 to 1.2.0
- use vault-url  param instead of vault in config.yaml for porter
- return vault uri from terraform instead of vault name for cloud-init script


LizaShak (approved):
LGTM!

resource_processor/run.sh
```diff
@@ -28,7 +28,7 @@ name = "aad_auth"
 plugin = "azure.keyvault"
 
 [secrets.config]
-vault = "${KEY_VAULT_NAME}"
+vault-url = "${KEY_VAULT_URL}"
```
tamirkamara, line 31: Since this is for porter v0 with an older plugin, I don't think you should change it.
porter v1 is using the config.yaml defined as a file.

resource_processor/shared/config.py
```diff
@@ -45,7 +45,7 @@ def get_config(logger_adapter) -> dict:
     config["porter_env"] = {
         "HOME": os.environ["HOME"],
         "PATH": os.environ["PATH"],
-        "KEY_VAULT_NAME": config["key_vault_name"],
+        "KEY_VAULT_URL": config["key_vault_url"],
```
tamirkamara, line 48: I'm not 100% we should change this one.
Where does porter use this environment variable?

tamirkamara (changes requested):

resource_processor/run.sh
anatbal: I reverted it to the same way that it was(the comment is up there)

tamirkamara (approved):

Issue #3299, opened 2023-03-02
Design and Adjust our CI/CD pipelines to support multiple clouds


LizaShak:
## Description

As a Azure TRE Developer
I want to understand what changes need to be done in CI/CD pipelines to deploy our solution  and apply them
So that we support multiple clouds in our pipelines.

Technical notes: 
- See what are the required changes and share with the team

## Acceptance criteria

- [ ] Deployment repo pipelines are able to deploy on gov cloud

PR #3317, opened 2023-03-06
Application Gateway FQDN should be sourced dynamically 

### LINKED ISSUE #3304: [GOV]  Application Gateway FQDN should be sourced dynamically
## Description

As a Azure TRE Developer
I want our code not to use the hardcoded ".cloudapp.azure.com" but to source it dynamically
So that it can be dynamically changed when the suffix changes.

anatbal:
# Resolves #3304 

## What is being addressed
changes to allow the code not to use the hardcoded ".cloudapp.azure.com" but to source it dynamically

-  tf changes:
    - [x] core
    - [x] airlock_notifier
    - [x] cyclecloud
    - [x] nexus
    - [x] base ws
    - [x] guac-windows
    - [x] guac-linux
- [x] e2e changes
- [x] scripts changes


devops/scripts/load_env.sh
```diff
@@ -40,5 +40,9 @@ else
   done < <(grep -v -e '^[[:space:]]*$' -e '^#' "$1" ) # feed in via Process Substition to avoid bash subshell (http://mywiki.wooledge.org/ProcessSubstitution)
 fi
 
+if [ -n "${TRE_ID:-}" ] && [ -n "${LOCATION:-}" ]; then
```
tamirkamara, line 43: Don't we always use load_and_validate_env before lod_env which makes TRE_URL always available?
Should you maybe have a condition on TRE_URL too?

core/terraform/appgateway/appgateway.tf
```diff
@@ -224,4 +224,11 @@ resource "azurerm_monitor_diagnostic_setting" "agw" {
   lifecycle { ignore_changes = [log_analytics_destination_type] }
 }
 
+# Once the deployment of the app gateway is complete, we can proceed to include the required DNS zone for Nexus, which is dependent on the FQDN of the app gateway.
+resource "azurerm_private_dns_zone" "non_core" {
```
tamirkamara, line 228: 1. The name of the resource isn't right
2. Non core dns zones have their own file. I think this should move there.
3. Have you tested the system starting with a version before this change and an installed nexus, then do the upgrade to the new version including this change? I think it might fail and needs attention as to how we do the migration.

tamirkamara (changes requested):

e2e_tests/config.py
```diff
@@ -6,6 +6,7 @@
 # Resource Info
 RESOURCE_LOCATION: str = config("RESOURCE_LOCATION", default="")
 TRE_ID: str = config("TRE_ID", default="")
+FQDN: str = config("FQDN", default="")
```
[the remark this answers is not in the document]
anatbal: Fixed

devops/scripts/load_env.sh
anatbal: As a result of our discussion - I removed this from load_env.sh and left it only in load_and_validate.sh which is being called each time before this will be called.

Makefile
```diff
@@ -261,7 +261,7 @@ bundle-register:
 	$(call target_title, "Registering ${DIR} bundle") \
 	&& . ${MAKEFILE_DIR}/devops/scripts/check_dependencies.sh porter,env \
 	&& az acr login --name $${ACR_NAME}	\
-	&& ${MAKEFILE_DIR}/devops/scripts/ensure_cli_signed_in.sh TRE_URL="$${TRE_URL:-https://$${TRE_ID}.$${LOCATION}.cloudapp.azure.com}" \
+	&& ${MAKEFILE_DIR}/devops/scripts/ensure_cli_signed_in.sh $${TRE_URL} \
```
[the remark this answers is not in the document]
anatbal: After our discussion - Added the task to have the TRE_URL param configured in the CI also in the public cloud as an acceptance criteria in this story https://github.com/microsoft/AzureTRE/issues/3299

core/terraform/appgateway/appgateway.tf
anatbal: 1. Changed the name to "nexus"
2. Moved to the right location
3. Fixed and tested in the migrartion.sh

tamirkamara (approved):

PR #3318, opened 2023-03-06
Add ARM_ENVIRONMENT as an environment variable

### LINKED ISSUE #3309: [GOV]  Porter templates installation environment includes cloud type
## Description

In order to deploy our templates with porter the azure terraform providers should include the following environment:
`backend "azurerm" {
    environment = "usgovernment"
  }`
or `public` if it is public azure cloud.
Alternatively it is possible to set an env var called `ARM_ENVIRONMENT ` read more about it here : https://developer.hashicorp.com/terraform/language/settings/backends/azurerm

When resource processor creates a subenv to install the template we need somehow to pass this env var there instead of passing it on each provider.  This way it can be done in one place. (See `ARM_USE_MSI ` as a reference, this one is [assed to the env)

As a Azure TRE Developer
I want to be able to pass ARM_ENVIRONMENT for porter install commands
So that we support multiple clouds and the change is in one place. 

## Acceptance criteria

- [ ] Porter is able to install templates in gov cloud
- [ ] Porter is able to install templates in public cloud

yuvalyaron:
# Resolves #3309

## What is being addressed

Added `ARM_ENVIRONMENT` as an environment variable both in the dev container and in the resource processor.
This environment variable is used by Terraform to set the Azure environment.
If the variable is absent, the default value will be public (the public Azure cloud)

## How is this addressed

- Added the variable to the config.yaml (along with JSON schema validation)
- Added arm_environment as variable in terraform, and from there, it is added to the `cloud-config.yaml` file, so it will be present in the resource processor as an environment variable.
- Added the variable to the `porter_env` dict in the config of the resource processor - in order to make it available in subprocesses of the resource processor


config.sample.yaml
```diff
@@ -3,6 +3,8 @@
 # run `make auth` to setup the Auth.
 tre_id: __CHANGE_ME__
 location: __CHANGE_ME__
+# Sets the Azure environment to be used. Possible values: public, usgovernment, china, german, stack
+# arm_environment: public
```
tamirkamara, line 7: This needs to be aligned with @LizaShak.

tamirkamara (approved):
Other than the one comment LGTM.

config.sample.yaml
LizaShak: Yes, we need it for the core deployment anyway, so from my perspective we will use this var to determine the cloud and propagate it to the terraform

PR #3320, opened 2023-03-07
remove admin firewall rules

### LINKED ISSUE #3286: Firewall fqdns should support multiple clouds
## Description

As a Azure TRE Developer
I want the fqdns added to firewall to be adjusted accordingly to the cloud it is deployed in
So that we can deploy our solution on multiple clouds easily.

Technical notes:
- A core example is the /workspaces/AzureTRE/templates/shared_services/firewall/terraform/rules.tf
- Try replacing those with ServiceTags
- Think of away to reduce the hardcoded parts

## Acceptance criteria

- [ ] FQDNS added to firewall are adjustable to the cloud it deployed in

tamirkamara:
# Resolves #3286 

## What is being addressed

The firewall shared service includes exceptions for domains previously used by the admin-vm but we don't seem to need them anymore.

## How is this addressed

- Remove redundant rules.

### DIFF
--- a/templates/shared_services/firewall/porter.yaml
+++ b/templates/shared_services/firewall/porter.yaml
@@ -1,7 +1,7 @@
 ---
 schemaVersion: 1.0.0
 name: tre-shared-service-firewall
-version: 1.0.0
+version: 1.0.3
 description: "An Azure TRE Firewall shared service"
 dockerfile: Dockerfile.tmpl
 registry: azuretre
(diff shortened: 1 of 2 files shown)


templates/shared_services/firewall/porter.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 schemaVersion: 1.0.0
 name: tre-shared-service-firewall
-version: 1.0.0
```
yuvalyaron, line 4: why do we skip 2 versions?

yuvalyaron (changes requested):

yuvalyaron (approved):

PR #3325, opened 2023-03-08
Pass additional params to clients

### LINKED ISSUE #3288: [GOV]  Azure credentials sourced based on the target cloud
## Description

As a Azure TRE Develope
I want the DefaultAzureCredential to source the authority based on the target cloud type
So that I can deploy our solution on multiple clouds

## Acceptance criteria

- [ ] DefaultAzureCredential uses the right authority

### LINKED ISSUE #3289: [GOV]  Management Clients use base_url and credential_scopes
## Description

As a Azure TRE Developer
I want the python Management Clients to set base_url and credential_scopes
So that it can be adjusted to the deployed cloud:

Technical notes:
Here is an example of CosmosDBManagementClient in US gov cloud: 

`CosmosDBManagementClient(
            credential,
            subscription_id=config.SUBSCRIPTION_ID,
            base_url=AZURE_US_GOV_CLOUD.endpoints.resource_manager,
            credential_scopes=[
                AZURE_US_GOV_CLOUD.endpoints.resource_manager + ".default"
            ])
        `
Additional clients needs fixing are: ResourceManagementClient and ComputeManagementClient

## Acceptance criteria

- [ ] ManagementClient getting the required params based on the target cloud type

LizaShak:
# Resolves #3288 
# Resolves #3289 

## What is being addressed

Pass additional parameters to clients sourced from cloud settings

## How is this addressed

- Update DefaultAzureCredential to source the authority from cloud endpoints
- Pass to  Management Clients  the management url from cloud endpoints.


api_app/core/cloud.py
```diff
@@ -0,0 +1,13 @@
+from msrestazure.azure_cloud import AZURE_PUBLIC_CLOUD, AZURE_US_GOV_CLOUD
+from core import config
+
+
+def get_cloud():
```
tamirkamara, line 5: Aren't you missing some of valid values?
What will happen if a match isn't found?
Should we do a case insensitive comparison? 
Is it possible to do a map instead of multiple if statements?

api_app/services/azure_resource_status.py
```diff
@@ -27,5 +27,9 @@ def get_azure_resource_status(resource_id):
 
 
 def get_azure_vm_instance_view(vm_name, resource_group_name) -> models.VirtualMachineInstanceView:
-    compute_client = ComputeManagementClient(credentials.get_credential(), config.SUBSCRIPTION_ID)
+    resource_manager_endpoint = cloud.get_cloud().endpoints.resource_manager
+    compute_client = ComputeManagementClient(credentials.get_credential(),
+                                             subscription_id=config.SUBSCRIPTION_ID,
+                                             base_url=resource_manager_endpoint,
+                                             credential_scopes=[resource_manager_endpoint + ".default"])
```
tamirkamara, line 34: Looks like we should create a method to return the credential scopes instead of doing this 3 times.

core/terraform/json-to-env.sh
```diff
@@ -101,6 +101,10 @@ jq -r '
         {
             "path": "event_grid_airlock_notification_topic_resource_id",
             "env_var": "EVENT_GRID_AIRLOCK_NOTIFICATION_TOPIC_RESOURCE_ID"
+        },
+        {
+            "path": "arm_environment",
```
tamirkamara, line 106: Why do we need this?

api_app/core/cloud.py
LizaShak: Do we want to note we support other values/clouds? I am not sure, we know already things are different on each cloud... So don't want to confuse ... The arm environment param is validated as part of the config schema validation , can contain only these values .. Will add the other checks you mentioned..

core/terraform/json-to-env.sh
LizaShak: The DefaultAzureCredential is used when no managed identity exists- so basically on local debugging in our use case, for this we need the TF output param to be added to private.env so the api will have it when debugging.  Reference: core/terraform/outputs.sh

api_app/core/cloud.py
yuvalyaron: perhaps we could raise an error if we don't support the cloud yet,
but if we don't support certain clouds maybe we should remove them from the schema of the config.yaml?

api_app/.env.sample
```diff
@@ -68,3 +68,8 @@ AZURE_CLIENT_SECRET=__CHANGE_ME__
 # Airlock configuration
 # ------------------------------------------
 AIRLOCK_SAS_TOKEN_EXPIRY_PERIOD_IN_HOURS=__CHANGE_ME__
+
+# Cloud configuration
+# -----------------
+# Sets the used Azure Cloud environment. Possible values: public, usgovernment, china, german
```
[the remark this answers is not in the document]
yuvalyaron: on a second thought, scratch that, I don't think we currently support stack, maybe we should remove it also from the `config_schema.json`

core/terraform/json-to-env.sh
tamirkamara: But I think this file and the outputs.sh one are used to deal with settings coming for terraform. Why do we need to "export" this one from terraform if it already exists as a regular environment sourced directly from the config?

api_app/core/cloud.py
tamirkamara: Currently our config allows all cloud types

LizaShak: Thinking now we better remove it from there too, WDYT? added the change, let me know in case you disagree

api_app/.env.sample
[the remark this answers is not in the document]
LizaShak: removed

api_app/core/cloud.py
```diff
@@ -0,0 +1,23 @@
...
+    arm_env = config.ARM_ENVIRONMENT.lower()
+    supported_clouds = {"public": AZURE_PUBLIC_CLOUD, "usgovernment": AZURE_US_GOV_CLOUD}
+    if arm_env in supported_clouds:
+        return supported_clouds[arm_env]
+    raise ValueError("Invalid arm environment. Got: " + arm_env + " .Supported envs are: public and usgovernment.")
```
yuvalyaron, line 10: nit :)
```suggestion
    raise ValueError(f"Invalid arm environment. Got: {arm_env}. Supported envs are: {', '.join(supported_clouds.keys())}.")
```

api_app/core/cloud.py
```diff
@@ -0,0 +1,27 @@
...
+        return supported_clouds[arm_env]
+    raise ValueError(f"Invalid arm environment. Got: {arm_env}. Supported envs are: {', '.join(supported_clouds.keys())}.")
+
+
+def get_authority():
+    return get_cloud().endpoints.active_directory.replace("https://", "")
+
+
+def get_aad_authority():
+    return get_cloud().endpoints.active_directory
```
tamirkamara, line 18: If I had to think what is the difference between authority and aad_authority....
Can we think about better names?

core/terraform/json-to-env.sh
LizaShak: Following our conversation, defined arm env as required param in the config. So it will be loaded to the env

yuvalyaron (approved):

tamirkamara (approved):

PR #3326, opened 2023-03-08
AAD INSTANCE source from cloud settings

### LINKED ISSUE #3297: [GOV]  AAD instance should be configurable in the api app
## Description

As a Azure TRE Developer
I want to able to override the `AAD_INSTANCE` in API app
So that I can configure it.

Technical notes: 
- Currently in AzureTRE/api_app/core/config.py the `AAD_INSTANCE` uses the default value but it is not passed from anywhere to it. We should be able to configure it and pass in the api deployments. 
- Think of a design that will allow to change this param: pass it to the api app, or check within the app the cloud type and set  it there

## Acceptance criteria

- [ ] AAD_INSTANCE is configurable and API app can work on multiple clouds

LizaShak:
# Resolves #3297 

## What is being addressed

 Remove hardcoded AAD_INSTANCE and source its value from cloud settings

### DIFF
--- a/api_app/core/cloud.py
+++ b/api_app/core/cloud.py
@@ -14,6 +14,10 @@ def get_authority():
     return get_cloud().endpoints.active_directory.replace("https://", "")
 
 
+def get_aad_authority():
+    return get_cloud().endpoints.active_directory
+
+
 def get_resource_manager_endpoint():
     return get_cloud().endpoints.resource_manager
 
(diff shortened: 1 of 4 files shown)


api_app/core/cloud.py
```diff
@@ -11,3 +11,7 @@ def get_cloud():
 
 def get_authority():
```
tamirkamara, line 12: Looking on the other method, should this be renamed to get_aad_authority maybe?

PR #3328, opened 2023-03-08
Query for available diagnostic categories

### LINKED ISSUE #3307: [GOV] Disable RuntimeAuditLogs and ApplicationMetricsLogs on Gov Cloud

tamirkamara:
# Resolves #3307 

## What is being addressed

Some clouds don't have all the diagnostic categories we have setup which causes a deployment failure.

## How is this addressed

- Like we already do elsewhere (mainly for app service), check what categories are available for the resource and turn on what we need out of them.


LizaShak (approved):
Just a question, so in case the ApplicationMetricsLogs is not supported the  log_category_types will not return it?

Issue #3285, opened 2023-03-01
[GOV]  Remove hardcoded cloud endpoints 


LizaShak:
## Description

As an Azure TRE Developer
I want to eliminate the hardcoded `login.microsoftonline.com` and  `graph.micorsoft.com`
So that I can deploy our solution in multiple clouds

Technical notes:
- Search for all the places where `login.microsoftonline.com` is passed as authority/issuer
- Consider the following design in your changes: https://github.com/microsoft/AzureTRE/issues/3310

## Acceptance criteria

- [ ] login.microsoftonline.com is sourced  based on the target cloud type
- [ ] graph.micorsoft.com is sourced  based on the target cloud type
- [ ] Documentation is updated and notes we support gov cloud

PR #3335, opened 2023-03-12
Remove hardcoded cloud endpoints

anatbal:
# Resolves [3285](https://github.com/microsoft/AzureTRE/issues/3285)

## What is being addressed

Remove hardcoded login.microsoftonline.com and graph.micorsoft.com from:

- [x] terraform
- [x] api 
- [x] e2e tests
- [x] UI - Added activeDirectoryUri in config.jsn which is being populated by the right cloud active directory uri


LizaShak (approved):

templates/shared_services/firewall/porter.yaml
```diff
@@ -49,6 +49,8 @@ parameters:
     type: string
     default: Standard
     description: The firewall and its policy SKU tier
+  - name: microsoft_graph_fqdn
```
tamirkamara, line 52: I wonder if we should add a default for all these types of params that will represent the value in the public cloud. This way we could allow easier install in the primary cloud a developer is likely to use.

anatbal: Done

PR #3336, opened 2023-03-12
Azure CLI should login to the right cloud

### LINKED ISSUE #3298: Switch cloud if needed before az login
## Description

As a Azure TRE Developer
I want to switch cloud if needed before az login
So that I am able to authenticate against multiple clouds

Technical notes:
- Deside if using `az cloud set --name AzureUSGovernment` or `az config set defaults.cloud=AzureUSGovernmen`
- Switch based on the `arm_environment`

## Acceptance criteria

- [ ] scripts using az login are authenticated against the required cloud

yuvalyaron:
# Resolves #3298

## What is being addressed

Set the azure cloud environment before every `az login` call - to support Azure environments other than the public cloud

## How is this addressed

- Added `AZ_CLOUD_ENVIRONMENT` to the resource processor - the value of this parameter is based on the ARM_ENVIRONMENT (this is used by porter v0 in `run.sh` and by the VM to download porter's image from ACR)

- Added `az cloud set` call in `load_and_validate_env.sh` to handle the `az` commands of the Makefile

- Added a special_porter_param - `az_cloud_environment` to the resource processor, to pass the cloud environment to the templates


devops/scripts/load_and_validate_env.sh
```diff
@@ -44,6 +44,16 @@ else
...
     export TRE_URL
+
+    # Set the cloud environment for Azure CLI (based on the ARM_ENVIRONMENT env var)
+    declare -A azure_environments
+    azure_environments["public"]="AzureCloud"
+    azure_environments["usgovernment"]="AzureUSGovernment"
+    azure_environments["china"]="AzureChinaCloud"
+    azure_environments["german"]="AzureGermanCloud"
+
+    az cloud set --name "${azure_environments[${ARM_ENVIRONMENT}]}"
```
tamirkamara, line 55: Not sure about doing it here.

devops/scripts/load_and_validate_env.sh
```diff
@@ -44,6 +44,16 @@ else
     export $(yq e "$GET_LEAF_KEYS|$TF_KEYS| $FORMAT_FOR_ENV_EXPORT" config.yaml)
     TRE_URL=${TRE_URL:-https://${TRE_ID}.${LOCATION}.cloudapp.azure.com}
     export TRE_URL
+
+    # Set the cloud environment for Azure CLI (based on the ARM_ENVIRONMENT env var)
+    declare -A azure_environments
+    azure_environments["public"]="AzureCloud"
+    azure_environments["usgovernment"]="AzureUSGovernment"
+    azure_environments["china"]="AzureChinaCloud"
+    azure_environments["german"]="AzureGermanCloud"
```
tamirkamara, line 53: Has the linter made you define it like that?
Isn't this possible?
```bash
azure_environments=( ["a"]="b" ["c"]=d )
```

tamirkamara (approved):
Can you update all versions bumps to just +1 on the build?
I know we don't usually do this but I think we will bump the minor one at one go when we merge all gov changes to main.

PR #3340, opened 2023-03-14
Release v0.9.0.7

dusan-ilic-mhra:
Put ENABLE_SWAGGER to true by default


tamirkamara:
@dusan-ilic-mhra this is not the right target repo for this PR.
Since it happen for you guys a couple of times already, I suggest you think of the reason... maybe the remotes for your repo aren't configured correctly?

PR #3343, opened 2023-03-15
RP_BUNDLE multiple values fix

### LINKED ISSUE #3341: Multiple RP_BUNDLE_ values causes RP initialisation failure
When specifying more than one value in `RP_BUNDLE_VALUES`, the terraform splits the key/value pairs out but when injecting them in to the yaml file each value after the first does not contain the correct spacing to be valid yaml - such:

```
stuff:
   rp_bundle_val1: value_1
rp_bundle_val2: value_2 <-- missing spacing to bring into correct block.
rp_bundle_val3: value_3
```

The terraform needs updating to include the space.

damoodamoo:
# Resolves #3341 

- Adds appropriate spaces to fix the autogenerated cloud_config.yaml file.
- Added some arbritrary rp_bundle values to the build in this repo to ensure it breaks in future if var handling changes

### DIFF
--- a/.github/actions/devcontainer_run_command/action.yml
+++ b/.github/actions/devcontainer_run_command/action.yml
@@ -185,6 +185,7 @@ runs:
             && inputs.CORE_APP_SERVICE_PLAN_SKU) || 'P1v2' }}" \
           -e WORKSPACE_APP_SERVICE_PLAN_SKU="${{ (inputs.WORKSPACE_APP_SERVICE_PLAN_SKU != ''
             && inputs.WORKSPACE_APP_SERVICE_PLAN_SKU) || 'P1v2' }}" \
+          -e TF_VAR_rp_bundle_values='{"tre_test_custom_val_1": "Value 1", "tre_test_custom_val_2": "Value 2"}' \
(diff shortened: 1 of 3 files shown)


.github/actions/devcontainer_run_command/action.yml
```diff
@@ -185,6 +185,7 @@ runs:
             && inputs.CORE_APP_SERVICE_PLAN_SKU) || 'P1v2' }}" \
           -e WORKSPACE_APP_SERVICE_PLAN_SKU="${{ (inputs.WORKSPACE_APP_SERVICE_PLAN_SKU != ''
             && inputs.WORKSPACE_APP_SERVICE_PLAN_SKU) || 'P1v2' }}" \
+          -e TF_VAR_rp_bundle_values='{"tre_test_custom_val_1": "Value 1", "tre_test_custom_val_2": "Value 2"}' \
```
tamirkamara, line 188: Don't you need to pass a real value here?
What this is meant to do?

damoodamoo: this is just passing in some arbitrary values to RP_BUNDLES - so that we can a) validate that my change doesn't prevent the RP starting, and b) highlight a future change that might break the way RP_BUNDLE values are dealt with, as this would break the build.

tamirkamara (approved):

Issue #3296, opened 2023-03-02
[GOV]  Use auth libraries to authenticate in E2E tests


LizaShak:
## Description

As a Azure TRE Developer
I want to eliminate hardcoded login endpoints calls and use azure-identity which we use in other places to authenticate
So that our code is easier to maintain.

Technical notes:

- Replace the REST calls to login.microsoftonline.com
- Consider the following as a replacement: https://learn.microsoft.com/en-us/python/api/azure-identity/azure.identity.clientsecretcredential?view=azure-python and https://learn.microsoft.com/en-us/python/api/azure-identity/azure.identity.usernamepasswordcredential?view=azure-python to replace the manual calls in :
`AzureTRE/e2e_tests/resources/workspace.py` and 
`AzureTRE/e2e_tests/helpers.py`

## Acceptance criteria

- [ ] e2e run successfully using azure-identity lib to authenticate

PR #3344, opened 2023-03-15
Use auth libraries to authenticate in E2E tests

anatbal:
# Resolves [#3296](https://github.com/microsoft/AzureTRE/issues/3296)
## What is being addressed

- Replace the REST calls to login.microsoftonline.com with  ClientSecretCredential, UsernamePasswordCredential

### DIFF
--- a/e2e_tests/cloud.py
+++ b/e2e_tests/cloud.py
@@ -5,5 +5,5 @@ def get_cloud() -> cloud.Cloud:
     return cloud.get_active_cloud()
 
 
-def get_authority_domain() -> str:
+def get_aad_authority_fqdn() -> str:
     return get_cloud().endpoints.active_directory.replace('https://', '')
--- a/e2e_tests/cloud.py
+++ b/e2e_tests/cloud.py
@@ -5,5 +5,5 @@ def get_cloud() -> cloud.Cloud:
     return cloud.get_active_cloud()
 
 
-def get_aad_authority_url() -> str:
-    return get_cloud().endpoints.active_directory
+def get_authority_domain() -> str:
+    return get_cloud().endpoints.active_directory.replace('https://', '')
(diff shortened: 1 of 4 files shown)


e2e_tests/cloud.py
```diff
@@ -5,5 +5,5 @@ def get_cloud() -> cloud.Cloud:
     return cloud.get_active_cloud()
 
 
-def get_aad_authority_url() -> str:
-    return get_cloud().endpoints.active_directory
+def get_authority_domain() -> str:
```
tamirkamara, line 8: Can you verify this is the same method name we used in the other apps?

e2e_tests/resources/workspace.py
```diff
@@ -41,24 +39,13 @@ async def get_workspace_auth_details(admin_token, workspace_id, verify) -> Tuple
...
-        try:
-            responseJson = response.json()
-        except JSONDecodeError:
-            raise Exception("Failed to parse response as JSON: {}".format(response.content))
-
-        if "access_token" not in responseJson or response.status_code != status.HTTP_200_OK:
-            raise Exception("Failed to get access_token: {}".format(response.content))
-
-        return responseJson["access_token"], scope_uri
+    return token.token, scope_uri
```
tamirkamara, line 51: This looks to be very similar to the block above. Is it possible to extract into a generic method like _get_token_ in the helpers class?

tamirkamara (approved):

e2e_tests/resources/workspace.py
anatbal: Done

e2e_tests/cloud.py
anatbal: Yes, the exact same one as in the api, I just added the 'str' for the return value to remain consistent

anatbal: FIixed

PR #3346, opened 2023-03-15
Add ARM_ENVIRONMENT as an environment variable in porter's installation containers

### LINKED ISSUE #3309: [GOV]  Porter templates installation environment includes cloud type

yuvalyaron:
# Resolves #3309 

## What is being addressed

Added ARM_ENVIRONMENT as a porter parameter so it will be added as an environment variables in porter's bundle installation containers.

**NOTE:**
the parameter was added to all templates except for: firewall, workspace-base, guac-service, guac-windows-vm.
As for these, the parameter will be added in the integration's PR

### DIFF
--- a/templates/shared_services/admin-vm/parameters.json
+++ b/templates/shared_services/admin-vm/parameters.json
@@ -39,6 +39,12 @@
       "source": {
         "env": "ADMIN_JUMPBOX_VM_SKU"
       }
+    },
+    {
+      "name": "arm_environment",
+      "source": {
+        "env": "ARM_ENVIRONMENT"
+      }
     }
   ]
 }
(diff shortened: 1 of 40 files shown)


yuvalyaron:
related to #3346 3318

tamirkamara (approved):

PR #3359, opened 2023-03-23
Allow default values in config_schema

### LINKED ISSUE #3331: Set default values for env vars on their load
## Description

As a Azure TRE Developer
I want to load env vars that have defaults even if not present in config.yaml 
So that all the env vars are present in my env and I can address them in my code.

## Acceptance criteria

- [ ] Config yaml doesn't specify a value for the default, yet it is available to the scripts/make commands with its default value

nyouens:
# Resolves #3331

## What is being addressed

Currently the config.yaml is a large file, some values could be defaulted in the config_schema.json to simplify the configuration. This will also support the Azure multi-cloud approach by defaulting Azure public cloud but allowing this to be over-written in the config file.

## How is this addressed

- Changed load_and_validate_env.sh to extract any node from config_schema.json with a default attribute
- Format as Key/Value pairs
- Export default environment variables
- Original processing will overwrite default variables with config.yaml entries if found. If not default values will remain.


LizaShak (approved):
LGTM :)

PR #3407, opened 2023-04-03
Remove AAD params from templates and configure in API.

### LINKED ISSUE #2628: Default workspace creation UI to `auto_create` and do not prompt the user.
When deploying a base workspace the user is prompted to enter Application ID and Secret. We want to default a custom template `auto_create` and not prompt for these values.

At present this need this file `https://github.com/microsoft/AzureTRE/blob/main/api_app/schemas/azuread.json` to be edited. This means that the API app needs to be rebuilt. I suggest the values in this file are moved to the Workspace `template_schema.json` so they can be customised without rebuilding the API image.

marrobi:
# Resolves #2628
## What is being addressed

At present each worksapce template specifies the Entra ID (Azure AD) app registration choices. This means that settings can be chosen that cannot be executed. 

## How is this addressed

- Use AUTO_ settings to configure workspace auth settings
- Remove references to manual setting of app reg creation in API and templates.
- Pass AUTO_ variables to API in terraform
- [ ] Ensure workflows work  - aad tests?


martinpeck:
Adding label spike as the code still has value.
Associated with issue #2628

PR #3419, opened 2023-04-04
Fix E2E tests in Gov Cloud

yuvalyaron:
## What is being addressed

The E2E tests polls the URL of the deployed guacamole.
Before these changes the guacamole URL was hardcoded, but in US Gov cloud, the root domain of the URL is `azurewebsites.us` and in the public cloud it's `azurewebsites.net`


tamirkamara (approved):

yuvalyaron:
Extended E2E tests passed:
https://github.com/microsoft/AzureTRE/actions/runs/4712278879/jobs/8357832710

PR #3420, opened 2023-04-04
Provide ability to copy URL to access an internally accessible resource.

### LINKED ISSUE #3122: Provide ability to copy URL to access an internally accessible resource.
As a researcher I can click on a connect button to access a TRE resource. If a resource is only available internally I cannot click connect. I do however need to be able to copy the URI. I cannot do this at present. The AML service shows a button that doesn't work, others there is no button.

Reproduce:
- Workspace with VMs and AML with no public access
- Login as Workspace Researcher (not owner)
- Access AML

Notes:
At present we have `connection_uri` and `internal_connection_uri`. We also have `is_exposed_externally`. Using a combination of these fields the appropriate behaviour should occur. If accessible, a button to click, if not a button to copy URI to the clipboard.

anatbal:
# Resolves #3122 
## What is being addressed
The connect button for the ws service and the user resource will now pop up a window referencing to clipboard when the guacamole is not exposed externally.


damoodamoo (approved):
like it!

PR #3443, opened 2023-04-16
Upgrade Guacamole deps 1.5.1

tamirkamara:
## What is being addressed

With Guacamole 1.5.* the base image of `guacd` has move to `alpine` which requires a bunch of changes in the bundle.

## How is this addressed

- Update all Guacamole dependencies to 1.5.1 (base image, war, jar - although it's `provided`).
- Use apk to install OS dependencies including openjdk11 (which has previously been manually installed).
- Generate host RSA keys as those aren't generated automatically in alpine
- Update s6 run script for guacd as it changed location
- Update terraform binary and providers


anatbal (approved):
LGTM

PR #3444, opened 2023-04-16
Add workflow default to RP_BUNDLE_VALUES

tamirkamara:
## What is being addressed

`RP_BUNDLE_VALUES` is an optional variable but its workflow default is an empty string (as is for any secret/var we use in the reusable workflow). That is invalid value in Terraform as the variable is defined as a map.

## How is this addressed

- Add `{}` as the workflow default for the param.


yuvalyaron (approved):

damoodamoo (approved):
sad face and many apologies from me :(

PR #3450, opened 2023-04-20
Include AzureML FQDN as a bundle output

### LINKED ISSUE #3448: AML workspace service fails to install and puts firewall into failed state
**Describe the bug**
When deploy the AML workspace service it fails to install as `aml_fqdn` is missing as an output and the firewall pipeline has a blank value.

```
Error message: expected "application_rule_collection.3.rule.0.destination_fqdns.1" to not be an empty string, 
```

LizaShak:
# Resolves #3448 

## What is being addressed

Output the Aml service fqdn as a bundle output. This is needed for the next step where it is added to the firewall rules.


marrobi (approved):
LGTM, providing the regex bit works :)

PR #3507, opened 2023-05-18
Adding code for supporting configurable SMTP port and SSL usage

migldasilva:
# Resolves the addition of configurations for supporting specific SMTP port and SSL usage 

## What is being addressed

So far SMTP port and SSL certificate usage are hard coded on the available files. This PR makes changes on the current files so that these configurations are exposed during deployment time.

## How is this addressed

- New variables are added
- Such variables can be defined in the payload sent to the Resource Processor

### DIFF
--- a/templates/shared_services/airlock_notifier/terraform/airlock_notifier.tf
+++ b/templates/shared_services/airlock_notifier/terraform/airlock_notifier.tf
@@ -51,7 +51,7 @@ resource "azurerm_resource_group_template_deployment" "smtp_api_connection" {
       value = var.smtp_server_enable_ssl
     },
     "serverPort" = {
-      value = var.server_port
+      value = var.smtp_server_port
     }
   })
 


migldasilva:
@microsoft-github-policy-service agree company="Medicines and Healthcare products Regulatory Agency (MHRA)"

templates/shared_services/airlock_notifier/porter.yaml
```diff
@@ -44,13 +44,20 @@ parameters:
   - name: smtp_server_address
     env: SMTP_SERVER_ADDRESS
     type: string
+  - name: smtp_server_port
```
tamirkamara, line 47: I think this should also include a default

templates/shared_services/airlock_notifier/template_schema.json
```diff
@@ -4,7 +4,7 @@
   "type": "object",
   "title": "Airlock Notifier Shared Service",
   "description": "A shared service notifying on Airlock operations",
-  "required": ["smtp_server_address", "smtp_username", "smtpPassword", "smtp_from_email"],
+  "required": ["smtp_server_address", "smtp_server_port", "smtp_username", "smtpPassword", "smtp_from_email"],
```
tamirkamara, line 7: If we add a default, can the port still be optional?

templates/shared_services/airlock_notifier/porter.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 schemaVersion: 1.0.0
 name: tre-shared-service-airlock-notifier
-version: 0.5.0
+version: 0.5.1
```
tamirkamara, line 4:
```suggestion
version: 0.7.0
```

tamirkamara (changes requested):

templates/shared_services/airlock_notifier/porter.yaml
migldasilva: Perfect. I'll add port 25.

templates/shared_services/airlock_notifier/template_schema.json
migldasilva: Yes, it can.

templates/shared_services/airlock_notifier/porter.yaml
```diff
@@ -44,13 +44,21 @@ parameters:
...
     env: SMTP_USERNAME
     type: string
   - name: smtpPassword
     env: SMTP_PASSWORD
     type: string
     sensitive: true
+  - name: smtp_server_enable_ssl
+    env: SMTP_SERVER_ENABLE_SSL
+    type: boolean
+    default: false
```
tamirkamara, line 61: I think this should be 
```suggestion
    default: true
```

tamirkamara (approved):

templates/shared_services/airlock_notifier/porter.yaml
migldasilva: Perfect. I have just accepted the suggestion.

PR #3513, opened 2023-05-23
include min memory for dev container to avoid hang issues

lordlinus:
## What is being addressed
Using dev container with less than 8 GB is causing issues during deployment.

## How is this addressed

Update host requirements spec with min 8 GB


tamirkamara:
@lordlinus Is this really not going to work with less than 8GB?
How did you came across to that kind of lowend machine?

tamirkamara:
@lordlinus Asking again since I worked on codespaces with just 4GB memory. Are you sure about this?

lordlinus:
@tamirkamara apologies for late reply, i did encounter both disk and memory issue when doing a full make `make all`

tim-allen-ck:
@lordlinus are you still finding this is a problem?

lordlinus:
Hi @tim-allen-ck i have set my default to 8 GB and no issues. If no one else has this issue you can leave it at the current default.
Thanks

PR #3519, opened 2023-05-28
Use host in windows mount script

### LINKED ISSUE #3518: Windows VM are missing shared storage mount
**Describe the bug**
A windows VM used to have the workspace shared storage automatically mounted but that doesn't happen anymore.

**Steps to reproduce**

1. Deploy Guacamole and a Windows VM
1. Login to the VM
1. Check the folder list and see that no shared storage is present

tamirkamara:
Resolves #3518 

## What is being addressed

The mount script fails to connect to shared storage.

## How is this addressed

- Use the `Host` instead of the `Endpoint` when calling the mount command as it expects a FQDN and not a URL.
- Upgrade versions (Terraform and providers)
- Use the native `templatefile` function instead of the `template_file` provider


templates/workspace_services/guacamole/user_resources/guacamole-azure-windowsvm/porter.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 schemaVersion: 1.0.0
 name: tre-service-guacamole-windowsvm
-version: 0.7.4
+version: 0.7.10
```
LizaShak, line 4: 0.7.5?

LizaShak (approved):
LGTM. Just a small comment

templates/workspace_services/guacamole/user_resources/guacamole-azure-windowsvm/porter.yaml
tamirkamara:
```suggestion
version: 0.7.5
```

PR #3534, opened 2023-06-05
Terraform handles peering sync natively

tamirkamara:
## What is being addressed

We correctly use AZCLI from within terraform to force network peering sync when there's an address space change.

## How is this addressed

- AzureRM 3.47 improved the ability for Terraform to handle this situation natively. So I've removed the script logic and all associated properties.


marrobi (approved):
As long as it works, LGTM!

PR #3541, opened 2023-06-06
Update doc and config schema for tre_id

### LINKED ISSUE #3536: tre_id can't have underscores
Documentation suggests that underscores are allowed in a tre_id https://github.com/microsoft/AzureTRE/blob/735ac3a538e2d1c88eaafdd739995e747c1b29fc/docs/tre-admins/setup-instructions/setup-auth-entities.md?plain=1#LL22C20-L22C20 

_**edit**_: also

https://github.com/microsoft/AzureTRE/blob/735ac3a538e2d1c88eaafdd739995e747c1b29fc/config_schema.json#LL16C16-L16C16

attempted deployment

```
│ Error: invalid value for name (Cosmos DB Account name must be 3 - 50 characters long, contain only lowercase letters, numbers and hyphens.)
│ 
│   with azurerm_cosmosdb_account.mongo,
│   on cosmos_mongo.tf line 2, in resource "azurerm_cosmosdb_account" "mongo":
│    2:   name                      = "cosmos-mongo-$***var.tre_id***"
│ 
```

t-young31:
# Resolves #3536 

## What is being addressed

The documentation and config.yaml schema suggests underscores can be used in a tre_id. This updates the. documentation and schema to only allow lowercase letters. Maybe hyphens are ok? (looks like all storage accounts use a replace on hyphens but not sure if there are other resources that assume the id only has lowercase letters)

### DIFF
--- a/config_schema.json
+++ b/config_schema.json
@@ -13,7 +13,7 @@
     "tre_id": {
       "description": "TRE unique identifier",
       "type": "string",
-      "pattern": "^[a-zA-Z 0-9\\_]*$",
+      "pattern": "^[a-z]*$",
       "maxLength": 11
     },
     "management": {
--- a/docs/tre-admins/environment-variables.md
+++ b/docs/tre-admins/environment-variables.md
@@ -21,7 +21,7 @@
 
 | <div style="width: 330px">Environment variable name</div> | Description |
 | ------------------------- | ----------- |
-| `TRE_ID` | A globally unique identifier. `TRE_ID` can be found in the resource names of the Azure TRE instance; for example, a `TRE_ID` of `mytre-dev` will result in a resource group name for Azure TRE instance of `rg-mytre-dev`. This must be less than 12 characters. Allowed characters: Alphanumeric and underscores|
+| `TRE_ID` | A globally unique identifier. `TRE_ID` can be found in the resource names of the Azure TRE instance; for example, a `TRE_ID` of `mytre-dev` will result in a resource group name for Azure TRE instance of `rg-mytre-dev`. This must be less than 12 characters. Allowed characters: lowercase letters|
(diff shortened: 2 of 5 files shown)


marrobi:
Thanks @t-young31. 

I'm sure we have had this discussion before, agree it is likely to restrict the ID up front. I think the issue is if people already have a `tre_id` with an underscore how do they proceed. with an upgrade. However given this mongo DB was added a while ago, not sure how many there are, if any.

Let us do a bit of research and see if we have any active deployments with `_` in the ID.

cc @itye-msft

t-young31:
I'd vote to reduce the friction for new adopters and anyone with an existing deployment, which I guess would be pre [core kv](https://github.com/microsoft/AzureTRE/commit/8b3f2d41076ef7d75062b93da846c5ed0f06fe87) needs to update their config schema

Hopefully the doc updates are useful even if the schema doesn't get updated!

tamirkamara:
We use numbers in our TREIDs so that would need to come back in too.

t-young31:
have done

"Use only lowercase letters" from the doc is then incorrect?

marrobi:
We definitely have numbers in TRE IDs, so need to update the docs.

marrobi:
@t-young31 actually the doc says "alphanumeric" rather than letters?

t-young31:
"Use only lowercase letters. Choose wisely!"

https://github.com/microsoft/AzureTRE/blob/0ed36d0280bb3081d6d61b590ae3496d0209e372/docs/tre-admins/setup-instructions/setup-auth-entities.md?plain=1#L22

t-young31:
feel free to do a find & replace `alphanumerics` -> `alphanumeric characters` on this branch

tamirkamara (approved):

PR #3543, opened 2023-06-06
Azure ML connection URI is an object, not string

### LINKED ISSUE #3486: Azure ML connection URI is an object, not string
When trying to use the UI to access an instance Azure ML not exposed externally we see the following:

08ec2236-f0ea-40c4-a71f-3de68fd5e73f

This is because the connection uri is an object. We should remove internal connection uri and have onyl connection uri.

bd8728d9-e856-4f07-89f2-55df213b9d1a

marrobi:
Resolves #3486

## What is being addressed
Removed `internal_connection_uri` and removed condition from `connection_uri`


marrobi:
/test-force-approve

Not tested in pipeline, and verified on deployment.

anatbal (approved):
LGTM, thanks for picking it up :)

PR #3558, opened 2023-06-09
Update docs to clarify cloning the repository

### LINKED ISSUE #3551: Provide guidance for cloning the AzureTRE-Deployment repo "correctly"
**Describe the bug**

Per #1395, when using Windows, the repo should be cloned to WSL instead of to a Windows file system which is then mounted in the container. 

However, there are no directions in the Quick Start at all about cloning the repo. Therefore, I think most Windows users will clone the repo in Windows, then open the folder in VS Code, and re-open in the dev container, which will cause the Windows file system to be mounted in the container.

marrobi:
# Resolves #3551

## What is being addressed
Add comment about cloning back in to docs and update for deployment repo.


marrobi:
@SvenAelterman can you read through this and let me know if the flow makes sense. Thanks.

SvenAelterman (approved):
Looks good to me.

There's a little word-smithing that could be done later.

PR #3568, opened 2023-06-13
Update eastus2 ADB udr

lordlinus:
Update host for eastus2 ADB UDR

## What is being addressed

Update ADB control plane hosts for eastus2 for firewall access

## How is this addressed

updated values from https://learn.microsoft.com/en-gb/azure/databricks/resources/supported-regions


marrobi (approved):
LGTM - Thanks!

PR #3573, opened 2023-06-14
Airlock Notifier - Change SMTP connector type

### LINKED ISSUE #3572: Change Airlock Notifier SMTP to a built-in connector
After SMTP connector became GA in Azure Logic Apps, terraform deployment can now be simplified to not use plain ARM for the smtp api connection deployment

## Description

As an Azure TRE Developer, I would like to use Logic Apps built in connector as the Airlock Notifier SMTP connector, so that the terraform will be simpler 

## Acceptance criteria

- [x] Replace managed connector with built in one
- [x] Remove ARM files that were used to deploy the managed connector
- [x] Change connections.json to point at the new service connection

SharonHart:
# Resolves #3572 

## What is being addressed

After SMTP connector became GA in Azure Logic Apps, terraform deployment can now be simplified to not use plain ARM for the smtp api connection deployment

## How is this addressed

- Remove api connector arm json definitions
- Change logic app connections.json to point at the new built in connector, configured with app settings.


.devcontainer/devcontainer.json
```diff
@@ -1,5 +1,10 @@
 {
   "name": "AzureTRE",
+  "features": {
```
SharonHart, line 3: Added for remotely connecting to GH codespaces (i.e. from Jetbrains Gateway)

tamirkamara (approved):
Thanks @SharonHart!

Issue #50, opened 2021-05-10
Define default logging retention policies 


christoferlof:
For auditing and troubleshooting purposes, all Azure resources provisioned should have logging and auditing enabled. 
At this moment, we don't have any policies explicitly defined. These needs to be defined, implemented and clearly documented.

_Originally posted by @deniscep in https://github.com/microsoft/AzureTRE/pull/49#discussion_r628223315_

marrobi:
We can do this via Azure Policy similar to https://github.com/microsoft/AzureTRE/pull/3623#issuecomment-1673318288

PR #3623, opened 2023-07-19
[DRAFT] Added support to specify tags in the config.yaml that will be applied to all resources deployed by TRE

### LINKED ISSUE #417: Azure resource tag management by cloud administrators
## Description

As a cloud administrator
I want to manage Azure tags applied to the TRE environment, Workspaces & Workspace services
So that I can us tags in governance e.i. cost center management

## Acceptance criteria

- [ ] Should be able to set the tags used to deploy the TRE environment
- [ ] Tags should be set for Workspaces and Workspace Services created by the Resource Processor
- [ ] Tags that should be ignore should be configurable

jlabhard-sg:
# Resolves #417 

## What is being addressed

We need all resources deployed by TRE (included management) to include specific tags upon creation to comply with corporate tagging policies. 

Important: The current PR contains a draft implementation that solves this issue. It is not fully complete and has not been tested yet. 

## How is this addressed
- Tags can be specified in the `config.yaml` file in a new `tags` field.
- Must be a string representation of JSON object with the following format: `'{"tag_key": "tag_value", "tag_key2": "tag_value2"}'`, similar to `rp_bundle_values`
- The tags are exported to environments variables similar to the other fields.
- Tags are added to `/devops/terraform/bootstrap.sh` directly in the command to create the resource group and storage account. This is done to comply with policies upon creation of resources.
- Tags are added to github actions on a best effort basis. We personally don't use them so I am less knowledgeable on their implementation.
- Tags are added as variable to the management terraform and applied to each resources. (This has been tested)
- Tags are added as variable to the core terraform and appended to the local `tre_core_tags` which is already applied to resources. Extension to modules is also considered.
- Similarily, for `shared_services`, `workspace_services`, `user_resources`, and `workspaces`:
    - Tags are added as terraform variable and appended to the respective locals (`tre_shared_service_tags`, `tre_workspace_service_tags`, `tre_user_resources_tags`, `tre_workspace_tags`)
    - The `porter.yaml` and `parameters,json` files have been updated for each template

#TODO: Tests, update documentation, increment template versions

### DIFF
--- a/config.sample.yaml
+++ b/config.sample.yaml
@@ -76,3 +76,7 @@ developer_settings:
 
 # Used by the API and Resource processor application to change log level
 #  debug: true
+
+# Specify here tags that should be applied to all resources deployed by the TRE
+# tag_name: "tag_value"
+#  tags: '{"tag_key": "tag_value"}'
--- a/.github/actions/devcontainer_run_command/action.yml
+++ b/.github/actions/devcontainer_run_command/action.yml
@@ -121,6 +121,9 @@ inputs:
     description: "JSON string containing key/value pairs to injet into the Resource Processor as ENV vars"
     required: false
     default: ""
+  TAGS:
+    description: "Tags to apply to all Azure resources"
+    required: false
 
 runs:
   using: composite
@@ -231,4 +234,8 @@ runs:
             && inputs.RESOURCE_PROCESSOR_NUMBER_PROCESSES_PER_INSTANCE) || 5 }}" \
           -e E2E_TESTS_NUMBER_PROCESSES="${{ inputs.E2E_TESTS_NUMBER_PROCESSES }}" \
           '${{ inputs.CI_CACHE_ACR_NAME }}${{ env.ACR_DOMAIN_SUFFIX }}/tredev:${{ inputs.DEVCONTAINER_TAG }}' \
+          -e TAGS='${{ (toJson(inputs.TAGS) != '""'
+            && inputs.TAGS) || '{}' }}' \
+          -e TF_VAR_tags='${{ (toJson(inputs.TAGS) != '""'
+            && inputs.TAGS) || '{}' }}' \
         bash -c "${{ inputs.COMMAND }}"
--- a/.github/workflows/deploy_tre_reusable.yml
+++ b/.github/workflows/deploy_tre_reusable.yml
@@ -243,6 +243,7 @@ jobs:
           TERRAFORM_STATE_CONTAINER_NAME: ${{ vars.TERRAFORM_STATE_CONTAINER_NAME }}
           MGMT_RESOURCE_GROUP_NAME: ${{ secrets.MGMT_RESOURCE_GROUP_NAME }}
           MGMT_STORAGE_ACCOUNT_NAME: ${{ secrets.MGMT_STORAGE_ACCOUNT_NAME }}
+          TAGS: ${{ vars.TAGS }}
 
       - name: ACR Login
         # failure in the first attempt indicates a new ACR, so we need to try again after it's been created
@@ -353,6 +354,7 @@ jobs:
           CORE_APP_SERVICE_PLAN_SKU: ${{ vars.CORE_APP_SERVICE_PLAN_SKU }}
           RESOURCE_PROCESSOR_NUMBER_PROCESSES_PER_INSTANCE: ${{ vars.RESOURCE_PROCESSOR_NUMBER_PROCESSES_PER_INSTANCE }}
           RP_BUNDLE_VALUES: ${{ vars.RP_BUNDLE_VALUES }}
+          TAGS: ${{ vars.TAGS }}
 
       - name: API Healthcheck
         uses: ./.github/actions/devcontainer_run_command
@@ -522,6 +524,7 @@ jobs:
           TRE_ID: ${{ secrets.TRE_ID }}
           LOCATION: ${{ vars.LOCATION }}
           BUNDLE_TYPE: ${{ matrix.BUNDLE_TYPE }}
+          TAGS: ${{ vars.TAGS }}
 
   register_bundles:
     name: Register Bundles
@@ -582,6 +585,7 @@ jobs:
           TRE_ID: ${{ secrets.TRE_ID }}
           LOCATION: ${{ vars.LOCATION }}
           BUNDLE_TYPE: ${{ matrix.BUNDLE_TYPE }}
+          TAGS: ${{ vars.TAGS }}
 
   register_user_resource_bundles:
     name: Register User Resource Bundles
@@ -632,6 +636,7 @@ jobs:
           LOCATION: ${{ vars.LOCATION }}
           BUNDLE_TYPE: ${{ matrix.BUNDLE_TYPE }}
           WORKSPACE_SERVICE_NAME: ${{ matrix.WORKSPACE_SERVICE_NAME }}
+          TAGS: ${{ vars.TAGS }}
 
   deploy_shared_services:
     name: Deploy shared services
@@ -662,6 +667,7 @@ jobs:
           TEST_ACCOUNT_CLIENT_SECRET: "${{ secrets.TEST_ACCOUNT_CLIENT_SECRET }}"
           TRE_ID: ${{ secrets.TRE_ID }}
           LOCATION: ${{ vars.LOCATION }}
+          TAGS: ${{ vars.TAGS }}
 
       - name: State Store Migrations
         uses: ./.github/actions/devcontainer_run_command
@@ -681,6 +687,7 @@ jobs:
           TERRAFORM_STATE_CONTAINER_NAME: ${{ vars.TERRAFORM_STATE_CONTAINER_NAME }}
           MGMT_RESOURCE_GROUP_NAME: ${{ secrets.MGMT_RESOURCE_GROUP_NAME }}
           MGMT_STORAGE_ACCOUNT_NAME: ${{ secrets.MGMT_STORAGE_ACCOUNT_NAME }}
+          TAGS: ${{ vars.TAGS }}
 
   deploy_ui:
     name: Deploy UI
@@ -711,6 +718,7 @@ jobs:
           MGMT_RESOURCE_GROUP_NAME: ${{ secrets.MGMT_RESOURCE_GROUP_NAME }}
           MGMT_STORAGE_ACCOUNT_NAME: ${{ secrets.MGMT_STORAGE_ACCOUNT_NAME }}
           SWAGGER_UI_CLIENT_ID: "${{ secrets.SWAGGER_UI_CLIENT_ID }}"
+          TAGS: ${{ vars.TAGS }}
 
   e2e_tests_smoke:
     name: "Run E2E Tests (Smoke)"
@@ -746,6 +754,7 @@ jobs:
           TRE_ID: ${{ secrets.TRE_ID }}
           IS_API_SECURED: false
           WORKSPACE_APP_SERVICE_PLAN_SKU: ${{ vars.WORKSPACE_APP_SERVICE_PLAN_SKU }}
+          TAGS: ${{ vars.TAGS }}
 
       - name: Upload Test Results
         if: always()
@@ -790,6 +799,7 @@ jobs:
           IS_API_SECURED: false
           WORKSPACE_APP_SERVICE_PLAN_SKU: ${{ vars.WORKSPACE_APP_SERVICE_PLAN_SKU }}
           E2E_TESTS_NUMBER_PROCESSES: ${{ inputs.E2E_TESTS_NUMBER_PROCESSES }}
+          TAGS: ${{ vars.TAGS }}
 
       - name: Upload Test Results
         if: always()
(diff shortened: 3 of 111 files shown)


jlabhard-sg:
@microsoft-github-policy-service agree company="SOPHiA GENETICS"

marrobi:
@jlabhard-sg approach looks good. I wonder if we can add anything to the terraform linter to ensure no services are missing the tags property.

Look forward to it being ready for review.

jlabhard-sg:
The PR has been tested with the deployment of a fresh TRE, shared services, base workspace and linux VM. The tags were successfully applied to all azure resources whenever possible, and complied with Azure Policies that requires tags on all resources.

I have looked into adding the tags to the terraform linter but it does not seem to support having the tags attribute, only having requirements for specific tags which is already in place.

PR is now ready for review.

marrobi:
@jlabhard-sg thanks, can you lint the terraform - `terraform fmt -recursive`

marrobi:
@jlabhard-sg can you resolve the conflicts please.

jlabhard-sg:
@marrobi The current failure of terraform lint happens because tags of resources in the management resource group do not include "tre_id". 

on `devops/terraform/main.tf`:
```
Notice: The resource is missing the following tags: "tre_id". (azurerm_resource_missing_tags)

  on main.tf line 10:
  10:   tags = merge(var.tags, {
  11:     project = "Azure Trusted Research Environment"
  12:     source  = "https://github.com/microsoft/AzureTRE/"
  13:   })
```

However, this is not something I have removed with this PR. The tre_id is also not a variable in the devops terraform so I'm not sure what should be done here.

marrobi:
@jlabhard-sg I think files only get linted when modified. Hopefully we can add some exclusions for that file similar to - https://github.com/terraform-linters/tflint/blob/master/docs/user-guide/annotations.md

If you have time to look, great, if not I will try to look next week.

jlabhard-sg:
@marrobi I've added annotation to exclude the tagging rule in `devops/terraform/main.tf` as suggested, we can run it to see if that works

config.sample.yaml
```diff
@@ -76,3 +76,6 @@ developer_settings:
 
 # Used by the API and Resource processor application to change log level
 #  debug: true
+
+# Specify here tags that should be applied to all resources deployed by the TRE in JSON string format
+#  tags: '{"tag_key":"tag_value"}'
```
marrobi, line 81: Can you add this to the docs, https://microsoft.github.io/AzureTRE/v0.13.0/tre-admins/environment-variables/,

Can you also update the CHANGELOG.md file please.

Noticed this page needs updating to explain how `config.yaml` maps to ENV vars etc. Will create a separate issue for that.

marrobi: I've added this to the docs, can you verify it's correct. Thanks.

marrobi:
@jlabhard-sg got an issue where tags aren't being updated for existing resource, I think this is due to:

```
```
being set on some (not all resources)

The reason this was added was so that we didn't replace corporate tags applied by Azure policy - hence a lot of TF churn on each TF deploy.

We could remove this, but it does also make me think, should these TRE wide tags be being applied via Azure policy rather than the TRE?

How does it work in your organisation?

My thinking at the moment is remove the ignore changes (that is typically a bad thing), and orgs need to add the corporate/policy applied tags to your new variable to prevent churn.

tamirkamara:
Removing the ignore tags is going to create issues on subscriptions where a higher level policy is used. For example, we have one that assigns an "creator" tag for each resource upon creating.

marrobi:
> Removing the ignore tags is going to create issues on subscriptions where a higher level policy is used. For example, we have one that assigns an "creator" tag for each resource upon creation.

Hmm, maybe a better solution is an Azure Policy that assigns tags based on `tre_id`?

Or alternatively we allow the tags to ignore to be specified in another variable.

jlabhard-sg:
> @jlabhard-sg got an issue where tags aren't being updated for existing resource, I think this is due to:
> 
> ```
> lifecycle { ignore_changes = [tags] }
> ```
> 
> being set on some (not all resources)
> 
> The reason this was added was so that we didn't replace corporate tags applied by Azure policy - hence a lot of TF churn on each TF deploy.
> 
> We could remove this, but it does also make me think, should these TRE wide tags be being applied via Azure policy rather than the TRE?
> 
> How does it work in your organisation?
> 
> My thinking at the moment is remove the ignore changes (that is typically a bad thing), and orgs need to add the corporate/policy applied tags to your new variable to prevent churn.

In our organization the Azure policy prevents the resource from being created if it does not include the necessary tags. I think this is due to the values of the tags being the responsibility of whoever creates these resources. In this case, I don't think tags automatically applied by an Azure policy would function. 

In our case, the ignore changes is not an issue since the policy affects only resources on creation. If it is an issue here I am not sure what the best solution is.

marrobi:
Nope, not possible. https://github.com/hashicorp/terraform/issues/24188

marrobi:
@jlabhard-sg how would you deal with the scenario "Now we need all resources to have this new tag?"

I think a policy might be abetter way forward I'm afraid, we could still add this to the TRE project to apply tags based on TRE_ID, although expect it would be relatively easy to do manually.

`Modify` operations are executed before `Deny`, so the tag would be added to the request before being blocked by the Deny.

https://learn.microsoft.com/en-us/azure/governance/policy/concepts/effects#order-of-evaluation

jlabhard-sg:
@marrobi Would this policy be created by the TRE/terraform, or is that something that should be created manually by our organization? I am a bit worried that the scope of the project would not have the necessary permission to create new policies. 

It is true that the scenario could be an issue in the future. Would it be possible to have a boolean variable determine whether the terraform should ignore tags or not?

marrobi:
It would be created by Terraform at subscription scope. I've been having a look to see if I can make it work... this is my current WIP `policy.tf` file.

```
resource "azurerm_policy_definition" "custom_tre_tags" {
  for_each     = var.tags
  name         = "custom_tre_tags_${each.key}"
  display_name = "${each.key} tag for TRE ${var.tre_id}"
  description  = "Add custom ${each.key} TRE tag to all resources on TRE ${var.tre_id}"
  policy_type  = "Custom"
  mode         = "Indexed"

  metadata = <<METADATA
  {
    "category": "TRE_Tags",
    "version": "1.0.0"
  }
  METADATA

  policy_rule = <<POLICY_RULE
  {
    "if": {
      "allOf": [
         {
          "field": "tags['tre_id']",
          "equals": "${var.tre_id}"
        }
      ]
    },
    "then": {
      "effect": "modify",
      "details": {
        "roleDefinitionIds": [
            "/providers/Microsoft.Authorization/roleDefinitions/b24988ac-6180-42a0-ab88-20f7382dd24c"
        ],
        "operations": [{
          "operation": "addOrReplace",
          "field": "tags[${each.key}]",
          "value": "${each.value}"
        }]
      }
    }
  }
  POLICY_RULE

}

resource "azurerm_subscription_policy_assignment" "custom_tre_tags" {
  for_each             = azurerm_subscription_policy_assignment.custom_tre_tags
  name                 =  "${each.key}_policy"
  policy_definition_id = each.value.id
  subscription_id      = data.azurerm_subscription.current.id
}

resource "azurerm_subscription_policy_remediation" "custom_tre_tags" {
  for_each             = azurerm_subscription_policy_assignment.custom_tre_tags
  name                 = "${each.key}_remediation"
  policy_assignment_id = each.value.id
  subscription_id      = data.azurerm_subscription.current.id
}
```

marrobi:
@jlabhard-sg this seems to work for me... are you ok to give it a go in your environment with the policy that blocks without certain tags?

```
resource "azurerm_policy_definition" "custom_tre_tags" {
  name         = "custom_tre_tags_${var.tre_id}"
  display_name = "Azure TRE ${var.tre_id}: Custom Tags"
  description  = "Add custom tags to all resources on TRE ${var.tre_id}"
  policy_type  = "Custom"
  mode         = "Indexed"

  metadata = <<METADATA
  {
    "category": "TRE_Tags",
    "version": "1.0.0"
  }
  METADATA

  policy_rule = jsonencode({
    "if" : {
      "allOf" : [
        {
          "field" : "tags['tre_id']",
          "equals" : var.tre_id
        }
      ]
    },
    "then" : {
      "effect" : "modify",
      "details" : {
        "roleDefinitionIds" : [
          "/providers/Microsoft.Authorization/roleDefinitions/b24988ac-6180-42a0-ab88-20f7382dd24c"
        ],
        "operations" : [
          for tag_key, tag_value in var.tags : {
            "operation" : "addOrReplace",
            "field" : "tags['${tag_key}']",
            "value" : tag_value
          }
        ]
      }
    }
  })
}

resource "azurerm_subscription_policy_assignment" "custom_tre_tags" {
  name                 = "custom_tre_tags_${var.tre_id}_assignment"
  policy_definition_id = azurerm_policy_definition.custom_tre_tags.id
  subscription_id      = data.azurerm_subscription.current.id
  location             = var.location

  identity {
    type = "SystemAssigned"
  }
}


resource "azurerm_role_assignment" "subscription_contributor" {
  scope                = data.azurerm_subscription.current.id
  role_definition_name = "Contributor"
  principal_id         = azurerm_subscription_policy_assignment.custom_tre_tags.identity[0].principal_id
}

resource "azurerm_subscription_policy_remediation" "custom_tre_tags" {
  name                 = "custom_tre_tags_${var.tre_id}_remediation"
  policy_assignment_id = azurerm_subscription_policy_assignment.custom_tre_tags.id
  subscription_id      = data.azurerm_subscription.current.id
}
```

I reverted the change in `local.tf`

```
  tre_core_tags = {
      tre_id              = var.tre_id
      tre_core_service_id = var.tre_id
  }
```
and added a dependency to the core RG in `main.tf`

```
   depends_on = [
    azurerm_subscription_policy_assignment.custom_tre_tags
  ]
```

Let me know if you can give it a go, and also the results. Lets leave the PR here until we know. Thanks.

marrobi:
I also note we haven't got ignore changes on all resources with tags defined.  :/

jlabhard-sg:
@marrobi Yes I will try to test the deployment tomorrow using this policy file instead of directly adding tags

marrobi:
@jlabhard-sg @tamirkamara I've created to PR here to add the missing lifecycle changes blocks (with a bit of help from GitHub copilot!)

jlabhard-sg:
@marrobi I've tested the deployment with the azure policy in the terraform and it seems to work, policy is complied with and did not encounter any permission issue. I think this the the way to go.

However, in our case, it is necessary to have the policy in the management resources terraform (`/devops` folder), and for this reason, also add `tre_id` to the variables and tags of the management resources. Would that be an issue for the TRE repo?

Specifically what I did:
* Add your code snippet to `/devops/terraform/tags_policy.tf`
* Reverted the tags changes and added `tre_id` to the variables and tags
* Added the `depends_on` on the management resources group __and__ the `tredev_purge` acr task
* Added the lifecycle ignore changes on the `tredev_purge` acr task
* Tested deployment of management resources and deployment of a VM without tags

If there is no issue, I can adapt the PR with these changes, adding the policy to the management resources, and removing the custom tags to all resources terraform.

marrobi:
@jlabhard-sg why do you need it in the `/devops` folder?

Initially the idea was multiple environments might share the same ACR and storage account, but in reality not sure that is the case anywhere. 

I don't see any real reason why policies couldn't go there, have you tagged the management resources with the `tre_id` too?

The other option is we do nothing, and say tagging policy is outside the remit of the repository, but we could provide an example in the docs on how to do it.

jlabhard-sg:
@marrobi Because the policy that requires the tags also applies on the management resources. The resource group and storage account could be created manually as they are then imported to the state, but the ACR is not. 
I did tag the management resources with the `tre_id`.

Doing nothing is also an option. the azure policies as they are defined here is not something that is necessary in either the core or devops terraform since it is at subscription scope. We could match it to something other than the `tre_id` to apply it to the management resources as well. 

Let me know what you think is the best approach. I actually didn't consider using policies to apply tags but it seems to solve the issue without needing changes to the repository.

marrobi:
Thanks, its actually similar to this one: https://github.com/microsoft/AzureTRE/issues/50

The management resources should probably have tags. @tamirkamara can you think of any reason why they shouldn;t be tagged with `tre_id`? To be honest they could sit in the main TRE resource group, but that's more effort from a migration perspective.

I'm in two minds about including the policies. It would make it easier for organisations to get up and running - and  people to want these things to work "out of the box", but is policies might also need tweaking to meet an organisations needs.

tamirkamara:
> The management resources should probably have tags. tamirkamara can you think of any reason why they shouldn;t be tagged with `tre_id`? To be honest they could sit in the main TRE resource group, but that's more effort from a migration perspective and is a chicken and egg situation with the state storage.

We don't use `tre_id` when creating the management resource group since at the time there was a notion that several TREs will use the same management RG.

marrobi:
I think there are two separate issues in play here. "how should we handle tagging", and "does the management resource group need a rethink". The latter needs a separate issue and longer discussion - for now you could tag in a policy filtering on the management resource group name.

Given you have a workaround, as much as I would like to get the PR in as you have put in effort, I think adding notes to the original issue might be the easiest next step, then if want to revisit later we can. 

We do need to get the ignore changes to tags PR in #3670 , otherwise TF is going to keep trying to remove any tags applied by policy.

PR #3647, opened 2023-08-02
Custom actions fail when action not defined in pipeline

### LINKED ISSUE #3646: Custom actions fail in release 0.12
When running custom actions, such as `renew` on resources with a pipeline defined they fail.

- deploy 0.12
- deploy cert service
- renew certificate - get a 503

marrobi:
# Resolves #3646

## What is being addressed

Fix bug where looking for custom actions in a pipleine when they do not exist.


Danny-Cooke-cCK:
We are experiencing this issue in 2 of our environments. one is 0.12 and the other is 0.11

api_app/service_bus/helpers.py
```diff
@@ -63,8 +63,8 @@ async def update_resource_for_step(operation_step: OperationStep, resource_repo:
 
     parent_template = await resource_template_repo.get_template_by_name_and_version(step_resource.templateName, step_resource.templateVersion, step_resource.resourceType, step_resource_parent_service_name)
 
-    # if there are no pipelines, no need to continue with substitutions.
-    if not parent_template.pipeline:
+    # if there are no pipelines, or custom action, no need to continue with substitutions.
+    if not parent_template.pipeline or primary_action not in parent_template.pipeline.dict():
```
SvenAelterman, line 67: So by adding this condition, we're avoiding an error on line 70 if `primary_action` doesn't exist in the collection of pipeline elements from the parent template?

Makes sense to me.

Would it be better to only make the call to `parent_template.pipeline.dict()` once?

marrobi: How's that?

SvenAelterman (approved):

SvenAelterman (approved):

PR #3655, opened 2023-08-07
Outputs of type boolean are stored as strings

### LINKED ISSUE #3432: Untyped Terraform variables cause issues
**Describe the bug**

Many templates use terraform input variables without defining their type which makes them string ones. This can cause problems with other types like bool are used in comparisons. 

A bool_var is defined in porter.yaml as bool.
```hcl
var.bool_var == false
```
This translates to something like `False == false` which makes the condition result in a **false** despite the logical expectation for **true**.

- [ ] #3657

**Steps to reproduce**

1. Install a Guacamole service without exposing it externally - be sure the uncheck the box.
1. This should have resulted in the web-app having a private endpoint, which is doesn't.

marrobi:
# Resolves #3432

## What is being addressed
- Amend resource Output objects to preserve Porter output type
- When converting Porter outputs to dict take account of the Output type, rather than just its contents.
- Update terraform variable descriptions and types for Guacamole workspace service

To test deploy a privately exposed Guacamole worksapce service. This PR will also unbolck the PR needed to ensure shared service URIs can be copied.


tamirkamara (approved):

marrobi:
/test-force-approve 

Passed: https://github.com/microsoft/AzureTRE/actions/runs/5787532772

PR #3656, opened 2023-08-07
Output is_exposed_externally, for all resources with URI

### LINKED ISSUE #3584: Copy URL for internal services does not work for shared services
If configure Gitea and Nexus to provide URI and `is_exposed_externally` to false, the URL gives a 403, rather than a copy dialogue.

marrobi:
# Resolves #3584

## How is this addressed

- add `is_exposed_externally` to restricted resource model
- Add `is_exposed_externally` with defaults for all resources with URIs

image


tamirkamara:
What is this for?

marrobi:
> What is this for?

Functionality was added to the UI to enable users to copy and paste a URI toa private resource. This relies on `is_exposed_externally`, As this isn't exposed for restricted view of a resource a connect button would go, and try access the URI over the internet hence give a 403. The PR exposes `is_exposed_externally` and applies it consistently across all templates with URIs.

martinpeck (commented):
Made one comment, then realised I don't really understand what this does so can't really give you much additional input. Soz.

api_app/models/domain/restricted_resource.py
```diff
@@ -9,7 +9,7 @@ class RestrictedProperties(AzureTREModel):
     description: str = ""
     overview: str = ""
     connection_uri: str = ""
-    internal_connection_uri: str = ""
+    is_exposed_externally: bool = True
```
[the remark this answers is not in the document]
marrobi: This is the default when creating the object, so if its not defined if it is exposed or not the UI will presume it is and provide a connect button if there is also a `connection_uri` specified, rather than the UI dialog to copy the URI. 

Not sure it really matters to be honest. Ideally bundles should define a `connection_uri` and `is_exposed_externally` and not use the default, but the default is more for backwards compatibility where `is_exposed_externally` has not been defined.

templates/shared_services/gitea/template_schema.json
```diff
@@ -26,6 +26,18 @@
       "description": "Long form description of the workspace service, in markdown syntax",
       "default": "The Gitea shared service is used for mirroring external Git repositories. For instructions on how to create Git mirrors see [https://docs.gitea.com/usage/repo-mirror](https://docs.gitea.com/usage/repo-mirror).",
       "updateable": true
+    },
+    "is_exposed_externally": {
+      "$id": "#/properties/is_exposed_externally",
+      "type": "boolean",
+      "title": "Expose externally",
+      "description": "Is the Gitea accessible from outside of the TRE network.",
+      "default": false
```
jjgriff93, line 35: This default is different to the model's default - intentional?

jjgriff93 (approved):

templates/shared_services/gitea/template_schema.json
marrobi: Yes, as its not configured in the terraform to enable it to be exposed externally.

PR #3654, opened 2023-08-04
Fix ability to debug resource processor locally

marrobi:
# Resolves #3426

## What is being addressed
- Since additional cloud support need Key Vault URI rather than name. Create an output and pass that through to RP.
- VS Code remote containers extension means `REMOTE_CONTAINERS_IPC` need setting when running Porter in debug session
- Add IP allow list to app insights for local debugging
- Give RP SPN access to read Key Vault secrets when local debugging

To test, follow:

https://microsoft.github.io/AzureTRE/v0.12.0/tre-developers/resource-processor/

PR #3659, opened 2023-08-08
Upgrade airlock review and unrestricted workspaces to base v0.12.0

### LINKED ISSUE #3617: "Airlock Import Review" and "unrestricted" workspace templates have hardcoded references to TRE versions (0.9.0 and 0.8.0)
**Describe the bug**

In line 12 of the Dockerfile.tmpl for the airlock-import-review workspace bundle, the Azure TRE version is hardcoded to 0.9.0.

See https://github.com/microsoft/AzureTRE/blob/8d31b90b1504132a276385da470b204f567a855a/templates/workspaces/airlock-import-review/Dockerfile.tmpl#L12C15-L12C15 

It seems this would be problematic because changes to the base workspace bundle won't make it to the airlock import review workspace. This is especially relevant with PR #3581

marrobi:
# Resolves #3617

## How is this addressed

- Upgrade base version
- Amend Porter config to work with new base versions

To test:
- Register new templates
- Deploy workspaces

image


templates/workspaces/airlock-import-review/porter.yaml
```diff
@@ -159,10 +157,6 @@ install:
         enable_local_debugging: ${ bundle.parameters.enable_local_debugging }
         register_aad_application: ${ bundle.parameters.register_aad_application }
         create_aad_groups: ${ bundle.parameters.create_aad_groups }
-        arm_tenant_id: ${ bundle.credentials.azure_tenant_id }
```
SvenAelterman, line 162: Why are these no longer needed?

SvenAelterman (commented):
Should this be two separate PRs: one for the local debugging fixes and one for the referenced versions?

templates/workspaces/airlock-import-review/porter.yaml
marrobi: We used to use the az cli within the base workspace bundle. This is no longer there hence these variables are not required.

marrobi:
> Should this be two separate PRs: one for the local debugging fixes and one for the referenced versions?

Doh, I've pulled the other PR into this branch. I've done it with another PR too. Let me try and revert them, unless fancy giving the #3654 a look!

SvenAelterman:
> > Should this be two separate PRs: one for the local debugging fixes and one for the referenced versions?
> 
> Doh, I've pulled the other PR into this branch. I've done it with another PR too. Let me try and revert them, unless fancy giving the #3654 a look!

I will look at #3654 but I would like to test that one - and I've never set up local debugging before, so might need some time.

SvenAelterman (approved):

PR #3668, opened 2023-08-10
Reduce logging noise

### LINKED ISSUE #2135: Logging Noise Seen from Resource Processor
When running an AppInsights query such as this...

``` kusto
traces
| where cloud_RoleName == "runner.py"
| where message !startswith "Retrying due to transient client"
| where message !startswith "No sessions for this process"
| where not (message startswith "Unable to find image" and strlen(message)<170)
| where severityLevel >1
| order by timestamp desc
```
... a number of messages that are simply `[stderr]` are reported...

image

Specifically, this query will return all such logs:

``` kusto
traces
| where cloud_RoleName == "runner.py"
| where message == "[stderr]"
```

This appears to be related to code in `runner.py` and `logging.py` within the resource processor.

We need to avoid this noise and, at the same time, make sure this noise doesn't mean we're dropping other more important logs.

marrobi:
# Resolves #2135

## How is this addressed
- In resource processor moved prefix into same log entry.
- Added LOGGERS_FOR_ERRORS_ONLY to reduce noise in API
- Configured logging before app startup to ensure statup errors are recorded and fitered.
- disable unwanted loggers prior to adding log handler
- ensure env vars are available for api app so that app insights is configured.


tamirkamara (approved):

resource_processor/shared/logging.py
```diff
@@ -140,8 +140,7 @@ def shell_output_logger(console_output: str, prefix_item: str, logger: logging.L
         logging.debug("Image not present locally, setting log to INFO.")
         logging_level = logging.INFO
 
-    logger.log(logging_level, prefix_item)
-    logger.log(logging_level, console_output)
+    logger.log(logging_level, f"{prefix_item} {console_output}")
```
tamirkamara, line 143: IIRC, this has been done like so because the log has been spread across multiple entries. 
It might not be like that now, but if it is still then this change is going to make it hard to understand.

marrobi: I tested this and the log analytics log were easier to follow with prefix on same line, otherwise the prefix was a separate entry (on its own) to the actual error which isn't useful.

PR #3674, opened 2023-08-15
Check docker return code before running commands 

### LINKED ISSUE #3673: Changing docker file permissions breaks docker outside the container
https://github.com/microsoft/AzureTRE/blob/c3f144f9593230cfb3a43a2b789dff800f5d25bc/devops/scripts/set_docker_sock_permission.sh#L7

This line changes the group of the `docker.sock` file _outside of the container_ (from `docker` to `citrixlog`). This means that I can't run docker outside the container anymore as my normal user isn't part of that group, so opening any new devcontainers or running `docker ps` fails with a permissions error (it works if I use `sudo`).

I think the reason it changes the group from `docker` to `citrixlog` is because the group number for docker inside the container is 1001 and that's the group number for `citrixlog` outside the container.

Is there a way to stop this acting on my normal system file outside the container and only affect the `/var/run/docker.sock` file _inside_ the container?

I'm using Ubuntu 20.04 and Docker version 24.0.5, build ced0996

marrobi:
# Resolves #3673

## What is being addressed
- Add check for docker return code before changing permissions


devops/scripts/set_docker_sock_permission.sh
```diff
@@ -4,5 +4,8 @@ set -e
 # By default the docker.sock file is not associated with docker group on codespaces or macOS
 # which causes a permission issue when docker is run without sudo.
 
-sudo chgrp docker /var/run/docker.sock
-sudo chmod g+rw /var/run/docker.sock
+if ! docker ps > /dev/null 2>&1; then
```
SvenAelterman, line 7: Why is this important? What happens if we just update the permissions like it has been?

SvenAelterman: Never mind, I see the customer issue

SvenAelterman (approved):
LGTM

marrobi:
/test-force-approve

Passed: https://github.com/microsoft/AzureTRE/actions/runs/5867301767

PR #3672, opened 2023-08-14
Add dependency on firewall deployment to rule collection

marrobi:
# Resolves #3671

Add dependancy on rule collection to Firewall. It is deploying before the Firewall has completed and is in a Succeeded state causing the issue referenced.

- [x] Update CHANGELOG.md if needed

PR #3682, opened 2023-08-18
Upgrade Airlock and Apply FW Rule for Airlock

### LINKED ISSUE #3664: Airlock Processor Unable to Start
We observed that a working/functioning deploy of the AzureTRE developed an issue where by airlock functionality stopped working. 

**Symptoms**

- The primary symptom that we observed was that storage containers for new import requests were not created. 
- In this state, messages were left unprocessed on `airlock-status-changed` service bus. 
- Additionally, in the azure portal, we saw an error in the portal when viewing the airlock processor function app that suggested that an SSL error had occurred. Also, the portal reported `error` for the function's runtime.

While this looked, at first, like an SSL issue further investigation showed the the airlock function was attempting to make outbound network calls during startup that were being blocked by the AzureTRE's firewall. Inspecting the firewall logs we discovered a number of DENY logs related to the function app runtime host attempting to connect to the following two endpoints:

- `functionscdn.azureedge.net`
- `func-airlock-processor-MYTRE.azurewebsites.net`

**Work Around**

Using the IP address of the function app opening up the firewall to allow outbound traffic has allowed the airlock function app to start up. The illustration below is clearly less constrained than it should be, but this patch to the firewall allows the function app to work. **Use with caution.**

image

**Longer Term Fix**

I suggest the longer term fix is to ensure that `functionscdn.azureedge.net` is added to the allow-list on the firewall so that traffic from the function app is not blocked I would not suggest we add `func-airlock-processor-MYTRE.azurewebsites.net` until we're sure we understand why the function app is trying to call itself.

**Root Cause Analysis**

I believe the airlock function app was initially working because the function is deployed BEFORE the firewall is deployed. So, the function app was able to make this call successfully. Then, for reasons I'm unclear about, the function app was redeployed. This might have been a scale-out event, or might have been as part of some action performed by another admin, but at any rate the function app was redeployed and attempted to make this network call on startup. This failed as the firewall now blocked the call.

The version of the Azure Functions runtime that we're using is 3.20.

In the source code for 3.20 we can see that the URL `https://functionscdn.azureedge.net/public` is defined here
https://github.com/Azure/azure-functions-host/blob/v3.20.0/src/WebJobs.Script/ScriptConstants.cs#L175

This url is used in the `ExtensionBundleManager`:
https://github.com/Azure/azure-functions-host/blob/v3.20.0/src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs#L37

This is then used to check for, and download, bundle extensions:
https://github.com/Azure/azure-functions-host/blob/v3.20.0/src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs#L173

https://github.com/Azure/azure-functions-host/blob/v3.20.0/src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs#L231

And, for context, the function's `host.json` specifies a bundle extension...

``` json
  "extensionBundle": {
    "id": "Microsoft.Azure.Functions.ExtensionBundle",
    "version": "[3.3.0, 4.0.0)"
  }
```

So, I believe that this is being called when the function is first deployed, and the bundle extension is being downloaded at this point. With the firewall in place it's not possible to download this bundle extension.

The bundle extension is what allows non-dotnet languages to register triggers:
https://github.com/Azure/azure-functions-extension-bundles/tree/v3.x

martinpeck:
# Resolves #3664

- upgrades the airlock processor function app to a newer runtime
- applies an firewall application rule for the function app processor so that the function runtime can be downloaded when it starts up
- also removes retention policies for *all* resources as these suddenly became deprecated and generated errors in new deployments (For details on the deprecation see https://learn.microsoft.com/en-us/azure/azure-monitor/essentials/migrate-to-azure-storage-lifecycle-policy)


marrobi:
E2E failed with firewall  - think this would resolve - https://github.com/microsoft/AzureTRE/pull/3672/

marrobi (approved):
LGTM. Tested airlock and containers are created.

marrobi:
/test-force-approve

Passed here: https://github.com/microsoft/AzureTRE/actions/runs/5964761486

PR #3695, opened 2023-08-31
update pypi firewall rules

### LINKED ISSUE #3694: Firewall is Incorrectly Configured. Calls to PyPI (via Nexus) and `pip install` will fail.
To fix #3651 the following PR changed the FQDNs that are registered with the Firewall #3677

This PR updated the FQDNs that the Nexus service registers with the firewall from including `*pypi.org` to `*.pypi.org`. 

Unfortunately, `*.pypi.org` doesn't work for the root domain, which is how Nexus first tries to contact the Python package index, and this results in two things...

1. `pip install` fails on all workspace VMs.
2. Errors are seen in the Nexus admin UI saying that SSL handshakes failed

The fix is to add `pypi.org` (the root domain) into the FQDNs along with the `*.pypi.org` rule.

The work around (until this is fixed) is to add this manually:

image

martinpeck:
# Resolves #3694 

## What is being addressed

## How is this addressed

- Update documentation

### DIFF
--- a/api_app/tests_ma/conftest.py
+++ b/api_app/tests_ma/conftest.py
@@ -422,7 +422,7 @@ def primary_resource() -> Resource:
         properties={
             "display_name": "test_resource name",
             "address_prefix": ["172.0.0.1", "192.168.0.1"],
-            "fqdn": ["*.pypi.org", "files.pythonhosted.org", "security.ubuntu.com"],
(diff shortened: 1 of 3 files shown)


martinpeck:
abandoning this PR in favour of another (smaller) PR

PR #3698, opened 2023-08-31
Add pypi.org into Allow List for Nexus Firewall

### LINKED ISSUE #3694: Firewall is Incorrectly Configured. Calls to PyPI (via Nexus) and `pip install` will fail.

martinpeck:
# Resolves #3694 

Adds `pypi.org` into the allow list of FQDNs for the Nexus service.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,42 @@ ENHANCEMENTS:
 BUG FIXES:
 
 
+## 0.14.1 (Unreleased)
+
+BUG FIXES:
+* Fix firewall config related to Nexus so that `pypi.org` is added to the allow-list  ([#3694](https://github.com/microsoft/AzureTRE/issues/3694))
+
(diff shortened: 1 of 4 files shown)


SvenAelterman (approved):
LGTM

PR #3717, opened 2023-09-21
Update get Shared Services route so full resource gets returned for admins

### LINKED ISSUE #3716: Shared Service Update does not work from shared-services page.
If try and upgrade a shared service form the shared services page, the form is filled out with template defaults, rather than actual values

image

If try do it from the actual service page, it works correctly:

image

marrobi:
# Resolves #3716

## What is being addressed

## How is this addressed

- Update get shared services route to return ful shared services resource for admins.
- Add tests to verify correct operation.


tamirkamara (approved):

marrobi:
/test-force-approve 

Passed: https://github.com/microsoft/AzureTRE/actions/runs/6272775982

PR #3723, opened 2023-09-22
Add missing upgrade outputs to Porter.yaml files.

### LINKED ISSUE #3685: Some templates have outputs defined for install but not upgrade
This means:

```
outputs:
  - name: ip
    type: string
    applyTo:
      - install
```
 
will only update on install, if the first deploy fails, and succeed on second attempt/update the fields never appear, so this should be:

```
outputs:
  - name: ip
    type: string
    applyTo:
      - install
      - upgrade  
```      

and should be listed under:

```
upgrade:
  - terraform:
     - outputs:
```

All `porter.yaml` files should be updated.

Sweetdevil144:
# Resolves #3685 

## What is being addressed
- Update `porter.yaml` templates to install and upgrade.
- Applied the same to all `porter.yaml` files, that is,
```yaml
outputs:
  - name: ip
    type: string
    applyTo:
      - install
```
was changed to 
```yaml
outputs:
  - name: ip
    type: string
    applyTo:
      - install
      - upgrade  
```


Sweetdevil144:
@microsoft-github-policy-service agree

Sweetdevil144:
@marrobi Can you review the PR ?

templates/shared_services/cyclecloud/porter.yaml
```diff
@@ -52,6 +52,7 @@ outputs:
     type: string
     applyTo:
       - install
+      - upgrade
```
tamirkamara, line 55: Please update the version number in all the bundles you've changed in this PR.

templates/workspace_services/mysql/porter.yaml
```diff
@@ -65,6 +65,7 @@ outputs:
     type: string
     applyTo:
       - install
+      - upgrade
```
marrobi, line 68: In this bundle upgrade isn't implemented and doesn't have an output defined as `mysql_fqdn`. So leave this one off. I suggest we open another issue to implement upgrade on the bundle.

marrobi (changes requested):
One small change on the mysql bundle, and as Tamir says the version of the bundles edited need incrementing so for example 1.2.3 goes to 1.2.4.

Thanks, nearly there :)

Sweetdevil144:
Hello @marrobi and @tamirkamara . Thanks for showing patience and apologies for late PR! Got a bit of personal reviews.
Leaving that aside, i have incremented version o the files I have changed. Along with that, I also modifed the mysql bundle as requested. 
Would like your opinion on these changes.

marrobi (approved):
LGTM Thanks.

PR #3724, opened 2023-09-22
TRE UI does not update when an action on a VM completes

### LINKED ISSUE #3353: TRE UI does not update when an action on a VM completes
For example, trying to **stop** a machine. The UI did not appear to register the request.

CleanShot 2023-02-14 at 14 00 46

then this only after refresh ....

CleanShot 2023-02-14 at 14 02 44

The same happened when I tried to deploy a new machine. The little blue spinner just told me the process was  pending, until I refreshed the browser page.

marrobi:
Fixes #3353

## How is this addressed

- Retrieve scope_id and pass to API call
- Return from components dependant on workspace context when the workspace is not set. Think there is a bigger issue around dependancies, the workspace context workspace property is initialised as '{}' meaning dependancies do not work correctly.

### DIFF
--- a/ui/app/src/components/shared/notifications/NotificationItem.tsx
+++ b/ui/app/src/components/shared/notifications/NotificationItem.tsx
@@ -16,7 +16,7 @@ import { useAppDispatch } from '../../../hooks/customReduxHooks';
 
 interface NotificationItemProps {
   operation: Operation,
-  showCallout: (o: Operation, r: Resource) => void
+  showCallout: (o: Operation, r: Resource) => void;
 }
 
 export const NotificationItem: React.FunctionComponent<NotificationItemProps> = (props: NotificationItemProps) => {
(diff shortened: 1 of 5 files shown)


jjgriff93 (approved):
Nice! David's more familiar with this than me so worth him checking too

ui/app/src/components/shared/notifications/NotificationItem.tsx
```diff
@@ -45,9 +45,9 @@ export const NotificationItem: React.FunctionComponent<NotificationItemProps> =
 
       try {
         if (op.resourcePath.indexOf(ApiEndpoint.Workspaces) !== -1) {
-          // we need the workspace to get auth details
           const wsId = op.resourcePath.split('/')[2];
-          ws = (await apiCall(`${ApiEndpoint.Workspaces}/${wsId}`, HttpMethod.Get)).workspace;
+          let scopeId = (await apiCall(`${ApiEndpoint.Workspaces}/${wsId}/scopeid`, HttpMethod.Get)).workspaceAuth.scopeId;
+          ws = (await apiCall(`${ApiEndpoint.Workspaces}/${wsId}`, HttpMethod.Get, scopeId)).workspace;
```
martinpeck, line 50: So, this calls the workspace endpoint with the scope. Makes sense.
What I'm less clear about are the lines of code that follow (which I realise aren't what you're changing, but it's hard to review this when there are other parts of this code that are confusing.

What do lines 52 to 55 do? The path is split. The number of parts are counted, and if there are 3 then we declare "we have a workspace!". Given this is how we determine we have a workspace, how do your changes work when we are **not** in the workspace context? 

Line 57 appears to say "if it **isn't** a workspace then get the resource using the workspace scope. Is this just poorly named variables? What will `ws.properties.scope_id` be equal to in this case? Is `ws`, in this case, not actually a workspace?

I guess I'm asking whether you've tested this...

- with a resource that is not a workspace
- with a resource that *is* a workspace
- with a resource that is a user resource within a workspace

marrobi: Got you, agree, needs more testing.

The question I need to answer is why are we getting the workspace before we actually know if we are in a workspace...

Counting segments doesn't seem great either, surely we can check for `/workspaces`. There is workspace context, but it gets set in the Workspace Provider which isn't yet loaded. I'm not sure if that could be used @damoodamoo @jjgriff93 ?

marrobi: I'm going to have another go at this, now got a better understanding, for each operation, need to say "is this a workspace", and if it is get the scope. Not sure the first API call is even needed.

marrobi: @martinpeck can you look at the update I've just pushed, does this make more sense?

damoodamoo: remember this is happening outside of any workspace context - the notifications can't use it the context as they span across all resources. We can't check for `/workspace` in the path as every path will start with `/workspace/`, regardless of what type of resource it is.

marrobi: Yup, get that now, check the latest changes.

damoodamoo (approved):
Pretty happy with this - so the core issue was that the TRE Researcher couldn't get the workspace details without the scope id. Remember this - and is why we added that `/scopeid` endpoint, so think this makes sense. Cheers, great bug to squash!

marrobi:
/test-force-approve

UI only.

PR #3727, opened 2023-09-25
Create KEY_VAULT_URL env var when a private.env file exists

### LINKED ISSUE #3711: Cannot build bundles after `make setup-local-debugging`
when i run make all 
tre produce this kind of error 

note   :  this type of error produce after run command >>> make setup-local-debugging
  

Runtime image build section isn't specified. Exiting...
could not render config file because ${secret.porter-db-connection-string} could not be resolved: rpc error: code = Unknown desc = could not get secret porter-db-connection-string: keyvault.BaseClient#GetSecret: Failure sending request: StatusCode=0 -- Original Error: Get "https://.vault.azure.net/secrets/porter-db-connection-string/?api-version=7.0": dial tcp: lookup .vault.azure.net: no such host
make[1]: *** [Makefile:189: bundle-build] Error 1
make[1]: Leaving directory '/workspaces/AzureTRE'
make: *** [Makefile:305: firewall-install] Error 2

marrobi:
# Resolves #3711

## How is this addressed

- Read vars form private.env when it exists and set KEY_VAULT_URL


devops/scripts/load_and_validate_env.sh
```diff
@@ -85,4 +85,13 @@ else
     export TRE_URL
 fi
 
+# if local debugging is configured, then set vars required by ~/.porter/config.yaml
+if [ -f "$DIR/../../core/private.env" ]; then
+  # shellcheck disable=SC1091
+  source "$DIR/../../core/private.env"
+  # shellcheck disable=SC2153
+  KEY_VAULT_URL=$KEYVAULT_URI
+  export KEY_VAULT_URL
```
tamirkamara, line 94: What is special about this one that it needs this special procedure?

marrobi: So to debug the resource processer the porter `config.yaml` is present. This requires the environment variable `KEY_VAULT_URL` to be set so the secret can be retrieved, this is set in the resource processor `config.py` code.

Without this env variable set the commands in the makefile that call porter fail as it can't resolve the keyvault to access the porter state. That make sense? If you have another way of doing this, do say so...

tim-allen-ck (approved):
LGTM

tim-allen-ck (approved):

Danny-Cooke-CK (approved):
approved

PR #3733, opened 2023-10-10
Update UI to better handle different roles

### LINKED ISSUE #3721: Workspace costs do not show in UI for a TRE User who is a workspace owner.
Log into the TRE with a user that isn't a TRE Admin, but is a Workspace Owner, they will not see the costs.

This is the TRE level token is being evaluated, not workspace level.

marrobi:
# Resolves #3721

## How is this addressed

- Update SecuredByRole to handle workpsace and app roles
- Update Cost tags to display for workspace serivces and user resources
- Hide History and Operations when user does not have access


ui/app/src/components/workspaces/UserResourceItem.tsx
```diff
@@ -10,30 +10,34 @@ import { useComponentManager } from '../../hooks/useComponentManager';
 import { ResourceBody } from '../shared/ResourceBody';
 
 interface UserResourceItemProps {
-  userResource?: UserResource
+  userResource?: UserResource;
```
damoodamoo, line 13: `nit` - pick `;` or `,` - or none, doesn't seem to matter which

damoodamoo (approved):

ui/app/src/components/workspaces/UserResourceItem.tsx
marrobi: Done

marrobi:
/test-force-approve 

UI only

PR #3739, opened 2023-10-12
Fix PivotItem/SecuredByRole render and workspace cost reloading

### LINKED ISSUE #3738: Nesting of PivotItems in SecuredByRole causes issues
As History and Operations PivotItems are wrapped by SecuredByRole they do not render.

marrobi:
Fixes #3738

## What is being addressed

- Cannot surround PivotItem by SecuredByRole element so using error message
- Costs relead when roles update

### DIFF
--- a/ui/app/package.json
+++ b/ui/app/package.json
@@ -1,6 +1,6 @@
 {
   "name": "tre-ui",
-  "version": "0.5.9",
+  "version": "0.5.10",
   "private": true,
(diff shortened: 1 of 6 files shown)


marrobi:
/test-force-approve

UI Only

damoodamoo (approved):
stupid fussy `<Pivot>` :)

PR #3745, opened 2023-10-16
Update porter version and on error getting porter outputs return dict

### LINKED ISSUE #3744: Porter command fails with `incomplete read of message header: context deadline exceeded`
Occasionally a porter command will fail with:

```
logging root    81      2023-10-16 12:51:24,402 WARNING [stderr] rpc error: code = DeadlineExceeded desc = invalid index specified: ([]mongo.IndexModel) (len=1 cap=1) {
 (mongo.IndexModel) {
  Keys: (primitive.D) (len=2 cap=2) {
   (primitive.E) {
    Key: (string) (len=9) "namespace",
    Value: (int64) 1
   },
   (primitive.E) {
    Key: (string) (len=4) "name",
    Value: (int64) 1
   }
  },
  Options: (*options.IndexOptions)(0xc0001e3800)({
   Background: (*bool)(<nil>),
   ExpireAfterSeconds: (*int32)(<nil>),
   Name: (*string)(0xc000061b20)((len=18) "namespace_1_name_1"),
   Sparse: (*bool)(<nil>),
   StorageEngine: (interface {}) <nil>,
   Unique: (*bool)(0xc00082ac58)(true),
   Version: (*int32)(<nil>),
   DefaultLanguage: (*string)(<nil>),
   LanguageOverride: (*string)(<nil>),
   TextVersion: (*int32)(<nil>),
   Weights: (interface {}) <nil>,
   SphereVersion: (*int32)(<nil>),
   Bits: (*int32)(<nil>),
   Max: (*float64)(<nil>),
   Min: (*float64)(<nil>),
   BucketSize: (*int32)(<nil>),
   PartialFilterExpression: (interface {}) <nil>,
   Collation: (*options.Collation)(<nil>),
   WildcardProjection: (interface {}) <nil>,
   Hidden: (*bool)(<nil>)
  })
 }
}
: connection(cosmos-mongo-xxxxx-westeurope.mongo.cosmos.azure.com:10255[-6]) incomplete read of message header: context deadline exceeded
rpc error: code = DeadlineExceeded desc = invalid index specified: ([]mongo.IndexModel) (len=1 cap=1) {
```

This has been seen retrieving outputs, and the status message fails as:
```
Service bus message is not formatted correctly: 1812c32e-b97e-4440-a4c1-dc868dfbadf5 Exception message: '1 validation error for ParsingModel[DeploymentStatusUpdateMessage] __root__ -> outputs value is not a valid list (type=type_error.list)'

```
```
{"operationId": "35405209-2031-48c4-9f99-17ab2e847f52", "stepId": "8fb74d96-8a80-4f60-8260-23eb37b82e1c", "id": "1812c32e-b97e-4440-a4c1-dc868dfbadf5", "status": "updated", "message": "1812c32e-b97e-4440-a4c1-dc868dfbadf5: upgrade action completed successfully.", "outputs": ""}
```

marrobi:
# Resolves  #3744

- Update porter to v1.0.15 as has updated mongo drivers
- If cannot get outputs return dict so message can be parsed as DeploymentStatusUpdateMessage


resource_processor/vmss_porter/runner.py
```diff
@@ -213,7 +213,7 @@ async def get_porter_outputs(msg_body: dict, message_logger_adapter: logging.Log
     if returncode != 0:
         error_message = "Error context message = " + " ".join(err.split('\n'))
         message_logger_adapter.info(f"{get_installation_id(msg_body)}: Failed to get outputs with error = {error_message}")
-        return False, ""
+        return False, {}
```
martinpeck, line 216: I guess this is fine, right?

image

This change as prompted more questions than it answers.

We appear to return `False`  in this situation, but on line 229 (where we get invalid JSON) we simply log this event and then return `True` on line 231.

Then, back on line 194, we appear to totally ignore the `True`/`False` flag, and we report that everything was successful.

This change will therefore further mask issues. I assume things are currently blowing up because the porter command failed, and your proposed change will hide this, and report success.

Perhaps I'm missing something that makes this an OK thing to do.

marrobi: Hmm, thanks for looking, correct. I was just concerned with the service bus message being valid.

I don't think the change makes things worse, because at the moment the operation just sticks in an updating state and the user isnt. However it does look like the True/False should be handled differently. 

Maybe needs a separate issue to ensure if it's false then the operation is marked as failed? Similar to:

https://github.com/microsoft/AzureTRE/blob/ad647cecb6b97220ade13a1195e2d81cebed4736/resource_processor/vmss_porter/runner.py#L179-L182

martinpeck: Agreed. I think the pattern you've highlighted is exactly what's needed.

Couldn't the empty dictionary be replaced with something that held and error status message, and the `success_message` be changed to say something else if the return value is False? If we did that in this PR, that would unblock the issue you see *and* make things easier to debug when they fail, right? 

As you say, your change doesn't make things worse. If we really need to make this change now, then I'll approve, but it doesn't feel like it would be much more work to make it better.

marrobi: @martinpeck I've made some changes, haven't tested yet, but value feedback.

martinpeck: Assuming testing is good, this looks good to me! Thanks!

martinpeck (approved):
(subject to testing) LGTM!

PR #3753, opened 2023-10-17
SecuredByRole fails if app roles are not loaded

### LINKED ISSUE #3752: SecuredByRole fails if app roles are not loaded
If the securedbyrole element renders prior to approles being loaded get an issues referencing an undefined variable.

Needs to update to: 

```
    (workspaceRoles?.some(x => props.allowedWorkspaceRoles?.includes(x)) || appRoles?.roles?.some(x => props.allowedAppRoles?.includes(x)))

```

marrobi:
Fixes #3752

Update so that app roles being undefined does not cause an issue.


marrobi:
/test-force-approve 

UI only.

damoodamoo (approved):

marrobi:
/test-force-approve

UI only.

Issue #3757, opened 2023-10-18
Problem creating Airlock Import request due to having +20 users/groups on a given App Registration


migldasilva:
**Describe the bug**
An error is raised when creating an Airlock Import request, if the given App Registration has more than 20 users/groups assigned to it. 

After tracking the code and log messages, it was possible to observe the error being raised from the line [248](https://github.com/microsoft/AzureTRE/blob/90b6faed6e6dd4453625dae83d0b295850989ec1/api_app/services/aad_authentication.py#L248), in the file `api_app/services/aad_authentication.py`.

Line [248](https://github.com/microsoft/AzureTRE/blob/90b6faed6e6dd4453625dae83d0b295850989ec1/api_app/services/aad_authentication.py#L248) makes a call to the Microsoft Graph JSON batch.

Following documentation posted [here](https://learn.microsoft.com/en-us/graph/json-batching), we can realize that a maximum of 20 request is in place.

I have been working in a workaround/patch for such code. However, is the original code created in such way deliberately?

**Steps to reproduce**

1. Have a TRE environment with at least 1 workspace available
2. Locate the App Registration for the a given Workspace, and add more than 20 users/groups
3. Create an Airlock Import request
4. Review logs from the Web App running the API

**Azure TRE release version (e.g. v0.14.0 or main):**

**Deployed Azure TRE components - click the (i) in the UI:**

marrobi:
Good catch. I expect it's never been tested with more than 20 requests. 

Something like the below (untested) should fix it...

```
def _get_user_emails(self, roles_graph_data, msgraph_token):
    batch_endpoint = self._get_batch_endpoint()
    batch_requests = self._split_batch_requests(self._get_batch_users_by_role_assignments_body(roles_graph_data))
    headers = self._get_auth_header(msgraph_token)
    headers["Content-type"] = "application/json"
    users_graph_data = []
    for batch_request in batch_requests:
        batch_response = requests.post(batch_endpoint, json=batch_request, headers=headers).json()
        users_graph_data += batch_response["responses"]
    return {"responses": users_graph_data}

def _split_batch_requests(self, batch_request_body):
    requests_list = batch_request_body["requests"]
    return [ {"requests": requests_list[i:i+20]} for i in range(0, len(requests_list), 20) ]
```

Can you test and create a PR? Thanks.

migldasilva:
@marrobi Did you read my mind? I even used list compressions, as you did. :)

[...]

PR #3759, opened 2023-10-18
Enabling +20 users/groups on batch endpoint payload

migldasilva:
# Enabling +20 users/groups on batch endpoint payload

## What is being addressed

An error is raised when creating an Airlock Import request, if the given App Registration has more than 20 users/groups assigned to it. As described in [this bug report](https://github.com/microsoft/AzureTRE/issues/3757).

## How is this addressed

- Changing Core API's Python code, so that +20 users/groups are supported

Also changed rsa package, not sure why but tests couldn't find existing one.


marrobi (approved):
LGTM

PR #3806, opened 2023-12-06
Move to using managed identity for auth to CosmosDB.

### LINKED ISSUE #345: [Task] API to use Managed Identity to authenticate against Cosmos DB

marrobi:
# Resolves #345

## What is being addressed
*Waiting on Logging PR #3762*

- Move to managed identity when connecting to the database when available.
- Use same connect method for health checks.
- Only attempt to create collection on bootstrap
- Split out control plane and data plane operations as have different permissions model.
- Create singleton database class to enable reuse of the cosmos client.
- Update Change Log


DragosDanielBoia:
@microsoft-github-policy-service rerun

damoodamoo (approved):

marrobi:
Ok, I've ended up refactoring the database connections to use a singleton class as the credential was being closed as per - https://github.com/Azure/azure-sdk-for-python/issues/32309

Can someone check and let me know if this is good/bad, both from a principle, and the actual Python. I had to rely on Stack Overflow!

As a result everything is running much faster! We were seeing this:

image

But as only trying to create the collections on bootstrap this has gone.

marrobi:
This now seems, but is a bit of a mess. @martinpeck if have some time in the new year, wouldn't mind talking through the intricacies of Python and async context mangers. I might try doing the PR from scratch using what I've learnt!

PR #3825, opened 2024-01-11
Refactor validate_template_version_patch method to use parent_service_template_name instead of parent_resource_id

### LINKED ISSUE #3824: Cannot upgrade user reources
When trying to upgrade a user resource to a new bundle we get an error similar to:
> Error: Template 'tre-service-guacamole-windowsvm' not found for resource type 'user-resource' with target template version '0.7.10'

this is as the following code in `api_app/db/repositories/resource_templates.py`

```python
    async def get_template_by_name_and_version(self, name: str, version: str, resource_type: ResourceType, parent_service_name: Optional[str] = None) -> Union[ResourceTemplate, UserResourceTemplate]:
        """
        Returns full template for the 'resource_type' template defined by 'template_name' and 'version'

        For UserResource templates, you also need to pass in 'parent_service_name' as a parameter
        """
        query = self._template_by_name_query(name, resource_type) + f' AND c.version = "{version}"'

        # If querying for a user resource, we also need to add the parentWorkspaceService (name) to the query
        if resource_type == ResourceType.UserResource:
            if parent_service_name:
                query += f' AND c.parentWorkspaceService = "{parent_service_name}"'
            else:
                raise Exception("When getting a UserResource template, you must pass in a 'parent_service_name'")
```

As  `parent_service_name` is being passed in as and ID rather than a name.

marrobi:
# Fixes #3824 

This pull request refactors the validate_template_version_patch method in order to use the parent_service_template_name instead of the parent_resource_id.  This was preventing user resource upgrades from working.


marrobi:
> @marrobi Can you confirm the right issue is linked? I don't see the relationship between Key Vault secrets restore and this PR.

It is now, sorry.

marrobi:
/test-force-approve

Passed: https://github.com/microsoft/AzureTRE/actions/runs/7499784309

SvenAelterman (approved):
LGTM, tests pass

PR #3841, opened 2024-02-08
fixing react router dom version and switching to absolute routing whe…

### LINKED ISSUE #3830: Airlock: Creating an import/export request causes a routing error
**Describe the bug**
Upon creating a new export/import request in Airlock, the URL route malforms and causes the UI to behave in an weird way. The issue is to do with routing when creating a new request. I have already forked the repo and fixed the issue. The issue to be exact is here: https://github.com/microsoft/AzureTRE/blob/31960894e15266f6ae76a4b7c1028cf9cc6bc459/ui/app/src/components/shared/airlock/Airlock.tsx#L258 The path is being navigated from a relative point of view and just appended rather then re-written. The fix I put in the fork is the following:
```typescript
    navigate(`../${newRequest.id}`, { relative: "path" });
``` 

**Steps to reproduce**

1. Open/create a new workspace and navigate to **Airlock** on the side panel.
2. Create a new import or export request.
3. Fill in the form and press create.

Expected:
The draft request to be created and the side panel to open automatically with the SAS URL information, with the URL `/requests/{id}`.

Actual:
The side panel closes and the path remains `/requests/new/{id}` but the side panel doesn't reopen with the SAS URL information. Double clicking requests simply appends the airlock id to the URL.

Work around: 
Exit and enter back to the requests list view and double clicking from there allows the side panel to open.

**Azure TRE release version**: v0.15.2, **UI version**: 0.5.17, (but can also see the issue on main).

yahya130:
Fixes #3830  

This PR fixes the following:

- Nails down the react-dom-router dependency to version 6.21.1.
- Updates the routing path to be more absolute when creating an airlock request.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 <!-- markdownlint-disable MD041 -->
 ## 0.17.0 (Unreleased)
 
+BUG FIXES:
+* Airlock: Creating an import/export request causes a routing error ([#3830](https://github.com/microsoft/AzureTRE/issues/3830))
+
 **BREAKING CHANGES & MIGRATIONS**:
 
 FEATURES:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,6 @@
 <!-- markdownlint-disable MD041 -->
 ## 0.17.0 (Unreleased)
 
-BUG FIXES:
-* Airlock: Creating an import/export request causes a routing error ([#3830](https://github.com/microsoft/AzureTRE/issues/3830))
-
 **BREAKING CHANGES & MIGRATIONS**:
 
 FEATURES:
@@ -19,6 +16,7 @@ BUG FIXES:
 * Fix issue with workspace menu not working correctly([#3819](https://github.com/microsoft/AzureTRE/issues/3819))
 * Fix issue with connect button showing when no uri([#3820](https://github.com/microsoft/AzureTRE/issues/3820))
 * Fix user resource upgrade validation: use the parent_service_template_name instead of the parent_resource_id. ([#3824](https://github.com/microsoft/AzureTRE/issues/3824))
+* Airlock: Creating an import/export request causes a routing error ([#3830](https://github.com/microsoft/AzureTRE/issues/3830))
 
 COMPONENTS:
 
--- a/ui/app/package.json
+++ b/ui/app/package.json
@@ -1,6 +1,6 @@
 {
   "name": "tre-ui",
-  "version": "0.5.19",
+  "version": "0.5.20",
   "private": true,
   "dependencies": {
     "@azure/msal-browser": "^2.35.0",
@@ -24,7 +24,7 @@
     "react-dom": "^18.2.0",
     "react-markdown": "^8.0.3",
     "react-redux": "^8.0.4",
-    "react-router-dom": "6",
+    "react-router-dom": "6.21.1",
     "remark-gfm": "^3.0.1",
     "typescript": "^5.1.6",
     "web-vitals": "^3.3.0"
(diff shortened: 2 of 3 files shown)


marrobi:
Thanks @yahya130 

@tamirkamara any idea why `flag_external_pr` is failing .

Could it be:
> This org has transitioned to a read-only GITHUB_TOKEN for GitHub Action workflows. This does break your builds. See https://aka.ms/github-token-perms-changes for how to resolve.

CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 <!-- markdownlint-disable MD041 -->
 ## 0.17.0 (Unreleased)
 
+BUG FIXES:
```
marrobi, line 4: Looks like we have two lots of bug fixes for this release, can you add lower down. Thanks.

marrobi:
> Thanks @yahya130
> 
> @tamirkamara any idea why `flag_external_pr` is failing .
> 
> Could it be:
> 
> > This org has transitioned to a read-only GITHUB_TOKEN for GitHub Action workflows. This does break your builds. See https://aka.ms/github-token-perms-changes for how to resolve.

Actually can see the GitHub Scripts action uses GITHUB_TOKEN.  Will have to look to het the workflows fixed.

tamirkamara:
@marrobi it is this. I'm adding a permissions block in a pr.

> > any idea why `flag_external_pr` is failing .
> > Could it be:
> > > This org has transitioned to a read-only GITHUB_TOKEN for GitHub Action workflows. This does break your builds. See https://aka.ms/github-token-perms-changes for how to resolve.
> 
> Actually can see the GitHub Scripts action uses GITHUB_TOKEN. Will have to look to het the workflows fixed.

jjgriff93 (approved):

benc-uk:
Random (ish) stranger here! 👋
Just a suggestion - it's normally good practice to add the update the updated package-lock.json whenever the package.json is updated.
And actually changes to dependency versions in the package.json might not even take place until the lockfile is updated too

yahya130:
> Random (ish) stranger here! 👋 Just a suggestion - it's normally good practice to add the update the updated package-lock.json whenever the package.json is updated. And actually changes to dependency versions in the package.json might not even take place until the lockfile is updated too

Hi @benc-uk  👋, not random-ish at all 😄 great suggestion and well noted, just to add more context, the app uses yarn as the package manager and the package.json file has been updated to reflect what it is in the yarn.lock file.

marrobi:
@yahya130 @benc-uk should there be an updated `yarn.lock` file then? There is one in the repo, but not this PR.

yahya130:
> @yahya130 @benc-uk should there be an updated `yarn.lock` file then? There is one in the repo, but not this PR.

It looks ok:
https://github.com/microsoft/AzureTRE/blob/31960894e15266f6ae76a4b7c1028cf9cc6bc459/ui/app/yarn.lock#L9395-L9408

@benc-uk what would be your opinion on this?

benc-uk:
Totally missed you're using yarn
Yep that looks good!

marrobi:
@tamirkamara looks like the bot is fixed. Thanks.

@yahya130 As this is UI and doesn't have tests, but will let it run through so we can ensure our builds are fixed, then merge. Thanks!

marrobi:
Looks like our tests are blocked by - https://github.com/getporter/porter/issues/2987

Will force approval given this only touches UI. But the bot seems to be fixed.

marrobi:
@tamirkamara seems still an issue, tests aren't being marked as complete. Looks like granted the right permissions here - https://github.com/microsoft/AzureTRE/blob/3835e5ad89209d55c131eaa8e6b4211fbcc6c9ff/.github/workflows/pr_comment_bot.yml#L21 , can't see any errors either.

Any ideas?

PR #3845, opened 2024-02-08
Add permissions to parent workflows

tamirkamara:
## What is being addressed

In a previous PR I've added permissions to reusable workflows and that needs to propagate up to the calling workflows too (if not, they assign default permissions which isn't enough).

Unfortunately this can't be tested until it reaches the main branch.


marrobi (approved):
LGTM

PR #3851, opened 2024-02-15
Update CodeQL actions

tamirkamara:
## What is being addressed

CodeQL v2 is deprecated which might cause problems in getting results.

## How is this addressed

- Update to v3.
- Also update the checkout action to the latest.


marrobi (approved):
LGTM

PR #3856, opened 2024-02-23
Add report check status to Test Results

marrobi:
Test results check not being marked as complete as the triggering sha is not the correct one. Use an action to mark the check as complete.


marrobi:
Tested here - https://github.com/marrobi/AzureTRE/actions/runs/8065472139/job/22031338742

damoodamoo (approved):

PR #3858, opened 2024-02-27
Move network rule set for Service Bus namespace

### LINKED ISSUE #3783: Terraform for Service Bus uses deprecated resource type `azurerm_servicebus_namespace_network_rule_set`
**Describe the bug**

When deploying TRE core (e.g., with `make deploy-core`), I observed this warning from Terraform:

 | Warning: Deprecated Resource
│ 
│   with azurerm_servicebus_namespace_network_rule_set.servicebus_network_rule_set,
│   on servicebus.tf line 78, in resource "azurerm_servicebus_namespace_network_rule_set" "servicebus_network_rule_set":
│   78: resource "azurerm_servicebus_namespace_network_rule_set" "servicebus_network_rule_set" {
│ 
│ The `azurerm_servicebus_namespace_network_rule_set` resource is deprecated and will be removed in version 4.0 of the AzureRM provider. Please use `network_rule_set` inside the `azurerm_servicebus_namespace` resource instead.

**Azure TRE release version (e.g. v0.14.0 or main):** v0.15.2

marrobi:
# Fixes #3783

- Update terraform for services bus to move network rules into namespace resource to avoid depreciation warning.
- Update setup_local_debugging.sh to use network_rule_sets.
- [x] Update changelog.


core/terraform/servicebus.tf
```diff
@@ -6,6 +6,24 @@ resource "azurerm_servicebus_namespace" "sb" {
   capacity            = "1"
   tags                = local.tre_core_tags
 
+  # Block public access
```
tamirkamara, line 9: Has this been tested on an existing environment?
Switching the way inside TF we accomplish the same Azure setup has caused issues in the past (required a migration or other mitigation).

tamirkamara (approved):

core/terraform/servicebus.tf
marrobi: Hmm, good point, no it hasn't. I'll try give it a test.

marrobi: Seems like some discussion here - https://github.com/hashicorp/terraform-provider-azurerm/issues/23954

Might need to add:

`terraform state rm azurerm_servicebus_namespace_network_rule_set.servicebus_network_rule_set`

to the `migrate.sh`.

marrobi: I've added the migration step. Deployed from main, then from this branch. All good, can see it being removed from state in the deployment step.

PR #3864, opened 2024-02-29
Destroy TRE script unbound variable fix

jonnyry:
Small bug in the `destroy_env_no_terraform.sh` script -

In the for loops that delete keyvault keys and certificates, the `${secret_id}` variable is used in both loops instead of `${key_id}` and `${certificate_id}` respectively.

This only manifests with an error ("unbound variable") if your key vault doesn't contain secrets (mine did due to another unrelated issue) so won't occur under normal circumstances, however the stdout of the script will be incorrect.


marrobi (approved):
LGTM

marrobi:
/test-force-approve not tested.

Thanks @jonnyry

PR #3868, opened 2024-03-01
Update Docker dependency update schedule

### LINKED ISSUE #3867: Add Docker to dependabot
We should have Docker enabled as a package source in dependabot.

marrobi:
# Fixes #3867

This pull request updates the Docker dependency update schedule in the dependabot.yml file. The schedule is now set to run monthly and will ignore any version updates that are only semver patches. This change ensures that the Docker dependencies are regularly updated while avoiding unnecessary pull requests for minor version updates.


imranckd (approved):
Reviewed

PR #3877, opened 2024-03-18
snyk Dockerfile python upgrades

nasaldrops:

## What is being addressed

## How is this addressed

- Update documentation


api_app/Dockerfile
tamirkamara, line 1: I'll put a few general comments here:
1. I think you should split the core application base image from the changes in the templates to 2 different PRs.
2. Are you sure a jump (in the apps) from python 3.8 to 3.13 doesn't require code changes? Why are you upgrading to a RC version of python - it should be a GA one.
3. You didn't upgrade the airlock function, why?
4. In the apps, I think you should increase the minor version and not just the build number.
5. We usually keep the same Debian version used across the entire solution - devcontainer, core apps and templates. You didn't include a devcontainer change which I think must be synced with the apps to enable good developer experience.

PR #3916, opened 2024-05-01
 Remove TLS1.0/1.1 support. Resolves #3914

### LINKED ISSUE #3914: Remove TLS1.0/1.1 support
Scanning the Azure TRE web portal shows TLS1.0 & 1.1 are enabled:

327127209-1ad2cbb2-ecb9-4d45-9eb5-91857e35aefb

TLS 1.0 and 1.1 are generally considered insecure and should be removed from internet exposed web endpoints (managed by the App Gateway), leaving TLS 1.2 and above available.

App Gateway TLS config documentation: https://learn.microsoft.com/en-us/azure/application-gateway/application-gateway-ssl-policy-overview

By default, the TRE is using 20150501 and should move to at least 20220101

The Azure App Gateway does not specify a TLS policy version explicitly:

https://github.com/microsoft/AzureTRE/blob/main/core/terraform/appgateway/appgateway.tf

And so relies on the default:

https://learn.microsoft.com/en-us/azure/application-gateway/application-gateway-ssl-policy-overview

Defaults:

    20150501 for Az API version < 2023-02-01
    20220101 for Az API version >= 2023-02-01

Given Azure resources are created by terraform, we are reliant on the Azure API version used by the Terraform Azure Provider to determine the above default.  Terraform does not easily reveal which API version it is using under the hood.

See Terraform Azure Provider ticket relating to this exact issue: https://github.com/hashicorp/terraform-provider-azurerm/issues/23995#issuecomment-1840538128

Therefore, suggest we explicitly set the TLS policy version in Terraform, to prevent picking up the default.

jonnyry:
# Resolves #3914

## What is being addressed

Removes TLS1.0/1.1 support.  Please see #3914.

## How is this addressed

Explicitly sets the SSL policy within the App Gateway to the one currently recommended by Microsoft `AppGwSslPolicy20220101`.

As this was not explictly set in Terraform it was using an older `AppGwSslPolicy20150501` which still supported TLS1.0/1.1.  See the issue for more details on this.

Once deployed, the Azure TRE portal no longer reports running 1.0/1.1:

Capture

(Please see the issue for the "before" report)

@marrobi PR as discussed


marrobi (approved):
LGTM.

tim-allen-ck (approved):
LGTM

jonnyry:
Looks like its failing on a version check.

@tim-allen-ck Do I just bump this to 0.9.9?

image

PR #3954, opened 2024-05-28
Improve documentation on TRE teardown

### LINKED ISSUE #3777: Improve documentation on TRE tear down
The existing documentation on TRE tear down doesn't make clear the difference between using `make tre-destroy` and `az group delete`:

https://microsoft.github.io/AzureTRE/v0.15.2/tre-admins/tear-down/

`az group delete` despite deleting the resource groups, leaves behind a number of resources such as the key vault, diagnostic log settings and others, which causes a subsequent deployment of the TRE to fail multiple times when deploying terraform.

`make tre-destroy` remedies this by seperately deleting the resources left behind as can be seen in [`destroy_env_no_terraform.sh`](https://github.com/microsoft/AzureTRE/blob/main/devops/scripts/destroy_env_no_terraform.sh).

The documentation would be more helpful to first time users, if the pitfalls of manually deleting the resources groups were pointed out.  I notice a number of tickets where people have fallen into this treap and then have deployed to a new TRE_ID to work around the issue.

jonnyry:
# Resolves #3777

## What is being addressed

Add a little info to the docs to nudge users to use `make tre-destroy` rather than `az group delete` since the latter will cause deployment errors if trying to re-deploy using the same TRE name.

### DIFF
--- a/docs/tre-admins/tear-down.md
+++ b/docs/tre-admins/tear-down.md
@@ -6,7 +6,7 @@ To remove the Azure TRE and its resources from your Azure subscription run:
 make tre-destroy
 ```
 
-Alternatively, you can delete the resource groups in Azure Portal or using the CLI:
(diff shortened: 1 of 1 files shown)


marrobi (approved):
LGTM

tim-allen-ck (approved):
LGTM

PR #3968, opened 2024-06-11
Fileshare linux mgmt

TonyWildish-BH:
## What is being addressed

The shared storage on Linux VMs is now mounted permanently, and under /shared-storage, not /fileshares/...

## How is this addressed

The only changes were in `templates/workspace_services/guacamole/user_resources/guacamole-azure-linuxvm/terraform/vm_config.sh`, where the automount stuff was commented out and code added to update `/etc/fstab` instead.


TonyWildish-BH:
...opened in the wrong place by mistake...

PR #3975, opened 2024-06-14
Expose FIREWALL_SKU as environment variable & support start/stop of SKU Basic Firewall

### LINKED ISSUE #3961: Cost optimisation - Firewall Basic SKU
Is it possible to use the Azure Firewall Basic SKU to reduce the running cost in non production environments?

I know this has previously been discussed here #2723 - it reads like the main blocker at the time was the inability to deallocate the firewall Basic SKU.

I've just tested out deallocation and allocation of the policy on a Basic SKU Firewall and it now appears to be possible:

image

@marrobi @tamirkamara Is there a finer point in the allocation/deallocation that I might have missed?  Were there any other blockers to using the Basic SKU?

jonnyry:
# Resolves #3961

## What is being addressed

1. Expose a FIREWALL_SKU environment variable into `config.yaml` and to be set in GitHub Actions as a variable.  If the variable is not present or set to an empty string, the build should default to a Standard SKU firewall.

2. Modify the start/stop scripts (both Powershell and `make tre-start/tre-stop`) to allow for the stopping and starting of the Basic SKU Firewall


marrobi (approved):
Looks good, do we need to point out the other option is Basic in the TRE docs and config, and maybe link to the Azure docs with the feature comparison?

jonnyry:
@marrobi thanks for review.  I've added the three options (Basic, Standard, Premium) with a link to the feature comparison to the docs - see most recent commit.

tim-allen-ck (approved):
LGTM

core/version.txt
```diff
@@ -1 +1 @@
-__version__ = "0.9.10"
+__version__ = "0.9.11"
```
tamirkamara, line 1:
```suggestion
__version__ = "0.10.0"
```

resource_processor/_version.py
```diff
@@ -1 +1 @@
-__version__ = "0.8.6"
+__version__ = "0.8.7"
```
tamirkamara, line 1:
```suggestion
__version__ = "0.9.0"
```

templates/shared_services/firewall/porter.yaml
```diff
@@ -45,7 +45,7 @@ parameters:
     type: string
     default: "W10=" # b64 for []
     description: "Network rule collection array"
-  - name: sku_tier
+  - name: firewall_sku
```
tamirkamara, line 48: I can't leave this comment in the right place since it's too far from the first change in the file, but I think you should also update this bundle's version on line 3.

templates/shared_services/firewall/terraform/locals.tf
```diff
@@ -15,4 +15,7 @@ locals {
   api_driven_network_rule_collection     = jsondecode(base64decode(var.api_driven_network_rule_collections_b64))
 
   firewall_policy_name = "fw-policy-${var.tre_id}"
+
+  default_firewall_sku   = "Standard"
+  effective_firewall_sku = coalesce(var.firewall_sku, local.default_firewall_sku)
```
tamirkamara, line 20: Why was this needed (vs using the default value)?

tamirkamara (approved):
Just a few versions changes as this adds support for a new feature the solution didn't allow before.

core/version.txt
jonnyry: Thanks - added in https://github.com/nwsde/nwsde-azuretre/commit/ad5ac13907869df2efb262dd0a6b993b7df2b504

resource_processor/_version.py
jonnyry: Thanks - added in https://github.com/nwsde/nwsde-azuretre/commit/ad5ac13907869df2efb262dd0a6b993b7df2b504

templates/shared_services/firewall/porter.yaml
jonnyry: Thanks - added in https://github.com/nwsde/nwsde-azuretre/commit/ad5ac13907869df2efb262dd0a6b993b7df2b504

templates/shared_services/firewall/terraform/locals.tf
jonnyry: It's to handle the FIREWALL_SKU parameter being present but its value being an empty string.  There was a situation where this occurred while I was testing... let me dig it out.

jonnyry: Ah OK... yes... it was to handle the situation where people had not set `FIREWALL_SKU` as a GitHub Actions variable.  I suppose I could have defaulted the GitHub Action input parameter to 'Standard' but it felt more appropriate to keep the default value in a single place & closer to the actual firewall terraform - since there are multiple routes for the parameter being passed in (from GitHub Actions variables, from config.yaml, via the resource processor).

https://github.com/nwsde/nwsde-azuretre/blob/ad5ac13907869df2efb262dd0a6b993b7df2b504/.github/actions/devcontainer_run_command/action.yml#L128

https://github.com/nwsde/nwsde-azuretre/blob/ad5ac13907869df2efb262dd0a6b993b7df2b504/.github/actions/devcontainer_run_command/action.yml#L241

tamirkamara: ok. thanks for the explanation

jonnyry:
@tim-allen-ck has the Smoke test failed?  I can't see it in the Actions tab

tim-allen-ck:
> @tim-allen-ck has the Smoke test failed? I can't see it in the Actions tab

I've run the extended tests, but they're currently failing atm, ill give you an update by the end of the day.

jonnyry:
@tim-allen-ck is this OK to merge in now?

(ps. i've just updated the branch with the latest changes, hence why the jobs have kicked off again)

jonnyry:
I think this PR may need recreating... I used the rebase option to pull this commit from main on top of the PR and looks like its not applied it (and broken the history):

https://github.com/microsoft/AzureTRE/commit/59d006a872f14af07b423957c7b1e0f9a1bed5fd

tim-allen-ck:
> I think this PR may need recreating... I used the rebase option to pull the commit below from main on top of the PR and looks like its not applied it:
> 
> 
> 
> https://github.com/microsoft/AzureTRE/commit/59d006a872f14af07b423957c7b1e0f9a1bed5fd
> 
> 
> 
> ---
> 
> 
> 
> Update - actually it does look OK, I can see the fix applied at the head of the branch:
> 
> 
> 
> https://github.com/nwsde/nwsde-azuretre/blob/jr/upstream-main/44-firewall-sku/templates/workspace_services/mysql/terraform/mysql.tf#L31
> 
> 
> 
> Though let me know if you'd rather not merge in a rebased branch as I am aware there are potential problems and I will recreate it.
> 
> 
> 
> @tim-allen-ck 

That's okay happy to merge

PR #3991, opened 2024-06-20
Core provider update

### LINKED ISSUE #3919: Update Core Terraform Provider Versions
Updating the provider versions for all core resources.

tim-allen-ck:
# Resolves #3919


marrobi (approved):
LGTM

core/terraform/.terraform.lock.hcl
marrobi, line 1: Why are you removing the lock file? Think is good to ensure the provider binaries actually match?

tim-allen-ck: added back in

Danny-Cooke-CK (approved):

PR #3994, opened 2024-06-21
MySQL Recreation Fix

### LINKED ISSUE #3993: MySQL database is destroyed if MySQL workspace service is updated
### Issue

If you update the MySQL workspace service it will destroy and recreate the attached MySQL database.

This also occurs with Gitea as it creates a MySQL database.

This occurs on [current main](https://github.com/microsoft/AzureTRE/tree/1ffb09baf37f4599adfd65b4259fdda7564da408)

### Example with the MySQL workspace service:

**1. Install the MySQL component:**

Porter logs show the following:

image

**2. Update the MySQL component (clicking the three dots, then Update in the UI):**

Porter logs show the destruction of the database:

image

tim-allen-ck:
# Resolves #3993

## What is being addressed

Added lifecycle rule for charset and collation to mysql resource to stop database recreating.


jonnyry:
@tim-allen-ck any chance you could fix in the Gitea shared service template as well?

This is where I discovered it first (and what I'm trying to get working at the moment).

Thanks

marrobi (approved):
LGTM

marrobi:
/test-force-approve

Not covered by tests

jonnyry:
Just tested the fix on my local instance - with Gitea rather than MySql component and still getting the issue:

image

For some reason its trying to force `utf8mb4` back to `utf8mb3`, even though both are valid charsets.

I can't tell whether its terraform or Azure that is causing the conversion - had a brief look at terraform but couldn't see anything.

marrobi:
Might be with adding an ignore_changes for those attributes of can't get it to settle.

tim-allen-ck:
> Might be with adding an ignore_changes for those attributes of can't get it to settle.

Changed to use a lifecycle rule, should cover newer versions of the charset/collation when they're introduced

Danny-Cooke-CK (approved):
approve

tim-allen-ck:
@jonnyry added fixes for gitea

PR #4022, opened 2024-07-03
Disable public network access to API App Service - traffic is routed through PE from App Gateway

### LINKED ISSUE #3986: Public network access to the API app service is enabled
Public network access is enabled to the API app service, even though it has a private endpoint which the app gateway uses (https://microsoft.github.io/AzureTRE/v0.16.0/assets/architecture-azure.png)

`public_network_access_enabled` is not defined here (and therefore defaults to `true`):

https://github.com/microsoft/AzureTRE/blob/1ffb09baf37f4599adfd65b4259fdda7564da408/core/terraform/api-webapp.tf#L20

image

jonnyry:
# Resolves #3986

## What is being addressed

Disable public network access to the API App Service - Web traffic should come via the App Gateway and not direct.  App Gateway uses a PE to communicate with the API App Service and therefore does not need a public endpoint.


marrobi (approved):
LGTM. Thanks.

tim-allen-ck (approved):
LGTM - not tested though 
Assuming won't break anything existing?

PR #4034, opened 2024-07-16
Remove AppServiceFileAuditLogs diagnostic setting

### LINKED ISSUE #4033: AppServiceFileAuditLogs diagnostic setting - remove?
The `AppServiceFileAuditLogs` diagnostic setting is enabled for App Services:

https://github.com/microsoft/AzureTRE/blob/43eb846c726073d454f8c0e71a767e771c58e7d1/core/terraform/locals.tf#L8-L11

It is a Premium SKU only setting, so you need a Premium App Service plan for it to be enabled.

There are zero logs for a 3 month time range in our core log analytics workspace, so there appears to be little benefit in having it enabled.

Unless there was an specific intention behind this setting, I propose it is removed, in order to remove the requirement on Premium SKU App Service plans.  This can also be applied to the Guacamole service and other App Service instances.

jonnyry:
# Resolves #4033

## What is being addressed

Remove `AppServiceFileAuditLogs` diagnostic setting from App Service plans as it prevents App Service from being on Dev/Test SKU plans, and is not providing a measurable benefit (see issue).


marrobi:
OK, I think we should look at moving all diagnostic settings like this to be applied via Azure policy.The policies can be applied to the resource grips we deploy. Would simplify the repo and help people with compliance.

Think might already be an issue for this.

marrobi (approved):

marrobi:
See https://github.com/microsoft/AzureTRE/issues/50

jonnyry:
Hi @tim-allen-ck would it be possible to get a merge on this one?  Thanks, Jonny

jonnyry:
Fab thank you very much :-)

PR #4058, opened 2024-08-06
Add route to retrieve users assigned to a workspace with UI

### LINKED ISSUE #4049: View users who are assigned roles within a workspace
TRE Admins and all workspace users should be able to retrieve which users are assigned to a workspace.

This can be done by adding an additional route to `workspaces_shared_router`, for example: `/api/workspaces/{workspace_id}/users/`.

The MS Graph API should be used to retrieve roles from the Enterprise Application. This can be done using the call described here - https://learn.microsoft.com/en-us/graph/api/serviceprincipal-list-approleassignedto?view=graph-rest-1.0&tabs=http , `GET /servicePrincipals(appId='{appId}')/appRoleAssignedTo`

This needs to appear in the Workspace Swagger docs under a section Users.

The API response should return a list of users, including their name (principalDisplayName), email (derived from another graph call using principalId), and the app roles assigned to the user.

The user interface needs a new section on the left hand menu under Airlock, called Users. This should display a list of users including their  name, email and role(s).

marrobi:
This pull request adds a new route to the `workspaces_shared_router` that allows users to retrieve the list of users assigned to a workspace. The route is `/api/workspaces/{workspace_id}/users/`.

The implementation uses the MS Graph API to retrieve the roles from the Enterprise Application. The API call `GET /servicePrincipals(appId='{appId}')/appRoleAssignedTo` is used to fetch the app roles assigned to the user.

The response of the API includes the name, email, and app roles of each user. The user interface also includes a new section called "Users" in the left-hand menu under Airlock. This section displays the list of users with their name, email, and role(s).

image

Also updated and refactored some of the repitition in the aad tests.

Fixes #4049


marrobi:
@jonnyry - or anyone else, if you fancy giving this a test, would welcome feedback. :)

jjgriff93:
I did the designs a while ago where users in a workspace were shown using an [avatar group](https://628d031b55e942004ac95df1-fbybvwbkdo.chromatic.com/?path=/story/components-avatargroup--default) on the workspace nav bar ([design here](https://www.figma.com/design/Ncl8TJsFf2WKf3frrqTbtq/AzureTRE-OSS-UI?node-id=427-21944&t=W1KegPUlWtAo4zzt-0)) and also on the workspace card ([here](https://www.figma.com/design/Ncl8TJsFf2WKf3frrqTbtq/AzureTRE-OSS-UI?node-id=425-21916&t=W1KegPUlWtAo4zzt-0).  Maybe not something you want to look at in this PR but could be a good way to display this

ui/app/src/components/workspaces/WorkspaceUsers.tsx
```diff
@@ -0,0 +1,140 @@
...
+import { ExceptionLayout } from '../shared/ExceptionLayout';
+import { User } from '../../models/user';
+import { Stack } from '@fluentui/react';
+
+interface IUser {
+  id: string;
+  name: string;
+  email: string;
+  role: string;
+  roles: string[];
```
jjgriff93, line 21: `role` and `roles` seems confusing?

jjgriff93 (approved):

ui/app/src/components/workspaces/WorkspaceUsers.tsx
marrobi: So for the grouped layout the user needs to appear once for each role in roles. Maybe I could remove roles at this point, but was thinking of doing something like this:

image

tim-allen-ck (approved):
LGTM

ui/app/src/components/workspaces/WorkspaceUsers.tsx
tim-allen-ck: can i resolve this?

PR #4087, opened 2024-09-05
Update unrestricted & airlock review workspace base version

jonnyry:
This is a follow up to #3970 PR which amended the base workspace for Azure SQL support.

Workspaces derived from base (i.e. unrestricted and airlock-import-review) both pull in the base workspace code from GitHub again using a git ref.

This PR amends the base version to 0.19.0 that the following workspace templates are built off:
- unrestricted
- airlock import review

This allows these workspaces to use the Azure SQL component.


tim-allen-ck (approved):
LGTM

jonnyry:
> @jonnyry can you update the branch? then we can merge it in

Done

PR #4098, opened 2024-09-26
Update Workspace limit in docs

### LINKED ISSUE #3920: TRE cannot scale beyond about 32 projects
## Description

In my Azure TRE deployment I am trying to test the limits of scalability, since we want to eventually run with up to 150 projects at a time. Yesterday, I created a large number of projects, and at about number 32, they started failing with the message `Subscription ******* already contains 250 storage accounts with Standard Dns endpoints in location uksouth and the maximum allowed is 250.`

From what I can understand, [if the storage endpoints were to use `AzureDnsZone` endpoints instead of `Standard`,](https://techcommunity.microsoft.com/t5/azure-storage-blog/public-preview-create-additional-5000-azure-storage-accounts/ba-p/3465466) that will raise the limit to 5000 endpoints, which should be enough for us?

My question is, is it sufficient to update the `storage.tf` in various places to add `dns_endpoint_type = AzureDnsZone`, or is there some reason that won't work?

## Steps
The steps I have tried are:
1. create a workspace
2. go to 1, until failure
3. look at the error message

## Code
n/a

tim-allen-ck:
# Resolves #3920 

## What is being addressed

Update Docs to reflect workspace limit and a workaround to extend.


marrobi (approved):
LGTM

Danny-Cooke-CK (approved):
approved

PR #4111, opened 2024-10-23
Expose APP_SERVICE_SKU build variable to allow enablement of App Gateway WAF

jonnyry:
## What is being addressed

Exposes a new `APP_GATEWAY_SKU` variable in config.yaml / CICD to allow the App Gateway SKU to be configured allowing the App Gateway WAF to be enabled.  (The Azure TRE currently deploys the Application Gateway with a Standard_v2 SKU which does not enable the WAF.)

image

When the `APP_GATEWAY_SKU` variable:
  - is not present, or is present and it's value is null/blank, or is equal to `Standard_v2` then a `Standard_v2` SKU App Gateway is provisioned (the same as before the change)
  - is equal to `WAF_v2` then a `WAF_v2` SKU App Gateway is provisioned along with a WAF Policy with default settings -  'Prevention' mode and the 'OWASP 3.2' ruleset

Notes about the change:

- I've added a `lifecycle { ignore_changes = [ policy_settings, managed_rules] }` statement so that the policy and rules are not managed in terraform after being set up to allow config of the rulesets to happen independently of this build
- There was an issue causing terraform to continually show changes in the `azurerm_application_gateway` around the `probe.match` field.  I've added this field explicitly as it was being readded on every run despite not being defined.
- The `azurerm_web_application_firewall_policy` resource requires a destruction provisioner as terraform does not properly unhook the WAF policy from the App Gateway before trying to destroy the WAF Policy.  This is needed if changing config from WAF_v2 > Standard_v2.


marrobi (approved):
Great work, haven't had chance to test, but providing it works as designed I'm good with it. 👍

jonnyry:
> Great work, haven't had chance to test, but providing it works as designed I'm good with it. 👍

Thanks.

I've run the following deployments manually to test the various permutations:

1. Local deployments (using make all/deploy-core)
  a. APP_GATEWAY_SKU variable not present
  b. APP_GATEWAY_SKU variable = Standard_v2
  c. APP_GATEWAY_SKU variable = WAF_v2
  d. APP_GATEWAY_SKU variable = Standard_v2 (to test destruction of WAF policy)

2. GitHub Actions deployments
  a. APP_GATEWAY_SKU variable not present
  b. APP_GATEWAY_SKU variable = Standard_v2
  c. APP_GATEWAY_SKU variable = WAF_v2
  d. APP_GATEWAY_SKU variable = Standard_v2 (to test destruction of WAF policy)

tim-allen-ck (approved):
LGTM

jonnyry:
Fixed lint issue (updated core version)

PR #4115, opened 2024-10-28
Change Key Vault to use RBAC instead of Access Policies

### LINKED ISSUE #4000: Key Vaults should use RBAC instead of access policies for access control
## Description

As a TRE Administrator
I want to use the recommended configurations for Azure resources
So that feature deprecations, etc. won't affect a production environment

## Acceptance criteria

- [ ] Key Vaults deployed by TRE use Azure RBAC for access control
- [ ] A migration path is documented for existing TRE instances

jonnyry:
# Resolves #4000

## What is being addressed

Change Key Vaults used in the TRE to use RBAC instead of Access Policies.

## How is this addressed

- Changes the Key Vault in the TRE code, and Workspace Key Vaults to use RBAC for authorisation instead of Access Policies
- Replaces Access Policies with RBAC Role Assignments:
  - 'Key Vault Administrator' role is used where full read/update/delete is required to key vaults - such as the deployment principal, and the resource processor
  - 'Key Vault Secrets User' is used where only read is required
- The PR does not to attempt to migrate manually create Access Policies outside of the terraform code, for example locally created access given to specific users.
- The following workspaces/shared services/workspace service templates have been updated:
  - Certificate shared service
  - Gitea shared service
  - Package Mirror (Sonatype Nexus) shared service
  - Gitea workspace service
  - Guacamole workspace service
  - ML Flow workspace service
  - OHDSI workspace service
  - Base workspace
  - Unrestricted workspace (TODO - see below)
  - Airlock import review workspace (TODO - see below)

## Follow up change required to unrestricted & airlock-import-review workspaces

Because of the mechanism the unrestricted & airlock-import-review workspaces pull code from the previous version of the Azure TRE, they will not receive this change until a new version is tagged, and their code is updated to pull the tag.

Therefore these two workspace templates will fail with Key Vault permissions until this work is completed.


jonnyry:
Fixed lint issues (terraform formatting)

tim-allen-ck (approved):
LGTM

marrobi:
@jonnyry This looks good, thank you. Has this been tested on an upgrade? Might be we should do that, can't see why it would cause any issues, but may do.

jonnyry:
> @jonnyry This looks good, thank you. Has this been tested on an upgrade? Might be we should do that, can't see why it would cause any issues, but may do.

@marrobi I haven't actually, only deployed to a fresh instance.  Let me attempt an upgrade and report back.

marrobi:
@jonnyry did you get chance to try the upgrade? If not I can give it a go today.

jonnyry:
@marrobi yes please if you get chance.  I didn't get round to it yesterday and not supposed to be working today :-D

marrobi (approved):
Upgrade works, tested workspace and VM deployment, code looks good. Thanks, great PR.

marrobi:
/test-force-approve 

passed here - https://github.com/microsoft/AzureTRE/actions/runs/11674140504

jonnyry:
Thanks @marrobi 

In order for unrestricted & airlock-import-review workspaces to pick up the change, will require a new version tagging and an update to those workspace templates.  Happy to put in the PR to update this, though I'm not able to create the tag myself.

PR #4104, opened 2024-10-04
Switch use of storage account keys by Terraform to use Azure AD authentication

marrobi:
Fixes issue #4103.

This pull request updates the Terraform code to switch from using storage account keys to using Azure AD authentication for operations on storage accounts. This change is necessary due to increasing policy restrictions on the use of storage account keys. The code changes include setting the `storage_use_azuread` flag to `true` in the provider block for the AzureRM provider. Additionally, the pull request includes updates to various modules and resources to enable Azure AD authentication. 

I've also removed old terraform migrations that delayed the testing, and a legacy cnab-state directory.

I've also included a helper script to increment all porter bundle versions in scenarios such as this.

Finally I fixed a bug with the health services bundle with the TF lock file that prevented it building. Again was needed for testing.

I've tested this myself, but would appreciate others giving it ago on a subscription with a policy banning the use of storage keys.

This does not fix issues of VM's using account keys to use storage and will have to be handled separately.

PR #4128, opened 2024-11-07
make the ARM_CLIENT_ID non mandatory in the bootstrap script

yuvalyaron:
# Tiny fix for https://github.com/microsoft/AzureTRE/pull/4104

## What is being addressed
Added a small fix to `bootstrap.sh` so that it doesn't require setting the non-mandatory `ARM_CLIENT_ID` environment variable in the config


marrobi (approved):
@yuvalyaron thanks, not sure how that slipped through.

marrobi:
/test-force-approve 

Changes only affect bootstrap and that passes successfully.

PR #4133, opened 2024-11-12
Update local debugging for RBAC key vaults

### LINKED ISSUE #4000: Key Vaults should use RBAC instead of access policies for access control

tim-allen-ck:
# Resolves #4000

## What is being addressed

Missed key vault policy in setup_local_debugging.sh

## How is this addressed

- keyvault_resource_id output added in core
- script changed to add user as `Key Vault Secrets User`


Danny-Cooke-CK (approved):

marrobi (approved):
LGTM, you could get the keyvault resource ID, from the `keyvault_name` that already exists in the outputs via the CLI? 

`az keyvault show -n kv-mr0110-ws-d8ab --query id`

tim-allen-ck:
> LGTM, you could get the keyvault resource ID, from the `keyvault_name` that already exists in the outputs via the CLI?
> 
> `az keyvault show -n kv-mr0110-ws-d8ab --query id`

yeah I thought about that, figured it was an unnecessary az command when we can get it for the tf

PR #4139, opened 2024-11-15
Enable storage account infrastructure encryption

### LINKED ISSUE #4001: Storage accounts should use infrastructure encryption
## Description

As a TRE Administrator
I want to deploy TRE in a manner compliant with common regulatory frameworks, like NIST SP 800-171 R2 and Microsoft's built-in compliance initiatives for those frameworks
So that research takes place in a compliant environment

## Acceptance criteria

- [ ] New TRE deployments' storage accounts use infrastructure encryption
- [ ] Existing TRE deployments are not modified

## Notes

Existing storage accounts cannot be updated to support infrastructure encryption. A feature flag at the core TRE level might be required so that existing storage accounts aren't attempted to be upgraded. Perhaps this could also be accomplished with Terraform ignore statements, but I don't know those very well.

jonnyry:
# Resolves #4001

## What is being addressed

Enable storage account infrastructure encryption.

Since storage account infrastructure encryption can only be turned on when a storage account is first created (unless you allow terraform to destroy the storage account and recreate it), this change is **only** applied on first deployment of a TRE (or a template component).  Storage accounts that exist within an existing TRE deployment will not be affected.

The terraform pattern used to acheive this behaviour is as follows:

```
resource "azurerm_storage_account" "stg" {

  ... existing attributes ...

  # changing this value is destructive, hence attribute is in lifecycle.ignore_changes block below
  infrastructure_encryption_enabled = true

  lifecycle { ignore_changes = [infrastructure_encryption_enabled ] }
}
```


tim-allen-ck (approved):
LGTM

tamirkamara (changes requested):
We have a CMK contribution coming in too. Are the two compatible with each other?
@yuvalyaron

yuvalyaron:
@tamirkamara yes, they are compatible

jonnyry:
Hey @tamirkamara do you have any more outstanding thoughts or feedback with your review?  Thanks

tamirkamara (approved):

jonnyry:
@tim-allen-ck this ok to merge now? thanks

Issue #4002, opened 2024-06-23
Azure resources should support customer-managed key for encryption at rest


SvenAelterman:
## Description

As a TRE Administrator
I want to deploy TRE in a manner compliant with common regulatory frameworks, like NIST SP 800-171 R2 and Microsoft's built-in compliance initiatives for those frameworks
So that research takes place in a compliant environment

## Acceptance criteria

CMK support for
- [ ] Storage accounts
- [ ] Resource processor VMSS
- [ ] ?

fortunkam:
We have a couple of options where to host the keys.
1.  Use the KeyVault deployed as part of the Instance?
2.  Add a new KeyVault (plus managed identity for rotation) to the instance resource group dedicated for keys
3.  Add a new KeyVault (plus managed identity for rotation) to the management resource group dedicated for keys

I tend to lean towards option 3 because it opens up options for customers that are using Managed HSM?

[The is the Landing Zone Guidance](https://learn.microsoft.com/en-us/azure/cloud-adoption-framework/ready/landing-zone/design-area/encryption-and-keys), it implies a KeyVault per application.

marrobi:
I think we need to allow for managed HSM. Could the KeyVault ID be configurable, defaulting to one in the management RG?

fortunkam:
Agreed, this would also lean us towards option 3.  Only downside is that I am not sure terraform for Managed HSM keys and Key Vault Keys are the same, need to investigate.

SvenAelterman:
I have not had a requirement to support HSM, so perhaps that might not need to considered until someone needs it?

There will need to be keys in the core for core VMs and storage, etc. For the workspaces, the keys should be in the workspace.

I am not sure if there's a need for a new Key Vault. CMK isn't a new application/workload. The workload, IMO, is the workspace.

marrobi:
We have a customer that needs this with managed HSM. :)

marrobi:
@yuvalyaron this is currently assigned to you, are you intending to complete the remaining tasks? Thanks.

yuvalyaron:
@marrobi
I've removed myself from the HSM task (https://github.com/microsoft/AzureTRE/issues/4146) and the TRE deployment blocking task (https://github.com/microsoft/AzureTRE/issues/4172). The HSM work is more complex than it initially appeared (Terraform's handling differs from what the Azure docs suggest)

PR #4154, opened 2024-11-26
Add CMK support for Cosmos DB (core)

### LINKED ISSUE #4144: CMK support for Cosmos DB (even though this is part of core, this case is listed separately as it requires special handling since the implementation is not straightforward)
Adding CMK support for Cosmos DB requires special handling, while CMK support for Cosmos is part of the core implementation and could have been implemented as part of  #4142, it requires special handling due to its complexity.

1. The Terraform implementation does not allow adding CMK to an existing Cosmos DB account without deleting the account. However, Azure documentation states that CMK can be added without requiring deletion.

2. AzureRM Cosmos DB currently only supports using FirstPartyIdentity when the account is created with CMK enabled. This means the account must initially be created with FirstPartyIdentity, and the identity type can only be changed after creation.
As seen in this example: https://github.com/Azure/terraform-azurerm-cosmosdb/blob/v1.0.0/examples/203-cosmosdb-customer-managed-key/main.tf

yuvalyaron:
# Resolves #4144 
Part of https://github.com/microsoft/AzureTRE/issues/4002

## What is being addressed

Added CMK support for Cosmos DB (in core).

## How is this addressed

- Using the Azure CLI command to add CMK support for Cosmos. Using the CLI is necessary because Terraform forces a resource recreation if the key_vault_key_id is set using Terraform (see https://github.com/hashicorp/terraform-provider-azurerm/issues/24781).


tim-allen-ck (approved):
LGTM

LizaShak (approved):
LGTM as well :)

PR #4161, opened 2024-12-01
Add CMK support for Base Workspace

### LINKED ISSUE #4145: CMK support for resources outside of core
As a complementary issue to #4142, we want to add CMK support for the remaining TRE resources, such as workspaces, workspace services, shared services, etc.

- make sure to update the CMK section in the docs accordingly.

yuvalyaron:
# Partially Resolves #4145 

## What is being addressed

Added CMK support for the base workspace

## How is this addressed

- Propagated the `ENABLE_CMK_ENCRYPTION` and `KEY_STORE_ID` environment variables to the resource processor, ensuring the workspace is automatically deployed with CMK encryption if set at the TRE level.

- Added an identity for the workspace, allowing it to access the Key Vault for creating the CMK.


templates/workspaces/base/porter.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 schemaVersion: 1.0.0
 name: tre-workspace-base
-version: 1.6.1
+version: 1.6.2
```
yuvalyaron, line 4: (it's not a patch) do we consider this as a major update?

templates/workspaces/base/terraform/airlock/storage_accounts.tf
```diff
@@ -296,3 +336,44 @@ resource "azurerm_role_assignment" "api_sa_data_contributor" {
   role_definition_name = "Storage Blob Data Contributor"
   principal_id         = data.azurerm_user_assigned_identity.api_id.principal_id
 }
+
+resource "azurerm_storage_account_customer_managed_key" "sa_import_approved_encryption" {
```
tamirkamara, line 340: Have you considered using a for_each here?

yuvalyaron: thanks, fixed

templates/workspaces/base/porter.yaml
```diff
@@ -116,6 +116,12 @@ parameters:
     default: true
   - name: arm_environment
     type: string
+  - name: enable_cmk_encryption
+    type: boolean
+    default: false
+  - name: key_store_id
```
LizaShak, line 122: I see the default is an empty string. If enable_cmk_encryption is true and the key is empty, will it work?

yuvalyaron: No, it won’t work.
The default is an empty string only to make the key optional when `enable_cmk_encryption` is false.

LizaShak (approved):
LGTM

tim-allen-ck (approved):
LGTM

yuvalyaron:
successful test run:
https://github.com/microsoft/AzureTRE/actions/runs/12243275845
https://github.com/microsoft/AzureTRE/pull/4173

PR #4170, opened 2024-12-09
Fix Dockerfile 'AS' casting

tamirkamara:
## What is being addressed

docker CLI runs some linting while building images. It alerts on some of the 'as' keyword casting.

## How is this addressed

- Use upper-case for 'FROM' and 'AS'


marrobi (approved):
LGTM

PR #4175, opened 2024-12-09
Block CosmosDB key authentication

### LINKED ISSUE #4174: Block key authentication in CosmosDB
As a system administrator I want cosmos db to disallow authentication with keys to increase security.

tamirkamara:
# Resolves #4174 

## What is being addressed

Using fixed keys for authentication in CosmosDB is a security risk. Although this solution doesn't currently use this method, the db still allows it.

## How is this addressed

- Block key authentication on the account level


marrobi (approved):

tim-allen-ck (approved):
LGTM

Issue #4142, opened 2024-11-19
CMK support for resources that support it in core (storage accounts, storage disks)


yuvalyaron:
Add CMK support for Storage accounts and VM storage disks for core TRE resources.
CMK support for the rest of the resources will be handled in a separate issue (#4145)

PR #4178, opened 2024-12-10
Add CMK encryption support for Azure Container Registry (ACR) and Service Bus

yuvalyaron:
# Part of #4142

## What is being addressed

- Added CMK for ACR and SB
- Renamed the encryption key to include the TRE ID


tamirkamara (approved):

LizaShak (approved):
LGTM

PR #4179, opened 2024-12-10
Extend validation environment TTL

### LINKED ISSUE #4176: Does `test-extended` always fail?
**Describe the bug**

Did anyone see test-extended actually succeed recently?

The process that stops a CI environment after 2 hours of in activity (either commit or comment) is not enough to cover the time it takes to run all tests not to mention if a deploy is also included. I think the first attempt is highly likely to fail (always) which is on its own a waste of time, it might be that following attempts are also likely to fail.

What do you think?

tamirkamara:
# Resolves #4176 

## What is being addressed

Validation environments are stopped before tests are completed.

## How is this addressed

- Stop time extended to 4 hours and workflow is scheduled to run every 1 hour

### DIFF
--- a/.github/workflows/clean_validation_envs.yml
+++ b/.github/workflows/clean_validation_envs.yml
@@ -4,7 +4,7 @@ name: Clean Validation Environments
 on:  # yamllint disable-line rule:truthy
   schedule:
     # Every 2 hours
-    - cron: "0 */2 * * *"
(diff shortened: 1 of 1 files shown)


tim-allen-ck (approved):
LGTM

marrobi (approved):

PR #4190, opened 2024-12-11
Change the way "inherited" workspaces retrieve the base workspace code

### LINKED ISSUE #4162: Change the way "inherited" workspaces retrieve the base workspace code
**Problem**

"Inherited" workspaces - airlock-import-review, unrestricted workspaces and our own custom workspaces - are built on top of the base workspace's code, to reduce duplication.

The base workspace code is pulled in in the inherited workspace's Dockerfile.tmpl file:

https://github.com/microsoft/AzureTRE/blob/b69ab016b6cae06818538a5d942ea42c3d6c6dde/templates/workspaces/airlock-import-review/Dockerfile.tmpl#L16-L19

The code is downloaded from a tagged version of the Azure TRE on GitHub, rather than by referencing the local copy of the base workspace, as the Docker build context does not allow access outside of the current working directory (the docker "build context").

This is unsatisfactory as every time the base workspace changes, for the change to be propagated to inherited workspaces, a versioned release of the Azure TRE is required, and then a subsequent commit is required to update this version number in inherited workspaces.  Consequently inherited workspaces typically lag behind the base workspace by several versions as this step is often missed.

**Proposed solution**

Rather than pulling the base workspace code from a tagged release on GitHub, use the local base workspace code in the `../base` workspace directory.

`docker build` now allows a `--build-context` parameter to be passed to containing additional named build contexts, which allows the referencing of files outside of the current build context.

https://github.com/docker/buildx/blob/master/docs/reference/buildx_build.md#-additional-build-contexts---build-context

This would then allow the COPY command to be used as follows:

```
COPY --from=my-build-context base/terraform/storage.tf .
```

In order for the above to be used a change to porter would be required to allow the passing of the `--build-context` parameter to the `docker build` command.

jonnyry:
# Resolves #4162

## What is being addressed

Change the way inherited workspaces retrieve the base workspace code - see solution description in #4162

## Notes

~~1. Currently requires the 'canary' release of [porter](https://www.github.com/getporter/porter).  Ideally this should be a pinned version, however we are waiting for a new version release which includes the --build-context parameter.~~
(Now resolved - Porter have release version v1.2.1 incorporating the change.)


tim-allen-ck:
so do you need to have built `base` first? Will it use the latest version of base?

jonnyry:
> so do you need to have built `base` first? Will it just use the latest version of `base`?

you don't need to have `base` built first, it pulls the latest files from the `../base` directory, by using the 'base' --build-context passed to porter build.

tim-allen-ck (approved):

tim-allen-ck:
@jonnyry https://github.com/microsoft/AzureTRE/actions/runs/12306981977/job/34350005999#step:3:1

jonnyry:
> @jonnyry https://github.com/microsoft/AzureTRE/actions/runs/12306981977/job/34350005999#step:3:1

Ah let me take a look.

jonnyry:
Made a change so that the `porter build --build-context` parameter is only used when required, i.e. for the following workspaces:
- airlock-import-review
- unrestricted

In all other cases, the `--build-context` parameter is not used.

The `bundle-build` make target does this by looking for a file named `porter-build-context.env` in the workspace folder, if it exists, it grabs the additional build context from the file and passes it to `porter build`, else it runs `porter build` without the parameter.

marrobi:
@jonnyry think we need to wait for a new Porter release for this? Not sure running on canary is wise given it could "update" at any time.

jonnyry:
> @jonnyry think we need to wait for a new Porter release for this? Not sure running on canary is wise given it could "update" at any time.

Yes indeed.  I have requested a patch release, though I don't think its quite as quick and simple as a release of the mixins - they are discussing and will get back to me.

jonnyry:
Porter have release version v1.2.1 incorporating the change. https://github.com/getporter/porter/releases/tag/v1.2.1

PR #4193, opened 2024-12-12
Fix Unix soft links support in shared filesystem

### LINKED ISSUE #4180: Shared filesystem doesn't support Unix soft links
The shared filesystem doesn't support soft links in a Unix VM. This greatly restricts the usefulness of the storage.

For example, we've been advising people to install software on the shared storage, so they can avoid customising the actual VMs, so they can destroy and create VMs as needed on a daily basis. A lot of our researchers use `conda`, and installing a basic conda environment will almost inevitably pull in `OpenSSL`, which creates a soft-link between certificate files.

**Steps to reproduce**
1. Create a workspace with shared storage.
2. Create a Linux VM in that workspace, log in to it.
3. Create a conda environment:
`conda create -p /path/to/shared-storage/my/env openssl`

This fails with `[Errno 95] Operation not supported: 'cacert.pem' -> '/shared-storage/conda/predict/ssl/cert.pem`.

Or, simpler:
```
cd /path/to/shared-storage
touch a
ln -s a b
ln: failed to create symbolic link 'b': Permission denied
```

This is only one example, there are lots of tools out there that make use of soft links.

I don't know if there's a filesystem that supports soft links and is still mountable on Windows, one that could be easily implemented in Azure? If not, it would be worth considering supporting a Linux-only shared filesystem in addition to the workspace-wide shared filesystem. This would allow maximum flexibility.

The only workaround we have is to tell people to write installation scripts for the tools they want to install, and to put them on the shared storage. Then, at least, when a new machine is booted, they can run the scripts and reproduce their environment. That's not ideal.

marrobi:
Fixes #4180

Enable Unix soft links support in shared filesystem for Unix VMs.

* **Autofs Configuration**
  - Add `uid=1000,gid=1000,mfsymlinks` to the `echo` statement in `templates/workspace_services/guacamole/user_resources/guacamole-azure-linuxvm/terraform/vm_config.sh` to support Unix soft links and set proper ownership and permissions.

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/microsoft/AzureTRE/pull/4193?shareId=7bdc1975-bbaa-4233-892a-c38109bb0e83).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ BUG FIXES:
 - Fix Dockerfile 'as' casting ([#4170](https://github.com/microsoft/AzureTRE/pull/4170))
 - Create policy to allow all user to configure color profiles to remove auth dialog. ([#4184](https://github.com/microsoft/AzureTRE/pull/4184))
 - Pre configure VS code option to prevent script failure ([#4185](https://github.com/microsoft/AzureTRE/pull/4185))
+- Enable symlinks to work on Linux VM shared storage ([#4180](https://github.com/microsoft/AzureTRE/issues/4180))
 
 COMPONENTS:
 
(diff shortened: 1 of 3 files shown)


marrobi:
/test-force-approve

Not tested.

Issue #4172, opened 2024-12-09
Block redeployment of TRE with CMK enabled if it has previously been deployed without it.


yuvalyaron:
We need to block redeployment of TRE with CMK enabled if it has previously been deployed without it, otherwise TRE will be stuck in an "in-between state" where some resources have CMK and some don't

PR #4195, opened 2024-12-13
CMK Encryption support for resources in templates

### LINKED ISSUE #4145: CMK support for resources outside of core

yuvalyaron:
# Resolves #4145

## What is being addressed

Added support for CMK encryption for resources in templates.

## How is this addressed

- Workspace services use the CMK and encryption identity of the workspace to enable encryption for supporting resources.
- Shared services use the CMK and encryption identity of the core TRE to enable encryption for supporting resources.


marrobi:
@yuvalyaron do we know what the upgrade behaviour is here? Might be worth incrementing bundles by a major version to prevent easy upgrade?

LizaShak (approved):
LGTM

LizaShak (commented):
Let me know when ready :)

yuvalyaron:
> @yuvalyaron do we know what the upgrade behaviour is here? Might be worth incrementing bundles by a major version to prevent easy upgrade?

Upgrading is not an issue as long as the `enable_cmk_encryption` variable remains unchanged, and we assume it will never change (though the exact approach to how to ensure this is yet to be determined):  https://github.com/microsoft/AzureTRE/issues/4172

templates/workspace_services/guacamole/user_resources/guacamole-azure-linuxvm/terraform/main.tf
```diff
@@ -73,3 +73,15 @@ data "azurerm_public_ip" "app_gateway_ip" {
   name                = "pip-agw-${var.tre_id}"
   resource_group_name = data.azurerm_resource_group.core.name
 }
+
+data "azurerm_key_vault_key" "ws_encryption_key" {
```
LizaShak, line 77: Maybe worth adding it in a data.tf file like it is everywhere else?

yuvalyaron: Done - created a data.tf file and moved everything

tamirkamara (approved):

yuvalyaron:
The workspace tests fail because the Terraform code from the base workspace in the old release does not include the newly added variables. This issue should be resolved once a new release is created.

PR #4219, opened 2024-12-23
Disable local/basic accounts

### LINKED ISSUE #4218: Secure storage accounts
## Description

As a TRE Administrator
I want to have storage accounts and other resources disable usage of local user accounts and fixed keys
So that the system will be more secure

tamirkamara:
## What is being addressed

Some policies keep changing less secure settings like local user access / deploy. TRE doesn't (shouldn't) use these options but it creates noise on Terraform runs.

## How is this addressed

- Disable storage account local user account
- Disable basic auth on web apps


LizaShak (approved):
LGTM

yuvalyaron (approved):

PR #4227, opened 2024-12-29
Disable storage access keys

### LINKED ISSUE #4218: Secure storage accounts

tamirkamara:
# Resolves #4218 

## What is being addressed

Storage accounts have the access key option enabled despite us not needing it. Some policies keep changing it and it can create noise in Terraform deployments.

## How is this addressed

- Disable access key usage on almost all storage account
- Add a comment where it appears we have an issue


core/terraform/airlock/airlock_processor.tf
```diff
@@ -28,7 +28,10 @@ resource "azurerm_storage_account" "sa_airlock_processor_func_app" {
   queue_encryption_key_type        = var.enable_cmk_encryption ? "Account" : "Service"
   allow_nested_items_to_be_public  = false
   cross_tenant_replication_enabled = false
-  tags                             = var.tre_core_tags
+  local_user_enabled               = false
+  # Function Host Storage doesn't seem to be able to use a User Managed ID.
```
yuvalyaron, line 32: what does this comment refer to?

core/terraform/airlock/airlock_processor.tf
```diff
@@ -62,8 +65,10 @@ resource "azurerm_linux_function_app" "airlock_function_app" {
   ftp_publish_basic_authentication_enabled       = false
   webdeploy_publish_basic_authentication_enabled = false
   storage_account_name                           = azurerm_storage_account.sa_airlock_processor_func_app.name
-  # consider moving to a managed identity here
+
+  # Function Host Storage doesn't seem to be able to use a User Managed ID.
   storage_account_access_key = azurerm_storage_account.sa_airlock_processor_func_app.primary_access_key
+  # storage_uses_managed_identity = true
```
yuvalyaron, line 71: why do we keep the 
`# storage_uses_managed_identity = true`?

yuvalyaron (approved):

LizaShak (approved):

core/terraform/airlock/airlock_processor.tf
tamirkamara: The line below

tamirkamara:
@jonnyry we have an Azure Files share which will never support AAD so the key has to be enabled there.

jonnyry:
Ah OK no problem :-) wasn't sure if you could use SAS keys

PR #4230, opened 2024-12-29
Amend destroy_env_no_terraform.sh to purge container repos

jonnyry:
## Problem

When using Defender to scan container images -

If the container registry is deleted as a whole (instead of deleting individual container repositories first), Defender does not remove scans of container images.

If redeploying a fresh TRE using the same name as previously, this leads to misleading results in Defender in dev/test environments with repeated scans 'stacking up' and making the results look larger than they actually are (see screen shots below).

Example of Defender results that are 'stacking up':

image

Example of Defender holding on to scans of deleted images (there is only 1 image in each repository):

image

## Diagnosis

According to this [Microsoft documentation](https://learn.microsoft.com/en-us/azure/defender-for-cloud/agentless-vulnerability-assessment-azure#if-i-remove-an-image-from-my-registry-how-long-before-vulnerabilities-reports-on-that-image-would-be-removed), container scans should be deleted within an hour, or maximum three days, however our experience is that container scans for deleted images (where the registry was deleted as a whole) can persist for up to two weeks.

> ## If I remove an image from my registry, how long before vulnerabilities reports on that image would be removed?
>
> Azure Container Registries notifies Defender for Cloud when images are deleted, and removes the vulnerability assessment for deleted images within one hour. In some rare cases, Defender for Cloud might not be notified on the deletion, and deletion of associated vulnerabilities in such cases might take up to three days.

It appears the 'hook' Defender uses to remove scans does not work when the ACR is deleted as a whole with images remaining in it, and only functions as intended when repositories/images are deleted individually first.

## Solution

This PR amends the `destroy_env_no_terraform.sh` script to individually delete container repositories, before the script continues as usual to delete the resource groups (including the container registry).

This ensures the 'hook' Defender uses to remove scans fires and scans are cleared up when a TRE is destroyed.


tamirkamara:
@jonnyry shouldn't the script just find the ACRs in the groups it's going to delete?

jonnyry:
> @jonnyry shouldn't the script just find the ACRs in the groups it's going to delete?

Yes good point, probably should - I'll amend.

jonnyry:
@tamirkamara Now updated to search for containter registries rather than requiring a separate parameter.

tamirkamara (approved):

tim-p-allen (approved):
LGTM

PR #4236, opened 2024-12-31
Fix public exposure in Guacamole service

### LINKED ISSUE #4199: Guacamole "exposed_external" flag not working correctly.
**Describe the bug**
Guacamole workspace service is exposed externally even when the "exposed externally" is not ticked.

main

ShakutaiGit:
# Resolves #4199

## What is being addressed

The current behavior does not allow for dynamic configuration of the `public_network_access_enabled` parameter in the Guacamole workspace template. 
This parameter is critical for determining whether the service should be exposed externally or restricted to internal access only.

## How is this addressed
- Introduced the `public_network_access_enabled` parameter with a default value of `false` to prevent public exposure by default.
- Updated `web_app.tf` to map the parameter to `var.is_exposed_externally`.
- Moved Azure Service Plan data retrieval to `data.tf` for better organization.
- Updated documentation and incremented the template version.


templates/workspace_services/guacamole/porter.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 schemaVersion: 1.0.0
 name: tre-service-guacamole
-version: 0.12.5
+version: 0.12.6
```
tamirkamara, line 4: please update the changelog file

tamirkamara (approved):

yuvalyaron (approved):

PR #4238, opened 2024-12-31
Add option for forced tunneling through TRE's Firewall

### LINKED ISSUE #4237: Add Option to Force Tunnel TRE's Firewall
Add an option to TRE's firewall to force tunnel traffic to another firewall.

**Describe alternatives you've considered**
Manual setup after deploying TRE (might be overridden by terraform)

yuvalyaron:
# Resolves #4237

## What is being addressed

Added the option for force tunnel TRE's Firewall to an external firewall

## How is this addressed

- Added `firewall_force_tunnel_ip` parameter to `rp_bundle_values`, when set, the following are created:
  * A route table to direct TRE traffic to the specified IP.
  * A public IP for firewall management.

After that, users have to manually connect TRE's VNet to the external firewall (e.g. through VNet Peering).


docs/tre-admins/configure-firewall-force-tunneling.md
```diff
@@ -0,0 +1,20 @@
...
+
+Forced tunneling ensures that all traffic from TRE is routed through a specific external firewall. This guarantees that all data passes through the firewall for inspection, control, or further processing before reaching its destination.
+
+To setup forced tunneling to an external firewall, follow these steps:
+
+## 1. Set the rp_bundle_values Parameter in  the config.yaml file
+Provide the external firewall's IP address:
+
+```json
+rp_bundle_values: '{"firewall_force_tunnel_ip":"10.0.0.4"}'
```
marrobi, line 11: Is there a reason we are using `rp_bundle_values` rather than exposing it as a parameter and giving it its own variable in a similar way to `firewal_sku`?

yuvalyaron: I can make that change, but from what I understand, `rp_bundle_values` are intended for passing parameters to the resource processor, which are used only by the bundles and not the processor itself. Should I change?

marrobi: If configured as a parameter for the bundle, the IP can be passed as part of the deployment message.

Hmm, I can see `firewall_sku` was set on the RP as an env var. @tamirkamara think you worked on that, why was it done this way, rather than passed as a property in `make firewall-install`?

jonnyry: that was me :-)  It may not have been the best way to pass it!

marrobi: Ok, my view is that we should be passing both firewall sku and the tunnel IP as props, as part of `firewall-install` here:

https://github.com/microsoft/AzureTRE/blob/c79d9e160fc404b246b4583a83ff5f5c5ad9caa7/Makefile#L304-L313

Not sure how easy that is to do... The `rp_bundle_values` will work, but believe it was intended for setting something that should not be passed via the API.

jonnyry: For ref, here's the change to expose the FIREWALL_SKU param:  https://github.com/microsoft/AzureTRE/commit/63ee9358b80e8769e796d61a9826bc4f7fe0387c

tamirkamara: We need to keep in mind that the firewall is deployed as a bundle and updated frequently. What is passed in the first install (i.e. `make firewall-install`) should be available in any subsequent action on the bundle (like adding rules). 
IIRC, the only way to do this today is via the rp_bundle* method.
Considering the SKU parameter again, including it in the RP .env makes that file include settings not meant for the RP operation, but for its bundles. It brings about this very discussion and make me question - do we want to add any such potential bundle param this way? It creates a situation where a new bundle that want to have a system wide setting require its own version of the RP. IMO that is less appropriate.

marrobi: If passed as a property to the API it will be stored in the firewall resource in cosmos, so is available for update/patch operations? The same as with any other resource?

In hindsight I don't think the `firewall_sku` approach is optimal, so we are aligned there. Both these bundle specific properties should be passed in when creating the shared service via the API.

tamirkamara: Do we have a method today that will take a parameter like this into a make command and pass it in the API request both interactively (developer) and automatically via github actions?
The method in the PR is still a viable option IMO that supports all deployment paths. It also doesn't block any future development to support the above with zero changes to this bundle.

marrobi: If want to use `RP_BUNDLE_VALUES` in CI, I think that will still work? I don't believe `RP_BUNDLE_VALUES` works for local development as they are only expanded in the terraform. I think they were introduced to aid CI. 

I've tried this:

```
firewall-install:
	$(call target_title, "Deploying Firewall shared service") \
  && . ${MAKEFILE_DIR}/devops/scripts/check_dependencies.sh env \
	&& $(MAKE) bundle-build bundle-publish bundle-register deploy-shared-service \
	DIR=${MAKEFILE_DIR}/templates/shared_services/firewall/ BUNDLE_TYPE=shared_service \
	PROPS="--firewall_sku $${FIREWALL_SKU}"
```

Seems to work for me (has to add the property to `template_schema.json` too). Agree the PR will work, although would rather follow the same path for Firewall SKU and forced tunnelling IP and have them explicitly set in config.yaml and work for local development.

I've updated to enabled upgrades to shared service properties too otherwise the the resource processor has to be recreated for the property to change.

https://github.com/microsoft/AzureTRE/compare/main...marrobi:AzureTRE:marrobi/test-firewall-props?expand=1

This should work for local development too? also means the properties are visible in the UI:
image

Thoughts?

tamirkamara: RP_BUNDLE_VALUES work in both CI and local development. In fact, Yuval tested it via the setting currently present in the config.yaml. So I don't any issue apart from this being less explicit. Having said that, this is somewhat rare settings... Yes we could make it nicer in the future.

marrobi: How did you test it? I can't see them.

image

tamirkamara: I meant that if you set these values and run `make tre-deploy` from your local machine the RP will get the values and will be able to use them (same for the CI). 
As for debug I don't think we tried. Are you sure you followed all steps to make these available after updating your config file? At any case, it works the 2 ways it counts :-)

marrobi: They aren't output to `private.env` so it's not going to work unless that's updated somehow. I should have been searching for `os.environ["RP_BUNDLE_custom_key_1"]` , but it's still not there. Saying that, it's the same for Firewall SKU.

IMHO they should be properties on the bundle so they can be configured via the TRE API/UI, are visible via the UI, without destroying and recreating the resource processor.

If you want to get it in as is without the changes suggested, then do it and I can amend it next week with a further PR.

tamirkamara: But they are params in the bundle so maybe we can add them to the template schema to all setting via the API. But TBH, passing it through the scripts isn't great and I don't think this is the right time to get into it.

marrobi: You are the one who added the code to allow the passing props into the shared services script ;-), all I showed above is where they need to be specified.

Lets start by adding them to the template schema.

tamirkamara (approved):

docs/tre-admins/configure-firewall-force-tunneling.md
marrobi, line 1: Think we should add a line or two about why leave the existing firewall in place. We often have conversations about this.

@tamirkamara  thoughts? 

Something like "To ensure workspace level rules can be created when TRE workspaces are created without manual intervention we highly recommend leaving the Azure TRE firewall in place. If you still need to send all traffic through a centralised enterprise firewall then forced tunnelling should be used, ad this firewall needs a superset of rules used by the TRE."

Thoughts?

tamirkamara: Yes, sounds good. Just TRE components rather than workspaces.

yuvalyaron: done

marrobi (approved):
LGTM. Thank you!

PR #4242, opened 2025-01-01
Use CMK blocks rather than CMK resources in Terraform

### LINKED ISSUE #4241: Use CMK blocks rather than CMK resources in Terraform
The current implementation uses a separate resource to configure CMKs. This causes Terraform to first create the azure resource and only then attach the CMK, which violates Azure policies that block unencrypted storage account creation (if one is set).
The CMK configuration should instead be implemented as a dynamic block directly inside the resource.

yuvalyaron:
# Resolves #4241

## What is being addressed

Currently, CMKs are configured using separate Terraform resources, where the Azure resource is created first, and the CMK is added as a separate resource later.

This approach is problematic for tenants with policies requiring CMK to be enabled, as the resource is initially created without the CMK, resulting in a policy violation even if the CMK is added later.

This PR addresses the issue by using CMK blocks within the resources themselves, instead of defining them as separate resources.


tamirkamara (approved):

marrobi (approved):
LGTM as long as its been tested.

PR #4245, opened 2025-01-02
Fix deletion of Guacamole VM in stopped state

### LINKED ISSUE #4135: Unable to delete a Guacamole User Resource if it is in the 'stopped' state
**Deleting a VM from the Guacamole service fails if the VM is in a stopped state**

Attempting to delete a 'Stopped' VM, results in a 'deleting_failed' status and displays the following error: 

```
compute.VirtualMachinesClient#
PowerOff: Failure sending request: StatusCode=0 -- 
Original Error: autorest/azure: Service returned an error. 
Status=<nil> 
Code="OperationNotAllowed" 
Message="Operation 'powerOff' is not allowed on VM 'linuxvm4fae' since the VM is either deallocated or marked to be deallocated.
``` 
Seems that the deletion request fails because it is trying to stop the VM, when it is already stopped. If I 'Enable' the resource, 'Start' it, then 'Disable' it, the resource can be deleted successfully. 

**Steps to reproduce**

1. Login to the TRE UI (as TRE Administrator)
2. Connect to a Workspace
3. Connect to the Virtual Desktops (Guacamole) service
4. Select 'Actions > Stop' from either a linux or windows VM
5. Select 'Disable' for the same resource
6. Select 'Delete' for the same resource

Although I'm currently seeing this error from a Custom Workspace and a Custom Guacamole User-Resource, I did also encounter this problem when using the tre-base-workspace (1.5.7) and the tre-service-guacamole-linuxvm (1.0.3)

**Azure TRE release version (e.g. v0.19.1):**

**Deployed Azure TRE components: tre-service-guacamole (0.10.9), tre-service-guacamole-linuxvm (1.0.3)**

marrobi:
Draft - what think about this approach with a script to ensure child resources are deleted? Need to do other VMs.

Add a configuration option to Guacamole VM templates to skip shutdown and force deletion, resolving the issue where VMs in a stopped state cannot be deleted. Update relevant Terraform and Porter files to implement this change and increment version numbers accordingly.

Fixes #4135

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ ENHANCEMENTS:
 * Enhance DPI of Linux display ([[#4200](https://github.com/microsoft/AzureTRE/issues/4200)])
 * Update Admin VM versions ([[#4217](https://github.com/microsoft/AzureTRE/issues/4217)])
 * Update devcontainer/RP/API package versions: base image, docker, az cli, YQ ([#4225](https://github.com/microsoft/AzureTRE/pull/4225))
+* Purge container repos individually in when using `make tre-destroy` ([#4230](https://github.com/microsoft/AzureTRE/pull/4230))
 * Upgrade Python version from 3.8 to 3.12 ([#3949](https://github.com/microsoft/AzureTRE/issues/3949))Upgrade Python version from 3.8 to 3.12 (#3949)
 * Disable storage account key usage ([[#4227](https://github.com/microsoft/AzureTRE/issues/4227)])
 * Update Guacamole dependencies ([[#4232](https://github.com/microsoft/AzureTRE/issues/4232)])
--- a/devops/scripts/destroy_env_no_terraform.sh
+++ b/devops/scripts/destroy_env_no_terraform.sh
@@ -153,11 +153,35 @@ if [ "${workspace}" != "0" ]; then
   | xargs -P 10 -I {} az rest --method delete --uri "{}?api-version=2020-08-01"
 fi
 
+# delete container repositories individually otherwise defender doesn't purge image scans
+function purge_container_repositories() {
+  local rg=$1
+
+  local acrs
+  acrs=$(az acr list --resource-group "$rg" --query [].name --output tsv)
+
+  local acr
+  for acr in $acrs; do
+    echo "Found container registry ${acr}, deleting repositories..."
+
+    local repositories
+    repositories=$(az acr repository list --name "$acr" --output tsv)
+
+    local repository
+    for repository in $repositories; do
+        echo "  Deleting: $repository"
+        az acr repository delete --name "$acr" --repository "$repository" --yes --output none
+    done
+  done
+}
+
 # this will find the mgmt, core resource groups as well as any workspace ones
 # we are reverse-sorting to first delete the workspace groups (might not be
 # good enough because we use no-wait sometimes)
 az group list --query "[?starts_with(name, '${core_tre_rg}')].[name]" -o tsv | sort -r |
 while read -r rg_item; do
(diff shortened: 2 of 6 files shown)


marrobi:
@tamirkamara @jonnyry attempt number 2, thoughts on this approach?

dram1964:
Hi @marrobi, I did test the original fix, but didn't realise that this left the os_disk deployed. Happy to test out the new code, but I was wondering whether this is the best approach. 
Given that the `virtual_machine {skip_shutdown_and_force_delete = true }` feature is designed to preserve sub-resources of the VM, the additional script is a 'fix for the fix'. 
Would it be better to add an `az vm start` operation to the 'delete' step, since the 'delete' operation appears to expect the vm to be started?

marrobi:
> Hi @marrobi, I did test the original fix, but didn't realise that this left the os_disk deployed. Happy to test out the new code, but I was wondering whether this is the best approach. Given that the `virtual_machine {skip_shutdown_and_force_delete = true }` feature is designed to preserve sub-resources of the VM, the additional script is a 'fix for the fix'. Would it be better to add an `az vm start` operation to the 'delete' step, since the 'delete' operation appears to expect the vm to be started?

Agree, that's another option, just worry about turning a VM on that has been intentionally stopped. Imagine it has a start-up script that does stuff, or you turn it off as it has some malware. I'd prefer to just get rid of it.

tamirkamara:
Wouldn't it be easier to use azure cli to delete the VM if it's not on and then let TF do its thing without any other modification?
We already have the VM resource ID available so just need to query if it exists in the right state and delete.

marrobi:
Ah, hadn't seen that - making a note to follow to https://github.com/hashicorp/terraform-provider-azurerm/pull/28412

I used this approach as thought it was generic and could be reused for other services where we get lingering resources. Providing they are tagged they will get deleted.

If it's going to be reverted, don't mind, as long as it's fixed :)

marrobi:
> > @dram1964 fancy testing this one?
> 
> Hi @marrobi - sorry I'm having trouble testing this. I've copied your changes into my repo, but can't build the linuxdsvm image (`make user_resource_bundle BUNDLE=guacamole-azure-linuxvm WORKSPACE_SERVICE=guacamole`).
> 
> The problem seems to be caused by this line in the Dockerfile.tmpl:
> 
> ```shell
> COPY --from=scripts --link . ${BUNDLE_DIR}/scripts/
> ```
> 
> And the error is:
> 
> ```shell
> failed to resolve source metadata for docker.io/library/scripts:latest
> ```
> 
> I have added the `porter-build-context.env` and created the `templates\scripts\delete_azure_resources.sh` - but I don't think the `scripts` env var is actually being set.
> 
> I guess this is probably because I'm on an earlier version of the TRE (v0.19.1), with older versions of linuxdsvm (1.0.3)?

@dram1964 ah, yes, could be the new porter support for build contexts has only recently been added so need to be on the latest version of porter. The latest is in the main branch devcontainer.

jonnyry:
@marrobi which resources are left lingering after the delete has applied? Is it just the `osdisk`?

If so `delete_os_disk_on_deletion` should do the trick according to [this comment](https://github.com/hashicorp/terraform-provider-azurerm/issues/28218#issuecomment-2580817235):

```
features {
        virtual_machine {
          skip_shutdown_and_force_delete = true
          delete_os_disk_on_deletion = true
        }
      }
```

marrobi:
Hmm, might be the nic too. Think you the one that hit the issue with the initial attempt in the E2E tests?

jonnyry:
> Hmm, might be the nic too. Think you the one that hit the issue with the initial attempt in the E2E tests?

Just had a look through the runs but unfortunately the logs have been purged already!

PR #4251, opened 2025-01-06

### LINKED ISSUE #4250: Core key vault firewall should not be set to "Allow public access from all networks"

jonnyry:
# Resolves #4250


tamirkamara:
@jonnyry How will a CI agent connect to the FW if this is set?

marrobi:
@jonnyry believe I remember you saying you a have a script that opens/closes the KV firewall when you do a deploy? Think that would need to be part of this and added to the workflows.

jonnyry:
@tamirkamara @marrobi Ah right, yes, I was too hasty with my PR...  the key vault secrets are read/written using the data plane.

I'll take another look.

jonnyry:
Closing this PR. Will open a new one once a solution in place for CI agent access.

PR #4252, opened 2025-01-06

### LINKED ISSUE #4250: Core key vault firewall should not be set to "Allow public access from all networks"

jonnyry:
Second PR to test build actions only... closing.


jonnyry:
Second PR to test build actions only... closing.

PR #4256, opened 2025-01-06
Add Service Bus SKU configuration to Terraform resources

marrobi:
Introduce a configurable SKU for the Azure Service Bus, allowing users to select between Standard and Premium options.

The Standard SKU does not support VNet integration hence is exposed publically, also message size is limited. however this will significantly reduce costs for development scenarios.

Have a feeling upgrade will fail due from prior version and needs testing.

TODO: Add  Service tags to firewall network rules.


marrobi:
Depends on https://github.com/microsoft/AzureTRE/pull/4342

PR #4260, opened 2025-01-07

### LINKED ISSUE #4250: Core key vault firewall should not be set to "Allow public access from all networks"

jonnyry:
# Resolves #4250

## What is being addressed

- Changes the core key vault firewall from *Allow public access from all networks* to *Allow public access from specific virtual networks and IP addresses*
- Adds an IP exception to the key vault firewall for the deployment machine's internet IP (or the `PUBLIC_DEPLOYMENT_IP_ADDRESS` variable if set) during deployment
- Removes the IP exception at the end of deployment (whether deployment succeeds or fails)

## How is this addressed

- A new script to add and remove the keyvault deployment IP exception:
  - `devops/scripts/kv_add_network_exception.sh`
- They are called from the following scenarios in order to provider access to KV:
  - `core/terraform/deploy.sh`
  - `core/terraform/scripts/letsencrypt.sh`
  - `devops/scripts/destroy_env_no_terraform.sh`
  - `core/terraform/destroy.sh`

- The script uses a bash trap so that it runs regardless of whether the preceeding code fails or not, to ensure the IP exception is removed

A bug in azurerm provider was encountered which required the use of a terraform provisioner:

1. A create provisioner on `azurerm_key_vault` was required to work around an [azurerm provider bug](https://github.com/hashicorp/terraform-provider-azurerm/issues/18970) which means if a key vault is being re-created (it was previously soft deleted), the network acls are not updated.  This can be removed when the bug is fixed, or a different workaround found.

---

Updates since inital commit (as discussed with @marrobi):

1. Remove use of tags and null provisioner to add tag.
2. Delete the following scripts as they're no longer used:
  - `devops/scripts/key_vault_list.sh`
  - `devops/scripts/set_contributor_sp_secrets.sh`
3. Refactor the add and remove scripts into a single script


core/terraform/keyvault.tf
```diff
@@ -1,14 +1,48 @@
...
+az keyvault network-rule add --name ${local.kv_name} --ip-address ${local.myip} --output none
+EOT
+  }
+}
+
+# provisioner required due to ignore_changes = [tags] in azurerm_key_vault.kv
+#
+resource "null_resource" "add_deployment_tag" {
+  triggers = {
+    always_run = timestamp()
```
marrobi, line 38: Why does this always need to run? Once it's added once, it shouldn't get removed?

core/terraform/scripts/letsencrypt.sh
marrobi, line 1: Can the Storage Account rules in this script be handles the same way?

core/terraform/keyvault.tf
jonnyry: The intention was so if the tag is removed in Azure, it will always be readded.

However as discussed, have removed the use of tags altogether, so the provisioner has been removed.

core/terraform/scripts/letsencrypt.sh
jonnyry: Yes certainly - planning to have a look at storage accounts after this.

CHANGELOG.md
tamirkamara, line 1: I don't have time to go over this the next few days and guess @marrobi is the same. Just wanted to point out we now have 2 vaults being used from the deployer point of view.

jonnyry: OK no problems.

Two vaults?  which is the second one?

jonnyry:
> 🤖 pr-bot 🤖
> 
> 🏃 Running tests: https://github.com/microsoft/AzureTRE/actions/runs/12671848713 (with refid `26f9d939`)
> 
> (in response to [this comment](https://github.com/microsoft/AzureTRE/pull/4260#issuecomment-2577708595) from @jonnyry)

Notes on test run starting with an empty environment:

KV exception added here:

https://github.com/microsoft/AzureTRE/actions/runs/12671848713/job/35314921879#step:3:432

```
Adding deployment network exception to key vault kv-***...
 Core resource group rg-*** not found
```

KV exception removed here:

https://github.com/microsoft/AzureTRE/actions/runs/12671848713/job/35314921879#step:3:8259

```
Removing deployment network exception to key vault kv-***...
 Deployment network exception removed
```

CHANGELOG.md
tamirkamara: When CMK is enabled another vault is created in the mgmt resource group

jonnyry:
> 🤖 pr-bot 🤖
> 
> 🏃 Running tests: https://github.com/microsoft/AzureTRE/actions/runs/12674163834 (with refid `26f9d939`)
> 
> (in response to [this comment](https://github.com/microsoft/AzureTRE/pull/4260#issuecomment-2578018990) from @jonnyry)

**Notes on test run starting with an existing TRE:**

KV exception added here:

https://github.com/microsoft/AzureTRE/actions/runs/12674163834/job/35322577601#step:3:456

```
 Adding deployment network exception to key vault kv-***...
 Keyvault kv-*** is now accessible
```

KV exception removed here:

```
 Removing deployment network exception to key vault kv-***...
 Deployment network exception removed
```

https://github.com/microsoft/AzureTRE/actions/runs/12674163834/job/35322577601#step:3:1181

CHANGELOG.md
tamirkamara, line 1: There's a pending ask to enable the deployer to access resources such as these keyvaults over private network only.
1. Will this make this approach obsolete?
2. If not, this means we will need to do all of this in a conditional way. Right?

jonnyry: 1. That would mean all TRE deployers would need to switch to private self hosted runners right?  If so, yes this would be obsolete.
2. If we want to support both deployment patterns - deployment from GitHub hosted runners + deployment from private self hosted runners with KV set to private networking - then yes we'd need to do it conditionally (in order to prevent the KV from being fully public).

I guess it depends on whether implementing keyvaults private networking only means switching off the ability to use Github hosted runners - is that the plan?

tamirkamara: The goal I'm referring to that resources will be accessed via private network only which means private runners.

The question for you, is weather this PR comes from a similar place but doesn't go as far yet and just limits which public IPs can access. Or, in a situation where private agents / network is done, will you still need this method of limiting public IPs

jonnyry: If the codebase is switching to deployment from private runners ONLY, then this won't be needed anymore.

tamirkamara: To clarify, in your deployment/usecase you would also like to use private runners?

jonnyry: Over the medium term we would like move to private runners.  Also means we can tighten some of the storage account firewall rules.

jonnyry: Until we've moved to private runners, adding a deployment IP exception to to the KV is preferable to public network access though?

tamirkamara: I'm just considering most orgs that are concerned with this might want to go all the way. This this solution might not be enough is those cases. 
Also considering the number of changes you had to do just for this resource... next up will be the ACR, storage account and yet another keyvault in the mgmt resource group. All in all it might not be worth the investment and complexities if the end goal is anyway to go private agents...

jonnyry: Agree private runners should be the objective, but when are they scheduled to be implemented?  Is this a reasonable stop gap until then?

marrobi: I think we need to ensure the solution can still be deployed without private runners. Most orgs start out testing out the solution etc, and don't have the infrastructure for private runners.

tamirkamara: My point is that we might not want to offer this middle way of opening the deployer IP in all currently defined public resources due to the complexities of supporting all the resources I mentioned above. 
It might be we offer public like today and secured via private runners. Testing out could be done via the current way and if you want to be secure then it would mean you need private agents.

jonnyry: Supporting all resources might be too much (e.g. the ACR) but the KV is one that stands out as worthy of tightening network access on.  However if the add & remove exception scripts were genercised, other resources could be supported with little change since the hooks to call them are in place.

marrobi: I think we should look to get this as tightens things up, then longer term think about accommodate private runners.

It comes up on Defender for Cloud, hence "this solution is not secure" is what people believe.

marrobi: Although are a lot of scripts to maintain...

Not sure there is another way?

jonnyry: Was thinking the add & remove scripts could be consolidated down to a single script, so there's only one script to call when you need KV access.

Also there's a couple of scripts I added the add/remove calls to that don't seem to be called from anywhere (they could possibly be removed?):

- [devops/scripts/key_vault_list.sh](https://github.com/microsoft/AzureTRE/blob/main/devops/scripts/key_vault_list.sh)
- [devops/scripts/set_contributor_sp_secrets.sh](https://github.com/microsoft/AzureTRE/blob/main/devops/scripts/set_contributor_sp_secrets.sh)

jonnyry: OK simplified version is checked in:

1. Single `add_kv_network_exception.sh` script which does both the add & remove on exit
2. Deleted old scripts key_vault_list.sh & set_contributor_sp_secrets.sh

Needs a good re-test.

@marrobi @tamirkamara WDYT?

jonnyry: Re testing, have run the following scenarios on a local CICD build:

1. Deploy to empty environment.
2. Deploy to existing TRE.
3. Certificate renewal.
4. Destroy TRE

marrobi (approved):
Tested locally, LGTM

core/terraform/locals.tf
```diff
@@ -46,4 +46,10 @@ locals {
 
   cmk_name                 = "tre-encryption-${var.tre_id}"
   encryption_identity_name = "id-encryption-${var.tre_id}"
+
+  # key vault variables
+  kv_name                          = "kv-${var.tre_id}"
+  kv_public_network_access_enabled = true
```
yuvalyaron, line 52: is this always true? if so it's redundant, if not, let's make it a variable with default value

yuvalyaron (approved):
Nice work!

core/terraform/locals.tf
jonnyry: It is always true yes - though the reason for declaring as a local was in order to make sure the same value gets used for the`public_network_access_enabled` attribute of `azurerm_key_vault` and in the create provisioner in the same resource.   Sadly within the provisioner you can't refer to `azurerm_key_vault.kv.public_network_access_enabled`, that would have been ideal - the only way was to move the value out into a local.

```
resource "azurerm_key_vault" "kv" {
...
public_network_access_enabled = local.kv_public_network_access_enabled
...
 provisioner "local-exec" {
    when    = create
    command = <<EOT
az keyvault update --name ${local.kv_name} --public-network-access ${local.kv_public_network_access_enabled ? "Enabled" : "Disabled"}
...
```

jonnyry:
> 🤖 pr-bot 🤖
> 
> 🏃 Running tests: https://github.com/microsoft/AzureTRE/actions/runs/13306353906 (with refid `26f9d939`)
> 
> (in response to [this comment](https://github.com/microsoft/AzureTRE/pull/4260#issuecomment-2656266258) from @jonnyry)

@marrobi another unexpected "resource already exists error".  Performed a `/test-destroy-env` just prior

```
 »»» 🤖 Creating resource group and storage account...
Location    Name
----------  -------------------
eastus2     ***
WARNING: A storage account with the provided name *** is found. Will continue to update the existing account.
ERROR: (StorageAccountAlreadyTaken) The storage account named *** is already taken.
Code: StorageAccountAlreadyTaken
Message: The storage account named *** is already taken.
make: *** [Makefile:34: bootstrap] Error 1
```

marrobi:
@jonnyry  I think it takes a bit of time for the storage name to become available again, it can get cached at the Azure platform side for a while. If run again it will likely work.

marrobi:
/test-force-approve

Tested locally.

jonnyry:
> @jonnyry I think it takes a bit of time for the storage name to become available again, it can get cached at the Azure platform side for a while. If run again it will likely work.

ah right, yes DNS issues.  I'll try again in 20 mins.

PR #4274, opened 2025-01-16
Add Windows image field to Admin VM template

jonnyry:
Add OS image field to Admin VM template in alignment with Guacamole VMs, to allow for easier OS image customisation.


tamirkamara:
Is this really needed @jonnyry ?

jonnyry:
> Is this really needed @jonnyry ?

It is useful for us to align our Windows images to a single base image.  However if you don't see value to the wider user base I can close.

marrobi (approved):

templates/shared_services/admin-vm/terraform/variables.tf
```diff
@@ -20,3 +20,12 @@ variable "enable_cmk_encryption" {
 variable "key_store_id" {
   type = string
 }
+
+variable "image_gallery_id" {
+  type    = string
+  default = ""
```
ShakutaiGit, line 26: Should we remove the default value ?

templates/shared_services/admin-vm/terraform/locals.tf
```diff
@@ -8,4 +8,14 @@ locals {
...
   encryption_identity_name = "id-encryption-${var.tre_id}"
+
+  # Load image details from porter.yaml
+  porter_yaml   = yamldecode(file("${path.module}/../porter.yaml"))
+  image_details = local.porter_yaml["custom"]["image_options"]
+
+  # Create local variables to support the VM resource
+  selected_image = local.image_details[var.image]
+  # selected_image_source_refs is an array to enable easy use of a dynamic block
+  selected_image_source_refs = lookup(local.selected_image, "source_image_reference", null) == null ? [] : [local.selected_image.source_image_reference]
```
ShakutaiGit, line 19: should we simplify to ?
selected_image_source_refs = coalesce(lookup(local.selected_image, "source_image_reference", null), [])

ShakutaiGit (approved):
LGTM
(very small/nit comments)

templates/shared_services/admin-vm/terraform/variables.tf
jonnyry: Ah so `image_gallery_id` is only passed if a deployment has set up a custom image gallery:

RP_BUNDLE_VALUES='{"image_gallery_id": "..."}'

So we need a default to prevent terraform from complaining about a missing value.

https://microsoft.github.io/AzureTRE/v0.20.0/tre-templates/user-resources/custom/#configuration

ShakutaiGit: i was thinking that the default value from the porter is enough 
https://github.com/microsoft/AzureTRE/pull/4274/files#diff-18bde907be01c5c2360c9ddffc1d247b39c194e2a01e10815ea12bf9803c1c09R71

templates/shared_services/admin-vm/terraform/locals.tf
jonnyry: it's a lot easier to read but i don't think it's equivalent - if the lookup succeeds and returns `"myimage"`...

the original will evaluate to `["myimage"]`
whereas the new version will evaluate to `"myimage"`

templates/shared_services/admin-vm/terraform/variables.tf
jonnyry: now fixed :-) https://github.com/microsoft/AzureTRE/pull/4274/commits/e31df9a0c7b59dc06c05fe894b83eddeac6566e9

jonnyry:
@marrobi @ShakutaiGit thanks for your reviews.

I've made a [small change](https://github.com/microsoft/AzureTRE/pull/4274/commits/cbdba29318faf683a1d32d8d80c9a3acaf1d62ce) to bring across the `secure_boot_enabled` and `vtpm_enabled` image config options that were [added to main](https://github.com/microsoft/AzureTRE/commit/fd89717d6e674c95cf90b20197b9ad2c478bc155) since this PR was originally submitted.

It ensures these values are configed per image, and not hardcoded to `true` as they were prior.  (The image you select may not be secure boot capable.)

marrobi:
@jonnyry not sure this will be covered by any of the tests?

PR #4276, opened 2025-01-19
Airlock Azure Function Host Storage Access with Managed Identity

tamirkamara:
## What is being addressed

The azure function that serves the airlock feature uses its host storage with a shared access key which is less secure.

## How is this addressed

- It took a while to find this undocumented way to direct the function to use the user-assigned managed identity it already has.


yuvalyaron (approved):

marrobi (approved):
LGTM. Took some time to start working.

PR #4289, opened 2025-01-29
Genericise Windows VM post install script to deal with different R versions

### LINKED ISSUE #4288: Windows R version must be 4.1.2 otherwise post install script doesn't update package mirror URL
The Guacamole Windows VM post install script which configures the R package mirror is hard coded to R version 4.1.2:

https://github.com/microsoft/AzureTRE/blob/7c88fe498de3dabc9adb2bfd0122d1f49f246a5b/templates/workspace_services/guacamole/user_resources/guacamole-azure-windowsvm/terraform/vm_config.ps1#L50-L57

The latest version of R is 4.4.2.

It would be preferable if the script looked for R versions and applied the package mirror config to the one it found / all of them, rather than only a specific version.

jonnyry:
# Resolves #4288

## What is being addressed

See #4288


tamirkamara (approved):

jonnyry:
/test-force-approve

not covered by CI test, tested locally

PR #4340, opened 2025-02-07
Fix Cosmos DB local debugging

marrobi:
Introduce functionality to assign the Cosmos DB Built-in Data Contributor role when configuring local debugging to so that cosmos data can be read.

Always use Entra ID credentials unless a key is manually specified.


ShakutaiGit (approved):
LGTM

devops/scripts/setup_local_debugging.sh
```diff
@@ -91,6 +91,19 @@ az role assignment create \
     --assignee "${LOGGED_IN_OBJECT_ID}" \
     --scope "${STATE_STORE_RESOURCE_ID}"
 
+ROLE_DEFINITION_ID=$(az cosmosdb sql role definition list \
```
yuvalyaron, line 94: Nice!

PR #4341, opened 2025-02-07
Update Azure Machine Learning workspace to use AD itegrated auth to storage

marrobi:
Refactor the Azure Machine Learning workspace resource to use the azapi provider so don;t have to use key based storage auth.


templates/workspace_services/azureml/terraform/main.tf
```diff
@@ -1,29 +1,61 @@
...
-  high_business_impact          = true
-  key_vault_id                  = data.azurerm_key_vault.ws.id
-  public_network_access_enabled = var.is_exposed_externally ? true : false
-  storage_account_id            = azurerm_storage_account.aml.id
-  tags                          = local.tre_workspace_service_tags
-
-  identity {
-    type = "SystemAssigned"
+resource "azapi_resource" "aml_workspace" {
+  type      = "Microsoft.MachineLearningServices/workspaces@2024-10-01-preview"
```
LizaShak, line 2: why not [2025-01-01-preview](https://learn.microsoft.com/en-us/azure/templates/microsoft.machinelearningservices/2025-01-01-preview/workspaces) ? or the [2024-10-01](https://learn.microsoft.com/en-us/azure/templates/microsoft.machinelearningservices/2024-10-01/workspaces) if you don't want a preview

templates/workspace_services/azureml/terraform/main.tf
```diff
@@ -1,29 +1,61 @@
...
+
+  dynamic "identity" {
+    for_each = var.enable_cmk_encryption ? [1] : []
+    content {
+      type         = "SystemAssigned, UserAssigned"
+      identity_ids = [data.azurerm_user_assigned_identity.ws_encryption_identity[0].id]
+    }
   }
 
-  lifecycle {
```
LizaShak, line 19: Why are you removing the lifecycle part?

templates/workspace_services/azureml/terraform/main.tf
marrobi: I need systemDatastoresAuthMode, is only in preview API, I did this before 2025-01-01-preview was out.

Will update and test.

templates/workspace_services/azureml/terraform/main.tf
marrobi: It's on line 8, but, yes, should add `imageBuildCompute`, not sure why public_network_access is on there, Will test it out. Thanks.

jonnyry (commented):
The changes to enable AD integrated auth LGTM, though there are two other features in there as well - adding logging to log analytics & a change to the API - are they related to this change?

marrobi:
> The changes to enable AD integrated auth LGTM, though there are two other features in there as well - adding logging to log analytics & a change to the API - are they related to this change?

Yeah, might split it out. The provider update requires log analytics workspace,but would be better seperate.

marrobi:
Blocked by #4451

martinpeck:
While I can't see anything obviously wrong here, I'm not sure I have enough knowledge around this change to review it. 
I'm untagging myself from the reviewers.

templates/workspace_services/azureml/terraform/main.tf
```diff
@@ -53,3 +85,14 @@ resource "azurerm_private_endpoint" "mlpe" {
   ]
 
 }
+
+resource "azurerm_application_insights" "ai" {
```
tamirkamara, line 89: A workspace has app insights already. Can't you use it?
If not, just consider that the AI is configured in a simplified approach without the private stuff we use in the workspace level. It might also suffer delete errors.

tamirkamara: @marrobi how will they access app insights? 
(the current workspace one is used by guacamole for example and it's true we don't want researchers to have access to it).

marrobi: So I ran though this, but struggled to test as the simple method of deploying online endpoints (which uses app insights) doesn't seem to be supported in the secure networking configuration. 

This isn't actually changed in this PR, I just moved the file it is sitting in. Suggest merge the PR as is, and if this comes up as an issue can be addressed separately.

tamirkamara: ok. will resolve for now.

tamirkamara (approved):

Issue #4243, opened 2025-01-02
Add docs around recovering from a failed firewall.


marrobi:
If the TRE firewall has to be deleted, or gets deleted by terraform but not recreated, then it can be hard to recover as the resource processor has no internet access.

To resolve the route table, `rt-<tre_id>` for the `ResourceProcessorSubnet` needs removing temporarily to allow internet access, then the firewall shared service can be updated/upgraded so that the firewall is recreated. Then the route table needs putting back on the `ResourceProcessorSubnet`.

PR #4342, opened 2025-02-07
Migrate Azure Firewall and route tables to core configuration

marrobi:
Move Azure Firewall Deployment to Core

Why: 
- Avoids the issue of the firewall and rules for core not being deployed when it gets deployed, this means if the shared shared service deployment hits an issue the system can still operate. ( https://github.com/microsoft/AzureTRE/issues/4243 )
- Resolves one of the issues to have a reduced service bus sku that requires traffic to be routed through the firewall ( https://github.com/microsoft/AzureTRE/issues/3953#issuecomment-2200325098 ).
- Opens up the option in future to deploy the system without a firewall (for dev, hence reducing cost)
- Removes the complexity of core config vars being used in shared service deployment.

Had to modify GitHub workflow for linting. Can someone verify this is correct?

Tested upgrade a few times, more wouldn't hurt.


martinpeck:
@marrobi please update with `main` and then publish for review

marrobi:
This would avoid https://github.com/microsoft/AzureTRE/actions/runs/14170065338/job/39697620186

core/terraform/firewall/import_state.sh
tamirkamara, line 1: Why do you import the state here if we usually do it in the migration script?

core/terraform/main.tf
```diff
@@ -92,6 +92,28 @@ module "network" {
...
+  log_analytics_workspace_id     = module.azure_monitor.log_analytics_workspace_id
+  firewall_management_subnet_id  = module.network.firewall_management_subnet_id
+  resource_processor_ip_group_id = module.network.resource_processor_ip_group_id
+  shared_services_ip_group_id    = module.network.shared_services_ip_group_id
+  web_app_ip_group_id            = module.network.web_app_ip_group_id
+  airlock_processor_ip_group_id  = module.network.airlock_processor_ip_group_id
+
+  depends_on = [
+    module.network,
+  ]
```
tamirkamara, line 114: I think this is redundant.

core/terraform/firewall/firewall.tf
tamirkamara, line 1: Does the FW really needs to be a module?

core/terraform/firewall/remove_state.sh
tamirkamara, line 1: Same here. I think it should be move to the migration script.

templates/shared_services/firewall/porter.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 schemaVersion: 1.0.0
 name: tre-shared-service-firewall
-version: 1.3.3
+version: 1.4.0
 description: "An Azure TRE Firewall shared service"
```
tamirkamara, line 5: I understand we can't change the name, but we can consider changing the description, template schema and docs to reflect the fact this is now just about managing the workspace level firewall policy rules.

templates/shared_services/firewall/terraform/locals.tf
```diff
@@ -1,30 +1,8 @@
...
-  ]
-  tre_shared_service_tags = {
-    tre_id                = var.tre_id
-    tre_shared_service_id = var.tre_resource_id
-  }
 
   api_driven_application_rule_collection = jsondecode(base64decode(var.api_driven_rule_collections_b64))
   api_driven_network_rule_collection     = jsondecode(base64decode(var.api_driven_network_rule_collections_b64))
 
   firewall_policy_name = "fw-policy-${var.tre_id}"
```
tamirkamara, line 7: I think it is now quite easy to provide this at run time as since it is set in the core it should be provided from there as a bundle variable.

templates/shared_services/firewall/terraform/rules.tf
```diff
@@ -215,8 +15,8 @@ resource "azurerm_firewall_policy_rule_collection_group" "dynamic_network" {
         for_each = network_rule_collection.value.rules
 
         content {
-          name = rule.value.name
-          # description      = rule.value.description
+          name             = rule.value.name
+          description      = rule.value.description
```
tamirkamara, line 19: Is it confirmed that network rules now support a description?

core/terraform/firewall/import_state.sh
marrobi: Hmm, I moved the file over from the firewall bundle, from when we went the other way. I can move this to the migrate script.

core/terraform/firewall/firewall.tf
marrobi: The reason was to leave the option to have it conditional in the future. If we don't want to deploy it for a dev environment for example.

core/terraform/main.tf
marrobi: Will remove.

templates/shared_services/firewall/porter.yaml
marrobi: Yup, primarily wanted to see if thought this works as a change in principal. If agreed, will go through and fix up.

templates/shared_services/firewall/terraform/locals.tf
marrobi: Not sure what you mean? The policy name?

templates/shared_services/firewall/terraform/rules.tf
marrobi: Will check. It deploys and works... can't remember why changed this.

marrobi: Good catch, no idea where this came from.

templates/shared_services/firewall/porter.yaml
marrobi: I've changed in the porter file, can't see any references in the docs.

templates/shared_services/firewall/terraform/locals.tf
tamirkamara: Yes, I meant the policy name. If you want it can come directly from the core deployment.

marrobi: You mean by adding `firewall_policy_id`  here: https://github.com/marrobi/AzureTRE/blob/7e532bd19b7b58d526cf84fa007060574b75ee8f/core/terraform/resource_processor/vmss_porter/locals.tf#L16 and passing through?

Can do I guess.

marrobi: Done, and tested an upgrade.

tamirkamara (approved):
LGTM but I didn't test myself.

jonnyry (approved):

PR #4344, opened 2025-02-07
Add validation to Template API endpoints

### LINKED ISSUE #1123: Validation Missing from API Endpoints
When calling the various POST endpoints it's possible to post a `workspace service` payload to the `workspace` api, and vice versa. Neither call will return an error. Both will register a template that's not useful.

In both cases, the payload sent to these APIs has a resourceType property that could have been used to validate the payload.

Validation could be added, and an error could be returned when inappropriate payloads are posted to these APIs.

marrobi:
Fixes #1123

Add validation to API endpoints to ensure correct resourceType in template registration payloads.

* **Workspace Service Templates**: Add validation in `api_app/api/routes/workspace_service_templates.py` to check if `resourceType` is `WorkspaceService` before registering the template. Return a 422 Unprocessable Entity error if the `resourceType` does not match.
* **Workspace Templates**: Add validation in `api_app/api/routes/workspace_templates.py` to check if `resourceType` is `Workspace` before registering the template. Return a 422 Unprocessable Entity error if the `resourceType` does not match.
* **User Resource Templates**: Add validation in `api_app/api/routes/user_resource_templates.py` to check if `resourceType` is `UserResource` before registering the template. Return a 422 Unprocessable Entity error if the `resourceType` does not match.
* **Shared Service Templates**: Add validation in `api_app/api/routes/shared_service_templates.py` to check if `resourceType` is `SharedService` before registering the template. Return a 422 Unprocessable Entity error if the `resourceType` does not match.
* **Common Error Message**: Add a common error message string `INVALID_RESOURCE_TYPE` in `api_app/resources/strings.py` for validation errors, including the expected and received resourceType.
* **Tests**: Add tests in `api_app/tests_ma/test_api/test_routes/test_workspace_service_templates.py`, `api_app/tests_ma/test_api/test_routes/test_workspace_templates.py`, `api_app/tests_ma/test_api/test_routes/test_user_resource_templates.py`, and `api_app/tests_ma/test_api/test_routes/test_shared_service_templates.py` to cover validation for each template registration

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/microsoft/AzureTRE/pull/4344?shareId=44ef0a01-0d1e-443a-995d-26eb865180ee).

### DIFF
--- a/api_app/resources/strings.py
+++ b/api_app/resources/strings.py
@@ -254,3 +254,6 @@ PARAMETERS_WITH_WRONG_TYPE = "Parameters with wrong type"
 
 # Value that a sensitive is replaced with in Cosmos
 REDACTED_SENSITIVE_VALUE = "REDACTED"
+
+# Common error message for invalid resource type
+INVALID_RESOURCE_TYPE = "Invalid resourceType. Expected '{}'. Received '{}'."
--- a/api_app/api/routes/shared_service_templates.py
+++ b/api_app/api/routes/shared_service_templates.py
@@ -33,6 +33,8 @@ async def get_shared_service_template(shared_service_template_name: str, is_upda
 
 @shared_service_templates_core_router.post("/shared-service-templates", status_code=status.HTTP_201_CREATED, response_model=SharedServiceTemplateInResponse, response_model_exclude_none=True, name=strings.API_CREATE_SHARED_SERVICE_TEMPLATES, dependencies=[Depends(get_current_admin_user)])
 async def register_shared_service_template(template_input: SharedServiceTemplateInCreate, template_repo=Depends(get_repository(ResourceTemplateRepository))) -> ResourceTemplateInResponse:
+    if template_input.resourceType != ResourceType.SharedService:
+        raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_ENTITY, detail=strings.INVALID_RESOURCE_TYPE.format('SharedService', template_input.resourceType))
     try:
         return await template_repo.create_and_validate_template(template_input, ResourceType.SharedService)
     except EntityVersionExist:
--- a/api_app/api/routes/user_resource_templates.py
+++ b/api_app/api/routes/user_resource_templates.py
@@ -1,4 +1,3 @@
-
 from typing import Optional
 from fastapi import APIRouter, Depends, HTTPException, status
 from pydantic import parse_obj_as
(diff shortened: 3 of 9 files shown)


marrobi:
This PR uncovered a few test bugs, however we are inconsistent with our resourceType naming. In the CI and makefile we use `workspace_service, in the Python API we use  `workspace-service`.

Going to leave this here, but should resolve this inconsistency. Suggest by amending the CI and makefile to use `-`.

marrobi:
Propose we should handle the naming difference in the Python code, and deal with the script update later, add notes to the release, then revert the Python  in the future.

martinpeck:
Marking as ready for review.
Can't be merged as-is.

api_app/resources/strings.py
```diff
@@ -254,3 +254,6 @@
 
 # Value that a sensitive is replaced with in Cosmos
 REDACTED_SENSITIVE_VALUE = "REDACTED"
+
+# Common error message for invalid resource type
+INVALID_RESOURCE_TYPE = "Invalid resourceType. Expected '{}'. Received '{}'."
```
martinpeck, line 259: I half wondered if printing the "received" string could be used to exfiltrate some information. I'm 99% sure that a string.format in python, using a `{}` placeholder won't allow a variable/environment variable to be printed...but if we wanted to be 100% sure then can we just remove it?

The sender should be able to determine what they sent. The more important thing is that what they send it not expected, and we tell them what is expected.

martinpeck (approved):
I've left a question/comment but LGTM either way.

api_app/resources/strings.py
marrobi:
```suggestion
INVALID_RESOURCE_TYPE = "Invalid resourceType. Expected '{}'."
```

So this? I believe I tested it, was a while ago.

marrobi:
> Propose we should handle the naming difference in the Python code, and deal with the script update later, add notes to the release, then revert the Python in the future.

@martinpeck what think of best way to sort this? The scripts send resources with `_` vs expecting `-` , see [/workspaces/marrobi-azuretre/.github/workflows/deploy_tre_reusable.yml for ](https://github.com/marrobi/AzureTRE/blob/973a02d7691a1e9272197d9159bb1699ec0a8e4a/.github/workflows/deploy_tre_reusable.yml#L561) for example. 

I don't want to break everyone's pipelines.

Could do a replace of `_` for `-` here: https://github.com/marrobi/AzureTRE/blob/973a02d7691a1e9272197d9159bb1699ec0a8e4a/devops/scripts/register_bundle_with_api.sh#L102-L103

martinpeck:
If we *did* patch up the problem in that script, I would suggest we do that only as a temp measure, generate a warning saying "WARNING: Fix this cos we won't patch this for you later", and then remove that patch code in a later release.

PR #4372, opened 2025-02-13
Add bundle target to Makefile for handling different types options in single command

guybartal:
## What is being addressed

Add bundle target to Makefile for handling different bundle types in single command


marrobi (approved):
LGTM

yuvalyaron (approved):

PR #4383, opened 2025-02-19
Add bastion deploy and sku configuration

### LINKED ISSUE #86: Enable Azure bastion free sku.
## Description

As an Azure administrator I want to chose is Azure bastion should be deployed so that I can save costs and remove a potential entry point to workspaces.

## Acceptance criteria

- [ ] Variable at deploy time to disable the deployment of bastion

marrobi:
Fixes #86

Add a variable to configure Azure Bastion SKU and enable/disable deployment.


marrobi:
Switching to draft given limited region availability.

jonnyry (approved):
LGTM

One small thing, should the bastion public ip also be conditional deployed? (or are there knock ons from making this conditional)

martinpeck (approved):
This looks fine to me, but I just wanted to check that the defaults make sense.

So, this is an optional configuration switch, but the default is true, and so we will always deploy bastion unless someone opts-out, and we will default to the Basic SKU.

Given this, do we expect Bastion to be deployed in most cases (i.e. will most of our testing be on deployments where Bastion is deployed)? Under what circumstances might someone want switch this to false? Why not always deploy Bastion?

I ask these questions as I think it'll help me understand if these defaults are correct, and if people need to understand that upgrading to the next release of TRE will require them to take action if they don't want Bastion.

marrobi:
> This looks fine to me, but I just wanted to check that the defaults make sense.
> 
> So, this is an optional configuration switch, but the default is true, and so we will always deploy bastion unless someone opts-out, and we will default to the Basic SKU.
> 
> Given this, do we expect Bastion to be deployed in most cases (i.e. will most of our testing be on deployments where Bastion is deployed)? Under what circumstances might someone want switch this to false? Why not always deploy Bastion?
> 
> I ask these questions as I think it'll help me understand if these defaults are correct, and if people need to understand that upgrading to the next release of TRE will require them to take action if they don't want Bastion.

It is often required for on the fly troubleshooting. So think the default should be true, otherwise people have to get bastion deployed.

Some customers don't, want it, my primary driver is to reduce dev and cd costs. Does that help?

martinpeck:
Makes sense. Will you call out "now deploys Bastion by default" in release notes?

marrobi:
> Makes sense. Will you call out "now deploys Bastion by default" in release notes?

It always has... so not sure need to, just now have an option to not.

martinpeck:
Sorry....yeah, my mistake.

I thought that this was changing when it was installed, or changing the sku that was used, but I now see that neither have changed (just the ability to opt out, or pick a sku).

marrobi:
/test-force-approve

Bastion part passed.

PR #4407, opened 2025-02-28
Fix: make show-core-output is failing due to lack of access to management storage account

### LINKED ISSUE #4404: `make show-core-output` is failing due to lack of access to management storage account
This isn't called during CICD, it only happens when running a build from the command line using `make all`, or directly calling `make show-core-output`.

Image

ashis-kar91:
# Resolves #4404

## What is being addressed

- The `make show-core-output` command in the make file was executing `terraform show` was failing due to lack of access to the management storage account. Because of this TRE deployments were failing while using the `make all` command. 
- There were too many redundant error messages being logged from `devops/scripts/mgmtstorage_enable_public_access.sh` during every attempt to verify the storage account access. 

## How is this addressed

- Added a new script file `core/terraform/show_output.sh`, which will be using `terraform_wrapper.sh` to execute the terraform show. `terraform_wrapper.sh` already has the script for management storage exception. 
- For the redundant error messages in `devops/scripts/mgmtstorage_enable_public_access.sh`, captured the underlying error in every attempt and printing it at the end if all the attempts fail. 
- Also removed the redundant access checks in `devops/scripts/mgmtstorage_enable_public_access.sh` before adding and removing network rules, since both the commands are idempotent.


marrobi (approved):
LGTM

jonnyry (approved):
Looking good, thanks very much :-)

PR #4418, opened 2025-03-10
Add malware scanning to workspace storage account for airlock exports

### LINKED ISSUE #4403: Airlock Malware Scanning not working for Export
Hi,

We deployed TRE with 'enable_airlock_malware_scanning: true' . Airlock import requests were not moving from submitted > in review and the import requests were not showing for the airlock manager.

We ran 'make deploy-core' with 'enable_airlock_malware_scanning: false' and the airlock import requests moved to 'in review' and were available to the airlock manager.

The malware scanning is a critical feature for our client and we would appreciate some assistance to establish how to get this enabled.

Many thanks,
Alan

SiobhanBaynes:
# Resolves #4403 

## What is being addressed

Fix to enable malware scanning at the workspace level, this ensures that malware scanning will be applied to airlock export requests as well as imports.

## How is this addressed

- Add `enable_malware_scanning` variable to Porter.yaml
- Add `enable_malware_scanning` variable to Airlock module
- Add `enable_malware_scanning` to Airlock variables
- Add code to enable malware scanning to `sa_export_inprogress` storage account


airlock_processor/BlobCreatedTrigger/__init__.py
```diff
@@ -32,7 +32,7 @@ def main(msg: func.ServiceBusMessage,
             logging.error("environment variable 'ENABLE_MALWARE_SCANNING' does not exists. Cannot continue.")
             raise
 
-        if enable_malware_scanning and constants.STORAGE_ACCOUNT_NAME_IMPORT_INPROGRESS in topic:
+        if enable_malware_scanning and constants.STORAGE_ACCOUNT_NAME_IMPORT_INPROGRESS in topic or constants.STORAGE_ACCOUNT_NAME_EXPORT_INPROGRESS in topic:
```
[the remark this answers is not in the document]
martinpeck: I ask because I think you might need brackets around the two OR'd terms...

```python
# Malware is disabled, and an export is in progress
malware_enabled = False
import_inprogress = False
export_inprogress = True

# current logic
if malware_enabled and import_inprogress or export_inprogress:
    print("Outcome 1")    # prints "Outcome 1"

# I think this is what you want
if malware_enabled and (import_inprogress or export_inprogress):
    print("Outcome 2")    # doesn't execute
```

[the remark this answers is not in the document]
SiobhanBaynes: Well spotted, thanks.

core/terraform/resource_processor/vmss_porter/variables.tf
```diff
@@ -97,3 +97,13 @@ variable "kv_encryption_key_name" {
...
   description = "Name of Key Vault Encryption Key (only used if enable_cmk_encryption is true)"
 }
+
+variable "enable_malware_scanning" {
+  type        = bool
+  description = "If False, Airlock requests will skip the malware scanning stage"
+}
+
+variable "scan_result_topic_name" {
+  type        = string
```
martinpeck, line 107: Is there any harm in setting a default for this? Presumably, if scanning isn't enabled it would make no difference.

templates/workspaces/base/terraform/airlock/storage_accounts.tf
```diff
@@ -230,6 +229,31 @@ resource "azurerm_private_endpoint" "export_inprogress_pe" {
...
+  resource_id = "${azurerm_storage_account.sa_export_inprogress.id}/providers/Microsoft.Security/defenderForStorageSettings/current"
+  method      = "PUT"
+
+  body = jsonencode({
+    properties = {
+      isEnabled = true
+      malwareScanning = {
+        onUpload = {
+          isEnabled     = true
+          capGBPerMonth = 5000
```
martinpeck, line 245: does this value need to be a parameter , or is `5000` an agreed value?

SiobhanBaynes: TBH this is copied straight from core where it was already in use.

core/terraform/resource_processor/vmss_porter/variables.tf
SiobhanBaynes: The name of the topic gets generated by `locals.tf` in core so I needed to output the generated name and pass it through to the workspace so we can use the same topic. The function that sits on the end of it already had code to cope with core or workspace so it just needed hooking up.

core/terraform/airlock/outputs.tf
```diff
@@ -17,3 +17,7 @@ output "event_grid_status_changed_topic_resource_id" {
 output "event_grid_airlock_notification_topic_resource_id" {
   value = azurerm_eventgrid_topic.airlock_notification.id
 }
+
+output "scan_result_topic_name" {
```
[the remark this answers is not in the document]
SiobhanBaynes: It gets passed through to the Porter bundle so that it can be applied to the `export-in_progress` storage account.

core/terraform/resource_processor/vmss_porter/cloud-config.yaml
```diff
@@ -63,6 +63,8 @@ write_files:
       LOGGING_LEVEL=${logging_level}
       ENABLE_CMK_ENCRYPTION=${enable_cmk_encryption}
       KEY_STORE_ID=${key_store_id}
+      ENABLE_MALWARE_SCANNING=${enable_malware_scanning}
```
tamirkamara, line 66: We generally want to pass only Resource Process params here. This is a param used in the bundles and should be passed via the rp_bundle_values. There was something we did to support named params provided in the config.yaml file to allow it to be more explicit. I will look it up.

tamirkamara (changes requested):
There's something fundamental about how we make config.yaml values available to bundles in this PR I feel we should improve.

core/terraform/resource_processor/vmss_porter/cloud-config.yaml
SiobhanBaynes: I take your point, but this is a system-wide setting that's used in core and _also_ needs to travel through to the workspace. Whether a workspace bundle uses that setting is dependent on the bundle?

SiobhanBaynes: I take your point, but the `enable_airlock_malware_scanning` value already exists in `config.sample.yaml` with a default of `true`. Currently the`rp_bundle_values` section is commented out in here, so we would have to uncomment this and also put the `enable_malware_scanning` value in here. This could then get out of sync with the value in core.

tamirkamara: I have a [PR ](https://github.com/microsoft/AzureTRE/pull/4437) that will help avoid making changes in the resource processor code for cases like this.
I will reply again after it's been merged.

tamirkamara: Oh, can you fix the title and/or description of this PR to reflect it is targeting the outgoing airlock part (and not the internal where the malware scanner works just fine)?

tamirkamara: The PR is now merged. IMO you should amend the following:
1. Revert all changes to the RP code and add config.yaml settings or other you want to be available for the bundle [here](https://github.com/microsoft/AzureTRE/blob/main/core/terraform/resource_processor/vmss_porter/locals.tf#L16)
2. Update the title and/or description of the PR as commented above.
3. Consider changing the name of params you make available and use in the bundle as I'm not sure it's clear enough in the bundle context to note it's related to the airlock feature.

SiobhanBaynes:
@microsoft-github-policy-service agree company="Microsoft"

templates/workspaces/base/porter.yaml
```diff
@@ -126,6 +126,14 @@ parameters:
...
     default: "GRS"
     description: "The redundancy option for the storage account in the workspace: GRS (Geo-Redundant Storage) or ZRS (Zone-Redundant Storage)."
+  - name: enable_airlock_malware_scanning
+    type: boolean
+    default: false
+    description: "Enable malware scanning on the workspace storage account"
+  - name: airlock_malware_scan_result_topic_name
+    type: string
+    default: ""
+    description: "The name of the topic to publish scan results to"
```
tamirkamara, line 136: Since both those params are internal and not up to user discretion, I think we should hide them in the UI. WDYT?
An example: https://github.com/microsoft/AzureTRE/blob/8812503011f73004960529acff20e76fb77fd164/templates/workspaces/airlock-import-review/template_schema.json#L165

tamirkamara (approved):

martinpeck (approved):
LGTM. Thanks @SiobhanBaynes and apologies for the delay in re-reviewing

marrobi:
/test-force-approve

Passed: https://github.com/microsoft/AzureTRE/actions/runs/14195182986

eyalanmegaw:
Hello,

I have just tested an airlock export request today:
export

But no record of malware scan in the function airlock processor invocations:
scan_result_invocations

Both scan results shown relate to imports not exports.

NB TRE was deployed from code on Friday 4th April.

Kind regards,
Alan

martinpeck:
Can you confirm how you are deploying the TRE?
If you're using the [azuretre_deployment ](https://github.com/microsoft/AzureTRE-Deployment) project, was is the value of `UPSTREAM_REPO_VERSION` in your `devcontainer.json`?

eyalanmegaw:
> Can you confirm how you are deploying the TRE? If you're using the [azuretre_deployment ](https://github.com/microsoft/AzureTRE-Deployment) project, was is the value of `UPSTREAM_REPO_VERSION` in your `devcontainer.json`?

Hi @martinpeck,

thanks for reverting.

we deployed manually via the quickstart (https://microsoft.github.io/AzureTRE/latest/tre-admins/setup-instructions/setup-auth-entities/) using the public template (https://github.com/microsoft/AzureTRE-Deployment). I cloned the files from the public repo on **3rd April**.

upstream_repo_version

kind regards,
Alan

martinpeck:
Thanks. The way that the AzureTRE-deployment repo works is that it pulls from the Releases of this repo. 

In your case, it's pulling release 0.21.0. You can see those releases here: https://github.com/microsoft/AzureTRE/releases

Release 0.21.0 was created in Feb, so it won't have any of these changes within it. When a new release is created this `UPSTREAM_REPO_VERSION` will be updated along with the creation of the release.

So, what you're seeing is expected. You're essentially testing the functionality of the current release, not a release that includes any of these malware scanning changes.

You have two options:

1. You can wait until a new release. This is by far the most reliable option. I suggest you do this.
2. You could, if you wanted to, change your UPSTREAM_REPO_VERSION so that instead of saying "v0.21.0" you specify `main`. This will pull the latest code from `main`. **WARNING**: be aware that you may well hit issues. While `main` is generally good, it might be broken, and you'll be picking up other changes too. Some of those changes won't have been tested in the way we test releases. Only do this if you're absolutely sure you know what you're doing, and you're happy to delete and redeploy if thing go wrong. 

If you go with option 2 (and, as I say, this has higher risk and I don't recommend it) then you can see more info on this here:

https://microsoft.github.io/AzureTRE/latest/tre-admins/upgrading-tre/#deploying-a-specific-version-of-azure-tre

(note: I've just spotted that the docs have a bug and refer to the old config values of `OSS_VERSION` and `OSS_REPO` which have changed to `UPSTREAM_REPO_VERSION` and `UPSTREAM_REPO`).

One obvious question you might have is "when will the next release appear?". Good question. Not sure.  Let me check.

martinpeck:
Raised #4479 to track the doc bug

eyalanmegaw:
> Thanks. The way that the AzureTRE-deployment repo works is that it pulls from the Releases of this repo.
> 
> In your case, it's pulling release 0.21.0. You can see those releases here: https://github.com/microsoft/AzureTRE/releases
> 
> Release 0.21.0 was created in Feb, so it won't have any of these changes within it. When a new release is created this `UPSTREAM_REPO_VERSION` will be updated along with the creation of the release.
> 
> So, what you're seeing is expected. You're essentially testing the functionality of the current release, not a release that includes any of these malware scanning changes.
> 
> You have two options:
> 
> 1. You can wait until a new release. This is by far the most reliable option. I suggest you do this.
> 2. You could, if you wanted to, change your UPSTREAM_REPO_VERSION so that instead of saying "v0.21.0" you specify `main`. This will pull the latest code from `main`. **WARNING**: be aware that you may well hit issues. While `main` is generally good, it might be broken, and you'll be picking up other changes too. Some of those changes won't have been tested in the way we test releases. Only do this if you're absolutely sure you know what you're doing, and you're happy to delete and redeploy if thing go wrong.
> 
> If you go with option 2 (and, as I say, this has higher risk and I don't recommend it) then you can see more info on this here:
> 
> https://microsoft.github.io/AzureTRE/latest/tre-admins/upgrading-tre/#deploying-a-specific-version-of-azure-tre
> 
> (note: I've just spotted that the docs have a bug and refer to the old config values of `OSS_VERSION` and `OSS_REPO` which have changed to `UPSTREAM_REPO_VERSION` and `UPSTREAM_REPO`).
> 
> One obvious question you might have is "when will the next release appear?". Good question. Not sure. Let me check.

thanks @martinpeck for taking the time to explain the release process, much appreciated. I think option 1 (waiting for next release) is the safest option. And yes my next question was exactly "when will the next release appear?" Is there an organised release schedule or does it vary due to the number of bugs and resources at hand? thanks again!

SiobhanBaynes:
Just re-tested in a new TRE deployment from main and I can see invocations for the export container:
Airlock Screenshot
So hopefully all will be well when you get this release.

Thanks for your input @martinpeck

PR #4422, opened 2025-03-17
Document Makefile Commands

### LINKED ISSUE #4296: Add documentation on make commands
## Description

As a Azure TRE Developer
I want to have a documentation on all the make commands
So that I know what they are ment for, what are the required arguments and all the information I need so I can easily find make commands that I need.

## Acceptance criteria

- [ ] All make commands are documented

LizaShak:
# Resolves  #4296  

## What is being addressed

This PR adds documentation for the commands in the Makefile to provide clear descriptions and usage examples for each target.


marrobi (approved):
LGTM

martinpeck:
What is this PR waiting on? Just another reviewer?

I think this is fine (and I'll review in more detail in a bit) but I'm not 100% sure we needed all of these to be documented. Or, put another way, I think some of the docs aren't adding much. 

For some commands it looks useful, but for the primary targets we already have descriptions that get shown when you type `make` ...

```console
➜ AzureTRE (lishakur/makefile-doc) ✗ make
all                       🚀 Provision all the application resources from beginning to end
tre-deploy                🚀 Provision TRE using existing images
images                    📦 Build and push all images
help                      💬 This help message :)
tre-start                 ⏩ Start the TRE Service
tre-stop                  ⛔ Stop the TRE Service
tre-destroy               🧨 Destroy the TRE Service
lint                      🧹 Lint all files
test-e2e-smoke            🧪 Run E2E smoke tests
test-e2e-extended         🧪 Run E2E extended tests
test-e2e-extended-aad     🧪 Run E2E extended AAD tests
test-e2e-shared-services  🧪 Run E2E shared service tests
test-e2e-custom           🧪 Run E2E tests with custom selector (SELECTOR=)
setup-local-debugging     🛠️ Setup local debugging
auth                      🔐 Create the necessary Azure Active Directory assets
db-migrate                🗄️ Run database migrations
```

Some of the targets in `Makefile` aren't listed here either because we don't expect them to be used directly/they're internal, but this PR adds comments to them. For example...

```
# Command: build-and-push-airlock-processor
# Description: Build and push Airlock Processor image
# Example: make build-and-push-airlock-processor
build-and-push-airlock-processor: build-airlock-processor push-airlock-processor
```

I'm not sure this comment adds much...but, equally, it's fine.

Makefile
```diff
@@ -14,32 +14,68 @@ E2E_TESTS_NUMBER_PROCESSES_DEFAULT=4  # can be overridden in e2e_tests/.env
...
+# Description: Migrate the firewall state from the core deployment to the shared services deployment.
+# This is a one-time operation and should only be run when you are moving from the core deployment to the shared services deployment.
+# This command will remove the firewall state from the core deployment and import it into the shared services deployment.
+# Example: make migrate-firewall-state
 migrate-firewall-state: prepare-tf-state
 
+# Command: bootstrap
+# Description: Bootstrap Terraform
+# Example: make bootstrap
 bootstrap:
```
martinpeck, line 63: should this have the `##` comment so that it appears when you run `make help`?

Makefile
```diff
@@ -14,32 +14,68 @@ E2E_TESTS_NUMBER_PROCESSES_DEFAULT=4  # can be overridden in e2e_tests/.env
...
 	$(call target_title, "Bootstrap Terraform") \
 	&& . ${MAKEFILE_DIR}/devops/scripts/check_dependencies.sh nodocker,env \
 	&& cd ${MAKEFILE_DIR}/devops/terraform && ./bootstrap.sh
 
+# Command: mgmt-deploy
+# Description: Deploy management infrastructure. This will create the management resource group (named <mgmt_resource_group_name> from the config.yaml file) with the necessary resources such as Azure Container Registry, Storage Account for the tfstate and KV for Encryption Keys if enabled.
+# Example: make mgmt-deploy
 mgmt-deploy:
```
martinpeck, line 71: Do we expect this to be run from the command line directly?
If so, we should add the `##` comment so that it appears when you run `make help`

Makefile
```diff
@@ -14,32 +14,68 @@ E2E_TESTS_NUMBER_PROCESSES_DEFAULT=4  # can be overridden in e2e_tests/.env
...
 	$(call target_title, "Deploying management infrastructure") \
 	&& . ${MAKEFILE_DIR}/devops/scripts/check_dependencies.sh nodocker,env \
 	&& cd ${MAKEFILE_DIR}/devops/terraform && ./deploy.sh
 
+# Command: mgmt-destroy
+# Description: Destroy management infrastructure. This will destroy the management resource group with the resources in it.
+# Example: make mgmt-destroy
 mgmt-destroy:
```
martinpeck, line 79: Do we expect this to be run from the command line directly?
If so, we should add the `##` comment so that it appears when you run `make help`

Makefile
```diff
@@ -65,12 +101,21 @@ $(call target_title, "Building $(1) Image") \
 	--cache-from ${FULL_IMAGE_NAME_PREFIX}/$(1):$${__version__} $${ci_cache:-} -f $(3) $(4)
 endef
 
+# Command: build-api-image
+# Description: Build API image using the build_image method.
+# Example: make build-api-image
```
martinpeck, line 106: Rather than repeat it lots, if we're giving examples of how to run it directly I assume we see this as something that can be run from the command line.

If that's the case, we should add the `##` comment so that it appears when you run `make help`

If it's not (and I think there are examples here that won't work if run directly) then we should not give examples.

martinpeck:
As per PR comments...

If it is valid to run the command directly from the command line, we should ensure that this command appears when `make help` is run.

If is it not valid (not expected) that the command will ever be run directly from the command line (l.e. the target is actually an internal thing, and requires other make targets to run before it) then we should not give any examples of how to run it directly.

ashis-kar91 (approved):

LizaShak:
Skipping tests because this is documentation only

PR #4447, opened 2025-03-26
Fix use of deprecated argument in jq

### LINKED ISSUE #4446: Use of deprecated argument in jq tool
The script file [`register_bundle_with_api.sh`](devops/scripts/register_bundle_with_api.sh) uses the [__jq__][jq] command line argument `--argfile`. This has been documented as _"Deprecated: Do not use"_ since 2015, and in 2023 with version 1.7 of jq, it was removed completely.

If the version of jq in the dev container is updated, the following error will occur when building a bundle.

```
jq: Unknown option --argfile
Use jq --help for help with command-line options,
or see the jq manpage, or online docs  at https://jqlang.github.io/jq
make[1]: *** [Makefile:284: bundle-register] Error 2
```

The replacement argument is `--slurpfile`. It works slightly differently, but simple to replace.

[jq]: https://jqlang.org/ "jq is a lightweight and flexible command-line JSON processor"

askew:
# Resolves #4446 

## What is being addressed

The default base image used by the devcontainer happens to include jq version 1.6. If this changes to upgrade jq to 1.7 or later the bug described in #4446 will occur.

## How is this addressed

There is only one script file this is used and it's a trivial change to fix it.

### DIFF
--- a/devops/scripts/register_bundle_with_api.sh
+++ b/devops/scripts/register_bundle_with_api.sh
@@ -99,7 +99,7 @@ fi
 acr_domain_suffix=$(az cloud show --query suffixes.acrLoginServerEndpoint --output tsv)
 explain_json=$(porter explain --reference "${acr_name}${acr_domain_suffix}"/"$(yq eval '.name' porter.yaml)":v"$(yq eval '.version' porter.yaml)" -o json)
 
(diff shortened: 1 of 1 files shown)


marrobi (approved):
LGTM, can you do the changelog, will run the tests too.

martinpeck (approved):
Should we also pin the version of `jq` that we're using in the container?
I assume we're not pinning it right now, and hence you've hit this.

That said...LGTM.

Issue #1073, opened 2021-11-12
Isolate workspaces into individual subscriptions, user resources into user resource groups, and service resources into a core resource group


cricex:
**Describe the solution you'd like**
As an Information Security Officer
I want each workspace deployed into their own spoke subscription, each user's resources deployed into their own resource group within the workspace's subscription, and service resources deployed into a service resource group within the workspace's subscription
So that we may improve security isolation, billing isolation, and overcome subscription resource limitations

![image](https://user-images.githubusercontent.com/7026048/141486727-4ed33e3c-ecaa-4eb9-b976-08fb8435f039.png)

marrobi:
Thanks, we would be limited to 800 resource groups per subscription. Is that an issue? 

The separate resource groups for services and user resources can be handled in the respective templates.  So no changes to the TRE APIs, or resources processor are required.

The subscription piece is doable, but work required.

cricex:
Is 800 a TRE or ARM limitation?  The documentation states that there is a limitation of 980 resource groups per subscription. (https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/azure-subscription-service-limits#subscription-limits)

Regardless, I believe 800 or 980 resource groups (and therefore 800 or 980 individual researchers) is sufficiently high limitation for an individual workspace.

Danny-Cooke:
This feature has been requested by a third party so will be looking at how to deliver this

marrobi:
@SvenAelterman @Danny-Cooke is it worth us getting on a call to discuss requirements? Then a design can be posted here and discussed?

SvenAelterman:
We can. Unfortunately, our project has not progressed.

marrobi:
This branch shows how this might be done - haven't tested across multiple subscriptions.

https://github.com/marrobi/AzureTRE/tree/marrobi/workspace-different-subscription

Each workspace service would have to use the parent `workspace_subscription_id` when configuring the provider.

If someone wants to pick it up happy to help.

TonyWildish-BH:
that looks interesting, thanks Marcus. We may be able to pick this up in the near future, we're mapping out our future strategy for scaling up, hence all the tickets recently.

TonyWildish-BH:
> I am working with a customer who also needs a solution for 300+ workspaces. The customer is working with a partner on implementation.
> 
> I would propose we all work together on the solution? LMK if you'd be open to that.
> 
[...]

PR #4451, opened 2025-03-27
Add ability to pass values to install stage on pipleine.

### LINKED ISSUE #4450: Pipeline substitution only works when values are marked as updatable so not for install.
When passing values into a resource in a pipeline, the code only allows input for values marked as updatable. This means if need to pass values into the app at install it does not work.

For example `subscription_id` into a workspace, needed by #1073 cc @askew 

And `log_analytics_workspace_name` as needed by #4341 

PR to come.

marrobi:
Fixes #4450

This allows values to be validated when the action is `install`.

It feels wrong using patch methods for install rather than update, but the database operation is a patch...

It is needed to enable:

For example subscription_id into a workspace, needed by https://github.com/microsoft/AzureTRE/issues/1073 cc @askew

And log_analytics_workspace_name as needed by https://github.com/microsoft/AzureTRE/pull/4341

https://github.com/microsoft/AzureTRE/blob/835891870427718eeac92de154d8ca9121439c07/templates/workspace_services/azureml/template_schema.json#L59-L75


marrobi:
@tamirkamara thoughts on this? The resource is created in the DB before the pipeline runs, hence is a patch, but when the action is install, we need to enable all fields to be updated as the resource is yet to e deployed.

api_app/tests_ma/test_db/test_repositories/test_resource_repository.py
```diff
@@ -380,7 +380,7 @@ def test_validate_patch_with_good_fields_passes(template_repo, resource_repo):
 
     # check it's valid when updating a single updateable prop
     patch = ResourcePatch(isEnabled=True, properties={'vm_size': 'large'})
-    resource_repo.validate_patch(patch, template_repo, template)
+    resource_repo.validate_patch(patch, template_repo, template, "update")
```
martinpeck, line 383: update, or upgrade?

marrobi: Good catch. The code only checks for install, so the test passed anyway.

 ```suggestion
    resource_repo.validate_patch(patch, template_repo, template, "upgrade")
```

File: api_app/service_bus/helpers.py [hunk 2]
```diff
@@ -83,6 +83,8 @@ async def update_resource_for_step(operation_step: OperationStep, resource_repo:
     for step in parent_template_pipeline_dict[primary_action]:
         if step["stepId"] == operation_step.templateStepId:
             template_step = parse_obj_as(PipelineStep, step)
+            if (template_step.resourceAction is None):
```
martinpeck, line 86: What does this do?
Why didn't we already have this?

api_app/db/repositories/resources.py
```diff
@@ -177,7 +177,7 @@ def validate_patch(self, resource_patch: ResourcePatch, resource_template_repo:
         update_template["required"] = []
         update_template["properties"] = {}
         for prop_name, prop in enriched_template["properties"].items():
-            if ("updateable" in prop.keys() and prop["updateable"] is True):
+            if (resource_action == "install" or ("updateable" in prop.keys() and prop["updateable"] is True)):
```
martinpeck, line 180: Passing comment: It's annoying that this talks about "updateable" when, I assume, this is checking for upgrades?

TBH...what is this code saying?
Your change is checking "the action is install", but what was the existing code doing, and is it a good idea to combine your code with this code?
Would it be cleaner (e.g. if other resource actions are supported later) to have a `elif` even if the code block associated is the same?

api_app/service_bus/helpers.py
marrobi: In the example above:

```
{ 
   "stepId": "12ba0dad-ea6c-4d0d-9255-d316212f5ffa", 
   "stepTitle": "Upgrade to ensure aware of address space", 
   "resourceType": "workspace", 
   "resourceAction": "upgrade", 
   "properties": [] 
 }, 
 { 
   "stepId": "main", 
   "properties": [ 
     { 
       "name": "log_analytics_workspace_name", 
       "type": "string", 
       "value": "{{ resource.parent.properties.log_analytics_workspace_name }}" 
     } 
   ] 
 }, 
 ```
 
The "main" step does not have an action defined, this defaults it to the action that is being taken on the resource the main step is for. As we want to enable none updatable fields to be updated on install, we need to know what action is being carried out on the main step. Which will always be the action on the primary resource. That make sense?

File: api_app/db/repositories/resources.py (hunk 3 above)
Comment (marrobi) [reply api_app/db/repositories/resources.py:180]: The existing code only allows properties marked as updateable to be updated in a pipeline. This PR adds the ability for them to be set if the action is install - be they marked updatable or not. Not sure an `elif` helps much, as it just repeats the next line that does the assignment.

File: api_app/service_bus/helpers.py (hunk 2 above)
Comment (martinpeck) [reply api_app/service_bus/helpers.py:86]: > That make sense? 

Not 100%. So, what will this value be (maybe give me an example) and in which part of this PR does this value then get used?

File: api_app/db/repositories/resources.py (hunk 3 above)
Comment (martinpeck) [reply api_app/db/repositories/resources.py:180]: It helps make it clear there are two possibile situations:

1. The resource action is "install"
2 Marker as updateable

I agree...logically this one-liner does the same thing, in fewer lines of code. I'm just saying that comparing two very different things (both which use different concepts) wasn't super clear to me. 

Also, I thought there were plans to extend this to other resource actions in the future. Is that not the case? If it is, then it makes it more obvious where we'd add those additional clauses.

It's not a big thing, but combination of an `if`, an `or`, and an `and` within the brackets, along with testing two different things, made this one-liner look messy.

File: api_app/service_bus/helpers.py (hunk 2 above)
Comment (marrobi) [reply api_app/service_bus/helpers.py:86]: The value set to the parent action: https://github.com/microsoft/AzureTRE/blob/835891870427718eeac92de154d8ca9121439c07/templates/workspace_services/azureml/template_schema.json#L58

I could possible set it to:
`if (template_step.resourceAction is None) and step["stepId"] == "main":`

But not sure that adds anything, and there is a potential scenario where we might want to do install on other steps.

An alternative would be to add `  "resourceAction": "install",`  to every main step in the repository, but this can be inferred if the resourceAction is not defined, it is the same as the primary action.
```
 { 
   "stepId": "main",
   "resourceAction": "install", 
   "properties": [ 
     { 
       "name": "log_analytics_workspace_name", 
       "type": "string", 
       "value": "{{ resource.parent.properties.log_analytics_workspace_name }}" 
     } 
   ] 
 }, 
 ```
It then gets used in the piece you have commented on further down.

api_app/db/repositories/resources.py
marrobi: Happy to change if think its clearer.

marrobi: For other resource actions, it needs to adhere to it being updatable or not. The only exception would be the first time a resource is deployed - hence `install`. I might be missing something, but don't think so.

martinpeck: Also, is the second clause basically doing...

```python
prop.get("updateable", False)
```

...?

api_app/_version.py
```diff
@@ -1 +1 @@
-__version__ = "0.23.0"
+__version__ = "0.23.1"
```
tamirkamara, line 1: Remember to bump this to 0.24 before merge.

api_app/service_bus/helpers.py
martinpeck: And is parent action always "install"?
Could the template be reordered so that "uninstall" was first?

api_app/db/repositories/resources.py
marrobi: Look that way :)

api_app/db/repositories/resources.py
```diff
@@ -100,7 +100,7 @@ async def validate_input_against_template(self, template_name: str, resource_inp
 
         return parse_obj_as(ResourceTemplate, template)
 
-    async def patch_resource(self, resource: Resource, resource_patch: ResourcePatch, resource_template: ResourceTemplate, etag: str, resource_template_repo: ResourceTemplateRepository, resource_history_repo: ResourceHistoryRepository, user: User, force_version_update: bool = False) -> Tuple[Resource, ResourceTemplate]:
+    async def patch_resource(self, resource: Resource, resource_patch: ResourcePatch, resource_template: ResourceTemplate, etag: str, resource_template_repo: ResourceTemplateRepository, resource_history_repo: ResourceHistoryRepository, user: User, force_version_update: bool = False, resource_action: str = 'upgrade') -> Tuple[Resource, ResourceTemplate]:
```
tamirkamara, line 103: Are there cases where resource_action won't be passed?
Should we just always pass it?

api_app/service_bus/helpers.py
marrobi: Yes, but then this would be set to "uninstall"? And the property adhere to not being updatable.

Confused what you are asking?

api_app/db/repositories/resources.py
tamirkamara: 1. Aren't there cases where a template has a default we don't want to allow changing in an install?
2. I agree the condition is somewhat confusing. I remember a discussion once to allow setting a property just once, maybe this is the case for that (too)?

api_app/service_bus/helpers.py
tamirkamara: I vote for the explicit way of having resourceAction specified in the json

api_app/db/repositories/resources.py
martinpeck:
```suggestion
            if (resource_action == "install" or prop.get("updateable", False) is True):
```

I assume that the `is True` is there because we're not interested in it being "truthy". Instead, we want it to be a boolean type, and specifically `True`. Otherwise, we could drop the `is True`.

api_app/service_bus/helpers.py
marrobi: The main step is always going to be the primary action though? Enabling it to be specified just confuses things, as when would main be upgrade in an install stage? You can always override it, this is in effect just a default?

api_app/db/repositories/resources.py
martinpeck: My suggestion maintains the specific semantics around `is True`. 
Yours would start to allow truthy values like "1" to be seen as `True`.

marrobi: This make you happier?

```suggestion
           if resource_action == "install" or prop.get("updateable", False) is True:
```

api_app/service_bus/helpers.py
martinpeck: > Confused what you are asking?

I'm not 100% familiar with this code, so I was trying to work out "what value does this get set to?".
The change you've made makes some assumptions. Specifically that the main step is "install", and because the string "install" is the same as the resource action string "install" that this will trigger the code changes elsewhere.

All of those assumptions are fine, but they're not obvious (to me, at least), so I wanted to understand what might break.

e.g. what happens if someone reorders the steps in a template schema, or what happens if we add other steps into the template schema in the future

e.g. The fact that the template step is called "install" (with that casing), and the resource action is called "install" means this works, but what if a future PR refactors the names of the steps?

If this code checked the parent action, checked "is it the install action", and then set resourceAction to "install", at least the logic, and the assumptions, would all be in that one chunk of code, and future changes wouldn't trip things up.

api_app/_version.py
marrobi:
```suggestion
__version__ = "0.24.0"
```

api_app/service_bus/helpers.py
marrobi: So:

```suggestion
            if (template_step.resourceAction is None and primary_action == "install"):
```

?

martinpeck: Yes....that's clearer, but also the next line can explicitly then set `template_step.resourceAction` to "install", rather than setting it to "primary_action" which just happens to be "install".

Resource actions and steps (primary actions) aren't, semantically, the same, right?

marrobi: Hmm, its the `install` of the bundle, guess one is the "primary resource action", and one is the "step action".  I don't mind, as long as the code works... 

I can change it.

stevehaigh:
When merged we'll need a minor update to the firewall-rules.md doc, as this mentions the pipeline steps.

api_app/service_bus/helpers.py
marrobi: Done.

api_app/db/repositories/resources.py
marrobi: Done

api_app/db/repositories/resources.py
marrobi: Done

api_app/db/repositories/resources.py
marrobi: Done

api_app/_version.py
marrobi: Done

martinpeck (approved):

tamirkamara (approved):

marrobi:
/test-force-approve 

Passed: https://github.com/microsoft/AzureTRE/actions/runs/14702963779

PR #4454, opened 2025-03-31
Skip removing keyvault rule when resource group is deleting

### LINKED ISSUE #4449: Resource Firewall rules cause cleanup workflow to fail
**Describe the bug**

The functionality of adding and removing access to specific resources (like KeyVault) to the current IP is causing CI cleanup failure.
Every single environment that is being cleaned up by the cleanup workflow will cause a failure.
The reason is that the workflow (via the script) doesn't wait for the resource group deletion to complete and then the trap to remove the added firewall rule for the KeyVault is firing, but it's not allowed as the resource group is in deleting state.

An example workflow run: https://github.com/microsoft/AzureTRE/actions/runs/14100752072/job/39496361256

(Unfortunately I get an email every time it happens)

tamirkamara:
# Resolves #4449 

## What is being addressed

When the CI cleans up (deletes) an environment is uses the "no-wait" option to delete resource groups. At the end of the script there's a trap to remove the keyvault IP access rule which will fail in this situation.

## How is this addressed

- Skip removing the rule if the resource group is in deleting state.


devops/scripts/kv_add_network_exception.sh
```diff
@@ -123,6 +128,15 @@ function does_kv_exist() {
   return 0
 }
 
+function is_core_rg_deleting() {
+  if [[ "$(az group show --name "${core_tre_rg:-}" --query "properties.provisioningState" -o tsv)" == "Deleting" ]]; then
```
jonnyry, line 132: Are there situations where ${core_tre_rg:-} is not set (since its suffixed with `:-`)?  If so should there be an if [[ -n "${core_tre_rg:-}" ]] before the az group show call?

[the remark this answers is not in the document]
tamirkamara: I don't see this whole procedure applies to the MGMT Keyvault. At the time I commented on this but seems it was left out.

tamirkamara: I was also thinking about it. I don't think it's possible that var will be empty, but then again you had put this in place: https://github.com/microsoft/AzureTRE/blob/main/devops/scripts/kv_add_network_exception.sh#L89.
Have you seen that @jonnyry ?

jonnyry: oh in fact it looks like it's only set in `destroy_env_no_terraform.sh` - so will need to find another way to determine the core RG name:

image

what about following the pattern used in `get_kv_name` - since we know that will resolve in all contexts:

https://github.com/microsoft/AzureTRE/blob/cba92f16d3d386ac2121cc45514023290f044033/devops/scripts/kv_add_network_exception.sh#L84

tamirkamara: updated the approach. please have a look.

jonnyry: @tamirkamara Nice approach, LGTM

jonnyry (approved):

LizaShak (approved):

PR #4455, opened 2025-03-31
Enable Workspace to deploy to separate subscription

### LINKED ISSUE #1073: Isolate workspaces into individual subscriptions, user resources into user resource groups, and service resources into a core resource group

askew:
# Partly resolves #1073

## What is being addressed

To enable workspaces to be deployed to separate subscriptions to the TRE core components, the workspace bundle needs take the workspace subscription id as a parameter and the Terraform needs to manage two sets of AzureRM providers settings, one for the workspace, and one core that it links to.

This PR adds these capabilities to the workspace bundle and Terraform configuration. __It does__ ___NOT___ resolve #1073 on it's own. It requires further changes to make the workspace subscription id available to the service and resource bundles.

## How is this addressed

- A subscription id field is available as an optional setting when creating a new workspace.
- The workspace subscription is passed to the Porter bundle, and the contained Terraform.
- The Terraform uses a second provider setting for the workspace subscription.


templates/workspace_services/guacamole/terraform/providers.tf
```diff
@@ -6,7 +6,7 @@ terraform {
     }
     azapi = {
       source  = "Azure/azapi"
-      version = "=1.15.0"
+      version = "= 2.3.0"
```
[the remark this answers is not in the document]
askew: The consistent approach in this project is to have a space after the `=`

api_app/resources/strings.py
```diff
@@ -143,6 +143,7 @@
 WORKSPACE_SERVICES_NEED_TO_BE_DELETED_BEFORE_WORKSPACE = "All workspace services need to be deleted before you can delete the workspace"
 WORKSPACE_DOES_NOT_HAVE_ADDRESS_SPACES_PROPERTY = "Workspace does not have address_spaces property"
 WORKSPACE_TEMPLATE_VERSION_EXISTS = "A template with this version already exists"
+WORKSPACE_SERVICE_TEMPATE_DOES_NOT_SUPPORT_SUBSCRIPTION = "The workspace service template does not support multi-subscription deployment."
```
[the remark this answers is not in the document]
askew: Good spot, Copilot, but actually this constant is not used so will remove.

templates/workspace_services/guacamole/terraform/variables.tf
```diff
@@ -70,3 +70,8 @@ variable "arm_environment" {
   type        = string
   description = "The ARM cloud environment"
 }
+variable "workspace_subscription_id" {
+  type        = string
+  description = "The id of the Azure subscription the workspace is deployed to"
+  default     = ""
```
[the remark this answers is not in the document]
askew: From [coalesce Function](https://developer.hashicorp.com/terraform/language/functions/coalesce):
 > `coalesce` takes any number of arguments and returns the first one that isn't null or an empty string.

Therefore "" will work, and needs to be consistent with an empty string being passed in.

marrobi:
@askew are they any docs for this PR? What do I need to do in the second subscription?

askew:
> @askew could do with something in the docs around the permissions needed to deploy to a second subscription.

@marrobi I've added some documentation to `docs/tre-admins/setup-instructions/installing-base-workspace.md`

api_app/services/cost_service.py
```diff
@@ -93,12 +93,30 @@ class CostService:
     def __init__(self) -> None:
         self.scope = "/subscriptions/{}".format(config.SUBSCRIPTION_ID)
         self.client = CostManagementClient(credential=credentials.get_credential())
-        self.resource_client = ResourceManagementClient(credentials.get_credential(),
-                                                        config.SUBSCRIPTION_ID,
-                                                        base_url=config.RESOURCE_MANAGER_ENDPOINT,
-                                                        credential_scopes=config.CREDENTIAL_SCOPES)
+        self.__resource_clients = {
```
LizaShak: Maybe we can  use the new get_resource_management_client method here and pass config.SUBSCRIPTION_ID to it?

LizaShak (commented):
Hi, I haven’t completed the full review yet, but I wanted to pause and ask a couple of questions before continuing:

As I understand it, not all workspace services support installation in a separate subscription — is that correct? What happens if a user tries to install a workspace service that doesn’t support it? Do we have a plan on supporting it in all WS ? If not maybe worth adding a documentation saying what is supported in this case

How would this be handled for services like Databricks, which depend on both a workspace component and a shared service (e.g., Databricks Auth)? Is cross-subscription deployment supported for these kinds of dependencies as well? I know it is not part of this PR, I just have a feeling this is where there might be a problem. Again if shouldn't be supported then maybe worth add a documentation on that

marrobi:
@LizaShak VNet peering work across subscriptions, so the shared services and DNS should still work.

@askew have you got time to comment on the other bits, know it's been a while, but keen to get this merged.

api_app/services/cost_service.py
askew: I think you are right, I'll look to update this.

marrobi: @askew would be great if you can, really keen to get this merged. @LizaShak if you can have another look once done would be appreciated.

LizaShak: hey, sorry for getting back to this only now, I was OOF, I was about to approve this PR, just wanted to check first if you decided to keep it they way it is or use the get_resource_management_client method ?

askew: @LizaShak, sorry I've been busy or away too. I've just pushed the change you suggested.

askew: @LizaShak I've made you [suggested code change](https://github.com/microsoft/AzureTRE/pull/4455/commits/5becab83ecdaaaff5f239c187504bbaf1c0b3680)

PR #4460, opened 2025-04-01
Pass workspace subscription id to services

### LINKED ISSUE #1073: Isolate workspaces into individual subscriptions, user resources into user resource groups, and service resources into a core resource group

askew:
# Partly Resolves #1073

## What is being addressed

PR #4455 enables the deployment of a base workspace to a separate subscription. Once this is done, any workspace service or resource now needs to know the Azure subscription id that the resources need deploying to.

## How is this addressed

- The workspace service create API now checks whether the workspace has a subscription id set. If it does it then checks to see if the service bundle has been updated to work with separate subscriptions. If it has, it will have a hidden parameter defined in the bundle template.
- If the service bundle does not support a separate subscription, the API returns an error as the bundle will not work.
- If the bundle is okay, the workspace subscription id is added to the bundle parameters so that it picks it up when deployed. 

This PR has the changes to the API plus the required changes to the Guacamole workspace service bundle, as an example.


templates/workspace_services/guacamole/template_schema.json
```diff
@@ -71,6 +71,12 @@
       "title": "Expose externally",
       "description": "Is the Guacamole service exposed outside of the vnet",
       "default": true
+    },
+    "workspace_subscription_id": {
```
marrobi, line 75: Know we discussed this, not sure it's needed as

https://github.com/microsoft/AzureTRE/blob/839548ccc7ed0d445d0324fa87f602aa45f6ebfe/devops/scripts/register_bundle_with_api.sh#L102-L103

uses porter explain t get all parameters and merges with template_schema.json. Worth a test.

tamirkamara (changes requested):
Blocking this for now as the method to pass the sub_id might change if/when this PR is merged: https://github.com/microsoft/AzureTRE/pull/4451

askew:
There are two options for passing the workspace subscription id to service/resource bundles.

1. Have the API add it by default, as per this PR
2. Use the pipeline definition in the template schema to get the reference from the parent, which would be enabled in #4451

In this case I feel option 1 is better for a few reasons: Firstly, every service or user resource will is deployed to an Azure subscription, so regardless of whether it is the same subscription as the core, or not, it is always relevant; Secondly, putting the reference in the pipeline template would require a lot of duplicate code in every bundle that just adds maintenance overhead.

PR #4470, opened 2025-04-08
Add dependency between Airlock processor and storage account private endpoint

### LINKED ISSUE #4433: Airlock Container not always created succesfully
When creating an airlock request the container is not always successfully created. This has been noticed before, but now seen in e2e tests:

https://github.com/microsoft/AzureTRE/actions/runs/13975180861/job/39177589907
```
INFO     e2e_tests.test_airlock:test_airlock.py:53 try #45 to upload a blob to container
 ....
INFO     e2e_tests.test_airlock:test_airlock.py:63 sleeping for 30 sec until container would be created
```

SiobhanBaynes:
Add dependency to avoid race condition between Airlock function app and storage account private endpoint

# Resolves #4433 

## What is being addressed

This PR adds a dependency between the private endpoint on the storage account used by the Airlock processor function app and the function app itself. This is to try and fix the intermittent problem of storage account containers not being created successfully for Airlock requests due to the function app not starting successfully.

## How is this addressed

- Add dependency between the private endpoint on the storage account used by the Airlock processor function app and the function app itself.
- Bump core version
- Update CHANGELOG.md

### DIFF
--- a/ui/app/package.json
+++ b/ui/app/package.json
@@ -1,6 +1,6 @@
 {
   "name": "tre-ui",
-  "version": "0.8.0",
+  "version": "0.8.1",
   "private": true,
   "type": "module",
   "dependencies": {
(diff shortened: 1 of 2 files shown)


marrobi (approved):
LGTM

PR #4360, opened 2025-02-11
TRE core should use private endpoint to access TRE management storage account #4353

ashis-kar91:
# Resolves #4353

## What is being addressed

- Access to TRE _management_ storage account was allowed from all networks, which is a security risk. TRE _core_ used to access the _management_ storage account via public network.
- TRE _management_ storage account should use a _private endpoint_ to restrict public network access and ensure data remains within a private network.
- Local TRE deployments will fail by making the storage account completely private. Hence _Allow public access from specific IP addresses_ during deployment.

## How is this addressed

The changes include the addition of a private endpoint for the TRE management storage account from the resource processor subnet, updates to network rules for the storage account to deny public access, and the introduction of a new script to handle adding and removing network exceptions for the management storage account during deployments.

### Security Enhancements:
* `core/terraform/resource_processor/vmss_porter/main.tf`: Add a new resource `azurerm_private_endpoint` for the TRE management storage account to use a private endpoint. This ensures secure access to the management storage account from TRE core resource processor subnet.
* Set default action to "Deny" and bypass to "AzureServices" for the TRE management storage account (`devops/terraform/bootstrap.sh`) [[1]](diffhunk://#diff-b8141bce138b6f8ec9bbe6df18ed9c586e940032a8ca17f6151dc943d4a9b0acR23-R24) [[2]](diffhunk://#diff-b8141bce138b6f8ec9bbe6df18ed9c586e940032a8ca17f6151dc943d4a9b0acR33-R35).
* Add network rules to deny public access and allow only specific IPs for the TRE management storage account (`devops/terraform/main.tf`).

### Deployment Script Improvements:
* Add script `mgmtstorage_add_network_exception.sh` to handle adding and removing network exceptions for the management storage account (`devops/scripts/mgmtstorage_add_network_exception.sh`).
[...]

PR #4495, opened 2025-04-18
Enable vnet exception for core key vault

### LINKED ISSUE #4463: Enable Core key vault access from private runners
## Description
As part of Issue #4250 and PR #4260 the public access to the key vault has been disabled and IP exception is added during the deployment. 
However, in certain scenarios while deploying TRE using CICD workflow from private runners, this approach doesn't work and the deployment fails, as those runners don't have public IPs.

## Proposals

1. Just in time enablement of public network access for the key vault during deployment, instead of public IP exception. This is a simpler solution to implement and consistent with Management Storage account and upcoming ACR network access behaviour. Management Storage account PR for reference: #4360 

2. Have a config flag to switch between allowlisting public IP and Vnet. This flag will be used in the script to enable Vnet exception for the private runners as shown in the image below. And if the flag is not set, the existing behaviour of public IP exception will be applied. 

Image

## Acceptance criteria

- [ ] Should be able to deploy TRE from private runners using CICD
- [ ] Should be able to deploy TRE from local dev machine
- [ ] Public access to the key vault should be in disabled state at the end of deployment

ashis-kar91:
# Resolves #4463

## What is being addressed

The public access to the key vault has been disabled and IP exception is added during the deployment.
However, in certain scenarios while deploying TRE using CICD workflow from private runners, this approach doesn't work and the deployment fails, as those runners don't have public IPs.

## How is this addressed
- This pull request introduces support for configuring an allowed subnet ID to enable virtual network (vnet) exceptions for accessing Core Key Vault from Github runners. 
- If allowed subnet ID is not provided in the config, it defaults to just in time public key vault access during deployment. This is to make it consistent with other private resource access patterns at the moment like management Storage account and ACR.


.github/actions/devcontainer_run_command/action.yml
```diff
@@ -145,6 +145,10 @@ inputs:
     description: "The Azure Resource ID for an external key store to use for CMK"
     required: false
     default: ""
+  ALLOWED_SUBNET_ID:
```
macromania, line 148: This is mainly to enable private runner's accessing TRE. Maybe we can just name it as `PRIVATE_GITHUB_RUNNER_SUBNET_ID`?

core/terraform/keyvault.tf
```diff
@@ -26,7 +26,7 @@ resource "azurerm_key_vault" "kv" {
     when    = create
     command = <<EOT
 az keyvault update --name ${local.kv_name} --public-network-access ${local.kv_public_network_access_enabled ? "Enabled" : "Disabled"} --default-action ${local.kv_network_default_action} --bypass "${local.kv_network_bypass}" --output none
-az keyvault network-rule add --name ${local.kv_name} --ip-address ${local.myip} --output none
+${local.allowed_subnet_id != "" ? "az keyvault network-rule add --name ${local.kv_name} --subnet ${local.allowed_subnet_id} --output none" : ""}
```
macromania, line 29: Do we need to have following changes if we check this local value and run the network rule?

https://github.com/microsoft/AzureTRE/pull/4495/files#diff-611a410355d8c35e8fffdbbc6df3fdca81b72e86f79899ed36d4d5661ca1cfe8R27

.github/actions/devcontainer_run_command/action.yml
marrobi: Maybe `PRIVATE_RUNNER_SUBNET_ID` or `PRIVATE_AGENT_SUBNET_ID`, not everyone uses GitHub to deploy.

core/terraform/keyvault.tf
ashis-kar91: Yeah, this is still required because of the comment mentioned in the line 23 above the provisioner.
`# create provisioner required due to https://github.com/hashicorp/terraform-provider-azurerm/issues/18970`

.github/actions/devcontainer_run_command/action.yml
ashis-kar91: `PRIVATE_AGENT_SUBNET_ID` sounds fine for all scenarios. Will rename to this.

marrobi (approved):
LGTM. Thanks.

jonnyry (commented):
Hi @ashis-kar91 

PR looks good, I just have a few questions:

1. Have you run a full TRE deployment with private runners?  Does a similar change (adding subnet ID to their firewalls) need to be made to the management storage account & the container registry for the whole deployment to work via private runner?

2. I can see that you set the subnet ID in the terraform for the private runner, but then you are also setting it in the bash script - whats the reason for needing to set it here as well?

3. Some docs would be really helpful for people wanting to use private runners (I’d like to do this at some point!) - adding `PRIVATE_AGENT_SUBNET_ID` to the existing docs, plus a brief page on how to get set up and deploy the TRE with private runners would be useful :-)

Thanks

macromania:
On @ashis-kar91 's absense, I can provide some answers to your qeustions @jonnyry 😌 

>Have you run a full TRE deployment with private runners? Does a similar change (adding subnet ID to their firewalls) need to be made to the management storage account & the container registry for the whole deployment to work via private runner?

Yes, this has been tested in the current private environment our TRE instance running and it works 👍🏼 

> I can see that you set the subnet ID in the terraform for the private runner, but then you are also setting it in the bash script - whats the reason for needing to set it here as well?

From the script comment: This logic is needed to avoid error, if there is a change in subnet after the initial deployment 😅 

> Some docs would be really helpful for people wanting to use private runners (I’d like to set them up soon) - adding PRIVATE_AGENT_SUBNET_ID to the existing docs, plus a brief page on how to get set up and deploy the TRE with private runners would be useful :-)

I have something cooking up for this and hopefully can share in the near future. Stay tuned 😏

macromania (approved):
LGTM

jonnyry:
Hi @macromania 

> On @ashis-kar91 's absense, I can provide some answers to your qeustions @jonnyry 😌
> 
> > Have you run a full TRE deployment with private runners? Does a similar change (adding subnet ID to their firewalls) need to be made to the management storage account & the container registry for the whole deployment to work via private runner?
> 
> Yes, this has been tested in the current private environment our TRE instance running and it works 👍🏼

OK great.  Just to clarify - you didn't need to make any changes to the management storage account or container registry in order for the private runner deployment to work?

> > I can see that you set the subnet ID in the terraform for the private runner, but then you are also setting it in the bash script - whats the reason for needing to set it here as well?
> 
> From the script comment: This logic is needed to avoid error, if there is a change in subnet after the initial deployment 😅

Not sure I fully understand that - as in change of subnet ID?  If the subnet ID changed wouldn't the terraform pick it up and apply it?

> > Some docs would be really helpful for people wanting to use private runners (I’d like to set them up soon) - adding PRIVATE_AGENT_SUBNET_ID to the existing docs, plus a brief page on how to get set up and deploy the TRE with private runners would be useful :-)
> 
> I have something cooking up for this and hopefully can share in the near future. Stay tuned 😏

OK great, though could we add `PRIVATE_AGENT_SUBNET_ID` to the existing docs as part of this PR? - there's a couple of pages that list the environment variables from memory.  Welcome the expanded doc when its ready :-)

thanks

core/terraform/locals.tf
```diff
@@ -62,6 +62,7 @@ locals {
   # key vault variables
   kv_name                          = "kv-${var.tre_id}"
   kv_public_network_access_enabled = true
-  kv_network_default_action        = var.enable_local_debugging ? "Allow" : "Deny"
+  kv_network_default_action        = "Allow" # Exception to allow public access during deployment. This will be changed to "Deny" in kv_remove_network_exception.sh
```
marrobi, line 65: Not sure why this is being set to Deny? This coud cause problems where policy ensures this is set to Deny. 

The KeyVault is created with via ARM, and should have the exception enabled on the VNet for data plane operations from the agent?

ashis-kar91: Setting the default action to `Allow` here to handle the edge case when the key vault is created for the first time and it's a deployment from non-private device with no subnet ID exception set (Existing deployment flow). 
If I change this to `Deny`, the newly created key vault won't be accessible by other resources as the network exception script is already executed before this terraform execution, and the deployment will fail.
Also, the network access default action is being changed to `Deny` in trap script executed at the end of core deployment. So, the policy scan on infra should see this as `Deny` which is expected.

marrobi: @ashis-kar91 so how was it working previously as it was set to Deny?

Some scenarios have policies that block actual deployment of resources if breach policy, trying to think of a good way forward.

ashis-kar91: @marrobi Previously `Deny` was working because we were adding the IP exception in the terraform. But now as per our discussion in the issue #4463, we are making it public just in time and changing back to private to make the behavior consistent with Storage account and ACR. So, setting this `Allow` makes it public during creation and the trap script makes it private at the end. 
I can make it conditional though, to `Deny` if private_agent_subnet_id is provided, else `Allow`. What do you think?

jonnyry:
Hi @ashis-kar91 

One follow up question below:

> > > > Have you run a full TRE deployment with private runners? Does a similar change (adding subnet ID to their firewalls) need to be made to the management storage account & the container registry for the whole deployment to work via private runner?
> > > 
> > > 
> > > Yes, this has been tested in the current private environment our TRE instance running and it works 👍🏼
> > 
> > 
> > OK great. Just to clarify - you didn't need to make any changes to the management storage account or container registry in order for the private runner deployment to work?
> > Also are you deploying a version of main that includes @ashis-kar91 change #4360 to the mgmt storage account?
> 
> Since the current implementation to access Storage account and Container registry do not reply on ip_rules and instead uses just in time public access, we are not facing any problem for them in the private runners. But I will create separate issues to add the private runner Vnet exception implementation for them in future to make the behavior consistent.

OK but this PR changes the KV to use just in time public access to align with the SA and ACR. If private runners work with just in time public access for SA and ACR, why do they not work with KV? (and why do you need to add Subnet ID to make them work?)

---

[removed the other portions for brevity] - thanks for follow up on the other two question, subnet ID (makes sense) and documentation (thank you!).

ashis-kar91:
> Hi @ashis-kar91
> 
> One follow up question below:
> 
> > > > > Have you run a full TRE deployment with private runners? Does a similar change (adding subnet ID to their firewalls) need to be made to the management storage account & the container registry for the whole deployment to work via private runner?
> > > > 
> > > > 
> > > > Yes, this has been tested in the current private environment our TRE instance running and it works 👍🏼
> > > 
> > > 
> > > OK great. Just to clarify - you didn't need to make any changes to the management storage account or container registry in order for the private runner deployment to work?
> > > Also are you deploying a version of main that includes @ashis-kar91 change #4360 to the mgmt storage account?
> > 
> > 
> > Since the current implementation to access Storage account and Container registry do not reply on ip_rules and instead uses just in time public access, we are not facing any problem for them in the private runners. But I will create separate issues to add the private runner Vnet exception implementation for them in future to make the behavior consistent.
> 
> OK but this PR changes the KV to use just in time public access to align with the SA and ACR. If private runners work with just in time public access for SA and ACR, why do they not work with KV? (and why do you need to add Subnet ID to make them work?)

@jonnyry Private runners work fine with just in time public access for KV as well. Allowing Subnet ID is just additional measure to improve the security posture, where we won't be using the just in time public access while deploying from runners by allowlisting the subnet ID. This specific change to allow subnet ID exception should be implemented for Storage account and ACR as well in future.
I think we have discussed this in the issue #4463 and this was the suggested approach from @marrobi https://github.com/microsoft/AzureTRE/issues/4463#issuecomment-2774693604
> 
> [removed the other portions for brevity] - thanks for follow up on the other two question, subnet ID (makes sense) and documentation (thank you!).

core/terraform/locals.tf
marrobi: @ashis-kar91 do you plan on making this change? We just seem to be always on Allow now.

We want to have the scenario where if private agent is in use then it will always be set to Deny. Thanks,

ashis-kar91: Yes, I am working on making this change. Hence not resolved it yet.

jonnyry (approved):

PR #4503, opened 2025-04-28
Add more error mesage parsing

### LINKED ISSUE #4502: Some error messages not displayed in side panel
The new error side panel only looks for the text `Error:` in the returned text. In certain circumstances this could be returned as `Error message:`.

Image

The code should also check for `Error message:`

SiobhanBaynes:
# Resolves #4502 

## What is being addressed

Change the way we detect whether errors were returned. Instead of parsing the error message we now use the deployment status for enhanced reliability.

## How is this addressed

- Use deployment status or step status to determine if an error has occurred.
- Update CHANGELOG.md


ui/app/src/components/shared/ResourceHistoryListItem.tsx
```diff
@@ -17,7 +17,7 @@ export const ResourceHistoryListItem: React.FunctionComponent<
     },
   };
   const [isErrorPanelOpen, setIsErrorPanelOpen] = React.useState(false);
-  const isError = typeof props.val === "string" && (props.val.includes("Error:") || props.val.includes("error:"));
+  const isError = typeof props.val === "string" && (props.val.toLowerCase().includes("error:") || props.val.toLowerCase().includes("error message:"));
```
marrobi, line 20: @SiobhanBaynes why not just search for `error` ? Or if its not an error, what gets returned?

SiobhanBaynes: I was trying to be super cautious, and not flag something as an error that wasn't. If it's not an error it gets displayed as normal.

marrobi: I think the only other values it might have are:

https://github.com/siobhanbaynes/azuretre/blob/84b2235fe2631d5c6577a0c7bd4acdf53b5d37d7/resource_processor/vmss_porter/runner.py#L216

https://github.com/marrobi/AzureTRE/blob/ee4ef3605b64822e7f2a980667d6829d9188bd83/api_app/resources/strings.py#L245-L249

Actually, probably want to display the error if the step has  one of the failed states?


RESOURCE_STATUS_AWAITING_DEPLOYMENT = "awaiting_deployment"
RESOURCE_STATUS_DEPLOYING = "deploying"
RESOURCE_STATUS_DEPLOYED = "deployed"
**RESOURCE_STATUS_DEPLOYMENT_FAILED = "deployment_failed"**

RESOURCE_STATUS_AWAITING_DELETION = "awaiting_deletion"
RESOURCE_STATUS_DELETING = "deleting"
RESOURCE_STATUS_DELETED = "deleted"
**RESOURCE_STATUS_DELETING_FAILED = "deleting_failed"**

RESOURCE_STATUS_AWAITING_UPDATE = "awaiting_update"
RESOURCE_STATUS_UPDATING = "updating"
RESOURCE_STATUS_UPDATED = "updated"
**RESOURCE_STATUS_UPDATING_FAILED = "updating_failed"**

SiobhanBaynes: Fixed to used status field instead of parsing the error message

martinpeck: Sorry to unresolve, but from the comment is says you're using a status field, but it still looks like you're parsing an error message. Is there a merge issue, or have I misunderstood?

SiobhanBaynes: No, good spot. I missed one reference where it was still using the error message, fixing now.

SiobhanBaynes: It turns out the history list never displays the message or status fields, so I've restored this file to the previous commit.

ashis-kar91 (approved):

PR #4507, opened 2025-04-29
Letsencrypt.yml fails with “Invalid reference in variable validation”

### LINKED ISSUE #4506: letsencrypt.yml fails with “Invalid reference in variable validation”
Run letsencrypt.yml GitHub Action on current main, and the following error occurs:

```
│ Error: Invalid reference in variable validation
│ 
│   on variables.tf line 234, in variable "enable_cmk_encryption":
│  234:       (try(length(var.external_key_store_id), 0) > 0 && try(length(var.encryption_kv_name), 0) == 0) ||
│ 
│ The condition for variable "enable_cmk_encryption" can only refer to the
│ variable itself, using var.enable_cmk_encryption.
```

This is caused by a combination of two things:

1. The following terraform validation being added here:

https://github.com/microsoft/AzureTRE/blame/1ae927455a0b1a46bef589a061568a3280c40e69/core/terraform/variables.tf#L232

2. The terraform version being used in [letsencrypt.yml being older than 1.9](https://github.com/microsoft/AzureTRE/blob/1ae927455a0b1a46bef589a061568a3280c40e69/.github/workflows/lets_encrypt.yml#L39). Version 1.9 introduces the ability for validation blocks to reference other variables than the variable being validated:

(https://github.com/hashicorp/terraform/blob/v1.9/CHANGELOG.md#190-june-26-2024)

=> letsencrypt.yml requires a terraform version bump

jonnyry:
# Resolves #4506

## What is being addressed

- Bumps the terraform version in `letsencrypt.yml` to align with the devcontainer terraform version (1.9.8).
- Updated `CHANGELOG`

## Tests

I have tested this change manually on my local TRE deployment branch.  It is not covered by automated tests.

### DIFF
--- a/.github/workflows/lets_encrypt.yml
+++ b/.github/workflows/lets_encrypt.yml
@@ -36,7 +36,7 @@ jobs:
       - name: Install Terraform
         uses: hashicorp/setup-terraform@v3
         with:
-          terraform_version: 1.2.9
+          terraform_version: 1.9.8
           terraform_wrapper: false
 
       - name: Renew Certificates
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ ENHANCEMENTS:
 * Format the error message in the Operations panel for enhanced readability ([#4493](https://github.com/microsoft/AzureTRE/issues/4493))
 
 BUG FIXES:
+* Letsencrypt.yml fails with "Invalid reference in variable validation" ([#4506](https://github.com/microsoft/AzureTRE/4506))
 
 ## 0.22.0 (April 20, 2025)
 


marrobi (approved):
LGTM

martinpeck (approved):
@marrobi are you just looking for a 2nd "tick"? In which case, done.

It would be better to create a test (even if it's just some notes on *how* to test this) as it's hard for me/others to validate this PR.

tamirkamara:
@marrobi, @marrobi IIRC it's a combination of letsencrypt rate limits and non-persisted agent storage that means that even requests for the same domain count as a new certificate. I haven't look into updated limits though so things might have changed.

martinpeck:
Agreed....I also have a vague recollection of rate limits being an issue here. I've found an internal chat from 2022 that confirms this.

I guess `make letsencrypt` acts as a manually-run test for this, right?

Issue #2851, opened 2022-11-09
UI: customise logo icon, text and theming


jjgriff93:
As a customer deploying the TRE UI, I would like to give it my organisation's look and feel through the ability to change the logo image, text and basic theming. This should be easy to customise and clearly documented.

**Implementation**
- Change logo from using an icon to using an image from the assets folder which could be easily swapped
- Add new field to the UI config to accept logo text
- Add another for a theme object which can be used with Fluent UI theming to customise the colours used: https://github.com/microsoft/fluentui/wiki/How-to-apply-theme-to-Fluent-UI-React-components

marrobi:
Some functionality covered by https://github.com/microsoft/AzureTRE/pull/4522 , suggest new issues opened if additional customisation required.

PR #4522, opened 2025-05-05
Ability to customise UI header and footer text

### LINKED ISSUE #4201: Ability to customise UI header and footer text
As a TRE administrator I would like to be able to customise the header and footer text in the TRE portal via deployment variables so that I can display organisational and environment specific information.

jonnyry:
# Resolves #4201

## What is being addressed

Allows the TRE portal header and footer text to be customised per deployment, using the ~`UI_PRODUCT_NAME`~`UI_SITE_NAME` and `UI_FOOTER_TEXT` variables, either in `config.yaml`, or as GitHub environment variables to be used with the CI/CD Actions.

image

## Testing

Tested via CI/CD and by calling `make build-and-deploy-ui` from the command line


.github/actions/devcontainer_run_command/action.yml
```diff
@@ -145,6 +145,14 @@ inputs:
     description: "The Azure Resource ID for an external key store to use for CMK"
     required: false
     default: ""
+  UI_PRODUCT_NAME:
```
marrobi, line 148: Can we change this to `UI_SITE_NAME` throughout please? Product is a bit of an overloaded term.

marrobi (approved):
LGTM

.github/actions/devcontainer_run_command/action.yml
jonnyry: Changed

marrobi:
@jonnyry just realised, need to add these to environment variables in the docs.

jonnyry:
> @jonnyry just realised, need to add these to environment variables in the docs.

Added.

.github/actions/devcontainer_run_command/action.yml
tamirkamara, line 1: I wonder if this is the right thing to pass UI customization as environment variables... This feels more like a UI customization asset / file sort of a thing that can be committed to the deployment repo in a way to be used to build the vanilla UI to each one needs.

marrobi: I think its ok for this basic customisation (maybe even a theme and logo)?

For example if want to identify Dev, UAT, Prod then don't want the hassle of a custom UI to build and maintain, agree for more significant customisation.

jonnyry: Hi @tamirkamara if we get into extensive customisation, theming etc, fully agree we don't want another 50 variables entering the build pipeline (also we would quickly hit the 100 per env limit).  Could hold the customisation file as a single GitHub variable potentially, although depends on how many KB it ends up as.

Can I propose that we start with this PR for basic customisation now?  I'm more than happy if the variables migrate into a fully fledged UI customisation solution down the line.  Our primary use case is to have a visual indicator of the environment that a user is logged into.

ashis-kar91:
> Resolves #4201

NIT: Resolves #2851 ?

ashis-kar91 (approved):
LGTM

PR #4575, opened 2025-05-30
Allow UI_SITE_NAME and UI_FOOTER_TEXT to be passed to deploy_tre_reusable.yaml

jonnyry:
## What is being addressed

This is a small follow up to #4201 allowing the UI_SITE_NAME and UI_FOOTER_TEXT to be passed as parameters to `deploy_tre_reusable.yaml`.

The current behaviour of this script reads these variables directly from the Actions variables (`var.UI_SITE_NAME` and `vars.UI_FOOTER_TEXT`), whereas this change allows you to pass the variables in to the `deploy_tre_reusable.yaml` workflow as parameters - allowing per dynamically calculated variables to be passed.

For example, if you want to append a version number or a date stamp to your `UI_FOOTER_TEXT`, the calling workflow of `deploy_tre_reusable.yaml` can perform this string manipulation and pass the value through.  This was more difficult before as the variable values were set statically in the Actions variables (updating Actions variables from a workflow is painful as the default GH API token does not have rights to perform this).

If values for `UI_SITE_NAME` and `UI_FOOTER_TEXT` are not passed as parameters, the Actions variables values are used instead.

### DIFF
--- a/.github/workflows/deploy_tre_reusable.yml
+++ b/.github/workflows/deploy_tre_reusable.yml
@@ -35,6 +35,16 @@ on: # yamllint disable-line rule:truthy
         description: ""
         type: string
         required: true
+      UI_SITE_NAME:
+        description: Change the header text in the TRE portal
+        type: string
+        default: ""
+        required: false
+      UI_FOOTER_TEXT:
+        description: Change the footer text in the TRE portal
+        type: string
+        default: ""
+        required: false
     secrets:
       AAD_TENANT_ID:
         description: ""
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy_tre_reusable.yml
```diff
@@ -772,8 +782,8 @@ jobs:
           MGMT_STORAGE_ACCOUNT_NAME: ${{ secrets.MGMT_STORAGE_ACCOUNT_NAME }}
           SWAGGER_UI_CLIENT_ID: "${{ secrets.SWAGGER_UI_CLIENT_ID }}"
           USER_MANAGEMENT_ENABLED: ${{ vars.USER_MANAGEMENT_ENABLED }}
-          UI_SITE_NAME: "${{ vars.UI_SITE_NAME }}"
-          UI_FOOTER_TEXT: "${{ vars.UI_FOOTER_TEXT }}"
+          UI_SITE_NAME: "${{ inputs.UI_SITE_NAME || vars.UI_SITE_NAME }}"
```
marrobi, line 785: Where are the default values set? Looking above, are they `""`, rather than `"Azure TRE"` etc?

marrobi (approved):
LGTM

.github/workflows/deploy_tre_reusable.yml
jonnyry: Here:

https://github.com/microsoft/AzureTRE/blob/bb4ff4efb17e68bf2b76e7eac1a28093192fd880/ui/app/src/components/shared/Footer.tsx#L49
https://github.com/microsoft/AzureTRE/blob/bb4ff4efb17e68bf2b76e7eac1a28093192fd880/ui/app/src/components/shared/TopNav.tsx#L23

marrobi: Thanks.

PR #4578, opened 2025-06-05
Fix CostTag API calls to reduce 429 errors

### LINKED ISSUE #4577: Each CostTag in the UI calls the cost service meaning get a 429 back from Azure
**Describe the bug**
Frequently see 429 errors due to rate limit exceeded.

marrobi:
CostTag previously made individual API calls for each tag in the event of an error, leading to frequent 429 errors due to rate limits. This update consolidates the API calls to minimize the number of requests and improve overall performance.

Fixes #4577


james-annages:
> @james-annages are you able to create a separate issue for that?
> 
> The fix in this issue does seem to have resolved the issue with the UI making too many calls.

@marrobi Apploges, I will open a new issuse now.

ui/app/src/components/shared/CostsTag.tsx
```diff
@@ -33,21 +33,6 @@ export const CostsTag: React.FunctionComponent<CostsTagProps> = (
...
-          )
-        ).workspaceAuth.scopeId;
-        const r = await apiCall(
-          `${ApiEndpoint.Workspaces}/${props.resourceId}/${ApiEndpoint.Costs}`,
-          HttpMethod.Get,
-          scopeId,
-          undefined,
-          ResultType.JSON,
-        );
-        costs = [{ costs: r.costs, id: r.id, name: r.name }];
```
tamirkamara, line 50: Can you explain why this isn't required anymore?

marrobi: It's already happening here - https://github.com/marrobi/AzureTRE/blob/7e532bd19b7b58d526cf84fa007060574b75ee8f/ui/app/src/components/root/RootLayout.tsx#L60-L110

Not sure why it was here too. Likely as a fall back, but if there was ever an error every single costTag would make the calls.

ashis-kar91 (approved):

ui/app/src/components/shared/CostsTag.tsx
marrobi: @tamirkamara can you take another look at this one?

PR #4614, opened 2025-06-30
Remove resource locks before deleting resource groups in destroy_env_no_terraform.sh

marrobi:
Ensure that any resource locks are removed prior to deleting resource groups to prevent deletion failures.


tamirkamara (approved):

marrobi:
/test-force-approve

tested locally.

ashis-kar91 (approved):

PR #4678, opened 2025-09-18
Add missing image_gallery_id parameter to portal.yaml for export review vm

jonnyry:
If you are using compute gallery images with your Guacamole VMs, then the `image_gallery_id` parameter is needed to carry the compute gallery reference.

It is present as a parameter in:

**Windows VM:**

https://github.com/microsoft/AzureTRE/blob/90b158b9bdf89af1c8e564e45617aef4976d299e/templates/workspace_services/guacamole/user_resources/guacamole-azure-windowsvm/porter.yaml#L77

**Windows Import Review VM:**

https://github.com/microsoft/AzureTRE/blob/90b158b9bdf89af1c8e564e45617aef4976d299e/templates/workspace_services/guacamole/user_resources/guacamole-azure-import-reviewvm/porter.yaml#L50

**Linux VM:**

https://github.com/microsoft/AzureTRE/blob/90b158b9bdf89af1c8e564e45617aef4976d299e/templates/workspace_services/guacamole/user_resources/guacamole-azure-linuxvm/porter.yaml#L69

But it is missing from the Export Review VM, which causes the following error:

```
Error: parsing "/images/imgdef-my-image-name": parsing the Image ID: the number of segments didn't match 
  
 Expected a Image ID that matched (containing 8 segments): 
  
 > /subscriptions/XXX/resourceGroups/example-resource-group/providers/Microsoft.Compute/images/imageValue 
  
 However this value was provided (which was parsed into 0 segments): 
  
 > /images/imgdef-my-image-name
  
 The following Segments are expected: 
  
 * Segment 0 - this should be the literal value "subscriptions" 
 * Segment 1 - this should be the UUID of the Azure Subscription 
 * Segment 2 - this should be the literal value "resourceGroups" 
 * Segment 3 - this should be the name of the Resource Group 
 * Segment 4 - this should be the literal value "providers" 
 * Segment 5 - this should be the name of the Resource Provider [for example 'Microsoft.Compute'] 
 * Segment 6 - this should be the literal value "images" 
 * Segment 7 - this should be the user specified value for this image [for example "imageValue"] 
```

This PR adds it in so its possible to use custom images with the Export Review VM.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ ENHANCEMENTS:
 BUG FIXES:
 * Fix disable public network access for stwebcertsTREID is still flagging in Defender ([#4640](https://github.com/microsoft/AzureTRE/issues/4640))
 * Fix error 'resource with the ID pip-fw-management already exists' during firewall migration ([#4661](https://github.com/microsoft/AzureTRE/issues/4661))
+* Add missing image_gallery_id parameter to porter.yaml for guacamole export review vm ([#4678](https://github.com/microsoft/AzureTRE/pull/4678)]
 
 ## 0.25.0 (July 18, 2025)
 **IMPORTANT**:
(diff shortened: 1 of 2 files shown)


jonnyry:
> Can you increment the bundle version?

Done.

PR #4708, opened 2025-10-14
fix pipeline template documentation

### LINKED ISSUE #4707: [Docs] Pipeline Template should be shared-service instead of shared_service
**Describe the bug**
I believe the pipeline schema docs are wrong and the resourceType should be a hyphen separated not underscore
https://microsoft.github.io/AzureTRE/latest/tre-templates/pipeline-templates/overview/
https://microsoft.github.io/AzureTRE/latest/tre-templates/pipeline-templates/pipeline-schema/

Image

**Steps to reproduce**

1. Try to deploy a pipeline using "resourceType": "shared_service"
2. It will fail

main

JC-wk:
# Resolves #4707
## What is being addressed

Pipeline documentation is incorrect 

## How is this addressed

Pipeline documentation has been corrected to replace underscores with hyphens as per the API spec.

api_app/resources/strings.py
```
# Resource Type
RESOURCE_TYPE_WORKSPACE = "workspace"
RESOURCE_TYPE_WORKSPACE_SERVICE = "workspace-service"
USER_RESOURCE = "user-resource"
RESOURCE_TYPE_SHARED_SERVICE = "shared-service"
```

### DIFF
--- a/docs/tre-templates/pipeline-templates/pipeline-schema.md
+++ b/docs/tre-templates/pipeline-templates/pipeline-schema.md
@@ -1,7 +1,7 @@
 # Pipeline Template Schema
 This document will help you write a valid `pipeline: {}` block in your template.
 
-> For a working example, see `./templates/shared-services/sonatype-nexus/template_schema.json`.
+> For a working example, see `./templates/shared-services/sonatype-nexus-vm/template_schema.json`.
 
 ## Schema
 ```json


docs/tre-templates/pipeline-templates/pipeline-schema.md
martinpeck, line 1: While here, could you fixed line 4 which says...

> For a working example, see `./templates/shared-services/sonatype-nexus/template_schema.json`.

... when that folder doesn't actually exist (it should be `sonatype-nexus-vm`)?

JC-wk: sure

PR #4720, opened 2025-10-17
remove ntp firewall rule

### LINKED ISSUE #4703: Remove NTP Outbound Firewall Rule
## Description

As Azure TRE Developer
I want to ensure outbound access on UDP port 123 is restricted.
The current rule is too permissive and when auditing the NTP traffic the rule is not actually needed at all.
The rule can be audited via log analytics:
```
AZFWNetworkRule 
| where Protocol == "UDP" and DestinationPort == "123" 
| summarize by SourceIp, DestinationIp
```

## Acceptance criteria

- [ ] Outbound NTP should be blocked by default as it is not required by the core TRE components

JC-wk:
# Resolves #4703 

## What is being addressed

Remove the network rule collection containing the default allow all rule to UDP 123

## Why
Workspaces by default block outbound UDP however the core subnets do not.
The linux vm's in the core subnet use chronyd to synchronize against the Azure host rather than an external NTP time source.
The rule being a wildcard also presents a risk of it being used for data exfiltration.

Users can audit NTP by running the following in their primary log analytics workspace (log-TREName)
```
AZFWNetworkRule 
| where Protocol == "UDP" and DestinationPort == "123" 
| summarize by SourceIp, DestinationIp
```
I have done this and only the admin Jumpbox was using it however it does not require it due to it also supporting the host sync method.

## TRE's requiring NTP
If a TRE requires NTP for non-core purposes such as to sync with domain controller this can be achieved in either of the following ways:
1. Create a shared service containing the required firewall rule in terraform.
2. Add the rule to the Template pipeline of the service they are deploying (see https://microsoft.github.io/AzureTRE/unreleased/tre-templates/pipeline-templates/pipeline-schema/)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 <!-- markdownlint-disable MD041 -->
 ## 0.26.0 (Unreleased)
 **BREAKING CHANGES**
-* The firewall nework rule collection nrc-general containing the "time" ntp firewall rule has been removed. If you require NTP you will need to specify your own firewall rules.([#4720](https://github.com/microsoft/AzureTRE/pull/4720))
+* The firewall network rule collection nrc-general containing the "time" ntp firewall rule has been removed. If you require NTP you will need to specify your own firewall rules.([#4720](https://github.com/microsoft/AzureTRE/pull/4720))
 * The updated `config_schema.json` will cause a validation error if you have the developer_settings configured with no items uncommented below it. To resolve this comment out developer_settings in your config.yaml (using #) if you do not have any developer settings configured. ([#4715](https://github.com/microsoft/AzureTRE/pull/4715))
 
 ENHANCEMENTS:


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 <!-- markdownlint-disable MD041 -->
 ## 0.26.0 (Unreleased)
 **BREAKING CHANGES**
+* The firewall nework rule collection nrc-general containing the "time" ntp firewall rule has been removed. If you require NTP you can specify your own rules via a template pipeline.([#4720](https://github.com/microsoft/AzureTRE/pull/4720))
```
marrobi: @JC-wk would you use a pipeline - this is in core?

JC-wk: yeah, it's a fair point

JC-wk: Updated to "The firewall network rule collection nrc-general containing the "time" ntp firewall rule has been removed. If you require NTP you will need to specify your own firewall rules."

core/terraform/firewall/rules.tf
tamirkamara, line 1: This seems like a big step back for most customers...
What is the alternative then? You wrote below one should add their own rules, but the method is unclear to me

james-annages: Im not sure how its a step backwords, as most systems get there time from the Azure host that is running them?

tamirkamara: Are they? Then why did someone needed this rule to begin with?
I doubt it was for a custom ntp, but maybe @marrobi remembers something else

tamirkamara: Adding some info:

> Historically, most Azure Marketplace images with Linux have been configured in one of two ways:
> 
> * No time sync service is running by default
> * ntpd is running as the time sync service, and synchronizing against an external NTP time source that is accessed over the network. For example, Ubuntu 18.04 LTS Marketplace images use ntp.ubuntu.com.
> To confirm ntpd is synchronizing correctly, run the ntpq -p command.
> 
> Some Azure Marketplace images with Linux are being changed to use chronyd as the time sync service, and chronyd is configured to synchronize against the Azure host rather than an external NTP time source. The Azure host time is usually the best time source to synchronize against, as it is maintained accurately and reliably, and is accessible without the variable network delays inherent in accessing an external NTP time source over the public internet.

tamirkamara: It might be that some of the Linux VM images deplyed today are still using default ntp servers and not the host method.
This will result in time errors over time that will cause issues.

jonnyry: The NTP rule is only for the core anyway, workspaces can't make external NTP calls.  The number of linux machines within the core network is relatively limited (and mostly under this repo's control) -

- Resource Processor - runs off the Ubuntu 22.04 LTS
- Shared service package mirror -  runs off the Ubuntu 22.04 LTS

While the standard Ubuntu 22.04 LTS used `ntpd`, if I understand correctly the azure flavour of the base image uses `chronyd` which uses host time sync.  I don't have a machine running currently to double check, perhaps someone could?

If someone introduces a new VM to the core that requires external time sync, they can add the rule back in.

@tamirkamara I'm not sure it's an issue, compared the alternative of leaving port 123 open to any dest?

JC-wk: @tamirkamara If someone wants to change or deploy core firewall rules my suggestion is they should create their own shared-service template with those rules inside a terraform file.
If they only need it for a small number of resources then they should use the pipeline template method defined inside those resources.
None of the core TRE resources require NTP. NTP was already blocked in the workspace NSG.

JC-wk: @tamirkamara As I see it this wide open rule is a security risk from an exfiltration point of view (albeit restricted by the workspace NSG limitations) and needs to be dealt with.
Several TRE users have confirmed they are ok with this change.
There is a simple log analytics query to audit NTP usage
```
AZFWNetworkRule 
| where Protocol == "UDP" and DestinationPort == "123" 
| summarize by DestinationIp
```

tamirkamara: Support from a couple of users is a good sign, but it shouldn't be the only or main reason for making a change.

Right now, the description of this PR and the linked issue doesn’t clearly explain what the change is or why it’s being made. Including the points discussed in this thread would have helped save time and made things easier to review.

Consider the following:

1. The goal is to remove a firewall rule that isn’t needed (or used - see point 3).
2. This only affects the core network, specifically the resource processor. The workspace network is already blocked by NSG, so this rule wouldn’t apply there anyway.
3. The resource processor is running Ubuntu 22, which uses the host time sync method by default (I confirmed this).
4. Given it's only the resource processor affected, not sure it's a breaking change.

Please update the issue and PR descriptions to clearly reflect this information.
Help the reviewer help you, and the rest of the community.

JC-wk: Hi @tamirkamara thanks for the feedback, apologies that I did not provide more detail here in the first instance. I have updated the issue and PR. Out of the box it's not a breaking change, I included it because it has the potential to be breaking for users that have domain joined vm's, they may have opened up the NSG to allow UDP out. I can change it so it says Important rather than breaking?

tamirkamara: That's great. Thank you @JC-wk

PR #4767, opened 2025-11-21
Disable public access on stweb storage account

### LINKED ISSUE #4766: Static Web storage account defaults to public access
**Describe the bug**
When running `make deploy-core` terraform sets the storage account to public access (this is the default)
when running `make build-and-deploy-ui` the storage account is first set to public then files are uploaded and it is set back to private (leaving it inconsistent since terraform apply would set it back to public.
As far as I can tell it doesn't need to be public and therefore setting
`public_network_access_enabled    = false` on `resource "azurerm_storage_account" "staticweb"` would prevent the state being inconsistent.

**Steps to reproduce**

1. run `make deploy-core` -> observe staticweb storage is set to allow public access
2. run `make build-and-deploy-ui` observe staticweb storage is set to disable public access
3. run `make deploy-core` -> observe staticweb storage is set to allow public access

main

Azure TRE
UI Version:
0.8.19
API Version:
0.25.4

JC-wk:
# Resolves #4766 

## What is being addressed

Disable public access on stweb storage account (in reality it will already be disabled in most cases since `make build-and-deploy-ui` disables it, but terraform will re enable it on an apply.

## How is this addressed

- set `  public_network_access_enabled    = false` on staticweb storage
- Updated CHANGELOG.md and version

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ BUG FIXES:
 * Disable public access on stweb storage account ([#4766](https://github.com/microsoft/AzureTRE/issues/4766))
 * Mark `auth_client_secret` variable as sensitive in terraform templates ([#4736](https://github.com/microsoft/AzureTRE/pull/4736))
 * Fix Azure Machine Learning workspace deployment failure in unrestricted workspace by aligning missing parameters and outputs with base workspace ([#4746](https://github.com/microsoft/AzureTRE/issues/4746))
+* Fix R configuration with incorrect quotes preventing package installation on Linux VMs ([#4657](https://github.com/microsoft/AzureTRE/issues/4657))
 
 COMPONENTS:
 
--- a/templates/workspace_services/guacamole/user_resources/guacamole-azure-linuxvm/porter.yaml
+++ b/templates/workspace_services/guacamole/user_resources/guacamole-azure-linuxvm/porter.yaml
@@ -1,7 +1,7 @@
 ---
 schemaVersion: 1.0.0
 name: tre-service-guacamole-linuxvm
-version: 1.3.4
+version: 1.4.0
 description: "An Azure TRE User Resource Template for Guacamole (Linux)"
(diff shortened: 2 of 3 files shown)


martinpeck:
* `make deploy-core` uses `deploy.sh`
* `deploy.sh` sources and uses `storage_enable_public_access.sh`

As such, I would expect `make deploy-core` to use the script I'm talking about, and I'd expect it to enabled and disable the public access. I might be mistaken, but looking at those files that's what I would expect.

martinpeck:
Thanks @JC-wk , @jonnyry. 

So, I guess the hypothesis is:

* there's no need for this storage account to have public access enabled outside of a small window
* we currently use a script to turn public access ON when we need it, and turn it off when we're done
* this script has hidden the issue because, when we do a full deploy, the script leaves us in the desired state with  public access disabled
* running a partial deploy (i.e. `deploy-core`) highlights that we haven't locked this storage down

In which case, the PR looks good. My only request is that we ensure we've tested/done the following:

1. A full deploy into a clean environment
2. A redeploy/update into an existing environment
3. We've looked at what lives in this storage account and validated that there's nothing present that *might* need public access, and that isn't otherwise solved elsewhere.

If we've done all of this, then I'm happy :) I'll approve the PR now on the basis that the above has been done or will be done.

Thanks @JC-wk.

Side note: 

The reason I wanted to dig into this, and make sure we weren't missing something, is that I get worried whenever we start to fix issues that have been present in the code base, but undetected, for some time. 

There are plenty of valid reasons why these situations might occur, but sometimes they highlight a bigger underlying design issue that needs to be considered. A quick fix might put sticky-paster over that design issue, when the correct fix might be to make a bigger change. 

Turns out that isn't the case for this issue, so apologies if this felt like an obvious fix, but I wanted to double check.

martinpeck (approved):
LGTM. Thanks @JC-wk

marrobi:
/test-force-approve 3c3c7e6

Passed.

Issue #3826, opened 2024-01-15
Granular Role Permissions


Danny-Cooke-cCK:
The standard AzureTRE roles don't fit some organisations roles and responsibilities or use cases. 

We would like to see more granular RBAC permissions within the TRE as follows :-
Researcher
   ++ Can log onto all VM's within a workspace
   ++ Can view workspace cost API
   -- Can't create airlock requests
   -- Can't view airlock request
   -- Can't create VM's
Workspace Owner
   ++ Can log onto all VM's within a workspace
   ++ Can create VM's within a workspace 
   ++ Can view workspace cost API
   -- Can't view airlock request
   -- Can't create airlock requests
Airlock Data Engineer
   ++ Can log onto all VM's within a workspace
   ++ Can create airlock requests
   ++ Can view airlock request
Airlock Manager
   ++ Can log onto all VM's within a workspace
   ++ Can create VM's within a workspace 
   ++ Can approve airlock requests
   -- Can't create airlock requests

We have considered approaching this separately via both organisations using their own forks but to ensure we have a consistent approach and so that it can benefit the community its best approach this centrally. 

We need to raise this Git request to ensure that organizations are aware of it and can incorporate it into their current decision-making and roadmap.

Danny-Cooke-cCK:
Interesting yes leave it with me. Would it be reasonable for a researcher in one organisation to have "Researcher" and in another they could have "researcher" + "Airlock requestor" ?

TonyWildish-BH:
adding some things we've discussed privately, for visibility:

- We'd like to be able to grant/deny access to some workspace templates, depending on the domain. This can help keep the template catalogue manageable when we have large numbers of templates but projects that don't span the entire space of possibilities.
- Within workspaces, we'd like the possibility to limit the resources that some users can deploy, but not others. E.g., not allowing someone to deploy an expensive VM if they're new to the team and not yet familiar with the implications.
- We'd also like the ability to group workspaces (and therefore their members) into teams, where teams may be able to share resources among their own workspaces, but not with other workspaces outside their team.

This should ideally be managed by roles and policies, not workflows and approvals. Approvals should only be needed for changing the level of access, so rarely, to minimise overhead.

marrobi:
@Danny-Cooke-cCK Maybe take a look at https://casbin.org/ - there are solutions out there to help with more complex permission models. 

We likely need something that maps roles, to permissions.

[...]

PR #4796, opened 2025-12-24
remove duplicate dependencies in API

### LINKED ISSUE #4797: Dependencies are duplicated which could cause auth checks to run twice within the API
**Describe the bug**
Dependencies are duplicated which would cause auth checks to run twice within the API
e.g get_current_tre_user_or_tre_admin in api_app/api/routes/shared_services.py
```
shared_services_router = APIRouter(dependencies=[Depends(**get_current_tre_user_or_tre_admin**)])
@shared_services_router.get("/shared-services", response_model=SharedServicesInList, name=strings.API_GET_ALL_SHARED_SERVICES, dependencies=[Depends(**get_current_tre_user_or_tre_admin**)])
```

>If you attach the same auth dependency both at the router level (via APIRouter(dependencies=[Depends(auth)])) >and again on the endpoint (either in dependencies=[Depends(auth)] or as a parameter user = Depends(auth)), >FastAPI will execute it twice because each Depends(...) occurrence is evaluated independently; router-level >dependencies are simply added to the route’s dependency list. [fastapi.tiangolo.com]

main

JC-wk:
## What is being addressed

Dependencies are duplicated which would cause auth checks to run twice

> If you attach the same auth dependency both at the router level (via APIRouter(dependencies=[Depends(auth)])) and again on the endpoint (either in dependencies=[Depends(auth)] or as a parameter user = Depends(auth)), FastAPI will execute it twice because each Depends(...) occurrence is evaluated independently; router-level dependencies are simply added to the route’s dependency list. [fastapi.tiangolo.com]

## How is this addressed

- Remove dependencies where they are the same as the router dependencies
- Updated CHANGELOG.md if needed
- Increment API version


marrobi:
@JC-wk thanks. Didn't realise this was an issue. In the past I've been in two minds if security dependencies are best on each route or the router. I thin they would actually be better on each route as it enables us to to custom logic for each route to enable something like:

https://github.com/microsoft/AzureTRE/issues/3826

Thoughts?

JC-wk:
@marrobi this one is just a small interim fix to prevent auth checks happening twice and that's how all of the other endpoints work currently so keeps it consistent.
I am planning to start work on #3826 soon though.

marrobi (approved):

marrobi:
@copilot please fix the merge conflicts in this pull request.

PR #4801, opened 2026-01-05
Github Actions version & permission updates

tamirkamara:
## What is being addressed

1. Bring all actions to their latest versions
2. Add default workflow level permission to all
3. Set job level permissions

### DIFF
--- a/.github/workflows/deploy_tre_reusable.yml
+++ b/.github/workflows/deploy_tre_reusable.yml
@@ -205,7 +205,7 @@ jobs:
           details_url: "https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}"
 
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
         with:
           persist-credentials: false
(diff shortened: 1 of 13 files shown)


.github/workflows/deploy_tre_reusable.yml
```diff
@@ -301,12 +301,12 @@ jobs:
           [
             build-and-push-api,
             build-and-push-resource-processor,
-            build-and-push-airlock-processor
+            build-and-push-airlock-processor,
```
JC-wk, line 304: maybe remove the added trailing comma

tamirkamara: I don't mind doing so but this is an auto format changed that happens automatically when I save the file. I think one of the formatters that we have installed in the devcontainer.

JC-wk: Ah, looks like this is a recent change that has defaulted to this behaviour
https://github.com/redhat-developer/vscode-yaml/issues/1112
There is now a checkbox for this setting which defaults to on in the red hat yaml extension
image

PR #4802, opened 2026-01-05
Fix pytest-asyncio deprecation warning by setting default loop scope

### LINKED ISSUE #4778: PytestDeprecationWarning: The configuration option "asyncio_default_fixture_loop_scope" is unset.
**Describe the bug**
/usr/local/lib/python3.12/site-packages/pytest_asyncio/plugin.py:208: PytestDeprecationWarning: The configuration option "asyncio_default_fixture_loop_scope" is unset.
The event loop scope for asynchronous fixtures will default to the fixture caching scope. Future versions of pytest-asyncio will default the loop scope for asynchronous fixtures to function scope. Set the default fixture loop scope explicitly in order to avoid unexpected behavior in the future. Valid fixture loop scopes are: "function", "class", "module", "package", "session"

  warnings.warn(PytestDeprecationWarning(_DEFAULT_FIXTURE_LOOP_SCOPE_UNSET))

**Steps to reproduce**

1. run Pytest
2. observe warning

main

**Solution**
add `asyncio_default_fixture_loop_scope` and `asyncio_mode` to `api_app/pytest.ini`, this is already present in `/workspaces/AxymTRE/e2e_tests/pytest.ini`
```
# Ensure pytest-asyncio has an explicit event loop scope to avoid deprecation warnings
asyncio_mode = auto
asyncio_default_fixture_loop_scope = function
```

see https://github.com/pytest-dev/pytest-asyncio/issues/924

CodeVishal-17:
# Resolves #4778

## Summary

This PR fixes a pytest-asyncio deprecation warning by explicitly configuring the asyncio mode and default event loop scope for test fixtures.

## Details

Pytest emits a `PytestDeprecationWarning` indicating that the configuration option
`asyncio_default_fixture_loop_scope` is unset. With newer versions of
`pytest-asyncio` (>= 1.3.0), this will become an error unless explicitly configured.

## How is this addressed

- Explicitly sets `asyncio_mode = strict` in `pytest.ini` (intentional and aligned with pytest-asyncio >= 1.3.0)
- Sets `asyncio_default_fixture_loop_scope = function`
- Updates async fixtures and test markers to be compatible with strict mode
- Ensures async client fixtures properly manage lifecycle using async context managers

All tests pass locally under `pytest-asyncio >= 1.3.0`.


JC-wk:
Hi @CodeVishal-17 have you been able to run `pytest ./api_app/` in the devcontainer with this config set, I'm seeing some tests fail when they are function scoped. I've been able to fix them by using @pytest.fixture instead of @pytest_asyncio.fixture for those tests setups Those test setups aren't async)
I think we can use strict for `asyncio_mode`
When I have set it to module scope the tests pass e.g
```
[pytest]
filterwarnings =
    error
asyncio_mode = strict
asyncio_default_fixture_loop_scope = function
```
Also I've added a pyproject.toml at the root to specify options for bandit and pytest in a single file as when running `pytest --ignore ./e2e_tests/` it seems to ignore the pytest.ini in the subfolder or at least still defaults to None for some of the tests.
See this branch https://github.com/JC-wk/AzureTRE/tree/fix/pytest-asyncio-deprecation and these changes https://github.com/microsoft/AzureTRE/compare/main...JC-wk:AzureTRE:fix/pytest-asyncio-deprecation

JC-wk (changes requested):
This works with pytest-asyncio v0.24.0 current devcontainer version but not with v1.3.0 (latest)
try with `pip install --upgrade pytest-asyncio`

CodeVishal-17:
@JC-wk  Thanks for the clarification!

I upgraded locally to pytest-asyncio==1.3.0 and reproduced the issue.

I’ve fixed the async fixtures to use @pytest_asyncio.fixture (including making the app fixture async) and verified that the pytest-asyncio deprecation / loop-scope issue is resolved under asyncio_mode=strict.

Local test collection failures are due to missing optional runtime dependencies (Azure SDKs / FastAPI / Starlette), which are provided in CI and unrelated to this change.

Please let me know if anything else needs adjustment.

api_app/tests_ma/test_api/conftest.py
```diff
@@ -127,11 +127,11 @@ def inner():
 
 
 @pytest_asyncio.fixture(scope='module')
```
JC-wk: my fix is just to change this to `@pytest.fixture(scope='module')` rather than making the method async (remove the _asyncio)

api_app/pytest.ini
```diff
@@ -1,3 +1,5 @@
 [pytest]
 filterwarnings =
     error
+asyncio_mode = auto
```
JC-wk: This currently defaults to strict so suggest setting it to strict (everything still passes for me)

JC-wk (changes requested):
the other change required is in, then all the tests pass for me
api_app/tests_ma/test_api/test_routes/test_airlock.py
change line 130 to `@pytest.fixture(autouse=True, scope='class')`
change line 306 to `@pytest.fixture(autouse=True, scope='class')`
see https://github.com/microsoft/AzureTRE/compare/main...JC-wk:AzureTRE:fix/pytest-asyncio-deprecation#diff-5013d846e8403adeed48f3a9b720e7c4b610b02a783a54854bc0432d84bcd22aL130-R130

api_app/tests_ma/test_api/test_routes/test_airlock.py
```diff
@@ -1,6 +1,6 @@
 import time
 import pytest
-import pytest_asyncio
+# import pytest_asyncio
```
JC-wk: delete the line rather than comment out if it's no longer needed

api_app/tests_ma/test_api/conftest.py
```diff
@@ -1,5 +1,7 @@
 import pytest
 import pytest_asyncio
+
+
```
JC-wk: lots of whitespace has been added in several files and it should be reverted

.github/workflows/build_validation_develop.yml
```diff
@@ -103,6 +103,12 @@ jobs:
           VALIDATE_TSX: true
           VALIDATE_TYPESCRIPT_ES: true
 
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+          cache: "pip"
```
tamirkamara: Is this needed?

CodeVishal-17: Thanks for flagging this — you’re right.
The CI Python setup change isn’t required for the pytest-asyncio fix, so I’ve removed it to keep the PR focused.

JC-wk:
Closing this mixed branch. A clean PR will be opened with only the pytest-asyncio loop-scope configuration change.

JC-wk:
moved to https://github.com/microsoft/AzureTRE/pull/5055 @CodeVishal-17

PR #4806, opened 2026-01-12
Update python dependencies

tamirkamara:
## What is being addressed

Update most python dependencies. The 2 areas left as is are: opentelemetry (as there seem to be a change there) and pytest and testing in general.

As part of this I also fixed:
1. The underlining issues that prevent the `extended` tests from running successfully (mostly github secrets and Azure)
2. httpx change that impacted a few unit tests
3. Probably a long standing failure in many unit tests in the API related to the way we setup logging that isn't available when testing. The solution was to make sure the setup function actually get the connection string we have in the config and not directly try to fetch it on its own from the environment variables.


JC-wk (approved):
happy with this one

Issue #4284, opened 2025-01-24
Template mechanism - passing params from workspace to workspace service?


jonnyry:
Does the current template mechanism support a way to pass outputs of a workspace install, as inputs to a workspace service install?

E.g.

A workspace creates Azure AD groups.  A workspace service (e.g. AzureSQL) wants to use these same groups for authentication/permissions management.  Is it possible to output the group names from the workspace, and for the resource processor to pass them as inputs to the workspace service install?

Appreciate you can recreate the group names from the workspace ID (which is passed in) but its rather brittle.

Thanks

marrobi:
This?

https://github.com/microsoft/AzureTRE/pull/3223

We need to update the docs.

jonnyry:
> This?
> 
> [#3223](https://github.com/microsoft/AzureTRE/pull/3223)
> 
> We need to update the docs.

Ah perfect thanks!

jonnyry:
Not tested yet, unfortunately ran out of time on that issue.

jonnyry:
Not having much luck.  Can't see to get either of the two params though into the terraform:

![Image](https://github.com/user-attachments/assets/36d28dfc-036f-4059-b143-7effac81eb98)

marrobi:
I have it working in this PR - https://github.com/microsoft/AzureTRE/pull/4341/files

For `log_analytics_workspace_name`. 

I had to update something to ensure it works for values that don't have updateable set to true.

will try get this in this week.

jonnyry:
Thanks :-)  I'll take at look at your PR and try again.

jonnyry:
For ref this was the piece I was missing this... 

![Image](https://github.com/user-attachments/assets/0d2b79ac-af96-4f63-be0e-2236874f2e5d)

I didn't realise you had to invoke the parent resource in the pipeline in order to access its properties.

marrobi:
Ah, ok, nor did I. 

Wonder if that requirement can be removed.

marrobi:
Hmm, I'm not sure it should be needed, I'm going to do some testing, looks like its pulled from the database.

https://github.com/microsoft/AzureTRE/blob/cba92f16d3d386ac2121cc45514023290f044033/api_app/service_bus/helpers.py#L53-L63

marrobi:
@jonnyry I did a test today and didn't seem to need the parent resource to be invoked, could get a workspace property into a workspace service deployment.

PR #4814, opened 2026-01-16
Fix type mismatch error in property substitution

### LINKED ISSUE #4813: when using property substitution types are converted to strings causes <ValidationError: "'True' is not of type 'boolean'">
**Describe the bug**
I have a parent (workspace) template that defines a bool `enable_backup`
```json
    "enable_backup": {
      "$id": "#/properties/enable_backup",
      "type": "boolean",
      "title": "Enable Backup",
      "description": "Enable backups for the workspace service (inherited from workspace)",
      "updateable": true
    },
```
I want to consume this property in a workspace service so backup config is inherited from the workspace
```json
    "enable_backup": {
      "$id": "#/properties/enable_backup",
      "type": "boolean",
      "title": "Enable Backup",
      "description": "Enable backups for the workspace service (inherited from workspace)",
      "updateable": true
    },
...
 {
        "stepId": "main",
        "properties": [
          {
            "name": "enable_backup",
            "type": "boolean",
            "value": "{{ resource.parent.properties.enable_backup }}"
          },
```
the property substitution tries to pass this through as a string "True"/"False" rather than a boolean value which the template expects.

This is because `api_app/service_bus/substitutions.py` is converting the property to a string on line 154 resulting in `<ValidationError: "'True' is not of type 'boolean'">`
```python
        # if the value to inject is actually an object / list - just return it, else replace the value in the string
        if isinstance(prop_to_get, dict) or isinstance(prop_to_get, list):
            return prop_to_get
        else:
            val = val.replace("{{" + t + "}}", str(prop_to_get))
````

this would be the same for an int

**Steps to reproduce**

1. create a template that accesses a parent property that is a bool or int (or maybe other non list types)
2. observe the error when trying to deploy or update the template

main
latest

JC-wk:
# Resolves #4813 #4284

## What is being addressed

Substituted properties will now retain their types unless they are part of a longer string see the issue for more detail

## How is this addressed

- return the substituted value as it's correct type instead of converting to string
- added tests
- Update documentation
- Update CHANGELOG.md
- Increment api version

### DIFF
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -15,7 +15,7 @@ permissions:
 jobs:
   deploy:
     name: Deploy Documentation
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-slim
     permissions:
       contents: write
     steps:
--- a/.github/workflows/deploy_tre_branch.yml
+++ b/.github/workflows/deploy_tre_branch.yml
@@ -35,7 +35,7 @@ permissions:
 jobs:
   prepare-not-main:
     name: Preparation
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-slim
     if: |
       github.ref != 'refs/heads/main'
     outputs:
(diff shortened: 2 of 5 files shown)


marrobi (approved):
LGTM

marrobi:
/test-force-approve passed other than UI due to Yarn issue

PR #4815, opened 2026-01-18
SuperLinter 8.3.2

tamirkamara:
## What is being addressed

Upgrade SuperLinter to the latest version 8.3.2


JC-wk (approved):
LGTM

tamirkamara:
@JC-wk Did you try any of these options?

> Recommended Fix: You need to ensure Super-Linter uses the configuration file located in ui/app. You can often do this by setting an environment variable in your workflow or Super-Linter configuration, for example:
> 
> * TYPESCRIPT_ES_CONFIG_FILE: ui/app/eslint.config.js
> * TSX_CONFIG_FILE: ui/app/eslint.config.js
> * Alternatively, configure the workflow to run npm install in ui/app and then run a localized lint command, rather than relying on the global
>   Super-Linter for this specific directory.

There's also a way to test locally with `make lint`

tamirkamara:
> Not tried the options yet since it would need to be done in the pipeline though @tamirkamara, I might be able to try locally with a docker image

@JC-wk IIRC, `make lint` simulates most of what happens in the ci environment.

PR #4830, opened 2026-01-28
Store initial form state to post new fields

matthew-a-dunlap:
This alongside previous form changes means that we will be able to send empty-but-changed values in forms


marrobi:
@matthew-a-dunlap how about I close the copilot generated PR and you target this one at main? I can review/approve it that way, but can't if I created it. :)

You good with that?

matthew-a-dunlap:
@JC-wk Thanks James for the thoughts. I'm definitely very naive and would love to discuss.

My instinct is to find a solution that works for all forms and all field types, to ensure fields that go from a value to empty during an update are always passed. Setting a default value for each field in each template seems cumbersome. I think this approach utilizing `ui:emptyValue` inside `ResourceForm.tsx` could work?

https://github.com/matthew-a-dunlap/AzureTRE/compare/main...matthew-a-dunlap:AzureTRE:fix-workspace-properties-blank-uiSchema?expand=1

PR #4847, opened 2026-02-04
Super linter pipeline fixes and fixes to mysql template and airlock_processor.tf

### LINKED ISSUE #4846: Mysql template ignores storage_mb param
**Describe the bug**
storage_mb param is not used in the tf module

**Steps to reproduce**

1. run `make tflint > tflint.txt`
2. observe error
```
Warning: [Fixable] variable "storage_mb" is declared but not used (terraform_unused_declarations)

  on variables.tf line 16:
  16: variable "storage_mb" {
```
3. investigated and found storage_mb is not used

main
latest

JC-wk:
# Resolves #4846

## What is being addressed
 - Fix TFLINT errors in pipeline
 - Fix for unused storage_mb param
 - Fix duplicated tf var `TOPIC_SUBSCRIPTION_NAME`
```
2026-02-04 13:38:25 [INFO]   Command output for TERRAFORM_TFLINT:
------
1 issue(s) found:

Warning: Duplicate key: "TOPIC_SUBSCRIPTION_NAME", first defined at airlock_processor.tf:83,5-30 (terraform_map_duplicate_keys)

  on airlock_processor.tf line 99:
  99:     "TOPIC_SUBSCRIPTION_NAME"         = azurerm_servicebus_subscription.airlock_processor.name

Reference: https://github.com/terraform-linters/tflint-ruleset-terraform/blob/v0.13.0/docs/rules/terraform_map_duplicate_keys.md

1 issue(s) found:

Warning: [Fixable] variable "storage_mb" is declared but not used (terraform_unused_declarations)

  on variables.tf line 16:
  16: variable "storage_mb" {

````
## How is this addressed

- Update CHANGELOG.md
- Increment versions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 <!-- markdownlint-disable MD041 -->
-## 0.27.0 (Unreleased)
+## (Unreleased)
+**BREAKING CHANGES**
+* _No changes yet_
+
+ENHANCEMENTS:
+* _No changes yet_
+
+BUG FIXES:
+* _No changes yet_
+
+## 0.27.0 (February 5, 2026)
 **BREAKING CHANGES**
 * Azure ML workspace service now requires auto group creation for RBAC; legacy service-principal role assignment fallback has been removed. ([#4687](https://github.com/microsoft/AzureTRE/pull/4687))
 * Fix missing arguments for airlock manager requests - change in API contract  ([#4544](https://github.com/microsoft/AzureTRE/issues/4544))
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: .github/linters/.tflint_core.hcl [hunk 1]
```diff
@@ -1,19 +1,25 @@
 # This is used for TRE tags validation only.
 
 config {
-  call_module_type = "all"
+  call_module_type = "none"
```
tamirkamara: Can you explain why you did this change?
I see that the "parent" tflint.hcl file uses "local" (which apparently I missed the last time), maybe we can just set like so to keep files consistent as much as possible?

tamirkamara (approved):

.github/linters/.tflint_core.hcl
marrobi: Good shout, I think they shoudl all be local? @JC-wk I can make that cahnge if needed.

marrobi: I've updated it, but will discuss wiht @JC-wk first.

marrobi:
/test-extended

(given airlock subscription change_

.github/linters/.tflint_core.hcl
JC-wk: @tamirkamara the reason the "tags" pipelines were failing is because there is no terraform init step before the checks so the modules do not get downloaded and with this being set to "all" it was expecting them to already exist. This is a change in behaviour in different versions of tflint. I was setting it to none as this step was just for checking tags then it should be as minimal as possible. Happy to leave it set at local since it's working.

PR #4883, opened 2026-03-06
Destroy and recreate service bus in Tre stop start scripts

### LINKED ISSUE #3953: 'Service Bus Namespace' Continues Running Even After `$ make tre-stop`
Hello team,

Is it expected behavior for the 'Service Bus Namespace' to keep running even after executing the `$ make tre-stop` command?

This screenshot was captured after running the `$ make tre-stop` command:
image

Given that the Premium tier of this service is not inexpensive, is there a way to turn it off or disable it when not needed?

JC-wk:
# Resolves #3953

## What is being addressed

When `make tre-stop` runs with DELETE_SERVICE_BUS=true destroy the service bus and stop function apps / web apps (to prevent log spam) to save costs.
Update: See additonal logging costs (right of chart) if the apps are not stopped
image

## How is this addressed

- Update control_tre to remove the service bus and stop related components
- Update documentation
- Update CHANGELOG.md

By including the terraform apply in the start script it allows a wider variety of things to be stopped / deleted and then recreated to save costs.


devops/scripts/control_tre.sh
```diff
@@ -70,9 +92,70 @@ if [[ "$1" == *"start"* ]]; then
...
 elif [[ "$1" == *"stop"* ]]; then
+
+  echo "Stopping Function Apps"
+  az functionapp list --resource-group "${core_rg_name}" --query "[?state=='Running'].name" -o tsv |
+  while read -r name; do
+    echo "Stopping Function App ${name}"
+    az functionapp stop --resource-group "${core_rg_name}" --name "${name}" &
+  done
+
+  echo "Stopping Web Apps"
```
SvenAelterman, line 107: Why are function apps and web apps being stopped?

devops/scripts/control_tre.sh
```diff
@@ -70,9 +92,70 @@ if [[ "$1" == *"start"* ]]; then
     az vm start --resource-group "${core_rg_name}" --name "${vm_name}" &
   done
 
+  echo "Starting Function Apps"
```
SvenAelterman, line 82: We're stopping web apps and function apps in the workspaces, but they are not being restarted.

CHANGELOG.md
```diff
@@ -1,7 +1,10 @@
 <!-- markdownlint-disable MD041 -->
 ## (Unreleased)
 
-* _No changes yet_
+ENHANCEMENTS:
+* Reduce costs by starting and stopping Service Bus and function apps via `tre-start` and `tre-stop` ([#3953](https://github.com/microsoft/AzureTRE/issues/3953))
```
SvenAelterman: Based on this note, I presume that the function apps and web apps are being stopped to save costs. However, stopping function apps and web apps does not affect the app service plan they are running on (which cannot be stopped) and thus costs will continue.

devops/scripts/control_tre.sh
```diff
@@ -70,9 +92,70 @@ if [[ "$1" == *"start"* ]]; then
...
+  done
+
+  echo "Stopping Web Apps"
+  az webapp list --resource-group "${core_rg_name}" --query "[?state=='Running'].name" -o tsv |
+  while read -r name; do
+    echo "Stopping Web App ${name}"
+    az webapp stop --resource-group "${core_rg_name}" --name "${name}" &
+  done
+
+  # Stopping all Function Apps in workspaces
```
SvenAelterman: I don't think this makes sense to do (see comment about no cost savings below), but moreover, this doesn't account for a multi-subscription scenario.

devops/scripts/control_tre.sh
JC-wk: They create lots of errors when the service bus goes missing
image

devops/scripts/control_tre.sh
JC-wk: I think I can remove the stopping of the workspace ones and just stop the core ones, the concern is the excessive logging of failures

devops/scripts/control_tre.sh
JC-wk: It might be unnecessary to stop the workspace ones? I'm not 100% sure if they will create the same logging errors.

JC-wk: removed the workspace ones

CHANGELOG.md
JC-wk: updated wording

devops/scripts/control_tre.sh
JC-wk: I was going by this comment https://github.com/microsoft/AzureTRE/issues/3953#issuecomment-2149129300

JC-wk: image

SvenAelterman: Makes sense

PR #4929, opened 2026-06-08
fix unbound variable issue in the airlock processor

JC-wk:
## What is being addressed

 If extract_properties() throws an exception (due to decoding errors or invalid structure) then an error 'request_properties' is not defined would be generated instead of reporting the actual error, masking the original root cause.

## How is this addressed

- Define the variables before the try: so they always exist
- Update CHANGELOG.md
- Increment version


maxmartin-cgi (commented):
Approved.

maxmartin-cgi (approved):

marrobi (approved):

PR #4930, opened 2026-06-08
Fix service bus client lifecycle

### LINKED ISSUE #4977: Inconsistent ServiceBusClient lifecycle management

**Describe the bug**
key message listeners repeatedly instantiate ServiceBusClient inside infinite loops but discard the instances without calling close() or wrapping them in context managers. This can leak connection sockets and AMQP channels under transient reconnect loops.

**Steps to reproduce**

1. this might be difficult to observe as it could take a while for enough sockets to be leaked

main
latest

JC-wk:
resolves #4977
## What is being addressed

inconsistent ServiceBusClient lifecycle management in deployment_status_updater.py, airlock_request_status_update.py, and runner.py.
key message listeners repeatedly instantiate  ServiceBusClient inside infinite loops but discard the instances without calling close()  or wrapping them in context managers. This can leak connection sockets and AMQP channels under transient reconnect loops.

## How is this addressed

- Prevents connection socket leaks: ServiceBusClient instances are now properly closed using async context managers
Prevents AMQP channel leaks: Ensures channels are cleaned up even under transient reconnect scenarios
Consistent pattern: All three files now follow the same best practice of using
- Update CHANGELOG.md
- Increment version


CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ ENHANCEMENTS:
 * Pin all GitHub Actions workflow steps to full commit SHAs to prevent supply chain attacks plus update to latest releases ([#4886](https://github.com/microsoft/AzureTRE/pull/4886))
 * Add Windows Server 2025 image support to Guacamole. ([#4890](https://github.com/microsoft/AzureTRE/issues/4890))
 
+BUG FIXES:
+* Fix inconsistent ServiceBusClient lifecycle management in deployment_status_updater.py, airlock_request_status_update.py, and runner.py to prevent connection socket and AMQP channel leaks
+
```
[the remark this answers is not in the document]
JC-wk: updated

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ ENHANCEMENTS:
 * Pin all GitHub Actions workflow steps to full commit SHAs to prevent supply chain attacks plus update to latest releases ([#4886](https://github.com/microsoft/AzureTRE/pull/4886))
 * Add Windows Server 2025 image support to Guacamole. ([#4890](https://github.com/microsoft/AzureTRE/issues/4890))
 
+BUG FIXES:
+* Fix inconsistent ServiceBusClient lifecycle management in deployment_status_updater.py, airlock_request_status_update.py, and runner.py to prevent connection socket and AMQP channel leaks
```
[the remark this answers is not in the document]
JC-wk: updated

api_app/service_bus/airlock_request_status_update.py
```diff
@@ -34,43 +34,46 @@ async def receive_messages(self):
...
+                                            else:
+                                                # could have been any kind of transient issue, we'll abandon back to the queue, and retry
+                                                await receiver.abandon_message(msg)
+
+                                await asyncio.sleep(10)
 
                 except OperationTimeoutError:
-                    # Timeout occurred whilst connecting to a session - this is expected and indicates no non-empty sessions are available
+                    # Timeout occurred whilst connecting - this is expected and indicates no non-empty sessions are available
                     logger.debug("No sessions for this process. Will look again...")
```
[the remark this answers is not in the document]
JC-wk: changed to messages

marrobi (commented):
From Opus 4.8

---

I found one correctness issue and one behavior risk:

- In `api_app/service_bus/airlock_request_status_update.py`, `ServiceBusClient` is now created once outside the inner polling loop and reused forever. That means if the connection drops or the client becomes unhealthy after a transient failure, the code can keep retrying with the same stale client instead of recreating it. The previous pattern re-established the client each iteration; this change may reduce leaks, but it also makes recovery less robust. Consider moving the `async with ServiceBusClient(...)` back inside the retry loop, or explicitly recreating the client after connection-related failures.

- In `deployment_status_updater.py`, the logic still keeps a single `ServiceBusClient` alive across the inner loop as well. If the intent is just to avoid leaks, that’s fine, but the current structure can hold the client open indefinitely while also suppressing reconnect churn. Make sure this is intentional and that there’s a clean exit/recreate path when the receiver or client encounters a fatal state.

- `resource_processor/vmss_porter/runner.py` now wraps `ServiceBusClient` in `async with`, but the exception handlers only sleep and continue. If `receive_message()` ever exits due to a non-fatal error, the outer `async with` will close the client and then immediately reopen it on the next `runner()` invocation, which is fine. Just verify the tests cover the `__aenter__/__aexit__` path for the new context-manager usage.

- The changelog entry mentions all three files, but the behavioral change in `airlock_request_status_update.py` is materially different from the others: it changes reconnection semantics, not just lifecycle management. That should be called out more explicitly in the PR description or handled with a narrower refactor.

api_app/service_bus/deployment_status_updater.py
marrobi, line 1: @JC-wk thoughts on:

deployment_status_updater.py gained an elaborate structure: an inner while True loop, an hourly client-recreation timer, and a nested try/except OperationTimeoutError. The other two files were not given equivalent treatment — they simply wrap the client and rely on the outer loop. The PR description promises a "consistent pattern," but the deployment updater now diverges significantly. Either simplify the deployment updater to match, or document why the session-based (NEXT_AVAILABLE_SESSION) consumer needs special handling.

ChrisChapman-gh (approved):
Assuming the conflicts are resolved and the tests pass - looks ok to me.

api_app/service_bus/deployment_status_updater.py
JC-wk: The API consumers now share the same structure

api_app/service_bus/airlock_request_status_update.py
```diff
@@ -34,43 +34,60 @@ async def receive_messages(self):
...
-                    current_time = time.time()
-                    polling_count += 1
-                    # Log a heartbeat message every 60 seconds to show the service is still working
-                    if current_time - last_heartbeat_time >= 60:
-                        logger.info(f"Queue reader heartbeat: Polled {config.SERVICE_BUS_STEP_RESULT_QUEUE} queue {polling_count} times in the last minute")
-                        last_heartbeat_time = current_time
-                        polling_count = 0
-
                     async with credentials.get_credential_async_context() as credential:
+                        # We keep a single ServiceBusClient alive across the inner loop to avoid excessive connection
```
marrobi: Do we need the comments?

api_app/service_bus/deployment_status_updater.py
```diff
@@ -43,43 +43,60 @@ async def receive_messages(self):
...
-                                        await receiver.complete_message(msg)
-                                    else:
-                                        # could have been any kind of transient issue, we'll abandon back to the queue, and retry
-                                        await receiver.abandon_message(msg)
-                            logger.info(f"Closing session: {receiver.session.session_id}")
-
-                except OperationTimeoutError:
-                    # Timeout occurred whilst connecting to a session - this is expected and indicates no non-empty sessions are available
-                    logger.debug("No sessions for this process. Will look again...")
+                        # We keep a single ServiceBusClient alive across the inner loop to avoid excessive connection
```
marrobi: Same.

marrobi (commented):
Can we reduce the verbosity of some of the comments?

marrobi (approved):

marrobi:
/test-force-approve 551fdfb82d647b1007fa423c8f00fdf9bc458632

E2E failed for unrelated reason.

PR #4932, opened 2026-06-09
Enhance service bus handling of invalid JSON in receive_message function

### LINKED ISSUE #4976: Service Bus can crash if it receives a malformed payload
**Describe the bug**
Currently the service bus will crash if it receives a malformed/non-JSON messages

**Steps to reproduce**

1. Send invalid json to the service bus and observe a crash

main

JC-wk:
resolves #4976 
## What is being addressed

Service bus will now skip malformed payloads instead of crashing

### Summary of Changes                                                
                                                                        
  1. Bug Fix in runner.py:                                         
      • Caught  json.JSONDecodeError  when deserializing the Service Bus message payload.                                                  
      • Sent the malformed message to the dead-letter queue via  await  receiver.dead_letter_message(msg, reason="InvalidJSON")  to clear it from the queue and aid in troubleshooting.
      • Added a  continue  statement inside the  except  block to prevent processing of empty/malformed message structures, avoiding downstream  TypeError  crashes.
  2. Added Unit Tests in test_runner.py:
      • Added a new unit test  test_receive_message_bad_json  to simulate receiving a malformed/non-JSON payload, asserting that:  
          • The message is successfully sent to the dead-letter queue.  
          • The message is not processed or marked complete (i.e. skipped).

- Updated CHANGELOG.md
- Increment version


marrobi (approved):
@JC-wk  Opus 4.8 review, let me kniw if agree:
---

A couple of things before merge:

2. **(Optional) Guard the dead-letter call.** If a message is *already* in a state where `dead_letter_message` can fail (e.g. lock lost / already settled), the raised exception would bubble to the outer `except Exception` and we'd re-enter the same situation. Consider wrapping the dead-letter in its own try/except so a settle failure is logged rather than aborting the batch, e.g.:

   ```python
   except json.JSONDecodeError as e:
       logger.error(f"Received bad service bus resource request message: {e}")
       try:
           await receiver.dead_letter_message(msg, reason="InvalidJSON", error_description=str(e))
       except Exception:
           logger.exception("Failed to dead-letter malformed message")
       continue
   ```

3. **(Nit) Test brittleness.** `test_receive_message_bad_json` asserts the exact exception string `"Expecting value: line 1 column 1 (char 0)"`. That text comes from the stdlib and could change across Python versions. Asserting on `reason="InvalidJSON"` (and maybe that `error_description` is non-empty) would be more robust, but this is minor.

```

PR #4943, opened 2026-06-30
 Fix intermittent workspace deletion failure on AMPLS private DNS zone group

marrobi:
Resolves #3194. Adds an azapi delete retry (+ timeouts) to the Azure Monitor private DNS zone group so transient AnotherOperationInProgress 409s during destroy are retried instead of failing.


marrobi:
@copilot update this branch from upstream main.

PR #4947, opened 2026-07-03
Issue 4940: Add skip deployment flag for PR bot tests

### LINKED ISSUE #4940: Skip deployment parameter for PR bot test commands
When we have an environment that is already deployed, there is no need to trigger all of the matrix jobs and container builds again. For example, if we want to make a change to test behaviour and not hte underlying code due to a transient failure, at present we have to run all of the prior steps if a new commit is pushed.

* Add a skip_deployment flag to the PR bot commands for /test and /test-extended
* Ensure this starts the workflow at e2e and smoke test jobs and skips all prior steps

**Describe alternatives you've considered**
Considered creating a separate E2E workflow just for running tests, however this may be more useful as we can drive it via pr bot commands.

**Additional context**

stuart-bass-cgi:
# Resolves #4940 

## What is being addressed

PR bot `/test` and `/test-extended` commands currently always run the full validation flow, including build and deployment jobs, even when the PR validation environment has already been deployed. This makes it slow to re-run smoke or extended E2E tests after transient test failures or test-only changes.

## How is this addressed

- Adds a `skip_deployment` flag for the `/test` and `/test-extended` PR bot commands.
- Passes the parsed flag from the PR comment bot into the reusable TRE deployment workflow.
- When `skip_deployment` is true, skips the deployment/build/register jobs and allows the smoke and requested custom E2E jobs to run against the existing deterministic PR validation environment.
- Keeps the existing external PR SHA validation in place, including when `skip_deployment` is supplied.
- Updates PR bot command documentation with the new syntax and fork PR examples.
- Adds unit test coverage for the new command parsing behaviour.


stuart-bass-cgi:
@microsoft-github-policy-service agree company="CGI"

marrobi:
@stuart-bass-cgi thanks for the PR.

A review with Opus 4.8 says:

I

---

**Overall:** Well-structured, focused change that solves a real developer pain point (re-running smoke/extended E2E tests without a full redeploy). Good test coverage and docs, plus a couple of nice opportunistic bug fixes. A few items to confirm before approval.

**🟢 Done well**
- Clean separation of `commandHasSkipDeploymentFlag()` and `getShaFromCommandParts()` — correctly supports passing a SHA *and* the flag in any order on fork PRs.
- Good unit-test coverage for internal `true`/default-`false`, fork `/test <sha> skip_deployment`, and `/test-extended`.
- Fixes real pre-existing bugs: `branchRefid` → `branchRefId` typo, and the debug echo that printed `prRefId` under the `branchRefId` label.
- Docs updated in step with the code, including fork examples.

**🟠 To address**
1. **Check status reporting on skip-deployment runs.** With `skipDeployment == true`, `deploy_management` is skipped, so the entire `needs` chain feeding `e2e_tests_smoke`/`e2e_tests_custom` is skipped. The `always() && (inputs.skipDeployment || ...)` guards make the tests run, which works — but please confirm the `summary` job still reports the `"Deploy PR / Run E2E Tests (Smoke)"` check back to the PR head SHA so the PR status check updates correctly. A short code comment explaining why `always()` + skipped upstreams produces the intended behavior would help future maintainers.
2. **`skip_deployment` is silently ignored by `/test-extended-aad` and `/test-shared-services`.** The flag is only wired into `/test` and `/test-extended`. A user appending it to the other commands gets a full deploy with no feedback. Consider warning on unsupported use, or confirm this asymmetry is intentional (docs already omit it for those commands).

**🟡 Nits**
- `skip_deployment` is a magic string in multiple places — consider a shared constant to avoid drift.
- `const { create } = require('domain');` (build.js) is unused dead code; low-risk cleanup while you're here.
- The string→bool round-trip (`== 'true'`) for the typed workflow input is correct — just flagging it's intentional.

**Verdict:** LGTM pending confirmation of (1) smoke-test check reporting and (2) the intentional flag asymmetry.

---

Let me know if you agree!

stuart-bass-cgi:
@marrobi - should all be resolved now

stuart-bass-cgi:
@marrobi - the review feedback should now be addressed in the latest commits.

- Added comments in `.github/workflows/deploy_tre_reusable.yml` to explain the `always()` guards: when `skipDeployment` is true, the deploy jobs in `needs` are intentionally skipped, but the smoke/custom E2E jobs still evaluate and run against the existing PR environment.
- Confirmed the summary job still reports the `Deploy PR / Run E2E Tests (Smoke)` check against `prHeadSha`; it only depends on the E2E jobs, so it still runs after skip-deployment runs even though deploy jobs are skipped.
- `skip_deployment` is now explicitly rejected for `/test-extended-aad`, `/test-shared-services`, and `/test-backups` with a bot warning rather than being silently ignored.
- Added a shared `SKIP_DEPLOYMENT_FLAG` constant, removed the unused `domain` import, and documented the intentional string-to-bool conversion in the workflow.
- Added unit coverage for unsupported `skip_deployment` usage and for external `/test-extended skip_deployment <sha>` ordering.

One caveat on live-testing the unsupported-command warning: `issue_comment` workflows run from the default branch, so the current live PR bot still loads `build.js` from `main`, not this PR. I tried `/test-extended-aad skip_deployment`; it started a full deploy because `main` does not yet include the new rejection logic. The behavior is covered by the PR unit tests and will apply to live comments once this parser change reaches `main`.

rudolphjacksonm (approved):

marrobi (approved):

PR #4969, opened 2026-07-14
change tsconfig `moduleResolution` to `bundler`

### LINKED ISSUE #4968: In tsconfig.json vscode warns `Option 'moduleResolution=node10' is deprecated and will stop functioning in TypeScript 7.0.`
**Describe the bug**
In tsconfig.json vscode warns Option 'moduleResolution=node10' is deprecated and will stop functioning in TypeScript 7.0.

```
Option 'moduleResolution=node10' is deprecated and will stop functioning in TypeScript 7.0. Specify compilerOption '"ignoreDeprecations": "6.0"' to silence this error.
  Visit https://aka.ms/ts6 for migration information.
Specify the module resolution strategy:

'node16' or 'nodenext' for modern versions of Node.js. Node.js v12 and later supports both ECMAScript imports and CommonJS require, which resolve using different algorithms. These moduleResolution values, when combined with the corresponding module values, picks the right algorithm for each resolution based on whether Node.js will see an import or require in the output JavaScript code.
'node10' (previously called 'node') for Node.js versions older than v10, which only support CommonJS require. You probably won't need to use node10 in modern code.
'bundler' for use with bundlers. Like node16 and nodenext, this mode supports package.json "imports" and "exports", but unlike the Node.js resolution modes, bundler never requires file extensions on relative paths in imports.
'classic' was used in TypeScript before the release of 1.6. classic should not be used.

There are reference pages explaining the theory behind TypeScript’s module resolution and the details of each option.

node: Deprecated, use "node10" in TypeScript 5.0+ instead
```

**Steps to reproduce**

1. Open tsconfig.json in vscode and see error under `    "moduleResolution": "node",`
main
latest

**Solution**

- Updated the module resolution strategy from  "node"  (deprecated in TS 5.0+, removed in TS 7.0) to  "bundler"  in tsconfig.json.
- Modern bundlers like Vite resolve modules using package exports, and  "bundler"  matches Vite's  resolution strategy.
- Verified the build via  npm run build , and it compiles successfully without deprecation errors or warnings.

JC-wk:
# Resolves #4968
## What is being addressed

The current moduleResolution setting of "node" (node10) is deprecated, since we are using vite as a bundler then moduleResolution should be updated to "bundler"

## How is this addressed

- change moduleResolution to bundler
- Updated CHANGELOG.md
- Increment ui version

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ ENHANCEMENTS:
 * Update dependencies to address Dependabot security alerts: `aiohttp` to 3.14.1, `Pygments` to 2.20.0, `esbuild`, `ws`, `js-yaml`, `@babel/core`, `flatted` (via vitest upgrade), and `react-router-dom`. ([#4950](https://github.com/microsoft/AzureTRE/issues/4950))
 
 BUG FIXES:
+* Fix UI TypeScript deprecation warning by updating `moduleResolution` to `bundler` in `tsconfig.json`. ([#4968](https://github.com/microsoft/AzureTRE/issues/4968))
 * Fix API timeout and name collision failures on workspace creation by checking storage account name availability and improved logging. ([#4946](https://github.com/microsoft/AzureTRE/pull/4946))
 * Fix error handling in airlock processor ([#4929](https://github.com/microsoft/AzureTRE/pull/4929))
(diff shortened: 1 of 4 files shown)


rudolphjacksonm (approved):
LGTM

marrobi (approved):

PR #5005, opened 2026-07-27
Update JSON Schema version to Draft 2020-12

### LINKED ISSUE #4795: Pipeline Template JSON Schema versions
## Description

I was just looking in to the differences within the [JSON Schema](https://json-schema.org/specification) and I could not see anywhere in the TRE Docs what schema are supported. We seem to be using Draft 07 in the templates are
`  "$schema": "http://json-schema.org/draft-07/schema",`
yet there are properties in the template that are newer such as [unevaluatedProperties](https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.9.3.2.4)
`api_app/models/domain/resource_template.py`

I think there needs to be documented standards around this and the templates may need updating to suit

JC-wk:
# Resolves #4795

## What is being addressed

Various schema are currently in use, this aligns templates to the latest.
Minor spelling and naming issues are also fixed
fixes issues in template schema docs

## How is this addressed

- Updated version numbers for several shared services and workspace templates to reflect new changes.
- Update documentation
- Update CHANGELOG.md
- Increment template versions


rudolphjacksonm (approved):
LGTM

marrobi:
@JC-wk think this needs extended tets?

JC-wk:
> @JC-wk think this needs extended tets?

They already passed https://github.com/microsoft/AzureTRE/actions/runs/30441451183

marrobi (approved):
If sure nothing breaks, for example switching resourceVersion to integer, all good,

JayDoubleu:
I reviewed `169b9e87`. The existing 764 API tests pass, but I reproduced two cases that need covering before merge:

### 1. Validation of older templates

Preserving `$schema` during registration changes validation behaviour for newly registered Draft 7 templates. Draft 7 ignores `unevaluatedProperties: false`.

Using the same template and inputs:

- Main rejects an unexpected property; this PR accepts it.
- Main rejects a PATCH to a property marked `updateable: false`; this PR accepts it.

Could we preserve these restrictions across supported schema versions and add regression tests for registration, create and PATCH?

### 2. UI support for the new schema metadata

I reproduced the outstanding review comment. `sanitiseTemplateForRJSF` drops `$schema`, `$id` and `$defs`. A property referencing `#/$defs/label` then fails with:

```text
Could not find a definition for #/$defs/label.
```

Retaining those fields also requires a compatible validator: the current default validator rejects the Draft 2020-12 schema URI. We should preserve support for existing templates and add a resource-form test covering rendering and valid/invalid submission through a local `$defs` reference.

### Version and validation

API `0.26.10` is now on main through #5043, so this needs another version bump after updating the branch.

After the fixes, I suggest focused create/edit form checks on an existing test TRE, including both legacy and updated templates, before approval.

JayDoubleu:
Rechecked at `a1bce356`. The API and bundle version increments are now addressed, but the two previously reported functional issues remain. We will push follow-up commits to this PR to address them, include regression tests and clarify compatibility guidance. We will first bring the branch up to date with `main`.

### 1. Preserve validation restrictions for older templates

Local checks using the production registration and validation methods confirmed that newly registered Draft 7 templates accept unexpected properties during resource creation and PATCH changes to properties marked `updateable: false`. Current main rejects both requests.

Preserving `$schema` selects the Draft 7 validator, which ignores `unevaluatedProperties: false`. We will preserve these restrictions across supported schema versions and cover registration, serialisation/reload, create and PATCH. The regression coverage will include older templates, Draft 2020-12 templates and templates without `$schema`.

### 2. Support the schema metadata in the UI

`sanitiseTemplateForRJSF` still removes `$schema`, `$id` and `$defs`. A form referencing `#/$defs/label` fails locally with:

```text
Could not find a definition for #/$defs/label.
```

Retaining those fields alone is insufficient: the current validator rejects the Draft 2020-12 schema identifier. We will update the UI schema handling and validator configuration while preserving legacy-template support, and add create/edit form tests covering rendering and valid/invalid submission through a local `$defs` reference.

### Validation

These findings concern application behaviour, rather than cosmetic schema changes. The existing metadata-persistence test does not cover either failure.

After the fixes are pushed and local regression checks pass, we will rerun extended tests against the updated revision. Focused create/edit checks with both legacy and updated templates should also form part of the validation. Passing the existing suite alone would not close these findings.

The reproductions above were local. No live Azure validation was performed.

JayDoubleu:
The previous extended-test run was blocked during deployment because the deleted Key Vault remains protected against purging, preventing reuse of its name.

I’ve started a [replacement validation run](https://github.com/microsoft/AzureTRE/actions/runs/35791141349) using a fresh CI environment and this PR merged with current `main` (`97ee23a6`).

This runs through a separate branch workflow. Deployment is underway, with smoke and extended test results still pending.

JayDoubleu:
Merging as approved by Marcus.

PR #5012, opened 2026-07-29
Update porter.yaml files to mask sensitive parameters

### LINKED ISSUE #5011: Improve handling of secrets in porter bundles
**Describe the bug**

Sensitive credentials are not consistently redacted from Terraform deployment logs when a deployment fails.

Some occurrences of the Terraform command are masked, while others expose values passed as command-line arguments, for example:

```text
-var auth_client_secret=<value>
-var client_secret=<value>
```

**Expected behavior**

Sensitive values should be redacted consistently in Porter, Terraform, shell, and pipeline logs.

This requires assigning `sensitive: true` in the `porter.yaml` of each template for the azure credentials to prevent leakage in to logs.

Or ideally handling secrets differently (Out of scope of this issue). Will log a separate issue to investigate whether secrets can avoid command-line arguments entirely in future, for example by using:

- Azure Key Vault references
- Protected environment variables such as `TF_VAR_*`
- A protected temporary variables file
- Managed identity or OIDC authentication instead of client secrets

**Steps to reproduce**

1. Configure a deployment with sensitive credentials.
2. Start a Terraform deployment.
3. Cause Terraform to fail.
4. Review all generated deployment and pipeline logs.
5. Compare repeated Terraform command/error entries.

main
latest

## Related issues
#2247

JC-wk:
# Resolves #5011

## What is being addressed

Porter contains params as secrets which can leak in to log files

## How is this addressed

- Set `sensitive: true` for all secrets
- Update CHANGELOG.md 
- Increment template versions


marrobi:
@JC-wk is this ready for review?

JC-wk:
> @JC-wk is this ready for review?

yes

marrobi (approved):
LGTM if it works.

JC-wk:
@marrobi airlock_request_sas_url is the one that just needs a check/test since that is also now set in terraform

marrobi:
> @marrobi airlock_request_sas_url is the one that just needs a check/test since that is also now set in terraform

Will test.

marrobi:
@JC-wk happy to merge, Opus did a small local check with some TF:

---

Tested the airlock_request_sas_url change specifically (the one flagged as also being set in Terraform). No live deployment — validated the Terraform sensitivity handling locally, since the concern was whether marking the variable sensitive breaks how the SAS URL is rendered into the review VM's bootstrap script.

What I checked:

terraform validate on the import-reviewvm root module and its vm submodule — both valid.

Sensitivity propagation — the SAS URL flows: var.airlock_request_sas_url (sensitive = true) → templatefile(download_review_data.ps1) → base64encode(join(compact([...]))) → VM custom_data. **I replicated this exact pattern in a throwaway config and confirmed:**

terraform apply succeeds with no "sensitive value" / invalid-context error.
The base64-decoded custom_data still contains the real SAS URL (incl. the sig= token). sensitive only masks it in plan/log output — it does not redact the value the VM actually needs to download the review data.
No leaks — neither review VM's outputs.tf exposes the SAS URL / review_data_script / custom_data.

Parity — porter sensitive: true matches Terraform sensitive = true for both import and export review VMs.

✅ Marking airlock_request_sas_url sensitive is safe — it validates, renders correctly, and the token still reaches the VM. Note this PR only added sensitive = true to the import reviewvm's variables.tf; the export one already had it, so both are now consistent.

I did not run a full end-to-end deploy (TRE + workspace + Guacamole + an in-review airlock import request provisioning an actual review VM). Happy to do that if we want live confirmation.

maxmartin-cgi (approved):

PR #5014, opened 2026-07-29
Fix workspace deletion when backup is enabled

### LINKED ISSUE #4962: Workspace deletion fails when backup is enabled because prepare_backup_for_destroy cannot disable Recovery Services Vault soft delete
**Describe the bug**

PR #4907 resolves the workspace creation issue caused by the Recovery Services Vault soft delete changes introduced by Azure. After implementing the changes from the PR, I can successfully create workspaces with backup enabled.

However, deleting a backup-enabled workspace consistently fails.

The failure occurs during the `prepare_backup_for_destroy` Terraform resource. The following Azure CLI command fails:

```
az backup vault backup-properties set \
  --name "$vault" \
  --resource-group "$rg" \
  --soft-delete-feature-state Disable
```

Azure returns:

```
BMSUserErrorDisablingSoftDeleteStateNotAllowed

Disabling soft delete or enhanced security state is not allowed for this vault.

```

Since the destroy script uses `set -euo pipefail`, the script exits immediately and the remaining cleanup is not executed.

Terraform later attempts to delete the storage resources but fails because the storage account lock still exists, resulting in:

```text
409 ScopeLocked
```

The workspace remains in the **Deleting** state until the operation eventually times out.

## Steps to reproduce

1. Deploy Azure TRE with the changes introduced in PR #4907.
2. Create a workspace with **Enable Backup** enabled.
3. Wait for the workspace deployment to complete successfully.
4. Delete the workspace.
5. Observe that the workspace remains in the **Deleting** state and eventually fails.

marrobi:
## What is being addressed

Workspace deletion fails when backup is enabled due to the inability to disable Recovery Services Vault soft delete. The deletion process hangs, leaving the workspace in a **Deleting** state.

## How is this addressed

- Introduced a `delete_backups_on_uninstall` flag to manage backup deletion behavior during workspace uninstallation.

- Updated the `remove_backup.sh` script to handle backup resource cleanup based on the new flag.

- Ensured that the soft delete feature remains enabled, as Azure does not allow disabling it.

- Enhanced documentation to clarify the behavior of the new flag and its implications.

- Updated the CHANGELOG and incremented the template version to reflect these changes.

### DIFF
--- a/docs/tre-templates/workspaces/base.md
+++ b/docs/tre-templates/workspaces/base.md
@@ -1,6 +1,6 @@
 # Azure TRE base workspace
 
-The base workspace template is the foundation that all other workspaces and workspace services are built upon. Alternative workspace architectures could be used. However, the templates provided in this repository rely on the specific architecture of this base workspace.
+The base workspace template is the foundation that all other workspaces and workspace services are built upon. Alternative workspace architectures could be used. However, the templates provided in this repository use the base workspace.
 
 The base workspace template contains the following resources:
 
@@ -33,8 +33,10 @@ When `enable_backup` is set (the default) the base workspace deploys a Recovery 
 
 Both flags are updateable, so the choice can be changed on an existing workspace before it is deleted.
 
+> **Note:** If [vault immutability](https://learn.microsoft.com/azure/backup/backup-azure-immutable-vault-concept) is enabled on the Recovery Services Vault (for example, enforced by an Azure Policy in your environment), backups cannot be deleted before they expire. In this case `delete_backups_on_uninstall = true` will not be able to remove the backups or vault, and workspace deletion may fail.
+
 ## Azure Trusted Services
-*Azure Trusted Services* are allowed to connect to both the key vault and storage account provisioned within the workspace. If this is undesirable additional resources without this setting configured can be deployed.
+*Azure Trusted Services* are allowed to connect to both the key vault and storage account provisioned within the workspace. If this is undesirable additional resources without this setting configured could be deployed by a custom workspace.
 
 Further details around which Azure services are allowed to connect can be found below:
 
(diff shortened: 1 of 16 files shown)


JC-wk:
@marrobi do the other workspaces need the same setting? (unrestricted and airlock-import-review)

marrobi:
> @marrobi do the other workspaces need the same setting? (unrestricted and airlock-import-review)

Hmm, they dont have backups enabled, but should check they deploy ok.

JC-wk:
> > @marrobi do the other workspaces need the same setting? (unrestricted and airlock-import-review)
> 
> Hmm, they dont have backups enabled, but should check they deploy ok.

The templates contains enable_backup though so I'd assumed they do have backups?

templates/workspaces/base/template_schema.json
JC-wk, line 1: I'd assumed this update would be needed on templates/workspaces/airlock-import-review/template_schema.json and templates/workspaces/base/template_schema.json since they both have enable_backup

marrobi: True, let me check.

marrobi: Yup, thanks, I'll update them too.

marrobi:
image

docs/tre-templates/workspaces/base.md
```diff
@@ -22,6 +22,17 @@ When deploying a workspace the following properties need to be configured.
...
 | `client_secret` | Valid client secret. | |
 
+## Backup
+
+When `enable_backup` is set (the default) the base workspace deploys a Recovery Services Vault into the workspace resource group and protects the workspace's shared storage (and any VMs configured for backup).
+
+| `delete_backups_on_uninstall` | Behaviour on workspace deletion |
+| ----------------------------- | ------------------------------- |
+| `false` (default) | **Retain the backups.** Protection is stopped but the recovery points are kept. The vault, its backup resources and the workspace resource group are removed from the Terraform state, so the **resource group is retained and continues to hold the Recovery Services Vault and its recovery points** after the workspace is deleted. |
+| `true` | **Delete the backups.** Protection is stopped with data deletion, the backup containers are unregistered and the Recovery Services Vault (with its recovery points) is deleted. Terraform then destroys the rest of the workspace, including its resource group, leaving nothing behind. |
```
SvenAelterman, line 32: I would add notes here about the impact of immutability.

SvenAelterman (approved):

JC-wk (approved):
Did a small test of this and it worked great

docs/tre-templates/workspaces/base.md
marrobi: Done

PR #5052, opened 2026-08-18
Show the user's TRE and workspace roles in the UI user menu

### LINKED ISSUE #5051: The UI does not show the roles that the user has
## Description

As a TRE user (Researcher, Workspace Owner, Airlock Manager or TRE Administrator)
I want the UI to show the roles that I have
So that I know why a function is not available, and I can give the correct data when I ask for help.

The UI knows the roles of the user, but it does not show them. It uses the roles only to select which elements to show:

- The core roles go into `AppRolesContext`. [`App.tsx`](https://github.com/microsoft/AzureTRE/blob/3b773995fb730b54b09028e3a5b6cc4a9b78bae1/ui/app/src/App.tsx#L39-L55) sets them from the access token, which [`useAuthApiCall.ts#L90-L95`](https://github.com/microsoft/AzureTRE/blob/3b773995fb730b54b09028e3a5b6cc4a9b78bae1/ui/app/src/hooks/useAuthApiCall.ts#L90-L95) decodes.
- The workspace roles go into `WorkspaceContext`. [`WorkspaceProvider.tsx#L50-L78`](https://github.com/microsoft/AzureTRE/blob/3b773995fb730b54b09028e3a5b6cc4a9b78bae1/ui/app/src/components/workspaces/WorkspaceProvider.tsx#L50-L78) gets them from the token and then sets them.
- These components read the `roles` field of the two contexts: [`SecuredByRole`](https://github.com/microsoft/AzureTRE/blob/3b773995fb730b54b09028e3a5b6cc4a9b78bae1/ui/app/src/components/shared/SecuredByRole.tsx), `LeftNav`, `RootLayout`, `ResourceCard`, `ResourceContextMenu`, `ResourceHistoryList`, `ResourceOperationsList`, `Airlock`, `WorkspaceProvider` and `WorkspaceUsers`.
- Each of these components uses the roles for one of two functions. It selects which elements to show, or it selects the correct token scope for an API call. No component shows the roles to the user.

The user menu shows only the name of the user and the Logout command ([`UserMenu.tsx#L27`](https://github.com/microsoft/AzureTRE/blob/3b773995fb730b54b09028e3a5b6cc4a9b78bae1/ui/app/src/components/shared/UserMenu.tsx#L27)).

This means that two different conditions look the same. `SecuredByRole` removes an element fully when the user does not have the role and the caller gives no error message. A Researcher who does not see the "Create new" button on the Workspaces page cannot find the cause. The cause can be the Researcher role, which is correct. The cause can also be a role assignment that is not yet in effect.

Only one part of the UI tells the user about the role of that user. [`WorkspaceProvider.tsx#L220-L225`](https://github.com/microsoft/AzureTRE/blob/3b773995fb730b54b09028e3a5b6cc4a9b78bae1/ui/app/src/components/workspaces/WorkspaceProvider.tsx#L220-L225) shows a message to a TRE Admin who opens a workspace. This message is useful. `SecuredByRole` can also show an "Access Denied" message. But that message gives the necessary role. It does not give the roles that the user has.

The condition also increases the support work. To find the cause of "I cannot see X", you must ask the user to look in Entra ID. As an alternative, an administrator must look at the app role assignments of that user. But the token in the browser already has this data.

Tests are also more difficult. Each role sees a different page, because `SecuredByRole` removes the elements that the role does not permit. Thus you must test each change to a controlled function with each role.

This test procedure gives no clear indication of the current role. You change the app role assignment in Entra ID. Then you open the UI again. Then you must look at the buttons to decide if the new role is in effect. If the correct element is not there, there are two possible causes. The change has a defect, or the browser has a token that it received before the change to the role.

A role indicator makes this check direct:

1. Remove the user from the role.
2. Sign in again.
3. Make sure that the indicator shows the new set of roles.
4. Start the test of the change.

This gives the person who does the test a clear indication that the system is in the correct condition. This is most important for the roles for which a test account is difficult to get.

The indicator must use the same token claims that `SecuredByRole` uses. It must not read Entra ID directly. If it read Entra ID, the indicator and the UI could show different data. This difference is important, because it shows that the browser has an old token. The UI does not show this condition now.

## Suggested solution

Show the roles that the client already has. A change to the API is not necessary, because the two sets of roles are already in the React contexts.

- Show the core roles (`TREAdmin`, `TREUser`) in the user menu. For example, show them as the secondary text of the `Persona`, or as a section of the menu.
- In a workspace, also show the roles that the user has in that workspace (`WorkspaceOwner`, `WorkspaceResearcher`, `AirlockManager`).
- Show an easy-to-read name for each role, not the value from the token. For example, show "TRE Administrator" for `TREAdmin`. A link to the [User roles](https://microsoft.github.io/AzureTRE/latest/azure-tre-overview/user-roles/) page is also useful.
- Show a message when the user has no roles in the current scope. For example, show "No roles assigned". This condition is the most difficult one to find now.

## Acceptance criteria

- [ ] The user can always see the core TRE roles in the UI. The user does not have to open a workspace first.
- [ ] When the user opens a workspace, the user can also see the roles that the user has in that workspace.
- [ ] The UI shows an easy-to-read name for each role. It does not show the value from the token.
- [ ] When the user has no roles in the current scope, the UI shows a message. It does not show an empty area.
- [ ] When a TRE Admin opens a workspace in which that user has no workspace role, the UI shows this condition correctly. The message in `WorkspaceProvider.tsx` refers to this condition.
- [ ] The UI gets the roles from the same token claims that `SecuredByRole` uses. Thus the indicator and the available functions always agree.
- [ ] When you change an app role assignment in Entra ID and then sign in again, the indicator shows the new roles. Thus the person who does the test can make sure that the new role is in effect.
- [ ] Unit tests examine the new code. The tests must include the condition in which the user has no roles. Put the tests with `UserMenu.test.tsx` and `SecuredByRole.test.tsx`.
- [ ] Update the documentation if you add a new element to the UI.

## Related issues

- #4376 asks for better descriptions of the role permissions in Entra ID and in the documentation. That issue tells the user what a role does. This issue tells the user which roles the user has.
- #3826 asks for more detailed role permissions. If the number of roles increases, a role indicator becomes more useful.
- #2861 (closed) added an error message for a user who does not have a necessary role. That message tells the user that a role is absent. This issue tells the user which roles the user has.
- #4049 (closed) added the Users tab of a workspace. That tab shows the roles of the other users. It does not show the roles of the user who signs in.
- #3809 (closed) added the TRE version and the deployment time to the UI. `Footer.tsx` shows this data. That change is similar, because it also adds status data to the UI.

## Additional context

A token can contain these roles:

| Scope | Value | Defined in |
| --- | --- | --- |
| App registration for the core API | `TREAdmin` | `devops/scripts/aad/create_api_application.sh` |
| App registration for the core API | `TREUser` | `devops/scripts/aad/create_api_application.sh` |
| App registration for a workspace | `WorkspaceOwner` | `templates/workspaces/base/terraform/aad/aad.tf` |
| App registration for a workspace | `WorkspaceResearcher` | `templates/workspaces/base/terraform/aad/aad.tf` |
| App registration for a workspace | `AirlockManager` | `templates/workspaces/base/terraform/aad/aad.tf` |

There is one more item, but it is not part of this request. The role `TRERole.AirlockAutomation = "TREAirlockAutomation"` is in [`api_app/auth/models.py#L10`](https://github.com/microsoft/AzureTRE/blob/3b773995fb730b54b09028e3a5b6cc4a9b78bae1/api_app/auth/models.py#L10). The script `create_api_application.sh` does not create this role. No route and no test uses it. I can make a different issue for this role if it is not necessary.

Azure TRE release version: main (0.29.0)

JayDoubleu:
# Resolves #5051

## What is being addressed

The UI knows the roles of the user, but it does not show them. The core roles are
in `AppRolesContext` and the workspace roles are in `WorkspaceContext`. The
components that read them use them only to select which elements to show.

Thus two different conditions look the same. `SecuredByRole` removes an element
fully when the user does not have the role and the caller gives no error message.
A Researcher who does not see the "Create new" button cannot find the cause. The
cause can be the Researcher role, which is correct. The cause can also be a role
assignment that is not yet in effect.

This also makes tests more difficult. Each role sees a different page, so you must
test a change to a controlled function with each role. Today you change the app
role assignment in Entra ID, open the UI again, and then look at the buttons to
decide if the new role is in effect.

## How is this addressed

The user menu now shows the roles that the user has.

- The menu button shows a summary line below the name. The user does not have to
  open the menu to see it.
- The menu lists `TRE roles`. On a workspace route it also lists
  `Workspace roles`.
- A scope with no roles shows `None assigned`. The UI does not show an empty
  area. This condition is the most difficult one to find today.
- The UI shows a display name for each role. `TREAdmin` becomes
  `TRE Administrator`. An unknown value keeps its raw text, so a new app role is
  still shown.

The button has room for one line. Thus the summary gives the name of one role and
counts the others: `Workspace Owner +2 more`. A token lists the roles in no
guaranteed order, so the roles are sorted first and the summary names the most
privileged one. The same order applies to the menu.

Inside a workspace the summary gives the workspace roles. A TRE Admin has no
workspace role there, so the summary falls back to the core roles instead of
`No roles assigned`.

`TopNav` takes the workspace roles as a property, because it is outside the
`WorkspaceContext` provider and cannot read them. The state is already in
`App`, and `WorkspaceProvider` clears it when it unmounts.

The menu reads the same token claims that `SecuredByRole` uses. It does not read
Entra ID. Thus the menu and the available functions always agree. A difference
would show that the browser has an old token, which is useful data.

## Screenshots

One role, and more than one role:

Role summary on the menu button

A TRE Admin in a workspace with no workspace role, and a user with no roles:

Empty role states

The menu open, for a user with all three workspace roles:

User menu open

The screenshots come from a preview harness that renders the component with a
stub in place of MSAL. The harness is not part of this change.

## Checks

- `npx vitest run`: 27 files and 250 tests pass. `UserMenu.tsx` and
  `roleNames.ts` are at 100% for statements, branches, functions and lines.
- `npx eslint .` is clean. `npx prettier --check src` is clean.
  `npm run build` passes.
- `ui/app/package.json` moves to 0.8.32. There is a CHANGELOG entry, and the
  AuthN + AuthZ section of `docs/tre-developers/ui.md` is updated.

## Open points

- The role entries are menu items with no action. A read-only section in a
  contextual menu is not a usual use of the control. I can move the roles to a
  callout or a panel if that is better.
- The persona is now `size40`, because Fluent does not show the secondary text
  below that size. The bar is 50px high, so the bar itself does not change.
- The workspace roles are empty for a moment while a workspace loads. The menu
  shows `None assigned` until the roles arrive.
- The role items carry a `Contact` icon. No icon shows at present, because the
  app does not call `initializeIcons()`. The `SignOut` icon on the Logout item
  has the same condition. This is not new, and it is not part of this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 **BREAKING CHANGES**
 
 ENHANCEMENTS:
+* Show the roles held by the signed-in user in the UI user menu, for both the TRE and the current workspace, so that a hidden feature can be told apart from a role assignment that has not taken effect ([#5051](https://github.com/microsoft/AzureTRE/issues/5051))
 
 ## (0.29.0) (August 14, 2026)
 **BREAKING CHANGES**
--- a/docs/tre-developers/ui.md
+++ b/docs/tre-developers/ui.md
@@ -36,6 +36,8 @@ When the user navigates into a Workspace, the client will request an access toke
 Workspace app registrations may be reused across multiple workspaces in development scenarios.
 From this access token we can find the Workspace-level roles the user is in (`WorkspaceOwner` / `WorkspaceResearcher`). These are in turn used to show/hide features of the UI.
 
+The roles from both token types are also shown to the user in the user menu in the top navigation bar, under `TRE roles` and `Workspace roles`. The workspace section only appears on a workspace route, and shows `None assigned` when the user holds no role in that workspace, which is the case for a `TREAdmin` viewing a workspace they are not a member of. The menu reads the same role claims that `SecuredByRole` gates on, so what it shows always matches what the UI lets the user do.
+
 ### React Contexts
 The React Context API is a clean way to handle a limited amount of global state, and is used for a few scenarios in this project:
 - TRE Roles Context: A context provides details of the base TRE roles a user is in, which can be consumed anywhere throughout the app
--- a/ui/app/package-lock.json
+++ b/ui/app/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "tre-ui",
-  "version": "0.8.31",
+  "version": "0.8.32",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "tre-ui",
-      "version": "0.8.31",
+      "version": "0.8.32",
       "dependencies": {
         "@azure/msal-browser": "^2.35.0",
         "@azure/msal-react": "^1.5.12",
--- a/ui/app/package.json
+++ b/ui/app/package.json
@@ -1,6 +1,6 @@
 {
   "name": "tre-ui",
-  "version": "0.8.31",
+  "version": "0.8.32",
   "private": true,
   "type": "module",
   "dependencies": {
--- a/ui/app/src/App.tsx
+++ b/ui/app/src/App.tsx
@@ -90,7 +90,7 @@ export const App: React.FunctionComponent = () => {
                   />
                   <Stack styles={stackStyles} className="tre-root">
                     <Stack.Item grow className="tre-top-nav">
-                      <TopNav />
+                      <TopNav workspaceRoles={workspaceRoles} />
                     </Stack.Item>
                     <Stack.Item grow={100} className="tre-body">
                       <GenericErrorBoundary>
--- a/ui/app/src/components/shared/TopNav.tsx
+++ b/ui/app/src/components/shared/TopNav.tsx
@@ -5,7 +5,13 @@ import { UserMenu } from "./UserMenu";
 import { NotificationPanel } from "./notifications/NotificationPanel";
 import config from "../../config.json";
 
-export const TopNav: React.FunctionComponent = () => {
+interface TopNavProps {
+  // Passed through to the user menu, which sits outside the WorkspaceContext
+  // provider and so cannot read the workspace roles itself.
+  workspaceRoles?: Array<string>;
+}
+
+export const TopNav: React.FunctionComponent<TopNavProps> = (props: TopNavProps) => {
   return (
     <>
       <div className={contentClass}>
@@ -29,7 +35,7 @@ export const TopNav: React.FunctionComponent = () => {
             <NotificationPanel />
           </Stack.Item>
           <Stack.Item grow>
-            <UserMenu />
+            <UserMenu workspaceRoles={props.workspaceRoles} />
           </Stack.Item>
         </Stack>
       </div>
--- a/ui/app/src/components/shared/UserMenu.test.tsx
+++ b/ui/app/src/components/shared/UserMenu.test.tsx
@@ -1,7 +1,10 @@
 import React from "react";
 import { describe, it, expect, vi, beforeEach } from "vitest";
 import { render, screen, fireEvent } from "@testing-library/react";
+import { MemoryRouter } from "react-router-dom";
 import { UserMenu } from "./UserMenu";
+import { AppRolesContext } from "../../contexts/AppRolesContext";
+import { RoleName, WorkspaceRoleName } from "../../models/roleNames";
 
 // Mock MSAL
 const mockLogout = vi.fn();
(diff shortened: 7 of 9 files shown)


marrobi:
@JayDoubleu thanks. I think below the name looks cluttered. Thoughts on these options:

image

I quite like number 2.

PR #5081, opened 2026-09-16
Fix UI router security advisories with React Router 7.18.4

JayDoubleu:
## What is being addressed

The UI locks React Router 6.30.4, which is covered by three remaining Dependabot alerts: [531](https://github.com/microsoft/AzureTRE/security/dependabot/531), [483](https://github.com/microsoft/AzureTRE/security/dependabot/483) and [482](https://github.com/microsoft/AzureTRE/security/dependabot/482).
A 6.x patch does not leave all reported vulnerable ranges.

## How is this addressed

- Upgrade `react-router-dom` and `react-router` to 7.18.4 and update their locked dependencies.
- Preserve the existing declarative router and application components.
- Add regression tests for nested navigation, relative panel dismissal, history and URL handling.
- Bump the UI version to 0.8.36 and update the changelog.

## Validation

Validated runtime candidate `bdb1e193d389b62227321780cde4ef81e9654361`.
The follow-up commit only adds this PR reference to the changelog; UI source and dependencies are unchanged.

- Node 24: clean install, 296 tests, lint, production build and dependency checks passed. Full npm audit reported zero vulnerabilities.
- Four external-navigation rejection tests fail on the old router and pass on 7.18.4. The 17 new application-navigation cases pass on both versions.
- Deployed the production UI build to an existing test TRE and verified the served artifact hashes.
- Automated Chrome checks passed for authenticated WorkspaceOwner navigation through workspace, service and user-resource pages, direct rendering/refresh, Back/Forward, query/fragment preservation and logout.
- No new routing crash was identified in those flows. The original UI was restored after testing.


maxmartin-cgi (approved):

JayDoubleu:
All reported CI checks are green. We also deployed the exact UI candidate and tested WorkspaceOwner navigation.

SvenAelterman (approved):
LGTM

marrobi (commented):
I believe we thought might be an issue with this...

"Could we check relative navigation under `requests/*` and `users/*` before merging? React Router v7 changes splat-path resolution, so calls such as `navigate("new")` and `navigate(item.id)` may append to an already-open child URL rather than navigate to a sibling. The new tests cover opening from the parent and closing panels, but not this case. Please add regression coverage and use explicit workspace-scoped paths where needed."

PR #5085, opened 2026-09-18
Prep for release v0.29.1

JayDoubleu:
Release tracking: #5084

## What is being addressed

The 19 PRs merged since v0.29.0 are still listed as unreleased. They include security patches, runtime fixes, GitHub automation improvements, the scheduled cleanup fix in #5087, and the case-study correction in #5072 that has not reached the website's `latest` documentation.

Prepare v0.29.1 as a maintenance release. No breaking TRE changes were identified in the reviewed range, v0.29.0 through 8e31021f6609236b87a7fc3a4fb7d1f13ae2dafd.

## How is this addressed

- Move the pending notes into a v0.29.1 section and add a fresh `Unreleased` section.
- Include all 19 merged PRs, including the missing PR bot, Dependabot and CLI metadata changes.
- Refresh all 28 entries from `devops/scripts/list_versions.sh` and add the API and resource processor versions from their source files.
- Add the full changelog comparison link. Leave the release date unset until publication.

The diff changes only `CHANGELOG.md`.

## Validation

- Markdown lint passes with the repository's `.markdownlint.json` configuration.
- `git diff --check` passes.
- All 19 merged PRs are linked in the v0.29.1 notes.
- All 30 component entries match the inventory script or their version files.
- The v0.29.0 and older changelog sections are unchanged.

## Before publication

Complete the remaining release checks in #5084. Deployment and end-to-end validation of the selected release commit are still required; the main deployment workflow is currently manually disabled.

Publishing a pre-release triggers the current documentation workflow and moves its `latest` alias immediately. This PR does not change that behaviour. Review the timing before publishing.

PR #5087, opened 2026-09-19
Defer scheduled CI cleanup while Actions runs are active

### LINKED ISSUE #5086: Scheduled CI cleanup stops PR environments while E2E tests are still running
**Describe the bug**

`Clean Validation Environments` can stop a PR environment while its deployment or E2E workflow is still active. It uses branch activity to decide whether to stop the environment, without first checking for active PR test runs.

This interrupted fixture teardown during extended testing for #5085.

**Evidence**

On 18 September 2026, all times UTC:

- **18:56:43:** All four selected extended tests had passed. Fixture teardown continued.
- **20:16:24:** Scheduled cleanup marked the PR environment for stopping because the branch had no recent activity.
- **20:16:55:** Cleanup began stopping the Application Gateway.
- **20:17:42:** E2E teardown began reporting connection timeouts.
- **20:19:17:** The test job failed after exhausting its retries.

The final result was **4 passed, 1 teardown error**, after approximately 3 hours 15 minutes.

Sources: [scheduled cleanup run](https://github.com/microsoft/AzureTRE/actions/runs/35390514610), [extended E2E job](https://github.com/microsoft/AzureTRE/actions/runs/35357600332/job/105687777224).

**Steps to reproduce**

1. Start `/test-extended` on an open PR.
2. Let the branch age exceed the configured cleanup stop threshold while the workflow remains active.
3. Run `Clean Validation Environments` while tests or fixture teardown are still running.

**Expected behaviour**

Cleanup should check workflow activity before stopping or deleting environments. Running and queued deployment/test workflows should prevent cleanup from interrupting them.

If GitHub workflow activity cannot be checked, cleanup should leave the environments untouched.

**Cause**

In [`clean_ci_validation_envs.sh`](https://github.com/microsoft/AzureTRE/blob/b57b6af0f4026b337211c0ace029fc27b79d58d5/devops/scripts/clean_ci_validation_envs.sh), the active-workflow check happens after PR environment cleanup and only protects workspace deletion on `main`.

Comment-triggered PR workflows also report the default branch, so matching only the PR branch would miss them.

**Azure TRE release version**

`main` at `b57b6af0`, while validating #5085.

**Deployed Azure TRE components**

Not applicable. This issue concerns the CI cleanup workflow and its environment-control script.

**Related issue**

#3482 reported cleanup deleting workspace resource groups during a deployment on `main` and is closed. This report covers stopping PR environments while their E2E workflow remains active.

JayDoubleu:
# Resolves #5086

## What is being addressed

Scheduled CI cleanup can stop a PR environment while its E2E workflow is still running. It checks branch age before checking workflow activity, and the existing workflow check only protects workspace deletion on `main`.

During extended testing for #5085, all four selected tests passed, but scheduled cleanup stopped the Application Gateway while fixture teardown was still polling the API. Teardown then failed with `httpx.ConnectTimeout`. The timeline and workflow logs are recorded in #5086.

## How is this addressed

- Check repository-wide workflow activity before any environment cleanup. This covers comment-triggered PR tests, which report the default branch.
- Defer cleanup for `requested`, `waiting`, `pending`, `queued` and `in_progress` runs. Exclude only the current cleanup run by its run ID.
- Follow all response pages and require each page to contain a `workflow_runs` array. Stop cleanup if the GitHub query, parsing or response validation fails.
- Repeat the check before deleting workspace resource groups left behind by tests on `main`.
- Add 15 mocked regression tests. Run them in Build Validation when the cleanup script, its tests or the validation workflow changes.
- Bump DevOps from `0.6.4` to `0.6.5` and record the version and PR reference in `CHANGELOG.md`.

## Before and after

| Scenario | Before | After |
| --- | --- | --- |
| An old PR branch still has an active E2E run | Cleanup can stop its environment before checking workflow activity. | The activity check defers the entire cleanup sweep, regardless of branch age. |
| A PR test was started by a comment | The run reports the default branch, and the existing check does not protect PR environment cleanup. | The repository-wide check catches the run without relying on its branch name. |
| A run is queued, waiting, pending or requested | These states are not checked. | Any other run in these states defers cleanup. |
| An active run is on a later API response page | The existing check reads only the first page. | The check follows all pages. |
| The workflow query fails or returns invalid data | The late check can allow main workspace deletion after an API failure. | The script exits with an error before further cleanup. |
| No other workflow runs are active | Branch age, PR state and branch existence determine cleanup. | The existing cleanup rules still apply. |
| A PR changes the cleanup script or its tests | Build Validation has no dedicated cleanup regression suite. | The 15-test suite runs for relevant changes; unrelated PRs skip it. |

## Validation

- All 15 regression tests passed locally at `14dfda3c` with Azure, GitHub, Git and date commands mocked.
- The malformed-response regression cases fail against the previous code and pass with this fix. Coverage includes invalid first and later pages, the second activity check, and valid empty arrays.
- The original script fails the regression covering all five protected workflow states because it stops the PR environment.
- ShellCheck, Flake8, Bash syntax, workflow YAML parsing and `git diff --check` passed.

Live Azure cleanup and E2E validation remain pending. The local tests do not contact Azure or GitHub.

## Trade-offs and limits

The guard defers the whole sweep, including cleanup of unrelated idle environments. Other workflows can therefore keep those resources running longer.

Both activity checks are snapshots. A workflow can start after either check, so this change does not serialise deployment and cleanup.


JayDoubleu:
Copilot feedback is addressed, including validation of malformed workflow responses. Build Validation passed on 14dfda3c, including all 15 cleanup regression tests and lint.

I am bypassing Azure deployment and smoke E2E because those jobs do not exercise the scheduled cleanup script. Dedicated mocked tests cover the cleanup guard. The new guard has not been tested against live Azure.

marrobi (approved):

ChrisChapman-gh (approved):

Issue #5083, opened 2026-09-18
OHDsi bundle build failing due to Debian Bullseye packages returning 404


ansarrafique:
I'm deploying Azure TRE v0.29.0 and the OHDsi workspace service is currently failing during the bundle build.

I deployed the same version around two weeks ago without any issues, but it has now started failing when installing packages from the Debian Bullseye repositories.

The OHDsi Dockerfile currently uses:

`FROM --platform=linux/amd64 debian:bullseye-slim`

The failure happens at:

`RUN apt-get update && apt-get install -y curl gnupg`

Several packages are now returning 404 Not Found, for example:

```
E: Failed to fetch http://deb.debian.org/debian-security/pool/updates/main/o/openssl/openssl_1.1.1w-0%2bdeb11u8_amd64.deb
404 Not Found

E: Failed to fetch http://deb.debian.org/debian-security/pool/updates/main/c/curl/libcurl4_7.74.0-1.3%2bdeb11u16_amd64.deb
404 Not Found

E: Failed to fetch http://deb.debian.org/debian-security/pool/updates/main/g/gnupg2/gpg_2.2.27-2%2bdeb11u3_amd64.deb
404 Not Found

E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
```

This eventually causes the OHDsi bundle build to fail.

As Debian 11 Bullseye reached the end of LTS support on 31 August 2026, I wonder if the OHDsi Dockerfile now needs to be updated to use a newer supported Debian version, along with the corresponding Microsoft package repository.

JayDoubleu:
Confirmed in a fresh GitHub-hosted bundle build: Debian Bullseye package downloads return `404 Not Found`, including OpenSSL and curl. [Build log](https://github.com/JayDoubleu/AzureTRE/actions/runs/35462657277/job/105949276797).

I will check the base image and Microsoft package installation to see whether a small update resolves this. The OHDSI fix will be separate from the proposed build-check workflow in #5089.

PR #5089, opened 2026-09-19
Add Dockerfile and Porter bundle build checks

### LINKED ISSUE #5088: Discuss scheduled Dockerfile and Porter bundle build checks
Draft implementation for discussion: #5089.

This proposal adds regular Dockerfile and Porter bundle builds to find broken image or package dependencies before a TRE deployment.

## Problem

External dependencies can break a build even when the repository has not changed. The OHDSI package-download failure in #5083 is an example. A regular build check can expose these failures before deployment validation runs.

## Proposed behaviour

- Build all 33 current targets each Monday at 06:00 UTC and on manual request: eight standalone Dockerfiles and 25 Porter bundle images.
- On pull requests, select affected targets, including bundles that consume shared source files. Changes to shared Porter tools select all bundles.
- Reject incomplete tracked Porter bundles during discovery. Each bundle must contain both `porter.yaml` and `Dockerfile.tmpl`. Complete bundle removals remain valid.
- Retry a failed target once after ten seconds, including Porter installation and the build. Keep both attempts in the log and fail the check if both fail.
- Report every selected target. Missing or invalid results, failed artifact uploads and cancelled jobs also fail the checks.
- Retain result artifacts for seven days. Replace a target's result when its job runs again, so a successful retry cannot leave a stale failure.

The workflow uses GitHub-hosted runners and read-only repository permissions. It needs no Azure deployment credentials and does not deploy resources or publish images. Scheduled runs start after the workflow is merged into the default branch. Existing deployment validation remains necessary to test deployed services.

## Current evidence

At commit `95c00347` on 21 September 2026:

- [All 33 image builds and the final report pass](https://github.com/microsoft/AzureTRE/actions/runs/35600971780), including OHDSI and the dev container.
- [Build Validation passes](https://github.com/microsoft/AzureTRE/actions/runs/35600971966), including all 155 GitHub-script tests and lint checks.
- The workflow previously detected the OHDSI failure. OHDSI now builds successfully with the fix from #5090, which is merged into `main` and included in the draft.

## Feedback requested

- Does the weekly schedule and selected pull request coverage fit the project's CI needs?
- Are workflow checks and summaries sufficient, or should scheduled failures also create or update a tracking issue?

Automatic issue creation is not included in the draft. If wanted, a separate reporting job would need `issues: write` and a way to update one existing issue without creating duplicates. Build jobs would retain read-only permissions.

JayDoubleu:
# Resolves #5088

Add regular builds for Dockerfiles and Porter bundle images so broken external dependencies can be detected before deployment. The current matrix contains eight standalone Dockerfiles and 25 Porter bundle images.

## Behaviour

- Build every target each Monday at 06:00 UTC and on manual request. Scheduled runs start after the workflow reaches the default branch.
- Select affected targets on pull requests to `main` or `feature/**`. Changes within shared build contexts select the source bundle and every consuming bundle, including when files are deleted.
- Select all bundles when shared Porter tools change, and all targets when this workflow or its scripts change.
- Reject incomplete tracked Porter bundles during discovery. Each bundle must contain both `porter.yaml` and `Dockerfile.tmpl`. Complete bundle removals remain valid.
- Retry a failed target once after ten seconds, including Porter setup and the build. Each Porter attempt uses a new installation directory. Keep both logs, fail after two failed attempts, and stop on cancellation.
- Report every selected target. Fail on repeated build failures, cancelled jobs, failed artifact uploads, or missing or invalid results.
- Retain result artifacts for seven days and replace them on job reruns. Key cancellation by PR number so unrelated forks cannot cancel each other's builds.

The workflow reuses the repository's Porter version pins and build helper. Build jobs have read-only repository permissions and need no Azure deployment credentials. They do not publish images or deploy resources. Existing deployment validation remains necessary to test deployed services.

## Why retry and replace results?

A Porter download failed with a connection reset and passed on retry. One automatic retry allows a temporary download failure to recover while repeated failures still fail the check.

That successful retry also exposed an old failed result artifact in the report. Replacing each target's artifact prevents the report from retaining that stale result.

## Validation

At `95c00347` on 21 September 2026:

- [All 33 image builds and the final report pass](https://github.com/microsoft/AzureTRE/actions/runs/35600971780).
- [Build Validation passes](https://github.com/microsoft/AzureTRE/actions/runs/35600971966), including all 155 GitHub-script tests and lint checks.
- Local checks pass for Actionlint, ShellCheck, YAML, Bash syntax, Markdown and the strict documentation build.
- Regression tests cover retries, repeated failures, cancellation, partial Porter installation, result validation and named-context dependencies. The dependency tests read the tracked context declarations and cover deleted source files. New tests cover incomplete bundle additions, partial deletions, untracked companions and complete bundle removals.
- A [job rerun and its dependent report](https://github.com/microsoft/AzureTRE/actions/runs/35594852796/attempts/2) passed after the artifact replacement fix.

The workflow caught the OHDSI failure from #5083. OHDSI now passes with #5090, which is merged into `main` and included here. The dev-container build also passes.

## Discussion

This remains a draft for feedback on the schedule, PR coverage and failure reporting. Automatic issue creation is not included. If wanted, a separate reporting job could maintain one tracking issue with `issues: write`; build jobs would retain read-only permissions.


JayDoubleu:
The conflicts are resolved and current `main` is included. OHDSI now passes with the fix from #5090, which was merged into `main`: [successful OHDSI build](https://github.com/microsoft/AzureTRE/actions/runs/35595860516/job/106320411726). This confirms that the workflow caught the original failure and passes with the corrected bundle.

The [latest run at `fa1c1428`](https://github.com/microsoft/AzureTRE/actions/runs/35595860516) passes all 33 image builds and the final report. All 138 GitHub-script tests also pass in [Build Validation](https://github.com/microsoft/AzureTRE/actions/runs/35595860406).

The dev-container difference is `INTERACTIVE=true` here, matching VS Code, versus the default `INTERACTIVE=false` in deployment CI. Deployment CI skips the GitHub CLI installer. The earlier failure followed an empty release lookup that the installer did not reject. The original API response was not logged. Subsequent dev-container builds passed without installer changes.

A Porter download connection reset passed on retry. The workflow now retries each failed target once after ten seconds, including Porter installation and the build. Both attempts stay in the log. Two failures fail the check. Result artifacts are replaced on job reruns so an earlier failure cannot remain in the report. The PR description explains this behaviour and its reason.

JayDoubleu:
Updated the draft and linked issue with the current behaviour and evidence at `95c00347`.

Recent changes and why they were needed:

- **One automatic retry after ten seconds:** a Porter download failed with a connection reset, then passed on retry. The retry covers setup and the build, uses a new Porter directory and keeps both attempt logs. Two failed attempts fail the check.
- **Replace result artifacts on reruns:** a successful Gitea retry left an older failed result in the report. Each target now replaces its result when its job runs again.
- **Build consumers of shared source files:** changes in the base workspace or shared Guacamole VM directories now trigger every affected bundle. Tests also cover deleted files and check the mapping against the tracked context declarations.
- **Reject incomplete Porter bundles:** deleting either `porter.yaml` or `Dockerfile.tmpl` could silently remove a broken bundle from the matrix. Discovery now checks tracked pairs before target selection. Regression tests cover partial deletions, incomplete additions, untracked companions and complete bundle removals.
- **Isolate cancellation by PR number:** different forks can use the same branch name. Their builds now have separate concurrency groups.

All three Copilot concerns identified so far are addressed. The two inline threads are resolved. The latest concern is fixed in `95c00347`. It appeared only in the review summary, so it has no thread to resolve. At this commit, [all 33 image builds and the report pass](https://github.com/microsoft/AzureTRE/actions/runs/35600971780), and [Build Validation passes with 155 tests](https://github.com/microsoft/AzureTRE/actions/runs/35600971966).

OHDSI now passes with the merged fix from #5090. The dev-container build also passes. The earlier GitHub CLI installer failure remains a separate weakness: an empty release lookup was not rejected, and deployment CI skips that installer.

The PR remains a draft for discussion. Feedback is welcome on the weekly schedule, PR coverage and whether scheduled failures need a tracking issue as well as workflow summaries. Automatic issue creation is not included.

ChrisChapman-gh (approved):

maxmartin-cgi (approved):
Approving, codex did raise this issue that does not relate to scheduled checks:

  - P1 — Build-input changes do not trigger validation. The workflow only triggers for Dockerfile/Porter manifest paths, while the matrix only selects a plain image when its Dockerfile changes. For example, api_app/Dockerfile copies and installs api_app/requirements.txt, but a PR changing that file matches neither the workflow paths nor the selection condition—so no Dockerfile build runs. The same applies to the root requirements.txt copied by the dev-container image and other files in image/bundle contexts. Expand the path filters and selection logic to cover each target’s complete build context (and Porter bundle contents). workflow line 16, selection logic line 109.

I think it would be worth taking a closer look at our suite of checks and seeing what we can consolidate at some point down the line.

marrobi (approved):
If adds value, great.

JayDoubleu:
Agreed @maxmartin-cgi this is to prevent domino effect from these failures and first attempt at making CI more robust and we should consolidate a lot of these checks as part of wider CI rewrite

